PR #46, opened 2021-04-29
Distribution WIP

dianaclarke:


conbench/entities/summary.py
```diff
@@ -148,6 +151,18 @@ def create(data):
             bulk.append(Time(result=x, summary_id=summary.id, iteration=i + 1))
         Time.bulk_save_objects(bulk)
 
+        distribution = get_distribution(repository, sha, 1000).first()
```
jonkeane: Am I reading this + following the logic in the function calls that this will get a distribution based on the last 1000 commits (as specified in our index)? 

What's the `.first()` doing there? I'm sure it's right — but curious what's going on with it?

conbench/entities/summary.py
```diff
@@ -148,6 +151,18 @@ def create(data):
...
+        if distribution:
+            values = dict(distribution)
+            Session.execute(
+                insert(Distribution.__table__)
+                .values(values)
+                .on_conflict_do_update(
```
jonkeane: OOH this is nice. I punted on the sql to do this, I'm glad sql alchemy (or our infrastructure?) makes this easy!

conbench/entities/summary.py
dianaclarke: I use `first` here because I know it will always be a list of size 1, and I don't want to have to always go `distribution 
 = distributions[0]`, first does that for me (with a `LIMIT 1`)

dianaclarke: PS. Yes, 1000 commits from the index.

jonkeane: Cool, more is almost certainly better, so we can leave it as is, but I was testing this with 100 commits and it was doing a decent job of capturing the variation. Happy to leave it at 1000, but if we run into any bottlenecks, feel free to drop that limit lower

PR #60, opened 2021-05-14
Backfill distribution table

dianaclarke:


migrations/versions/0d4e564b1876_backfill_distribution.py
```diff
@@ -0,0 +1,186 @@
...
+    runs_by_id = {r["id"]: r for r in runs}
+    commits_by_id = {c["id"]: c for c in commits}
+    machines_by_id = {m["id"]: m for m in machines}
+
+    seen = {
+        f'{d["sha"]}{d["case_id"]}{d["context_id"]}{d["machine_id"]}'
+        for d in distributions
+    }
+
+    summaries = connection.execute(summary_table.select())
```
ElenaHenderson: @dianaclarke If we only select summary rows with run_ids for runs with name like `commit: %`, we can cut down loop iterations (`for summary in summaries`) by more than 50%:


```
select count(1) 
from summary

2288167

select count(1) 
from summary, machine
where summary.machine_id = machine.id and
machine.name in ('ursa-i9-9960x', 'ec2-t3-large-us-east-2', 'ec2-t3-xlarge-us-east-2', 'ursa-thinkcentre-m75q') 

1525400


select count(1) 
from summary, machine, run
where summary.machine_id = machine.id and
machine.name in ('ursa-i9-9960x', 'ec2-t3-large-us-east-2', 'ec2-t3-xlarge-us-east-2', 'ursa-thinkcentre-m75q') and
summary.run_id = run.id and
run.name like 'commit: %'

993577


select count(1) 
from summary, run
where summary.run_id = run.id and
run.name like 'commit: %'

1036257
```

PR #133, opened 2021-07-29
Drop sha and repository

dianaclarke:


migrations/versions/847b0850ea81_drop_sha_and_repository.py
```diff
@@ -0,0 +1,60 @@
...
+        "distribution", "commit_id", existing_type=sa.VARCHAR(length=50), nullable=False
+    )
+    op.drop_index("distribution_repository_index", table_name="distribution")
+    op.drop_index("distribution_sha_index", table_name="distribution")
+    op.drop_index("distribution_index", table_name="distribution")
+    op.create_index(
+        "distribution_index",
+        "distribution",
+        ["case_id", "context_id", "commit_id", "machine_hash"],
+        unique=True,
```
ElenaHenderson, line 30: @dianaclarke will you please also drop `distribution_case_id_index` since composite index `distribution_index` can be used by Postgres to do the same?

PR #165, opened 2021-09-10
[API]: Adding api feature to support delete run executions

davisusanibar:
Adding api feature to support delete run executions

### DIFF
--- a/conbench/api/runs.py
+++ b/conbench/api/runs.py
@@ -1,7 +1,10 @@
+import flask_login
+
 from ..api import rule
 from ..api._endpoint import ApiEndpoint, maybe_login_required
 from ..entities._entity import NotFound
 from ..entities.run import Run, RunSerializer
+from ..entities.summary import Summary
 
 
 class RunEntityAPI(ApiEndpoint):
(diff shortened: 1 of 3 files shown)


conbench/tests/api/test_runs.py
```diff
@@ -76,3 +77,14 @@ def test_run_list_filter_by_sha_and_machine_no_match(self, client):
         args = urllib.parse.urlencode(args)
         response = client.get(f"/api/runs/?{args}")
         self.assert_200_ok(response, [])
+
+
+class TestRunDelete(_asserts.DeleteEnforcer):
+    url = "api/runs/{}/"
+
+    def test_delete_run(self, client):
```
dianaclarke, line 88: Can you make this test follow the same pattern as:

```
    def test_delete_benchmark(self, client):
        self.authenticate(client)
        summary = _fixtures.create_benchmark_summary()

        # can get before delete
        Summary.one(id=summary.id)

        # delete
        response = client.delete(f"/api/benchmarks/{summary.id}/")
        self.assert_204_no_content(response)

        # cannot get after delete
        with pytest.raises(NotFound):
            Summary.one(id=summary.id)
```

Perhaps:

```
    def test_delete_run(self, client):
        self.authenticate(client)
        summary = _fixtures.create_benchmark_summary()
        run_id = summary.run_id

        # can get before delete
        Run.one(id=run_id)

        # delete
        response = client.delete(f"/api/runs/{run_id}/")
        self.assert_204_no_content(response)

        # cannot get after delete
        with pytest.raises(NotFound):
            Run.one(id=run_id)
```

conbench/api/runs.py
```diff
@@ -34,6 +37,30 @@ def get(self, run_id):
...
+            "404": "404"
+        parameters:
+          - name: run_id
+            in: path
+            schema:
+                type: string
+        tags:
+          - Runs
+        """
+        summaries = Summary.all(run_id=run_id)
```
dianaclarke, line 57: As mentioned in slack, we should cascade on delete rather than first deleting each benchmark result in a run.

https://docs.sqlalchemy.org/en/14/orm/cascades.html#delete

This can be done in a follow-up pull request though.

conbench/tests/api/test_runs.py
davisusanibar: Changed

PR #166, opened 2021-09-10
[UI]: Adding ui feature to support delete run executions

davisusanibar:
Adding user interface feature to support delete run executions

### DIFF
--- a/conbench/app/runs.py
+++ b/conbench/app/runs.py
@@ -13,6 +13,8 @@ from ..config import Config
 class RunPlot(AppEndpoint, ContextMixin, RunMixin):
     def page(self, benchmarks, baseline_run, contender_run, form):
         compare_runs_url = None
+        if not flask_login.current_user.is_authenticated:
+            delattr(form, "delete")
         if baseline_run and contender_run:
             compare = f'{baseline_run["id"]}...{contender_run["id"]}'
             compare_runs_url = f.url_for("app.compare-runs", compare_ids=compare)
--- a/conbench/app/runs.py
+++ b/conbench/app/runs.py
@@ -11,7 +11,7 @@ from ..config import Config
 
 
 class RunPlot(AppEndpoint, ContextMixin, RunMixin):
-    def page(self, benchmarks, baseline_run, contender_run, form):
+    def page(self, benchmarks, baseline_run, contender_run, form, run_id):
         compare_runs_url = None
         if not flask_login.current_user.is_authenticated:
             delattr(form, "delete")
@@ -25,6 +25,7 @@ class RunPlot(AppEndpoint, ContextMixin, RunMixin):
             title="Run",
             benchmarks=benchmarks,
             compare_runs_url=compare_runs_url,
+            run_id=run_id,
             form=form,
         )
 
@@ -33,9 +34,6 @@ class RunPlot(AppEndpoint, ContextMixin, RunMixin):
             return self.redirect("app.login")
 
         contender_run, baseline_run = self.get_display_run(run_id), None
-        if contender_run is None:
-            self.flash("Error getting run.")
-            return self.redirect("app.index")
         if contender_run:
             baseline_url = contender_run["links"].get("baseline")
             if baseline_url:
@@ -50,7 +48,7 @@ class RunPlot(AppEndpoint, ContextMixin, RunMixin):
         for benchmark in benchmarks:
             augment(benchmark, contexts)
 
-        return self.page(benchmarks, baseline_run, contender_run, DeleteForm())
+        return self.page(benchmarks, baseline_run, contender_run, DeleteForm(), run_id)
 
     def _get_benchmarks(self, run_id):
         response = self.api_get("api.benchmarks", run_id=run_id)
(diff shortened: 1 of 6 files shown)


conbench/templates/run.html
```diff
@@ -51,6 +52,7 @@
                 {% endfor %}
             </tbody>
         </table>
+        {{ wtf.quick_form(form, id="run-form", button_map={'delete': 'primary'}) }}
```
dianaclarke: This button should only show up if you're authenticated, right now it shows up either way. The API layer won't actually let you delete it if you're unauthenticated, but it should still be hidden.

For benchmark delete, this was accomplished like so:

```
class Benchmark(AppEndpoint, BenchmarkMixin, RunMixin, TimeSeriesPlotMixin):
    def page(self, benchmark, run, form):
        if not flask_login.current_user.is_authenticated:
            delattr(form, "delete")
```

I suspect something similar can be done for deleting runs.

conbench/tests/app/test_runs.py
```diff
@@ -5,8 +5,44 @@
...
+        self.assert_page(response, "Run")
+        assert f"{run_id}</li>".encode() in response.data
+
+        data = {"delete": ["Delete"], "csrf_token": self.get_csrf_token(response)}
+        response = client.post(f"/runs/{run_id}/", data=data, follow_redirects=True)
+        self.assert_page(response, "Home")
+        assert b"Run deleted." in response.data
+
+        response = client.get(f"/runs/{run_id}/", follow_redirects=True)
+        print(response.data)
```
dianaclarke: remove `print(response.data)`

conbench/templates/run.html
davisusanibar: Thanks, changed

conbench/tests/app/test_runs.py
davisusanibar: Changed

conbench/tests/app/test_runs.py
```diff
@@ -5,8 +5,43 @@
 class TestRunGet(_asserts.GetEnforcer):
     url = "/runs/{}/"
     title = "Run"
-    redirect_on_unknown = False
```
dianaclarke: This needs to be added back.

The run UI should not redirect to the home page if it doesn't find the run UUID. This is because runs UUIDs are created outside of Conbench by the orchestration layer when a benchmark run kicks off and those run URLs are immediately posted as a GitHub comment. Users should immediately be able to click on those run links, even if benchmark results for that run UUID haven't started coming in yet. Initially, the run page will be an empty datatable, but results will eventually start trickling in.

conbench/app/runs.py
```diff
@@ -41,15 +50,38 @@ def get(self, run_id):
...
+        if form.delete.data:
+            if form.validate_on_submit():
+                response = self.api_delete(
+                    "api.run",
+                    run_id=run_id,
+                )
+                if response.status_code == 204:
+                    self.flash("Run deleted.")
+                else:
+                    self.flash("Error deleting runs.")
```
dianaclarke: `"Error deleting runs."` -> `"Error deleting run."`

conbench/app/runs.py
```diff
@@ -20,13 +25,17 @@ def page(self, benchmarks, baseline_run, contender_run):
...
             compare_runs_url=compare_runs_url,
+            form=form,
         )
 
     def get(self, run_id):
         if self.public_data_off():
             return self.redirect("app.login")
 
         contender_run, baseline_run = self.get_display_run(run_id), None
+        if contender_run is None:
```
dianaclarke: This code needs to be removed.

```
        if contender_run is None:
            self.flash("Error getting run.")
```

As mentioned below, the run page should still render even if the run entry in the database doesn't exist yet.

conbench/templates/run.html
```diff
@@ -66,6 +68,12 @@
 
 $(document).ready(function() {
     $('#unit-tooltip').tooltip()
+    $("#run-form").find("#delete").attr("type", "button");
+    $("#run-form").find("#delete").attr("data-toggle", "modal");
+    $("#run-form").find("#delete").attr("data-target", "#confirm-delete");
+    $("#run-form").find("#delete").attr("data-form-id", "#run-form");
+    $("#run-form").find("#delete").attr("data-href", "{{ url_for('app.run', run_id=benchmarks[0].run_id) }}");
```
dianaclarke: I know it's odd, but you can't assume a run has benchmarks.

If a benchmark doesn't have runs, you'll get an error like this:

`jinja2.exceptions.UndefinedError: list object has no element 0`

You should be able to wrap in in something like this:

```
        {% if benchmarks %}
        ...
        {% endif %}
```

Better, perhaps pass the run_id to the HTML via the controller, otherwise we won't be able to delete a run that has zero benchmarks.

``` 
        return self.render_template(
            "run.html",
            title="Run",
            benchmarks=benchmarks,
            compare_runs_url=compare_runs_url,
            form=form,
            run_id=contender_run["id"],
        )
```

dianaclarke: Actually, you won't necessarily have a populated `contender_run` at that point, so instead you'll need to pass `run_id` from:

```
    def get(self, run_id):
        if self.public_data_off():
            return self.redirect("app.login")
```

into:

```
    def page(self, benchmarks, baseline_run, contender_run, form):
        compare_runs_url = None
```

And then:

```
        return self.render_template(
            "run.html",
            title="Run",
            benchmarks=benchmarks,
            compare_runs_url=compare_runs_url,
            form=form,
            run_id=run_id,
        )
```

Or something like that.

PR #201, opened 2021-10-22
Adding version tracking

drabastomek:
This PR adds support for version tracking in conbench via `setuptools_scm`. It also adds content to README.md that helps to log in to `psql`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ repository, and the results are hosted on the
     $ brew services start postgres
 
 On Linux this would be
+
     $ sudo service postgresql start 
 
 
--- a/README.md
+++ b/README.md
@@ -89,6 +89,9 @@ repository, and the results are hosted on the
 ### Start postgres
     $ brew services start postgres
 
+On Linux this would be
+    $ sudo service postgresql start 
+
 
 ### Create databases
     $ psql
@@ -102,6 +105,20 @@ repository, and the results are hosted on the
      * Environment: development
      * Debug mode: on
      * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)
+
+If this renders an authentication error, set the password for your user.
+
+    $ psql
+    # ALTER USER <username> PASSWORD '<password>';
+
+Then, set an environmental variable `DB_PASSWORD` to reflect the chosen password.
+
+    $ export DB_PASSWORD=<password>
+
+Alternatively, if you don't want to set the `DB_PASSWORD` you can change the password in postgres to `postgres`.
+
+    $ psql
+    # ALTER USER <username> PASSWORD 'postgres';
 
 
 ### Test app
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ repository, and the results are hosted on the
 
 On Linux this would be
 
-    $ sudo service postgresql start 
+    $ sudo service postgresql start
 
 
 ### Create databases
--- a/conbench/__init__.py
+++ b/conbench/__init__.py
@@ -1,5 +1,19 @@
 import os
 
+# set the version
+try:
+    import importlib.metadata as importlib_metadata
+except ImportError:
+    # TODO: remove this when Python 3.7 support is dropped
+    import importlib_metadata
+
+try:
+    __version__ = importlib_metadata.version(__name__)
+except Exception:
+    __version__ = importlib_metadata.version("conbench")
+
+del importlib_metadata
+
 
 def create_application(config):
     import flask as f
--- a/conbench/__init__.py
+++ b/conbench/__init__.py
@@ -1,4 +1,3 @@
-# set the version
 import importlib.metadata as importlib_metadata
 import os
 
--- a/conbench/__init__.py
+++ b/conbench/__init__.py
@@ -1,11 +1,7 @@
 import os
 
 # set the version
-try:
-    import importlib.metadata as importlib_metadata
-except ImportError:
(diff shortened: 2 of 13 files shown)


conbench/__init__.py
```diff
@@ -1,5 +1,19 @@
 import os
 
+# set the version
+try:
+    import importlib.metadata as importlib_metadata
+except ImportError:
+    # TODO: remove this when Python 3.7 support is dropped
```
dianaclarke: This try accept can be removed as conbench only works for Python 3.8 and greater.

drabastomek: Removing now.

drabastomek: Done.

conbench/api/index.py
```diff
@@ -10,6 +10,17 @@
 from ..api._endpoint import ApiEndpoint
 from ..db import Session
 
+# get the version
```
dianaclarke: You can do the following, rather than getting it again from `importlib_metadata`.

`from .. import __version__`

conbench/app/index.py
```diff
@@ -3,6 +3,16 @@
 from ..app.benchmarks import RunMixin
 from ..config import Config
 
+# set the version
```
dianaclarke: Same here, you can just import the version you already calculated.

`from .. import __version__`

requirements-build.txt
```diff
@@ -24,3 +24,4 @@ python-dotenv
 PyYAML
 requests
 SQLAlchemy
+setuptools_scm
```
dianaclarke: Let's keep it simple and just use i`mportlib_metadata`, rather than both `importlib_metadata` + `setuptools_scm`. Apologies for sunk cost – I very much appreciate your work.

drabastomek: The `setuptools_scm` is purely to track the versions from github tags. Just clarifying you want to revert back to `version=1.33.0` in `setup.py`?

conbench/api/index.py
drabastomek: done

conbench/app/index.py
drabastomek: done

requirements-build.txt
dianaclarke: Yup, revert back to `version=1.33.0` in `setup.py`, and remove the `setuptools_scm` dependency. We can consider using `setuptools_scm` later, but for now I don't think we need it.

dianaclarke:
Hmmm... does this build[1] error go away if you add `pip install -e .` before this line here:

https://github.com/conbench/conbench/blob/main/.github/workflows/actions.yml#L28

`importlib.metadata.PackageNotFoundError: No package metadata was found for conbench`
https://github.com/conbench/conbench/runs/3979062990?check_suite_focus=true

requirements-build.txt
drabastomek: This is done.

conbench/api/index.py
```diff
@@ -10,6 +10,9 @@
...
```
dianaclarke: minor: this comment can go

conbench/app/index.py
```diff
@@ -3,6 +3,9 @@
...
```
dianaclarke: minor: this comment can go

conbench/cli.py
```diff
@@ -32,6 +32,19 @@ def list_benchmarks():
     print(json.dumps(benchmarks, indent=2))
 
 
+@conbench.command(name="version")
```
dianaclarke, line 36: you can just import version here too:

`from . import __version__`

And then trim this to:

```
@conbench.command(name="version")
def version():
    print(f"conbench version: {__version__}")
```

drabastomek:
> Hmmm... does this build[1] error go away if you add `pip install -e .` 

Yes. My understanding of the `importlib.metadata` is that the package needs to be built/installed first before the metadata can be read. Unlike `python setup.py --version` that just reads the version from the `setup.py`.

https://docs.python.org/3/library/importlib.metadata.html

Let me add this to the build.

dianaclarke:
> > Hmmm... does this build[1] error go away if you add `pip install -e .`
> 
> Yes. My understanding of the `importlib.metadata` is that the package needs to be built/installed first before the metadata can be read. Unlike `python setup.py --version` that just reads the version from the `setup.py`.
> 
> https://docs.python.org/3/library/importlib.metadata.html
> 
> Let me add this to the build.

My bad. It should probably be: `pip install .` (no editable mode)

conbench/cli.py
drabastomek: Done.

conbench/api/index.py
drabastomek: done

conbench/app/index.py
drabastomek: done

README.md
```diff
@@ -89,6 +89,9 @@ repository, and the results are hosted on the
 ### Start postgres
     $ brew services start postgres
 
+On Linux this would be
```
dianaclarke, line 92: This isn't rendering quite right. 

https://github.com/conbench/conbench/blob/40e395b5844c7f0f829704a0f965ce7ad57dea62/README.md#start-postgres

You need a newline after `On Linux this would be`. And you have a trailing space here: `sudo service postgresql start `.

```
### Start postgres
    $ brew services start postgres

On Linux this would be

    $ sudo service postgresql start
```

drabastomek: Should be fixed now.

PR #294, opened 2021-11-17
Fix not null migration

dianaclarke:


ElenaHenderson:
Hi @dianaclarke, 

Let me know if I should merge it.

dianaclarke:
@ElenaHenderson Sorry for not responding sooner. I didn't see your comment until now.

I was trying to migrate my old ursa pqsl dump to see where the null info_ids were coming from, but I gave up migrating it. It will take days (and I already tried to migrate it a couple of times).

Anyhoo, I think this is pretty safe. And I'm trying the following to see if the progress actually shows up in datadog.

`logger = logging.getLogger("alembic.runtime.migration")`

Merging now :)

dianaclarke:
Oh! That worked! I was able to see the log messages in datadog :)

I had a typo in the log messages, but alembic ignored that error and the migration still succeeded.

PR #306, opened 2022-02-10
Support benchmarking on clusters (WIP)

ElenaHenderson:
Some benchmarks are run on machine clusters instead of one machine. 

This PR adds support for capturing benchmark results for clusters.

If we want to rename `machine` table to hardware in the database, I will create a separate PR (after this PR is merged and deployed) to rename `machine` table, foreign keys and etc.

Note that I tested the migration against a copy of Conbench production db.

Here is how UI looks like with new change:

- Runs
Screen Shot 2022-02-14 at 4 43 49 PM

- Benchmark
Screen Shot 2022-02-14 at 4 47 02 PM
44be4e.png">

- Compare Benchmarks
Screen Shot 2022-02-14 at 4 49 14 PM


conbench/entities/hardware.py
```diff
@@ -0,0 +1,182 @@
...
+            "gpu_product_names": self.gpu_product_names,
+            "links": {
+                "list": f.url_for("api.hardwares", _external=True),
+                "self": f.url_for("api.hardware", hardware_id=self.id, _external=True),
+            },
+        }
+
+
+class Cluster(Hardware):
+    info = Nullable(postgresql.JSONB)
```
jonkeane, line 104: 🎉 I was going to ask if info was arbitrary JSON, but this looks like it is. Awesome! And I see below that it's in the hash so basically we can bump the hash by changing the info, yeah?

conbench/entities/distribution.py
```diff
@@ -17,7 +17,7 @@ class Distribution(Base, EntityMixin):
     case_id = NotNull(s.String(50), s.ForeignKey("case.id", ondelete="CASCADE"))
     context_id = NotNull(s.String(50), s.ForeignKey("context.id", ondelete="CASCADE"))
     commit_id = NotNull(s.String(50), s.ForeignKey("commit.id", ondelete="CASCADE"))
-    machine_hash = NotNull(s.String(250))
+    hardware_hash = NotNull("machine_hash", s.String(250))
```
jonkeane, line 21: I'm a bit surprised by `"machine_hash"` here. Should this be `"hardware_hash"`? Or are we keeping the `machine_hash` name around for some of these bits? I'm sure there's a good reason — I just can't figure it out so far

conbench/entities/hardware.py
```diff
@@ -0,0 +1,182 @@
...
+
+    __mapper_args__ = {"polymorphic_identity": "cluster"}
+
+    def generate_hash(self):
+        return (
+            self.name
+            + "-"
+            + hashlib.md5(
+                json.dumps(self.info, sort_keys=True).encode("utf-8")
+            ).hexdigest()
```
jonkeane, line 124: This might be YANGNI, but would it be possible to have two info json columns? One that is included in the hash and one that is not?

This way for informational not-distribution disturbing info, we could add it to the non-hash-altering info store (and for those, we shouldn't need to run a migration either if we only care about that info going forward...)

jonkeane (commented):
This is looking fantastic, thanks for all this work.

A few questions / comments for you

conbench/entities/hardware.py
ElenaHenderson: Yep, `Info` field will store arbitrary JSON since definition of a cluster can vary a lot.
`Hash` is generated based on this JSON and any changed in this JSON (except order of keys) will cause a new cluster to be created and thus new hash.

conbench/entities/distribution.py
ElenaHenderson: This is a SQLAlchemy way of allowing me to rename all fields in code (`machine_hash` to `hardware_hash`, `machine` to `hardware`) without actually renaming this field in the actual database.

Renaming machine table and all its foreign keys, indexes, constraints at the db level will need to be handled in a separate migration and PR. I am still deciding if I should do this though since this will be giant. It is doable though.

conbench/entities/hardware.py
ElenaHenderson: Yep. On it.

conbench/entities/distribution.py
jonkeane: Aaah I see! Yeah I agree on pushing that migration out if we can. Maybe a comment here explaining that would be helpful to remind us that's what's going on?

ElenaHenderson: Added comments in 3 places where `machine_hash` comes up.

conbench/entities/hardware.py
ElenaHenderson: Done.

conbench/entities/hardware.py
jonkeane: You're the best!

conbench/entities/hardware.py
```diff
@@ -0,0 +1,203 @@
...
+            "links": {
+                "list": f.url_for("api.hardwares", _external=True),
+                "self": f.url_for("api.hardware", hardware_id=self.id, _external=True),
+            },
+        }
+
+
+class Cluster(Hardware):
+    info = Nullable(postgresql.JSONB)
+    distribution_info = Nullable(postgresql.JSONB)
```
jonkeane: This might be a bit bike shedding (so tell me to go away if it is!), but I want to get these names right since they'll be with us / require a migration to change otherwise...


What about `info` (for the hash one) and `optional_info`? We avoid `distribution` which might be confusion with things like linux distributions. And it makes it a bit clearer that the optional one is a little more ephemeral / won't disrupt history?

ElenaHenderson: Done.

README.md
```diff
@@ -590,6 +590,36 @@ Benchmark result:
 }
 ```
 
+### Example simple benchmarks executed on machine cluster instead of one machine
+
+If your benchmark is executed on a machine cluster instead of one machine, you
+can capture cluster info in the following manner.
+Note that a benchmark will have a continuous history on a specific cluster as long as cluster's `name` and `info` do not change.
```
jonkeane, line 597:
```suggestion
Note that a benchmark will have a continuous history on a specific cluster as long as cluster's `name` and `info` do not change.
There is also an `optional_info` field for information that should not impact the cluster's hash (and thus disrupt the distribution history), but should still be recorded. 
```

jonkeane (approved):
Looks good to me, thank you!

PR #308, opened 2022-03-02
Send floats as floats in JSON responses

### LINKED ISSUE #187: Send numerics as numerics in JSON responses?
While working on the 5.0.0 -> 6.0.0 report I noticed that the json responses used strings of numbers (often with rounding). Is it possible to have a parameter or always send numbers with higher precision + as numerics?

ElenaHenderson:
Note that I tested that my change did not break  Conbench UI and graphs.


dianaclarke (approved):
LGTM 😸

PR #309, opened 2022-03-02
Store and display benchmark errors

### LINKED ISSUE #176: How to handle error reporting payloads?
If I'm reading [the flow for r benchmarks correctly](https://github.com/ursacomputing/benchmarks/blob/c7ba6389aaf7f5d740f33a4b3c0263fe42730dc5/benchmarks/_benchmark.py#L182-L203), if there is an error of some sort, we call `_handle_error()` which helpfully creates a JSON with the stdout/stderror, etc.  It looks like this doesn't get recorded with conbench (I don't see a call to [`record()`](https://github.com/ursacomputing/benchmarks/blob/c7ba6389aaf7f5d740f33a4b3c0263fe42730dc5/benchmarks/_benchmark.py#L86-L109), but I might be missing something there. Would it be possible to send those to conbench? I know that that might also require some UI work to be able to display them.

Additionally, if there is a standard (I could grab an example of what's currently produced), I could extend {arrowbench} to produce error JSONs of this shape in cases where we catch an error and provide a more helpful message. I don't think we would want to totally remove the `_handle_error()` for {arrowbench} benchmarks, but in cases where {arrowbench} knows that something is wrong, we could create the error payload in {arrowbench} and have that uploaded to conbench.

ElenaHenderson:
This PR :
- implements storing benchmark errors as JSON values
- implements displaying benchmark errors in UI
- adds JSON `summary.error` column to store benchmark errors
- adds Boolean `run.has_errors` column to mark a run with benchmarks with errors so we can: 
     - display green checkmark or red exclamation mark next to each run on runs UI page
     - avoid making an expensive SQL join to summary table to see if a run has any benchmarks with errors
- adds `populate_local_conbench.py` script so it is easier to test conbench UI locally
- updates `/compare`, `/benchmarks` endpoints to include errors in JSON responses
- updates `/runs` endpoints to include `has_errors` value in  JSON responses
- updates conbench runner to capture and publish benchmark error with a stack trace by default. Note that benchmark error will not be captured if benchmark overrides `benchmark()` or `record()` methods without considering errors.

**UI Changes**
- Runs:
Screen Shot 2022-03-03 at 3 41 26 PM

- Run with a benchmark that has error. Error icon is clickable and goes to Benchmark page:
Screen Shot 2022-03-03 at 3 43 18 PM

- Compare runs. Error icon is clickable and goes to Benchmark page:
Screen Shot 2022-03-03 at 3 44 31 PM

- Benchmark with error:
Screen Shot 2022-03-03 at 3 45 13 PM


conbench/api/_examples.py
```diff
@@ -23,20 +23,29 @@ def _api_user_entity(user):
...
 
 def _api_benchmark_entity(
-    summary_id,
-    case_id,
-    info_id,
-    context_id,
-    batch_id,
-    run_id,
-    name,
+    summary_id, case_id, info_id, context_id, batch_id, run_id, name, error=None
```
jonkeane, line 26: I believe it that this is a `black` style thing but whoa I would never have guessed this is what would happen by adding `, error=None`

conbench/entities/_comparator.py
```diff
@@ -109,13 +111,18 @@ def change(self):
         if self.baseline is None or self.contender is None:
             return 0.0
 
+        if self.baseline.error or self.contender.error:
+            return 0.0
```
jonkeane, line 115: I don't think we should change these now, but in principle these should probably be something like a missing value since they aren't really defined. But we can deal with that if it's an issue later.

conbench/entities/summary.py
```diff
@@ -204,18 +224,19 @@ def _dump(self, summary):
...
-                "q3": float(summary.q3),
-                "iqr": float(summary.iqr),
+                "min": to_float(summary.min),
+                "max": to_float(summary.max),
+                "mean": to_float(summary.mean),
+                "median": to_float(summary.median),
+                "stdev": to_float(summary.stdev),
+                "q1": to_float(summary.q1),
+                "q3": to_float(summary.q3),
+                "iqr": to_float(summary.iqr),
```
jonkeane, line 234: Oh, actually I think this will do what I was talking about above about getting `NA`s out. Cool!

jonkeane (approved):
This looks fantastic. Most of my (few) comments are really questions for my own edification. Thanks for this!

conbench/api/_examples.py
ElenaHenderson: Yep, it is a `black` thing.

conbench/entities/_comparator.py
ElenaHenderson: Yep, I did not want to mess with this code too much but I had to do this because something was breaking (I forgot what it was)

PR #316, opened 2022-03-29
Better formatting for traceback in runs that error

### LINKED ISSUE #312: Better formatting for traceback in runs that error
Maybe the newlines are being stripped? Or it just needs to be blockquote formatted or something? 

e.g. https://conbench.ursa.dev/benchmarks/eba01676640441f58a374feacb26b9d6/

image

Christian8491:
Closes #312 

Before this PR an error looks (which don't recognize end of lines `'\n'`)
conbench_error_before

After this PR
conbench_error_after

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -33,6 +33,7 @@ jobs:
       - name: Migrations
         run: |
           alembic upgrade head
+        continue-on-error: true
         env:
           DB_USERNAME: "postgres"
           DB_HOST: "localhost"
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -33,7 +33,6 @@ jobs:
       - name: Migrations
         run: |
           alembic upgrade head
-        continue-on-error: true
         env:
           DB_USERNAME: "postgres"
           DB_HOST: "localhost"
--- a/conbench/templates/benchmark-entity.html
+++ b/conbench/templates/benchmark-entity.html
@@ -65,7 +65,8 @@
             {% for k,v in benchmark.error.items() %}
               <li class="list-group-item" style="overflow-y: auto;">
                 <b>{{ k }}</b>
-                <div align="right" style="display:inline-block; float: right;">{{ v }}</div>
+                <br></br>
+                <div align="left" style="display:inline-block; white-space: pre-line; float: center;">{{ v }}</div>
               </li>
             {% endfor %}
           {% else %}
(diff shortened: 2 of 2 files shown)


conbench/templates/benchmark-entity.html
```diff
@@ -65,7 +65,8 @@
             {% for k,v in benchmark.error.items() %}
               <li class="list-group-item" style="overflow-y: auto;">
                 <b>{{ k }}</b>
-                <div align="right" style="display:inline-block; float: right;">{{ v }}</div>
+                <br></br>
+                <div align="left" style="display:inline-block; white-space: pre-line; float: center;">{{ v }}</div>
```
jonkeane:
```suggestion
                <div align="left" style="display:inline-block; white-space: pre; float: center;">{{ v }}</div>
```

I believe that this will also preserve the tabs/spaces in the strings as well. Could you try it out and see if that's true?

Christian8491: You're right. See the image 
conbench_error_aft2

jonkeane: Awesome, thanks!

conbench/templates/benchmark-entity.html
```diff
@@ -65,7 +65,8 @@
             {% for k,v in benchmark.error.items() %}
               <li class="list-group-item" style="overflow-y: auto;">
                 <b>{{ k }}</b>
-                <div align="right" style="display:inline-block; float: right;">{{ v }}</div>
+                <br></br>
```
jonkeane:
```suggestion
                 <br />
```

One last, minor suggestion

jonkeane (approved):
Looks good, one minor thing and I'll merge when CI is green

Christian8491:
As additional, passing just the `e` (exception as str) we can have something this:
conbench_error_short

jonkeane:
The CI failure looks like it's related to https://github.com/maxcountryman/flask-login/issues/584 (flask-login is not compatible with the new werkzeug).

jonkeane:
For now, could we add `continue-on-error: true` to the migration step so that the other tests run + I'll make an issue to remove that when flask-login is released?

We need to add it: https://github.com/conbench/conbench/blob/3d38c7132f7fb6fd5e06b72a7ea344b30e4e029e/.github/workflows/actions.yml#L33-L35

jonkeane:
I've made https://github.com/conbench/conbench/issues/317 — I'll watch the CI

jonkeane:
Welp, I was hopeful the tests would run without hitting this, but they have not! Maybe instead, we should either install flask-login from git or pin werkzeug to the older version that doesn't have the breaking change?

.github/workflows/actions.yml
```diff
@@ -33,6 +33,7 @@ jobs:
       - name: Migrations
         run: |
           alembic upgrade head
+        continue-on-error: true
```
jonkeane:
```suggestion
```

jonkeane:
I'm going to merge this, and we can deal with the test failures + dependencies separately

Issue #299, opened 2021-12-07
Add machines page to UI so users can view all machines and their specs in one place


ElenaHenderson:

jonkeane:
Questions from @Christian8491 :

> Where can I find all the machines we are using?

The machines are all in the hardware table (the hardware entity is defined in https://github.com/conbench/conbench/blob/main/conbench/entities/hardware.py). Likely what you'll want to do is create a hardware page in https://github.com/conbench/conbench/tree/main/conbench/app that displays the info in that table

> The information we want to capture is the same that "hardware" label holds ?

This ticket is mostly for displaying the info that we already capture. We can add more info to hardware in our runners if we want, but let's display what we have now, and then we can make sure it is extensible + will display additional extra info when we get there.

Christian8491:
> The machines are all in the hardware table (the hardware entity is defined in https://github.com/conbench/conbench/blob/main/conbench/entities/hardware.py). Likely what you'll want to do is create a hardware page in https://github.com/conbench/conbench/tree/main/conbench/app that displays the info in that table

Sorry but what hardware table are you talking about ? What I can see from https://github.com/conbench/conbench/blob/main/conbench/entities/hardware.py#L37-L52 is only the specs from a Machine.  How a user can see all the machines ? (if it is what this PR is supposed to).

jonkeane:
Hardware is the parent class. There are two kinds: machines (for single machine infrastructure) and clusters (for multi machine infrastructure).

So for this ticket we will want one page that lists all the available hardware (both machines and clusters) in a table. And then also a page that shows all the detailed information about the specific piece of hardware. This is similar to how we have a runs page that lists all runs and then you click through and see the benchmarks that ran as that particular rub.

PR #321, opened 2022-03-30
Add machines page to UI so users can view all machines and their specs in one place

Christian8491:
PR related to https://github.com/conbench/conbench/issues/299


conbench/app/cluster.md
```diff
@@ -0,0 +1,9 @@
+# Cluster
```
dianaclarke: I could be wrong, but I think the mission is to display all the Conbench machines dynamically in the UI, rather than hardcoded in markdown.

I've merged a starting point here: https://github.com/conbench/conbench/pull/322/files

Let me know if you have any questions.

Cheers!

jonkeane: Exactly! Thank you Diana

dianaclarke: There's still lots to do. (my machine name is `diana`)

Screen Shot 2022-03-30 at 9 52 24 PM
Screen Shot 2022-03-30 at 9 52 32 PM

Christian8491: So glad to read your thought on this @dianaclarke . Thanks!

Christian8491: A first glance .. 

conbench_hardware_entity_1

jonkeane: That's looking good!

Christian8491:
@jonkeane I will need your help here to review this PR. What else do you have in mind about the scope of this PR? As you have noticed @dianaclarke already did an awesome job. 

This is how it looks now (from the Hardware page)

hardware_info

conbench/templates/hardware-entity.html
```diff
@@ -10,8 +10,113 @@
...
+          <td>gpu_count</td>
+          <td><div class="ellipsis-500">{{ hardware.gpu_count }}</div></td>
+        </tr>
+        {% endif %}
+        {% if hardware.gpu_product_names %}
+        <tr>
+          <td>gpu_product_names</td>
+          <td><div class="ellipsis-500">{{ hardware.gpu_product_names }}</div></td>
+        </tr>
+        {% endif %}
```
jonkeane: Do flask templates offer an option  to do this in a loop so that you don't have to repeat all of this boiler plate html? something like (this won't work exactly, but I'm sure there's a way to do something like it!):

```
  {% for key in hardware%}
  <tr>
    <td>{{ key }}</td>
    <td><div class="ellipsis-500">{{ hardware.key }}</div></td>
  </tr>
  {% endfor %}
```

jonkeane (changes requested):
One suggestion about using a for loop instead of hardcoding.

Also, we should consider what might happen when the hardware is a cluster. The kinds of info is slightly different. See https://github.com/conbench/conbench/blob/62201e4f8985e10b313c3c18faaa5be846e1c3d3/conbench/entities/hardware.py#L103-L134 for the information that clusters have (note that both `info` and `optional_info` are json fields, so you'll want to convert them from json into something that looks ok in the app.

ElenaHenderson:
@Christian8491 Will you please rebase?

conbench/templates/hardware-entity.html
Christian8491:
```
{% for k, v in hardware.items() %}
        <tr>
          <td>{{ k }}</td>
          <td><div class="ellipsis-500">{{ v }}</div></td>
        </tr>
        {% endfor %}
```
Works well! 
Do we want https://github.com/conbench/conbench/blob/main/conbench/entities/hardware.py#L96 (`links`  info) be part of the details showed ?

Christian8491:
> Also, we should consider what might happen when the hardware is a cluster. The kinds of info is slightly different.

Do you recommend me to do some runs in a cluster?

conbench/templates/hardware-entity.html
jonkeane: No, we should exclude the links

jonkeane:
You shouldn't need to do actually runs with a cluster to be able to test out how you would handle the `info` and `optional_info` json blobs.

There's some example data in https://github.com/conbench/conbench/blob/1d63b2bf244a7d9639c38d2a10ee2e1f6ed6ec4f/conbench/api/_examples.py#L382-L391 which we should expand a little bit with a few more made up values in the `info` json (and also add an `optional_info` json bit). That should be enough (plus populating your local conbench with those examples https://github.com/conbench/conbench#to-populate-local-conbench-with-sample-runs-and-benchmarks) to be able to get that working.

ElenaHenderson (changes requested):
May I please see how UI page looks for clusters?
Also, the test is failing.

Christian8491:
using a cluster (after populating local conbench) 

cluster_hardware

I will review about the test failures.

ElenaHenderson (approved):
Thank you!

PR #323, opened 2022-03-31
Move data and times into summary table

ElenaHenderson:
`summary_id` indexes for `data` and `time` tables got too large (~2GB) so they can no longer be used by SQL statements below on RDS instance with 4GB of RAM. These SQL statements are taking > 3 minutes to complete and cause conbench HTTP requests to timeout.

```sql
SELECT summary.id AS summary_id, summary.case_id AS summary_case_id, 
summary.info_id AS summary_info_id, summary.context_id AS summary_context_id, 
summary.run_id AS summary_run_id, summary.unit AS summary_unit, 
summary.time_unit AS summary_time_unit, summary.batch_id AS summary_batch_id, 
summary.timestamp AS summary_timestamp, summary.iterations AS summary_iterations, 
summary.min AS summary_min, summary.max AS summary_max, 
summary.mean AS summary_mean, summary.median AS summary_median, 
summary.stdev AS summary_stdev, 
summary.q1 AS summary_q1, summary.q3 AS summary_q3, summary.iqr AS summary_iqr, 
summary.error AS summary_error, case_1.id AS case_1_id, 
case_1.name AS case_1_name, case_1.tags AS case_1_tags, info_1.id AS info_1_id, info_1.tags AS info_1_tags, 
context_1.id AS context_1_id, context_1.tags AS context_1_tags, data_1.id AS data_1_id, 
data_1.summary_id AS data_1_summary_id, data_1.iteration AS data_1_iteration, 
data_1.result AS data_1_result, time_1.id AS time_1_id, time_1.summary_id AS time_1_summary_id, time_1.iteration AS time_1_iteration, time_1.result AS time_1_result 
FROM summary LEFT OUTER JOIN "case" AS case_1 ON case_1.id = summary.case_id 
LEFT OUTER JOIN info AS info_1 ON info_1.id = summary.info_id 
LEFT OUTER JOIN context AS context_1 ON context_1.id = summary.context_id 
LEFT OUTER JOIN data AS data_1 ON summary.id = data_1.summary_id 
LEFT OUTER JOIN time AS time_1 ON summary.id = time_1.summary_id 
WHERE summary.run_id = '8bbe71ee7abf4c37b94de660d28254bf'
```
Here is the execution plan for the above SQL statement that shows `summary_id` indexes for `data` and `time` tables are no longer used:

image

This PR moves `data` and `time` tables data into `summary` table to avoid the need to join `summary` table with these tables and thus avoid performance issues caused by slow SQL statements above. 

@jonkeane I will drop `data` and `time` tables and rename `summary` table to `benchmark_result` in the next two PRs.


conbench/tests/app/test_users.py
```diff
@@ -142,6 +143,7 @@ def test_user_create_post_unauthenticated(self, client):
 
     def test_user_create_post_no_csrf_token(self, client):
         self.authenticate(client)
-        response = client.post("/users/create/", data={})
+        data = {"name": "New Name", "email": "test@test.com"}
```
ElenaHenderson, line 146: I had to fix this because some kind of package upgrade causes this too fail with 400 now when data = `{}`

dianaclarke:
This is why I ❤️ you!

(I've had too much 🍻 to actually review, but I 💯 trust you)

Thanks!

migrations/versions/88040ddc0b08_move_data_and_times.py
```diff
@@ -0,0 +1,52 @@
...
+# revision identifiers, used by Alembic.
+revision = "88040ddc0b08"
+down_revision = "e06cede49f65"
+branch_labels = None
+depends_on = None
+
+
+def set_data_and_times():
+    connection = op.get_bind()
+    # meta = sa.MetaData()
```
dianaclarke: remove these comments?

ElenaHenderson:
@dianaclarke Thank you! Glad you are having good time. 
I tested the migration against db's copy (completes in about 40 mins) but I will be retesting it one more time tomorrow morning before I do anything.

I will look why tests are failing tomorrow morning too. They are not failing on Buildkite or when I run them locally.

migrations/versions/88040ddc0b08_move_data_and_times.py
ElenaHenderson: done

conbench/entities/summary.py
```diff
@@ -19,12 +17,12 @@
 from ..entities.case import Case
 from ..entities.commit import Commit, get_github_commit, repository_to_url
 from ..entities.context import Context
-from ..entities.data import Data
+from ..entities.data import Data  # noqa: F401
```
jonkeane, line 20: TIL about F401 and being able to disable _specific_ qas

conbench/tests/api/_expected_docs.py
```diff
@@ -128,38 +130,38 @@
...
+                                    0.099094,
+                                    0.037129,
+                                    0.036381,
+                                    0.148896,
+                                    0.008104,
+                                    0.005496,
+                                    0.009871,
+                                    0.006008,
+                                    0.007978,
+                                    0.004733,
```
jonkeane, line 164: 🙏 for all these strings to floats

conbench/tests/app/test_users.py
jonkeane: Thanks for these!

migrations/versions/88040ddc0b08_move_data_and_times.py
```diff
@@ -0,0 +1,50 @@
...
+depends_on = None
+
+
+def set_data_and_times():
+    connection = op.get_bind()
+    connection.execute(
+        text(
+            """UPDATE summary
+           SET data = ARRAY(SELECT result FROM data WHERE summary_id = summary.id ORDER BY iteration),
+           times = ARRAY(SELECT result FROM "time" WHERE summary_id = summary.id ORDER BY iteration);
```
jonkeane, line 26: I don't think we need to change this, because we aren't using this data in this way: but technically what we would want is to construct the array based on the actual value of iteration so if one was missing (e.g. we have times for iterations 1, 2, 4, 5 but 3 has no time we would want `[x, x, NULL, x, x]`.

But again, let's leave this be so that it's simpler and if we get to the point where some iterations fail, we can deal with that later. And as far as I know we don't have any examples of that at all now (and even if we did we don't look at those in that way regardless)

jonkeane (approved):
This looks great! A few comments, but nothing that needs to change IMO

conbench/entities/summary.py
ElenaHenderson: These lines will go away in the next PR when I drop these tables. I need them here because of tests and how drop_all() works.

ElenaHenderson: These lines will go away in the next PR when I drop these tables. I need them here because of tests and how drop_all() works.

migrations/versions/88040ddc0b08_move_data_and_times.py
ElenaHenderson: I will look into this before merging PR.

jonkeane: I just ran a check to see if we have any occurences of this at all, and we currently do not. Both of the queries below returned no rows:

```
select n, max_iter, summary_id from (select count(*) n, max(iteration) max_iter, summary_id 
from public.time
group by summary_id) as tbl
where n != max_iter;

select n, max_iter, summary_id from (select count(*) n, max(iteration) max_iter, summary_id 
from public.data
group by summary_id) as tbl
where n != max_iter;
```

So I don't think we need to worry about this at all as part of the migration.

ElenaHenderson: I did the same!  and did not find anything.
```sql
select summary_id, count(1), max(iteration)
from data
group by summary_id
having by count(1) != max(iteration)

select summary_id, count(1), max(iteration)
from time
group by summary_id
having by count(1) != max(iteration)
```

PR #325, opened 2022-03-31
Rename summary table to benchmark_result

ElenaHenderson:
This PR 
- renames `summary` table to `benchmark_result` since we are now storing details in this table.
- deletes tests for migrations since we already ran the migrations being tested and updating these migrations to work with table rename will not be productive and will not be solving any issues

Note that 
- Arrow-Benchmarks-CI will be paused during this migration
- I tested this migration against conbench db copy and it completes within few seconds.


conbench/tests/migrations/test_1fed559406c5_recalculate_distribution3.py
```diff
@@ -1,245 +0,0 @@
-import decimal
```
jonkeane, line 1: Mostly out of curiosity: this clean up of these tests/migrations/... isn't related to the rename, just cleaning up stuff that we should have cleaned up before, yeah?

jonkeane (approved):
This looks great, thanks! One question (for my own curiosity)

dianaclarke (approved):
LGTM - much better name!

conbench/tests/migrations/test_1fed559406c5_recalculate_distribution3.py
ElenaHenderson: We should have removed these tests once we were done with migrations.

PR #329, opened 2022-04-13
Baseline run should be last run for parent commit with at least one benchmark and context pair matching contender's

ElenaHenderson:
This PR changes logic used for finding a baseline run for a contender run. 
Baseline run is now set to the last run for a parent commit with at least one benchmark/context pair matching contender's instead of all benchmark and context pairs matching contender's.


jonkeane (approved):
Looks good, thank you!

PR #331, opened 2022-04-14
api/compare/... endpoints > Send baseline_z_score, contender-z_score and threshold_z as floats

ElenaHenderson:


conbench/entities/_comparator.py
```diff
@@ -187,8 +188,8 @@ def formatted(self):
             "regression": self.regression,
             "improvement": self.improvement,
             "threshold_z": fmt(self.threshold_z),
```
jonkeane: Should we also send the threshold_z as a float too?

jonkeane (approved):
This looks fantastic, just one possible addition. Thanks!

conbench/entities/_comparator.py
ElenaHenderson: Fixed. I checked that this change will not affect UI.

PR #332, opened 2022-04-14
Format z_score to 3 decimal places in UI

ElenaHenderson:


jonkeane (approved):
This looks great, thanks for this! I've opened #333 to discuss possible doing more like this for things like `display_mean` (we might not be able to, but we can figure that out there when we decide to do it — I don't want to clutter this PR)

PR #339, opened 2022-05-05
Add search arg as a query string to routes with search in UI and update urls when search value is updated

### LINKED ISSUE #311: Let users focus on specific language
For example https://conbench.ursa.dev/compare/runs/9292568e6e4b4c1882d0564bf8ebc3e7...9aef9573adc0475d88c70056788a4875/ is only relevant to the JS tests since there are no changes in the other directories. It would be great to easily filter to JS only.

ElenaHenderson:
This PR 
- enables `search` query string parameter for all UI pages with `Search` box
- makes urls for UI pages with `Search` box to be updated when a user enters values into Search checkbox
- updates `conbench/tests/populate_local_conbench.py` so local conbench has benchmarks for various languages

This is how query string looks like in page URL:
Screen Shot 2022-05-06 at 1 48 02 PM

### DIFF
--- a/conbench/templates/index.html
+++ b/conbench/templates/index.html
@@ -99,7 +99,7 @@
 {{super()}}
 <script type="text/javascript">
   var table = $('#runs').dataTable( {
-    "responsive": true,
+      "responsive": true,
       "order": [[0, 'desc']],
       "autoWidth": false,
       "columnDefs": [
--- a/conbench/templates/index.html
+++ b/conbench/templates/index.html
@@ -99,7 +99,6 @@
 {{super()}}
 <script type="text/javascript">
   var table = $('#runs').dataTable( {
-    "search": {"return": true},
     "responsive": true,
       "order": [[0, 'desc']],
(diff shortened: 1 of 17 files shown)


conbench/templates/index.html
```diff
@@ -99,7 +99,7 @@ <h3>Language-independent Continuous Benchmarking (CB) Framework</h3>
 {{super()}}
 <script type="text/javascript">
   var table = $('#runs').dataTable( {
-      "responsive": true,
+    "responsive": true,
```
jonkeane: This looks like the list below is misaligned, yeah?

jonkeane (approved):
This looks great, one very minor (style) comment

dianaclarke (approved):
Very cool!!

Love this too:  `python -m conbench.tests.populate_local_conbench`

conbench/templates/index.html
ElenaHenderson: Thank you! Fixed.

PR #340, opened 2022-05-09
Update README To populate local conbench with sample runs and benchmarks section

ElenaHenderson:
This PR updates REDAME so it is clear that conbench app needs to be running when running `conbench.tests.populate_local_conbench`


boshek (approved):
LGTM. And it worked too :)

PR #341, opened 2022-05-13
Compare view to see what has changed

### LINKED ISSUE #315: A compare view to see what has changed?
Our benchmark comparison view is fantastic when the info, context, and hardware align. But sometimes those will shift due to external changes (changes to the software under benchmark, changes to the running environment).

It would be great if we could have a view that allowed us to see that info, context, and  hardware for two benchmarks (and runs, maybe?) side by side so that it would be easy to see what changed (and thus, what disrupted the history one might be expecting).

ElenaHenderson:
This PR 
- updates `/compare/benchmarks` UI page to show differences between 2 benchmarks
- updates `conbench/tests/populate_local_conbench.py` so local conbench has benchmarks with various info, context, tags and hardware fields that make it possible to test above UI changes locally.

This is how `/compare/benchmarks` UI page looks like when info, context, tags and hardware fields do not match between 2 benchmarks. Please let me know if you want me to adjust colors or something else.

Screen Shot 2022-05-12 at 7 45 47 PM
Screen Shot 2022-05-12 at 7 45 54 PM
Screen Shot 2022-05-12 at 7 46 04 PM

### DIFF
--- a/conbench/templates/compare-entity.html
+++ b/conbench/templates/compare-entity.html
@@ -163,16 +163,16 @@
             </li>
           {% endfor %}
           <li class="list-group-item active">Tags</li>
-          {% for tag in tags_fields %}
-            {% if tag != "id" and (tag not in baseline.tags or tag not in contender.tags or baseline.tags[tag] != contender.tags[tag]) %}
+          {% for field in tags_fields %}
+            {% if field != "id" and (field not in baseline.tags or field not in contender.tags or baseline.tags[field] != contender.tags[field]) %}
               <li class="list-group-item list-group-item-danger" style="overflow-y: auto;">
             {% else %}
                <li class="list-group-item" style="overflow-y: auto;">
             {% endif %}
-              {% if tag in baseline.tags %}
-                <b>{{ tag }}</b>
-                {% if baseline.tags[tag] is not none %}
-                  <div align="right" style="display:inline-block; float: right;">{{ baseline.tags[tag] }}</div>
+              {% if field in baseline.tags %}
+                <b>{{ field }}</b>
+                {% if baseline.tags[field] is not none %}
+                  <div align="right" style="display:inline-block; float: right;">{{ baseline.tags[field] }}</div>
                 {% endif %}
               {% else %}
                 <b>*</b>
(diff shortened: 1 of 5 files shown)


conbench/templates/compare-entity.html
```diff
@@ -163,34 +163,72 @@
...
+            {% else %}
+               <li class="list-group-item" style="overflow-y: auto;">
+            {% endif %}
+              {% if field in baseline.tags %}
+                <b>{{ field }}</b>
+                {% if baseline.tags[field] is not none %}
+                  <div align="right" style="display:inline-block; float: right;">{{ baseline.tags[field] }}</div>
+                {% endif %}
+              {% else %}
+                <div class="invisible">*</div>
```
ElenaHenderson: Using `<div class="invisible">*</div>` because I could not figure out how to make an empty cell to have the same height as non-empty cell.

I tried this but it messed up JSON fields:
```
<li class="list-group-item" style="height: 40px;overflow-y: auto;">
``` 
Let me know if you have better ideas.

conbench/templates/compare-entity.html
```diff
@@ -163,34 +163,72 @@
...
-          {% for k,v in baseline_run.hardware.items() %}
-            <li class="list-group-item" style="overflow-y: auto;">
-              <b>{{ k }}</b>
-              <div align="right" style="display:inline-block; float: right;">{{ v }}</div>
+          {% for field in hardware_fields %}
+            {% if field != "id" and (field not in baseline_run.hardware or field not in contender_run.hardware or baseline_run.hardware[field] != contender_run.hardware[field]) %}
+              <li class="list-group-item list-group-item-danger" style="overflow-y: auto;">
+            {% else %}
+               <li class="list-group-item" style="overflow-y: auto;">
+            {% endif %}
```
jonkeane: I see this logic in a few places, any way that we could abstract it? It's not a huge deal if not, but it's finicky enough that it would be nice not needing to maintain it in multiple places.

It also might be nice to comment explaining what it's doing too.

jonkeane (changes requested):
Thanks for this! One small(ish?) change

conbench/templates/compare-entity.html
ElenaHenderson: Will look into this!

ElenaHenderson: Done. 
- Used jinja macros to reduce repetitive code.
- Added comments to macros.

conbench/templates/compare-entity.html
```diff
@@ -1,5 +1,40 @@
 {% extends "app.html" %}
 
+{# This macro displays Tags, Hardware, Context and Info fields in comparative view so
+matching fields between baseline and contender are aligned and unmatching fields and/or values are highligted in red #}
```
jonkeane:
```suggestion
matching fields between baseline and contender are aligned and mismatched fields and/or values are highlighted in red #}
```

Very very minor language tweak

jonkeane (approved):
This is awesome, thank you!

PR #342, opened 2022-05-14
Allow to compare runs in UI

ElenaHenderson:
This PR adds ability to select baseline run to compare in UI on `compare/runs` page:
Screen Shot 2022-05-16 at 5 43 53 PM

Screen Shot 2022-05-16 at 5 43 24 PM


jonkeane:
This looks great to me. Is it possible to put this up on a staging server to poke at it? (I can also pull locally of course, but if it's easy to deploy there that would be fantastic

jonkeane (approved):
This is fantastic! thank you

PR #347, opened 2022-06-03
Changed Reason column on home page to Name

### LINKED ISSUE #336: Add nightly reason (in the list view) when the run comes from a "nightly"
Currently there are only two options captured `commit` and `pull request` for the Reason. Note that the reason for individual-page shows   `commit: <commit hash> nightly`, so we can extract nightly tag from here.

austin3dickey:
Fixes #336.

This PR changes the "Reason" column on the `conbench` home page to "Name," displaying the full run name, so users can tell at a glance why the run was triggered.

### DIFF
--- a/conbench/app/index.py
+++ b/conbench/app/index.py
@@ -9,20 +9,12 @@ from ..config import Config
 
 class Index(AppEndpoint, RunMixin):
     def page(self, runs):
-        reasons = {r["display_name"] for r in runs if r["display_name"]}
-        commits = {r["commit"]["url"] for r in runs if r["commit"]["url"]}
-        authors = {
-            r["commit"]["author_name"] for r in runs if r["commit"]["author_name"]
-        }
         return self.render_template(
             "index.html",
             application=Config.APPLICATION_NAME,
             title="Home",
             version=__version__,
             runs=runs,
-            has_reasons=len(reasons) > 0,
-            has_authors=len(authors) > 0,
-            has_commits=len(commits) > 0,
             search_value=f.request.args.get("search"),
         )
 
--- a/conbench/tests/populate_local_conbench.py
+++ b/conbench/tests/populate_local_conbench.py
@@ -14,8 +14,13 @@ def generate_benchmarks_data(
     benchmark_language,
     timestamp,
     hardware_type,
+    is_nightly,
     mean=16.670462,
 ):
+    run_name = f"commit: {commit}"
+    if is_nightly:
+        run_name += " nightly"
+
     data = {
         "batch_id": uuid.uuid4().hex,
         "context": {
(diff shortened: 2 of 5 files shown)


conbench/tests/populate_local_conbench.py
```diff
@@ -14,8 +14,13 @@ def generate_benchmarks_data(
     benchmark_language,
     timestamp,
     hardware_type,
+    is_nightly,
```
austin3dickey, line 17: I couldn't find an appropriate place to add a unit test for this behavior, because it looks like we don't unit test how the HTML displays. So I modified this script and checked that the behavior works as expected locally.

ElenaHenderson (commented):
Posted comment into review

ElenaHenderson (commented):
We should not have logic that parses run names in conbench and figures out a reason.

Run names can be anything. The only requirement we have for a run name is to start with "commit: " if its benchmark results should be included in history.

We should

- add run.reason column
- get CI to send run reason in addition to run name
- update Conbench runs pages to display both run name and reason
- update CI that runs benchmarks to send reason (nightly, commit, etc) in addition to run name

Or we can just replace run reason with run name column on runs page.

@jonkeane @austin3dickey Let me know what you think.

austin3dickey:
Agreed; as I was digging into this issue I thought the `reason` attribute could use better treatment than parsing a string. I'm willing to add special treatment of `reason` to this repo and the CI.

jonkeane:
Agreed with @ElenaHenderson and @austin3dickey that having a separate reason attribute more widely. We also might (re)consider the decisions that the contents of the runname be how we decide if something should be included in history or not. It's worked well enough for us, but we might want to separate that out and make it (more) explicit.

One thing we could do in the short term (i.e. repurpose this PR to do this) would be to expose the `run_name` column in the webapp for now, which includes that information and then follow up with the more complete + nicer solution of separating the reason logic from the naming logic.

conbench/app/index.py
```diff
@@ -9,20 +9,12 @@
...
-        }
         return self.render_template(
             "index.html",
             application=Config.APPLICATION_NAME,
             title="Home",
             version=__version__,
             runs=runs,
-            has_reasons=len(reasons) > 0,
-            has_authors=len(authors) > 0,
-            has_commits=len(commits) > 0,
```
austin3dickey, line 25: These are unused.

austin3dickey:
@ElenaHenderson @jonkeane I repurposed this PR to switch the Reason column to displaying the full run name. Here's what it looks like now. I believe this still fixes #336, and I'll move on to handling first-class `reason` and `include_in_history` attributes later.

image

ElenaHenderson:
> @ElenaHenderson @jonkeane I repurposed this PR to switch the Reason column to displaying the full run name. Here's what it looks like now. I believe this still fixes #336, and I'll move on to handling first-class `reason` and `include_in_history` attributes later.
> 
@austin3dickey Sounds great! Thank you! Will you please create 2 separate issues for first-class `reason` and `include_in_history` attributes?

ElenaHenderson (approved):
Looks great! Thank you!

jonkeane (approved):
Looks good! Thanks!

austin3dickey:
done, added #349 and #350 !

PR #351, opened 2022-06-08
Added reason as first-class requestable attribute of Run

### LINKED ISSUE #349: Add a first-class `reason` attribute to runs
It would be useful for each benchmark run to have a `reason` string attribute that stores a short reason why the run was kicked off (e.g. automated commit, someone asked a bot nicely on a PR, or a manual run). This would be optionally submitted by the user, along with run name and run ID. This attribute could be displayed in a column on the home page.

_Originally posted by @ElenaHenderson in https://github.com/conbench/conbench/issues/347#issuecomment-1147983741_

austin3dickey:
Fixes #349.

This PR adds a `reason` attribute to the Run entity. When posting benchmarks, users may now add `"run_reason": "..."` to the body, which takes the place of parsing the run name to try to figure out why a run was kicked off. This also exposes the new Reason on the app home page as a new column.

### DIFF
--- a/conbench/entities/run.py
+++ b/conbench/entities/run.py
@@ -12,6 +12,7 @@ class Run(Base, EntityMixin):
     __tablename__ = "run"
     id = NotNull(s.String(50), primary_key=True)
     name = Nullable(s.String(250))
+    reason = Nullable(s.String(250))
     timestamp = NotNull(s.DateTime(timezone=False), server_default=s.sql.func.now())
     commit_id = NotNull(s.String(50), s.ForeignKey("commit.id"))
     commit = relationship("Commit", lazy="joined")
@@ -85,6 +86,7 @@ class _Serializer(EntitySerializer):
         result = {
             "id": run.id,
             "name": run.name,
+            "reason": run.reason,
             "timestamp": run.timestamp.isoformat(),
             "commit": commit,
             "hardware": hardware,
--- a/conbench/static/app.css
+++ b/conbench/static/app.css
@@ -21,7 +21,6 @@ table.dataTable thead .sorting_desc {
 }
 
 table.dataTable td {
-  padding: 6px !important;
   font-size: 13.4px;
 }
 
--- a/conbench/templates/index.html
+++ b/conbench/templates/index.html
@@ -35,6 +35,7 @@
                     <th scope="col">Commit</th>
                     <th scope="col">Hash</th>
                     <th scope="col">Hardware</th>
+                    <th scope="col">Reason</th>
                     <th scope="col">Name</th>
                 </tr>
             </thead>
@@ -71,6 +72,12 @@
                      <td>{{ run.commit.sha }}</td>
                      <td><div class="ellipsis-175">{{ run.hardware.name }}</div></td>
 
+                     {% if run.reason %}
+                       <td><div class="ellipsis-175">{{ run.reason }}</div></td>
+                     {% else %}
+                       <td></td>
+                     {% endif %}
+
                      {% if run.name %}
                        <td><div class="wordbreak-175">{{ run.name }}</div></td>
                      {% else %}
@@ -101,7 +108,6 @@
   var table = $('#runs').dataTable( {
       "responsive": true,
       "order": [[0, 'desc']],
-      "autoWidth": false,
       "columnDefs": [
           {
               "targets": [4],
@@ -111,16 +117,7 @@
       ],
       "language": {
           "searchPlaceholder": "commit hash works too",
-      },
-      "columns": [
-         { "width": "15%" },
-         { "width": "25%" },
-         { "width": "10%" },
-         { "width": "30%" },
-         { "width": "0%" },
-         { "width": "10%" },
-         { "width": "10%" },
-      ],
+      }
   } );
 
 enable_search_query_string('#runs');
--- a/conbench/tests/api/_expected_docs.py
+++ b/conbench/tests/api/_expected_docs.py
@@ -305,6 +305,63 @@
                 },
                 "description": "OK",
             },
+            "CompareBenchmarkResult": {
+                "content": {
+                    "application/json": {
+                        "example": {
+                            "commits": {
+                                "baseline": {
+                                    "author_avatar": "https://avatars.githubusercontent.com/u/1299904?v=4",
+                                    "author_login": "bkietz",
+                                    "author_name": "Benjamin Kietzman",
+                                    "id": "some-baseline-commit-id",
+                                    "message": "ARROW-11767: [C++] Scalar::Hash may segfault",
+                                    "parent_sha": "6d703c4c7b15be630af48d5e9ef61628751674b2",
+                                    "repository": "https://github.com/apache/arrow",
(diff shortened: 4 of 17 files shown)


conbench/static/app.css
```diff
@@ -21,7 +21,6 @@ table.dataTable thead .sorting_desc {
 }
 
 table.dataTable td {
-  padding: 6px !important;
```
austin3dickey, line 24: With the additional column, depending on screen size, sometimes the `Name` column now gets cut off and wrapped to the "responsive" plus sign icon on the left side of the DataTable. I noticed that the plus sign gets uglily overlapped by the timestamp unless we get rid of this padding.

conbench/templates/index.html
```diff
@@ -111,16 +117,7 @@ <h3>Language-independent Continuous Benchmarking (CB) Framework</h3>
...
-      },
-      "columns": [
-         { "width": "15%" },
-         { "width": "25%" },
-         { "width": "10%" },
-         { "width": "30%" },
-         { "width": "0%" },
-         { "width": "10%" },
-         { "width": "10%" },
-      ],
```
austin3dickey, line 123: This was hard to get "just right" for different screen sizes and I think the autoWidth is fine enough.

conbench/tests/api/_expected_docs.py
```diff
@@ -305,6 +305,63 @@
                 },
                 "description": "OK",
             },
+            "CompareBenchmarkResult": {
```
austin3dickey, line 308: I'm assuming it's okay that some of these are reordered; just an artifact of how swagger works

migrations/versions/bc0ac747cec6_add_run_reason_column.py
```diff
@@ -0,0 +1,32 @@
...
+    op.add_column("run", sa.Column("reason", sa.String(length=250), nullable=True))
+
+    # use some custom logic to backfill the column
+    op.execute("UPDATE run SET reason = 'commit' WHERE name LIKE 'commit%'")
+    op.execute("UPDATE run SET reason = 'pull request' WHERE name LIKE 'pull request%'")
+    op.execute(
+        "UPDATE run SET reason = 'manual commit' WHERE name LIKE 'manual commit%'"
+    )
+    op.execute("UPDATE run SET reason = 'merge commit' WHERE name LIKE '%merge-commit'")
+    op.execute("UPDATE run SET reason = 'nightly' WHERE name LIKE '%nightly'")
```
austin3dickey: Looking for feedback on whether this backfill code is necessary. I think it would be nice

austin3dickey:
Also, I chose not to add `run_reason` to the compare or history apis. Let me know if you disagree

alistaire47:
> Also, I chose not to add `run_reason` to the compare or history apis. Let me know if you disagree

I think @boshek and I might need it in the history one to separate results when generating reports, assuming this will replace the prefixing

boshek:
> > Also, I chose not to add `run_reason` to the compare or history apis. Let me know if you disagree
> 
> I think @boshek and I might need it in the history one to separate results when generating reports, assuming this will replace the prefixing

This is a _real_ nice addition. 

I don't think we _need_ it but at least for history api, I think it would ideal to add it. We can certainly get around it by filtering for nightly elsewhere and then joining but if you are in here mucking around, it may be worth adding right now. Given that compare doesn't even return run_name, it seems safe to skip that one though.

migrations/versions/bc0ac747cec6_add_run_reason_column.py
jonkeane: I think it would be nice too. I'm sure @ElenaHenderson will have opinions on it! But I'm glad you included it

conbench/static/app.css
jonkeane: Yeah, we don't (and shouldn't) do this here, but it would be nice to make the front page slightly more responsive without resulting in a pretty unusable UI (I find the pluses, especially on mobile) not suuppper useable.

conbench/entities/run.py
```diff
@@ -12,6 +12,7 @@ class Run(Base, EntityMixin):
     __tablename__ = "run"
     id = NotNull(s.String(50), primary_key=True)
     name = Nullable(s.String(250))
+    reason = Nullable(s.String(250))
```
jonkeane, line 15: I don't think there is, but is there value in validating the reasons? Ideally there would be a relatively small number of reasons and it isn't a free-form field (that's more or less what name is for...).

Maybe describing that in the docs is good enough?

jonkeane (approved):
Looking good. I like what I see, though @ElenaHenderson definitely should have a look before we merge it

conbench/tests/api/_expected_docs.py
ElenaHenderson: Yep.

migrations/versions/bc0ac747cec6_add_run_reason_column.py
ElenaHenderson: Yep, we should do this.

ElenaHenderson (approved):
Looks great to me! Thank you!

Please note that we should update CI to send run reason in benchmark results before merging and deploying this PR.

If we deploy this PR, backfill reason column, all the new runs will have empty reason and we will need to backfill again.

alistaire47:
> This is a _real_ nice addition.
>
> I don't think we _need_ it but at least for history api, I think it would ideal to add it. We can certainly get around it by filtering for nightly elsewhere and then joining but if you are in here mucking around, it may be worth adding right now. Given that compare doesn't even return run_name, it seems safe to skip that one though.

Actually maybe what we really want is the ability to filter by `run_reason` in the APIs, though that's beyond the scope of this story. And agreed, this PR is 😍

jonkeane:
> what we really want is the ability to filter by run_reason in the APIs

😍 This would be fantastic, would someone mind making an issue here for that?

alistaire47:
> 😍 This would be fantastic, would someone mind making an issue here for that?

Brief, but...sufficient? https://github.com/conbench/conbench/issues/352

conbench/entities/run.py
austin3dickey: good point; I'll flesh out the docs a little more. I don't want to be prescriptive about what possible reasons could be, but I'll include a note like "runs with the same reason will probably be compared with each other so try to keep the cardinality down"

austin3dickey:
Okay, with the advent of #352 I won't add `run_reason` to the History APIs. If we decide it's necessary in that issue we can do it then.

austin3dickey:
When fleshing out the docs I realized I had forgotten to add `--run-reason` to the CLI. Take another look if you want!

migrations/versions/bc0ac747cec6_add_run_reason_column.py
austin3dickey: I decided to wait on the backfill code for now. When we merge and deploy this PR, it will add an empty Reason column. Then we can change all pipelines to start publishing reasons. And finally I'll come back and add a migration that backfills any rows with empty reasons. Thanks for the idea @ElenaHenderson .

PR #357, opened 2022-06-13
Fix get_distribution. Benchmark results with errors should not be used to compute distribution

ElenaHenderson:


boshek (approved):
Thanks! LGTM.

PR #360, opened 2022-07-15
Returning benchmark language

### LINKED ISSUE #359: Return language field from benchmark result
I'd like for the benchmark result endpoint to return a dedicated language field. My thinking is to add:

```python
"language": benchmark_result.context.tags.get("benchmark_language", "unknown"),
```

after this: 

https://github.com/conbench/conbench/blob/269c8fb062f0a0d0d2cf11c3f1290d97f5b6c002/conbench/entities/benchmark_result.py#L186

This works locally but I was wondering if there is anything else I should be doing to add this.

boshek:
Closes #359

It would be helpful to have language accessible in benchmarks at this level. 

I think some eyes would be helpful on the tests I wrote. I'm not certain I approached that correctly.

### DIFF
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -38,6 +38,7 @@ class BenchmarkResult(Base, EntityMixin):
     context = relationship("Context", lazy="joined")
     run = relationship("Run", lazy="select")
     unit = Nullable(s.Text)
+    language = Nullable(s.Text)
     time_unit = Nullable(s.Text)
     batch_id = Nullable(s.Text)
     timestamp = NotNull(s.DateTime(timezone=False))
@@ -184,6 +185,7 @@ class _Serializer(EntitySerializer):
             "run_id": benchmark_result.run_id,
             "batch_id": benchmark_result.batch_id,
             "timestamp": benchmark_result.timestamp.isoformat(),
+            "language": benchmark_result.context.tags.get("benchmark_language", "unknown"),
             "tags": tags,
             "stats": {
                 "data": [to_float(x) for x in benchmark_result.data],
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -185,7 +185,9 @@ class _Serializer(EntitySerializer):
             "run_id": benchmark_result.run_id,
             "batch_id": benchmark_result.batch_id,
             "timestamp": benchmark_result.timestamp.isoformat(),
-            "language": benchmark_result.context.tags.get("benchmark_language", "unknown"),
+            "language": benchmark_result.context.tags.get(
+                "benchmark_language", "unknown"
+            ),
             "tags": tags,
             "stats": {
                 "data": [to_float(x) for x in benchmark_result.data],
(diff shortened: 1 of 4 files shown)


austin3dickey (changes requested):
I validated this change with my local conbench; looks good. I think you did the unit tests correctly.

After running `pytest` locally, the `conbench/tests/api/_expected_docs.py` file changes because you changed the docs. Could you please commit the changes to that file?

boshek:
> I validated this change with my local conbench; looks good. I think you did the unit tests correctly.
> 
> After running `pytest` locally, the `conbench/tests/api/_expected_docs.py` file changes because you changed the docs. Could you please commit the changes to that file?

Done. Thanks for that.

austin3dickey (approved):

conbench/entities/benchmark_result.py
```diff
@@ -38,6 +38,7 @@ class BenchmarkResult(Base, EntityMixin):
     context = relationship("Context", lazy="joined")
     run = relationship("Run", lazy="select")
     unit = Nullable(s.Text)
+    language = Nullable(s.Text)
```
ElenaHenderson, line 41: Language is not benchmark result so it should not be stored in benchmark results

ElenaHenderson (changes requested):
We should include `context` dict (that includes `language` field) in benchmark results json the same way as we include `tags` dict.

conbench/entities/benchmark_result.py
```diff
@@ -184,6 +185,9 @@ def _dump(self, benchmark_result):
             "run_id": benchmark_result.run_id,
             "batch_id": benchmark_result.batch_id,
             "timestamp": benchmark_result.timestamp.isoformat(),
+            "language": benchmark_result.context.tags.get(
```
ElenaHenderson, line 188: We should include `context` dict (that includes `language` field) in benchmark results json the same way as we include `tags` dict.

dianaclarke: Conbench already has a concept of benchmark language. It expects it to be called `benchmark_language`, and that it is sent in the `context`.

- https://github.com/conbench/conbench/search?q=benchmark_language
- https://github.com/conbench/conbench/search?q=display_language

Screen Shot 2022-07-19 at 12 35 13 PM

dianaclarke: You can also filter by the benchmark language. Hope that helps, cheers!

Screen Shot 2022-07-19 at 12 43 14 PM

Screen Shot 2022-07-19 at 12 42 28 PM

boshek: Sorry for the tardy response. Great thought about using the contexts endpoint. I am largely interacting with this via the api so a join on the context id would be just a fruitful here and avoids overloading a benchmark result with _all_ of the context. Right now benchmarks returns a link to the context. Certainly I can process that to extract the id but I wonder if it might be useful to return `context_id` here. I can modify the PR to do that instead which is maybe a bit cleaner?

austin3dickey:
I'm going to close this stale PR for now -- but @boshek if you'd like to open a new PR adding `context_id` to the `GET /api/benchmarks` endpoint, that would be welcome!

PR #366, opened 2022-09-02
Aggressively raise errors if there are extra kwargs that are not processed

ericdill:
@jonkeane this is almost certainly overly aggressive error raising, but i think it is a decent conversation starter. You and I had been working on some code that was aiming to use `Conbench.record(...)` more programmatically. In trying to pass run_reason and run_name to the `record` function some of these kwargs were getting silently ignored in the call to `self.publish`, causing a substantial amount of confusion about why we weren't properly passing run_reason and run_name to conbench.

I don't think this change will have any impact on existing code, though it might show some other ways in which `record` is being called incorrectly.

Another (maybe better) option could be to add a lengthy docstring to record explaining that you can just pass run_reason and run_name to the `options` kwarg instead and then the existing function will handle them properly.

### DIFF
--- a/conbench/runner.py
+++ b/conbench/runner.py
@@ -224,7 +224,7 @@ class Conbench(Connection, MixinPython, MixinR):
 
     def record(self, result, name, error=None, publish=True, **kwargs):
         """Record and publish an external benchmark result."""
-        tags, info, context, github, options, cluster_info, output = self._init(kwargs)
+        tags, info, context, github, options, cluster_info, output, remaining_kwargs = self._init(kwargs)
 
         tags["name"] = name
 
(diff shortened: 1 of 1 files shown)


conbench/runner.py
```diff
@@ -260,13 +260,16 @@ def record(self, result, name, error=None, publish=True, **kwargs):
...
 
-        run_reason = options.get("run_reason")
-        if run_reason is not None:
-            benchmark["run_reason"] = run_reason
+        benchmark["run_reason"] = options.get("run_reason") or remaining_kwargs.pop('run_reason', None)
+        if benchmark["run_reason"] is None:
+            raise RuntimeError("conbench requires run_reason to be set via --run-reason CLI or passing run_reason kwarg to `record()` func")
+
+        if remaining_kwargs:
+            raise RuntimeError(f"There are remaining kwargs that Conbench.record(...) isn't processing: {remaining_kwargs}")
```
jonkeane, line 272: Could we make this a warning (for now, at least) since this is used in a handful of places including some outside of our immediate control. Then we can follow up (or likely do this in the work Ed's working on...) to make it stricter once we ensure that no one was relying on this "feature"

conbench/runner.py
```diff
@@ -260,13 +260,16 @@ def record(self, result, name, error=None, publish=True, **kwargs):
         else:
             benchmark["machine_info"] = self.machine_info
 
-        run_name = options.get("run_name")
-        if run_name is not None:
-            benchmark["run_name"] = run_name
+        benchmark["run_name"] = options.get("run_name") or remaining_kwargs.pop('run_name', None)
+        if benchmark["run_name"] is None:
+            raise RuntimeError("conbench requires run_name to be set via --run-name CLI or passing run_name kwarg to `record()` func")
```
jonkeane, line 265: I don't think this will actually impact anyone, but for context: the original reason these weren't marked as required like this is that really they are only required _if one is publishing results_. And the CLI run-conbench-manually-locally continues if there's a `POST` failure.

I don't think very many people are doing this run-manually (I know of one instance in the past few weeks, but very few-to-none other than that). The work that Ed's doing to separate the runners-bits out from the conbench bits will also totally change that interaction paradigm (and so we will avoid this issue entirely).

I'm fine to have these be errors, though maybe we should also update the docs to basically say "if you aren't sending results to conbench, make something up, it doesn't matter, but it has to be non-empty" https://github.com/conbench/conbench/blob/35fde9954e2b365e36472f16b46fe3067e9dc455/conbench/cli.py#L128-L141

conbench/runner.py
jonkeane: And, separately: should we have a test to confirm this does what we think it does + we don't accidentally break it in the future?

jonkeane (commented):
Thanks for this! Two pretty minor comments

austin3dickey:
I'm going to close this stale PR since this behavior will be in the new CLIs, but we'll revisit if necessary!

PR #369, opened 2022-09-07
Lightweight adapters

alistaire47:
Adds a new python package (currently called `conbencher`, though that's flexible) that defines lightweight adapters that can take benchmark results produced by a runner dispatched via the command line, transform the results into what conbench needs, and sends them to the conbench server. Users creating a new adapter should only have to inherit from `_BenchmarkAdapter` and then define the `command` attribute (i.e. how to call benchmarks) and `.transform_results()` (i.e. how to convert the results into `BenchmarkResult` instances).

This PR contains the abstract class, one for [Google Benchmark](https://github.com/google/benchmark), one for [archery](https://arrow.apache.org/docs/developers/benchmarks.html), and one for [folly](https://github.com/facebook/folly). Later we'll likely add a generic adapter for results from runner packages that already adhere to the schema required.

It also makes the conbench server POST API more flexible such that stats do not need to be precomputed; raw values can be passed in and stats can be computed server-side. This change is to prevent lots of users from calculating things slightly differently when not using the conbench package to run/publish benchmarks.


conbench/entities/benchmark_result.py
```diff
@@ -57,8 +58,41 @@ def create(data):
...
+            "q3",
+            "iqr",
+        ]
         if has_error:
             benchmark_result_data = {"error": data["error"]}
+        # calculate stats if any missing
+        # explicit `is None` because `stdev` is often 0
+        elif any([data["stats"].get(field) is None for field in stats_fields]):
+            dat = data["stats"]["data"]
+            fmt = "{:.6f}"  # not really sure why we're doing this, but now here for backwards consistency
```
alistaire47: I'm happy to not round everything to 6 decimal places and convert numbers to strings if we're ok with it; this is just what we were doing before for whatever reason

jonkeane: Oooph, yeah IMO not rounding, or at least not rounding to the point of JSON precision would be nice

conbench/entities/benchmark_result.py
```diff
@@ -1,5 +1,6 @@
 import flask as f
 import marshmallow
+import numpy as np
```
jonkeane, line 3: Glad this dependency is here and doesn't need to be in all the runners!

conbench/entities/benchmark_result.py
```diff
@@ -57,8 +58,41 @@ def create(data):
...
+            "stdev",
+            "q1",
+            "q3",
+            "iqr",
+        ]
         if has_error:
             benchmark_result_data = {"error": data["error"]}
+        # calculate stats if any missing
+        # explicit `is None` because `stdev` is often 0
+        elif any([data["stats"].get(field) is None for field in stats_fields]):
```
jonkeane: Am I reading this correctly that this will actually overwrite any sent field so long as a single field is `None`? Could we make it so that it fills in each field only if that field is `None`?

conbencher/README.md
```diff
@@ -0,0 +1,7 @@
+# conbencher
```
jonkeane: I know this'll be lots of renaming files and things, but what about `conbenchconnector`? or `conbenchadapter`? `conbencher` is totally fine, it's just a bit generic and not totally obvious that _this_ is the one connector piece that you should depend on (as opposed to `conbench` itself or some other package)

conbencher/python/conbencher/adapters/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .archery import ArcheryAdapter
+from .gbench import GoogleBenchmarkAdapter
```
jonkeane: Nice, I like the separation here, makes it a bit easier to point to what writing a new one would look like

conbencher/python/conbencher/adapters/archery.py
```diff
@@ -0,0 +1,43 @@
...
+import uuid
+
+from ..result import BenchmarkResult
+from .gbench import GoogleBenchmark, GoogleBenchmarkAdapter
+
+
+class ArcheryAdapter(GoogleBenchmarkAdapter):
+    """A class for running Apache Arrow's archery benchmarks and sending the results to conbench"""
+
+    command = ["archery", "benchmark", "run"]
```
jonkeane: Should this be in an `__init__` function so that it's accessible on the instance of the class?

conbencher/python/setup.py
```diff
@@ -0,0 +1,38 @@
...
+setuptools.setup(
+    name="conbencher",
+    version=__version__,
+    description="Utilities for Conbench Runs",
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    packages=setuptools.find_packages(),
+    classifiers=[
+        "Programming Language :: Python :: 3.8",
+        "License :: OSI Approved :: MIT License",
```
jonkeane: I think we want this to be apache2 to match all of the other modules (I know conbench itself is something else, but all new things should be apache2 ideally)

conbencher/python/setup.py
```diff
@@ -0,0 +1,38 @@
...
+    ],
+    python_requires=">=3.8",
+    maintainer="Apache Arrow Developers",
+    maintainer_email="dev@arrow.apache.org",
+    url="https://github.com/conbench/conbencher",
```
jonkeane: I'm not totally sure this is right, it might be better as conbench developers (and have VoltronData given credit too)

jonkeane (commented):
Thanks for this! Liking what I'm seeing so far. A handful of relatively minor comments

conbench/entities/benchmark_result.py
alistaire47: We can, but I worry we'll get things differently calculated (e.g. the many types of percentiles), so if we recalculate it we get more consistency. Frankly I'd rather we not accept precalculated stats at all, but for now we need it for backwards compatibility.

conbencher/README.md
alistaire47: Oh I'd like to change this; I just don't know to what yet. `conbenchadapters` has nice parallelism with a package Austin once made, though it is a bit long.

conbencher/python/conbencher/adapters/archery.py
alistaire47: Just changed this for every other class 😆 For archery, it is a CLI so it should always be available, presumably, provided it's installed and on PATH. It's fine to override the attribute after initialization if, say, you want to point to a particular archery binary.

conbencher/python/setup.py
alistaire47: Ah, I definitely copied this from somewhere...maybe benchmarks or conbench 😬 Will switch

alistaire47: Changed this and added a license file. Do we need a `NOTICE` file? Apache suggests yes, but nobody else seems to have one.

conbencher/python/setup.py
alistaire47: Changed these so they're less wrong, but I'm not sure they're what we want yet

conbencher/README.md
alistaire47: hm, maybe mirror `benchalerts` and call it `benchadapt`?

conbench/entities/benchmark_result.py
austin3dickey: Can we always compute stats here no matter whether they were given or not? We don't have to reject POSTs with stats in them (for now) but we can just document that posting stats is soft-deprecated and will be ignored

conbench/entities/benchmark_result.py
```diff
@@ -57,8 +58,40 @@ def create(data):
...
+            dat = data["stats"]["data"]
+            q1, q3 = np.percentile(dat, [25, 75])
+
+            benchmark_result_data = {
+                "data": [x for x in dat],
+                "times": [x for x in data["stats"].get("times", [])],
```
austin3dickey: Any reason you're doing `[x for x in mylist]` here instead of just `mylist`?

conbencher/README.md
austin3dickey: I like that, or `benchadapters`

conbencher/python/conbencher/adapters/_adapter.py
```diff
@@ -0,0 +1,86 @@
...
+
+        Parameters
+        ----------
+        kwargs
+            Passed through to `run()`
+        """
+        self.run(**kwargs)
+        self.post_results()
+
+    def run(self, params: list[str] = None) -> list[BenchmarkResult]:
```
austin3dickey: If we're supporting python 3.8 (which I think we should continue) then I think this needs to be `typing.List`

conbencher/python/conbencher/adapters/gbench.py
```diff
@@ -0,0 +1,226 @@
...
+    """A class for running Google Benchmarks and sending the results to conbench"""
+
+    def __init__(self, command: list[str]) -> None:
+        """
+        Parameters
+        ----------
+        command : list[str]
+            A list of strings defining a shell command to run the benchmarks
+        """
+        self.result_file = NamedTemporaryFile().name
```
austin3dickey: I might be completely wrong here, but I was under the impression that if you did something like
```
file = NamedTemporaryFile()
filename = file.name
```
then the file wouldn't be cleaned up until the `file` variable is out of the scope of the interpreter and garbage-collected. So when you do `NamedTemporaryFile().name`, since there's no variable storing the `NamedTemporaryFile` object, it could be garbage-collected sooner? To be safe, you should put that object in a separate variable.

austin3dickey: Oh yeah I just validated this is the case.
```
>>> import tempfile
>>> x = tempfile.NamedTemporaryFile().name
>>> x
'/var/folders/fz/ydy9vm4d6qbgxw11ys6j_kfh0000gn/T/tmpq2myi2ed'
>>> with open(x) as f:
...  print(f.read())
...
FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/fz/ydy9vm4d6qbgxw11ys6j_kfh0000gn/T/tmpq2myi2ed'
>>>
>>> file = tempfile.NamedTemporaryFile()
>>> x = file.name
>>> with open(x) as f:
...  print(f.read())
...

>>>
```

conbencher/python/conbencher/adapters/_adapter.py
```diff
@@ -0,0 +1,86 @@
...
+        ----------
+        params : list[str]
+            Additional parameters to be appended to the command before running
+        """
+        command = self.command
+        if params:
+            command += params
+
+        res = subprocess.run(args=command)
+        res.check_returncode()
```
austin3dickey: nit: If you want you can just use `check=True` in `subprocess.run()` instead of the second line here!

conbencher/python/conbencher/adapters/_adapter.py
```diff
@@ -0,0 +1,86 @@
...
+
+        return self.results
+
+    @abc.abstractmethod
+    def transform_results(self) -> list[BenchmarkResult]:
+        """
+        Method to transform results from the command line call into a list of
+        instances of `BenchmarkResult`.
+        """
+        raise NotImplementedError
```
austin3dickey: nit: Since this is an `abstractmethod` you don't need this line; the interpreter will throw an error when constructing a child class if it's not implemented. Usually I just have the docstring and that's enough to satisfy the syntax.

(this doesn't matter unless you're striving for 100% coverage someday 🚀)

conbencher/python/conbencher/adapters/gbench.py
```diff
@@ -0,0 +1,226 @@
...
+    def _parse_gbench_json(raw_json: dict) -> tuple[dict, list]:
+        """Parse gbench result json into a context dict and a list of grouped benchmarks"""
+        gbench_context = raw_json.get("context")
+
+        # Follow archery approach in ignoring aggregate results
+        non_agg_benchmarks = [
+            b for b in raw_json["benchmarks"] if b["run_type"] != "aggregate"
+        ]
+
+        benchmark_groups = groupby(
```
austin3dickey: TIL groupby

conbencher/python/conbencher/adapters/folly.py
```diff
@@ -0,0 +1,71 @@
...
+                )
+
+                parsed_benchmarks.append(parsed_benchmark)
+
+        return parsed_benchmarks
+
+
+if __name__ == "__main__":
+    folly_adapter = FollyAdapter(
+        ["echo", "hello"], Path("~", "Downloads", "tmpn4lr6v57").expanduser()
```
austin3dickey: hello! 😉 

(maybe remove this)

conbencher/python/conbencher/machine_info.py
```diff
@@ -0,0 +1,301 @@
+import os
```
austin3dickey: Is the goal to eventually remove this from `conbench`?

conbencher/python/conbencher/result.py
```diff
@@ -0,0 +1,93 @@
...
+    context: Dict[str, Any] = None
+    github: Dict[str, Any] = field(default_factory=github_info)
+    error: str = None
+
+    def __post_init__(self):
+        if not self.machine_info and not self.cluster_info:
+            self.machine_info = machine_info(host_name=None)
+
+        if self.run_reason:
+            assert self.run_reason in {"commit", "merge", "nightly"}
```
austin3dickey: The current conbench instances I know about have these reasons:
```suggestion
            assert self.run_reason in {"commit", "manual", "nightly", "pull request"}
```
I'm still [moderately against](https://github.com/conbench/conbench/pull/351#discussion_r893615937) enforcing a strict set of reasons, because people might want to use this package for a "weekday" or "release" run (or something else in the future) without making a PR to this code.

conbencher/python/tests/adapters/test_folly.py
```diff
@@ -0,0 +1,96 @@
...
+
+class TestFollyAdapter:
+    folly_bms = [
+        *folly_jsons["velox_benchmark_feature_normalization.json"],
+        *folly_jsons["velox_benchmark_basic_selectivity_vector.json"],
+    ]
+
+    @pytest.fixture(scope="class")
+    def folly_adapter(self):
+        tempdir = Path(TemporaryDirectory().name)
```
austin3dickey: same thing here

austin3dickey (changes requested):
Lookin' really good!! This is all small stuff

conbench/entities/benchmark_result.py
jonkeane: We don't need to decide here, but I like the flexibility of providing a stats if the client knows best. I might be imagining a future that won't happen, but frameworks like Google bench and folly could calculate those underlyingly but not expose the iteration timings, in which case we would want to be able to send those along without requiring that they are always calculated on the server

conbench/entities/benchmark_result.py
alistaire47: Oops, I edited out the silly formatting we were doing, but forgot to unwind this; will fix

conbencher/python/conbencher/adapters/_adapter.py
alistaire47: fixed this everywhere (I think)

conbench/entities/benchmark_result.py
boshek: A +1 for at least having the ability for pre-computed stats. I completely understand the desire but if we make this decision, we will hear about it. I wonder if there is a place for both conbench and user generated stats.

conbencher/python/conbencher/adapters/gbench.py
alistaire47: It's mostly just to generate a path here—the actual file needs to be written by the benchmark, not this class. Actually, I think maybe we need to make users specify it; the `--output` bit below this was left over from archery

conbencher/README.md
boshek: `benchclamp`?

conbencher/python/LICENSE.txt
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
boshek:
```suggestion
   Copyright (c) 2022, Voltron Data.
```

conbencher/python/conbencher/adapters/gbench.py
alistaire47: Ok, fixed this, and where we do need tempfiles, switched to `tempfile.mktemp()`, because the context manager was just distracting here

conbencher/python/conbencher/adapters/_adapter.py
alistaire47: oh very nice!

conbencher/python/conbencher/machine_info.py
alistaire47: Yeah, this shouldn't live in conbench itself; it always needs to run on the benchmarking machine, not the server. It will live in both till everything gets migrated and we can simplify conbench.

conbencher/python/conbencher/result.py
alistaire47: Yeah, the more I think about it, the less I'm convinced we should validate here; users could easily have weekly runs or whatever. If we want to keep cardinality down, the right way is probably to validate server-side against a table that users can populate via an endpoint.

boshek: If we go the route of having a prescribed value, I wonder "test" might also be useful.

boshek (approved):
Nice! Just a few comments.

conbencher/python/LICENSE.txt
alistaire47: I think that's part of the license that supposed to be filled in when copied elsewhere. I'm unclear on how it's supposed to look, though; for Apache projects themselves they put [a NOTICE file in the same dir as the LICENSE file](https://github.com/apache/arrow/blob/master/NOTICE.txt) and [this is left as a template](https://github.com/apache/arrow/blob/master/LICENSE.txt#L190). But none of the Python packages I see have an equivalent NOTICE, whereas LICENSE is explicitly mentioned in the packaging docs.

tl;dr: 🤷 i'm confused

conbencher/python/conbencher/result.py
alistaire47: Ripped it out for now

conbench/entities/benchmark_result.py
alistaire47: k, changed this to only compute missing stats when data is available

conbencher/python/conbencher/machine_info.py
austin3dickey: :nice:

benchadapt/python/benchadapt/adapters/archery.py
```diff
@@ -9,9 +12,12 @@ class ArcheryAdapter(GoogleBenchmarkAdapter):
     """A class for running Apache Arrow's archery benchmarks and sending the results to conbench"""
 
     def __init__(self) -> None:
-        super().__init__(command=["archery", "benchmark", "run"])
+        result_file = NamedTemporaryFile()
```
austin3dickey: ah shoot you know what? I think this should be an attribute too (could be hidden). Or else it could get garbage collected after exiting `__init__()` and subsequent writes to the file might fail. 

Unless you're planning never to write to this file if the benchmarks are supposed to override it

conbench/entities/benchmark_result.py
austin3dickey: Makes sense to me!

austin3dickey (approved):
Looks great! One (probably insignificant) thing

benchadapt/python/benchadapt/adapters/archery.py
alistaire47: Oops, missed this one. Changed it to `tempfile.mkfile()` too; the super init will store it as an attribute.

benchadapt/python/setup.py
```diff
@@ -0,0 +1,38 @@
...
+    long_description_content_type="text/markdown",
+    packages=setuptools.find_packages(),
+    classifiers=[
+        "Programming Language :: Python :: 3.8",
+        "License :: OSI Approved :: Apache Software License",
+    ],
+    python_requires=">=3.8",
+    maintainer="Voltron Data",
+    maintainer_email="conbench@voltrondata.com",
+    url="https://github.com/conbench/benchadapt",
```
ElenaHenderson: Is this correct url?

alistaire47: It doesn't work yet, but I think after merge that will point to what's currently here: https://github.com/conbench/conbench/tree/edward/conbencher-py/benchadapt

alistaire47: oh wait no, that's not right, I think it should be https://github.com/conbench/conbench/tree/main/benchadapt

alistaire47: pushed

ElenaHenderson (approved):
Looks great! Easy to read. Thank you!
Just one question but it should not block the merge.

conbench/entities/benchmark_result.py
jgehrcke: Update: with https://github.com/conbench/conbench/pull/1127 we made self-calculated aggregates take precedence, see my detailed opinion here: https://github.com/conbench/conbench/pull/1127#discussion_r1169138841

PR #370, opened 2022-09-15
Pass ArcheryAdapter output file

alistaire47:
A quick bugfix PR; I realized I missed moving specifying the output file in the archery command back when cleaning up the gbench one. More minorly, also got the logic backwards when warning about publishable results.

### DIFF
--- a/benchadapt/python/benchadapt/result.py
+++ b/benchadapt/python/benchadapt/result.py
@@ -78,7 +78,7 @@ class BenchmarkResult:
                 "Result not publishable! `machine_info` xor `cluster_info` must be specified"
             )
 
-        if bool(res_dict["stats"]) != bool(res_dict["error"]):
+        if bool(res_dict["stats"]) == bool(res_dict["error"]):
             warnings.warn(
                 "Result not publishable! `stats` xor `error` must be be specified"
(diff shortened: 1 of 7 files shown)


alistaire47:
Also realized we need to export `BenchmarkAdapter` if we want users to inherit from it, so renamed it to remove the `_` prefix

austin3dickey (approved):
nice catches

benchadapt/python/benchadapt/result.py
```diff
@@ -78,7 +78,7 @@ def to_publishable_dict(self):
                 "Result not publishable! `machine_info` xor `cluster_info` must be specified"
             )
 
-        if bool(res_dict["stats"]) != bool(res_dict["error"]):
```
jonkeane, line 81: Do we have a test that covers this?

jonkeane (approved):
Looks good, one small comment about (probably?) adding a test for something we caught here

benchadapt/python/benchadapt/result.py
alistaire47: Oops, there's not even a file that tests it at all directly. Added.

PR #372, opened 2022-09-16
Allow overriding of result defaults in benchadapt

alistaire47:
We need to be able to override parts of the `BenchmarkResult` metadata at runtime in {benchadapt}, so this PR adds ways to 

1. override values passed to `BenchmarkResult`'s init method (e.g. for `cluster_info` so `machine_info` doesn't get auto-populated)
2. append values after instantiation where we don't want to override all auto-populated values (e.g. to add more tags). 

I don't _love_ this approach—it seems a little complicated and heavy-handed, so if y'all have suggestions of how to make this more elegant, please pipe up!


benchadapt/python/benchadapt/adapters/_adapter.py
```diff
@@ -17,16 +17,42 @@ class BenchmarkAdapter(abc.ABC):
...
+        self,
+        command: List[str],
+        result_defaults_override: Dict[str, Any] = None,
+        result_defaults_append: Dict[str, Any] = None,
+    ) -> None:
         self.command = command
 
+        if not result_defaults_override:
+            result_defaults_override = {}
+        self.result_defaults_override = result_defaults_override
```
austin3dickey: nit: instead of
```py
if not x:
    x = y
self.x = x
```
you can do this nice lil truthy thing!
```py
self.x = x or y
```

benchadapt/python/benchadapt/adapters/_adapter.py
```diff
@@ -17,16 +17,42 @@ class BenchmarkAdapter(abc.ABC):
...
+        Appended after instantiation. Useful for appending extra tags or other
+        metadata in addition to that gathered elsewhere. Only applicable for dict
+        attributes. For each element, will override any keys that already exist,
+        i.e. it does not append recursively.
     results : List[BenchmarkResult]
         Once `run()` has been called, results from that run
     """
 
     command: List[str]
+    result_defaults_override: Dict[str, Any] = None
```
austin3dickey: `result_defaults_append` should go here too right?

benchadapt/python/benchadapt/adapters/_adapter.py
```diff
@@ -64,6 +90,28 @@ def transform_results(self) -> List[BenchmarkResult]:
...
+        A method to create instances of `BenchmarkResult` with defaults filled with
+        any specified on init in ``result_defaults``.
+
+        Parameters
+        ----------
+        kwargs
+            Named arguments to parameters of `BenchmarkResult`. Takes priority over
+            arguments from ``result_defaults`` and `BenchmarkResult`'s defaults.
+        """
+        res = BenchmarkResult(**{**self.result_defaults_override, **kwargs})
```
austin3dickey: 🤩

benchadapt/python/benchadapt/adapters/_adapter.py
```diff
@@ -17,16 +17,42 @@ class BenchmarkAdapter(abc.ABC):
     command : List[str]
         A list of args to be run on the command line, as would be passed
         to `subprocess.run()`.
+    result_defaults_override : Dict[str, Any]
+        A dict of default values to be passed to `BenchmarkResult`. Useful for
+        specifying metadata only available at runtime, e.g. build info. Overrides
+        `BenchmarkResult` defaults; is overridden by values passed in
```
austin3dickey: One thing I'm confused about is that someone implementing an adapter needs to know to always call `self.curried_benchmark_result()` on each benchmark result before returning them from `self.transform_results()`. Could we automatically do that in `run()` instead?

benchadapt/python/benchadapt/adapters/_adapter.py
```diff
@@ -17,16 +17,42 @@ class BenchmarkAdapter(abc.ABC):
...
+        ``transform_results()``. Values of ``None`` do not unset defaults, just as
+        when passed to ``BenchmarkResult`'s init method.
+    results_defaults_append : Dict[str, Any]
+        A dict of default values to be appended to `BenchmarkResult` values.
+        Appended after instantiation. Useful for appending extra tags or other
+        metadata in addition to that gathered elsewhere. Only applicable for dict
+        attributes. For each element, will override any keys that already exist,
+        i.e. it does not append recursively.
```
austin3dickey: hm, so even after reading the docs, source, and tests, I still don't have a good intutition for everything that will be put into these defaults. Could you add concrete examples to the docs? Like "put cluster info and XYZ here" and "put extra tags here"? We can always modify the docs later if the status quo changes.

Having said that, I wonder if it would be better to scope these new variables to just modifying cluster info and tags if that's all we need to modify right now. Unless there will be other attributes to set defaults for, and you worry it'd get out of control? As this PR stands, these new attributes seem too generic to me to reason about.

austin3dickey (changes requested):
A smattering of comments

benchadapt/python/benchadapt/adapters/_adapter.py
alistaire47: Yeah, they need use use the method instead of the class. (Well technically they don't have to, but if they use `BenchmarkResult` directly users won't be able to override anything.) I don't love this idiom, but I don't have a better idea at the moment.

benchadapt/python/benchadapt/adapters/_adapter.py
alistaire47: It's not just those; most of the fields are metadata fields. `tags`, `info`, `context`, `github`, and `machine_info`/`cluster_info` are all dicts of metadata; fields like `run_reason` may be overridden this way at runtime too.

I was trying to avoid the previous approach of lots of `extra_tags` params flying around and coupling between classes generally; if we change the structure of `BenchmarkResults` in the future, I don't want to have to overhaul this. But the UX feels clunky. I can write more examples, but I feel like there must be a better way to do this...

austin3dickey: Yeah, if you anticipate putting all 6 of those fields in one of these two `result_defaults_` dicts in future implementations, then that would be way too many to first-class here. I'm sure I'll understand better once I see the Arrow adapter implementation.

benchadapt/python/benchadapt/adapters/_adapter.py
austin3dickey: Oh I meant [here](https://github.com/conbench/conbench/blob/7d565be48044a7770bb4a14842fc043e23e12366/benchadapt/python/benchadapt/adapters/_adapter.py#L82) could you do something like 
```py
self.results = [self.curried_benchmark_result(res) for res in self.transform_results()]
```
? Then we'll take one step off the person implementing a custom adapter

alistaire47: That would only work for changing fields after-the-fact, not before, which is useful to prevent fields from getting auto-populated, e.g. if we specify `cluster_info`, `machine_info` will stay `None` (as it should so it can get posted) instead of getting auto-populated.

And we can't necessarily pass in a partially pre-populated instance, because some fields like `timestamp` we really do want to be generated at runtime (though this probably doesn't matter too much). 

But I think you might be right; the path forward here might be only editing after the fact, and improving `BenchmarkResult` so setting `cluster_info` will unset `machine_info` (which probably involves moving it out of `__post_init__` and figuring out how to make dataclasses and properties play nice...lordy, it's attrs all over again). I'll play a little and see what I can do.

benchadapt/python/benchadapt/adapters/_adapter.py
alistaire47: yep, realized this later...not quite sure how this was working?

benchadapt/python/benchadapt/adapters/_adapter.py
alistaire47: This seems like one of those things that's mostly only intuitive to people who spend all day in Python, but it is short, so why not

benchadapt/python/benchadapt/adapters/_adapter.py
alistaire47: Ok, adjusted `BenchmarkResult` so we can run everything after-the-fact. Had to add a little wackiness (there's now `cluster_info`, `_cluster_info`, and `__cluster_info` 😬), but the `__post_init__` is now gone and post-init setting of `cluster_info` is now handled properly. I moved the abstract method from `transform_results()` to `_transform_results()`; the former now calls the latter plus updates the results with any metadata set on init. 

Much happier with this version; there's still some hackiness, but writers of adapters don't have to worry about it, and users of adapters still get all the functionality.

benchadapt/python/benchadapt/adapters/_adapter.py
alistaire47: Kept this UI, aside from renaming `result_defaults_*` to `result_fields_*` (since they're now applied after, not as defaults). Still not sure I love this UI (or the name), but it is powerful and easy to maintain. We could add some syntactic sugar methods to build the dicts after init, though I'm not really sure that's better, and it would make it easier to call things in the wrong order and end up in a weird state.

benchadapt/python/benchadapt/adapters/_adapter.py
austin3dickey: True, though I think it's a pretty popular pattern for iterator defaults since they're mutable.

benchadapt/python/benchadapt/adapters/_adapter.py
austin3dickey: Technically you don't need to declare any of these attributes at all, at least if you set them before you try to read them. But it's nice for type hinting and so devs can keep track of what attributes you're planning to use in methods.

benchadapt/python/benchadapt/adapters/_adapter.py
```diff
@@ -57,13 +76,46 @@ def run(self, params: List[str] = None) -> List[BenchmarkResult]:
...
+    def _transform_results(self) -> List[BenchmarkResult]:
+        """
+        Method to transform results from the command line call into a list of
+        instances of `BenchmarkResult`. The results of this method will be
+        updated to apply runtime metadata values specified on init.
         """
 
+    def update_benchmark_result(self, result: BenchmarkResult) -> BenchmarkResult:
+        """
+        A method to update instances of `BenchmarkResult` with vallues specified on
```
austin3dickey:
```suggestion
        A method to update instances of `BenchmarkResult` with values specified on
```

jonkeane:
```suggestion
        A method to update instances of `BenchmarkResult` with values specified on
```

benchadapt/python/benchadapt/adapters/folly.py
```diff
@@ -12,19 +12,39 @@ class FollyAdapter(BenchmarkAdapter):
...
+        result_fields_override : Dict[str, Any]
+            A dict of values to override on each instance of `BenchmarkResult`. Useful
+            for specifying metadata only available at runtime, e.g. build info. Applied
+            before ``results_field_append``.
+        results_fields_append : Dict[str, Any]
+            A dict of default values to be appended to `BenchmarkResult` values after
+            instantiation. Useful for appending extra tags or other metadata in addition
+            to that gathered elsewhere. Only applicable for dict attributes. For each
+            element, will override any keys that already exist, i.e. it does not append
+            recursively.
```
jonkeane, line 38: Is there any way we could depend on these docs being in one place? Or are they subtly different from one another?

benchadapt/python/tests/adapters/test_adapter.py
```diff
@@ -0,0 +1,69 @@
...
+        result_fields_override = {"cluster_info": {"size": "v big"}}
+
+        fake_adapter = FakeAdapter(
+            command=["echo", "hello"],
+            result_fields_override=result_fields_override,
+        )
+
+        res_list = fake_adapter.transform_results()
+        assert res_list[0].cluster_info == result_fields_override["cluster_info"]
+        assert res_list[0].machine_info is None
```
jonkeane: A comment that `machine_info` is default there (and thus this test is also confirming that goes away) might be nice?

jonkeane (approved):
Thanks for this! 

IIRC, we will circle back to this once we have our PRs adding this to the few places we use it, but it would great to have an example in a readme of what this overall looks like + examples in these over-ride methods for how to use them. But I totally get wanting to et the UX right first, and coming back to them

benchadapt/python/benchadapt/result.py
```diff
@@ -59,15 +59,23 @@ class BenchmarkResult:
...
     context: Dict[str, Any] = None
     github: Dict[str, Any] = field(default_factory=github_info)
     error: str = None
 
-    def __post_init__(self):
-        if not self.machine_info and not self.cluster_info:
-            self.machine_info = machine_info(host_name=None)
+    @property
+    def _cluster_info(self) -> Dict[str, Any]:
+        return self.__cluster_info
```
austin3dickey: wow yeah this is ugly haha! But I can't think of a better way to avoid overriding the dataclasses magic.

Maybe it would be a little easier to understand if you renamed `_cluster_info` -> `_temp_cluster_property` and `__cluster_info` -> `_cluster_info_cache`? (unless you somehow need the [name mangling](https://docs.python.org/3/tutorial/classes.html#private-variables) of the double underscore)

benchadapt/python/tests/adapters/test_adapter.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def test_append_results(self) -> None:
+        results_defaults_append = {"tags": {"price": "$15.99"}}
```
austin3dickey: nit: change this test variable name too

austin3dickey (approved):
I like this a lot better. Just some small things.

benchadapt/python/benchadapt/adapters/_adapter.py
austin3dickey: Oops I should have refreshed before posting my review 😄

benchadapt/python/benchadapt/adapters/_adapter.py
austin3dickey: I like this!

benchadapt/python/benchadapt/adapters/folly.py
alistaire47: They're identical. I'm always torn when writing Python docs; there are ways in Sphinx to inherit sections, I think? but not a direct equivalent of roxygen's `@inheritParams` that I've seen. Nor are we actually rendering any of these docs anywhere, so I'm presuming they'll mostly get used by people reading docstrings in the source or their editor, in which case I really want these to actually be documented on the classes users touch, not just the parent class for devs. I'd love a general solution beyond copying and pasting if anyone's got one, but I haven't.

PR #373, opened 2022-09-20
Use repository.lower() when looking up existing commit and creating new one

ElenaHenderson:


conbench/entities/benchmark_result.py
```diff
@@ -125,7 +125,7 @@ def create(data):
             repository = repository_to_url(data["github"]["repository"])
 
         # create if not exists
-        commit = Commit.first(sha=sha, repository=repository)
+        commit = Commit.first(sha=sha, repository={repository.lower() if repository else None})
```
austin3dickey: I don't think you need the set
```suggestion
        commit = Commit.first(
            sha=sha, repository=repository.lower() if repository else None
        )
```

austin3dickey (commented):
Are there other places we need to wrap with `lower()`? For example should we store all the repos as lowercase on the commit entity?

conbench/entities/benchmark_result.py
ElenaHenderson: Thank you! I have no idea what I was thinking.

ElenaHenderson:
> Are there other places we need to wrap with `lower()`? For example should we store all the repos as lowercase on the commit entity?

Looked. Don't really see anything similar situations for other tables. I would also be careful to lower() anything unless we look for data on all conbench instances and make sure we have correct migrations.

austin3dickey:
Hm, just thought of something. If a server is set up to track a repo with an uppercase letter, e.g. `https://github.com/my_org/MyRepo`, won't something strange happen?

1. someone posts a benchmark result for commit sha `"abc"`
2. a new Commit is created with `"sha": "abc", "repository": "https://github.com/my_org/MyRepo"`
3. someone posts another benchmark result on the same commit
4. this PR code doesn't find any matches for `sha="abc", repository="https://github.com/my_org/myrepo"`
5. it tries to create another Commit with `"sha": "abc", "repository": "https://github.com/my_org/MyRepo"`
6. that fails due to the unique index

That's why I was thinking we potentially have to store the repo as `.lower()`.

ElenaHenderson:
> Hm, just thought of something. If a server is set up to track a repo with an uppercase letter, e.g. `https://github.com/my_org/MyRepo`, won't something strange happen?
> 
> 1. someone posts a benchmark result for commit sha `"abc"`
> 2. a new Commit is created with `"sha": "abc", "repository": "https://github.com/my_org/MyRepo"`
> 3. someone posts another benchmark result on the same commit
> 4. this PR code doesn't find any matches for `sha="abc", repository="https://github.com/my_org/myrepo"`
> 5. it tries to create another Commit with `"sha": "abc", "repository": "https://github.com/my_org/MyRepo"`
> 6. that fails due to the unique index
> 
> That's why I was thinking we potentially have to store the repo as `.lower()`.

You are right. I should have thought about this!

austin3dickey (approved):
Nice, looks good. One more thing; would you mind putting an uppercase repo in [these tests](https://github.com/conbench/conbench/blob/main/conbench/tests/entities/test_commit.py#L29)? Thanks!!

ElenaHenderson:
> Nice, looks good. One more thing; would you mind putting an uppercase repo in [these tests](https://github.com/conbench/conbench/blob/main/conbench/tests/entities/test_commit.py#L29)? Thanks!!

Done.

PR #374, opened 2022-09-21
Add more logging to benchadapt for CI visibility

alistaire47:
It's currently very hard to see what's running (and what's not) in CI; this should help

### DIFF
--- a/benchadapt/python/benchadapt/adapters/_adapter.py
+++ b/benchadapt/python/benchadapt/adapters/_adapter.py
@@ -1,9 +1,10 @@
 import abc
+import json
 import subprocess
 from typing import Any, Dict, List
 
 from ..client import ConbenchClient
-from ..log import fatal_and_log
+from ..log import fatal_and_log, log
 from ..result import BenchmarkResult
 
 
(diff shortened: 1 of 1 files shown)


benchadapt/python/benchadapt/adapters/_adapter.py
```diff
@@ -120,17 +126,23 @@ def post_results(self) -> list:
         """
         Post results of run to conbench
         """
-        client = ConbenchClient()
```
alistaire47, line 123: Moved this below the erroring out if results don't exist

benchadapt/python/benchadapt/adapters/_adapter.py
```diff
@@ -120,17 +126,23 @@ def post_results(self) -> list:
...
+
+        log.info("Posting results to conbench")
         res_list = []
         for result in self.results:
             result_dict = result.to_publishable_dict()
+            log.debug(
+                f"Posting benchmark result to conbench: `{json.dumps(result_dict)}`"
+            )
             res = client.post(path="/benchmarks", json=result_dict)
             res_list.append(res)
```
jonkeane, line 145: It might be nice to also log a non-failing response here (well in the client.post method really...) too if we've got this level of debugging elsewhere

alistaire47: The client does `log.debug("POST <url> <json>")` already, though our default level is set to INFO. I don't see the number of requests here being prohibitively enormous, but it's still probably easier to have users change the logging level to DEBUG if they want to see it all

jonkeane: Ah I missed that, great

PR #375, opened 2022-09-22
Bugfix: Coerce string-likes to strings for logging

alistaire47:
The `command` attribute is a list passed to `subprocess.run()`, but it will sometimes contain instances of `pathlib.Path()`, which `.run()` will coerce properly, but we have to coerce manually for logging.

### DIFF
--- a/benchadapt/python/tests/adapters/test_adapter.py
+++ b/benchadapt/python/tests/adapters/test_adapter.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 from typing import List
 
 from benchadapt.adapters import BenchmarkAdapter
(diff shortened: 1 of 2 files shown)


benchadapt/python/tests/adapters/test_adapter.py
```diff
@@ -42,6 +43,12 @@ def test_transform_results(self) -> None:
         res_list = fake_adapter.transform_results()
         assert res_list[0] == BenchmarkResult(**RESULTS_DICT)
 
+    def test_run(self) -> None:
+        fake_adapter = FakeAdapter(command=["echo", Path(".").resolve()])
+
+        res_list = fake_adapter.run()
+        assert res_list[0] == BenchmarkResult(**RESULTS_DICT)
```
jonkeane, line 50: Thanks for the test here!

jonkeane (approved):

PR #377, opened 2022-09-23
Show more details when HTTP request fails

alistaire47:
Bugfixes and cleanup:

- Print details when HTTP request fails
- Add default for `info` and `context`
- Add more things to docs that I've learned
- Remove duplicative logging
- Format JSON logging in client as JSON

### DIFF
--- a/benchadapt/python/benchadapt/result.py
+++ b/benchadapt/python/benchadapt/result.py
@@ -27,7 +27,8 @@ class BenchmarkResult:
     stats : Dict[str, Any]
         Measurement data and summary statistics
     tags : Dict[str, Any]
-        Many things. Determines history runs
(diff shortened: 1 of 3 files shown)


benchadapt/python/benchadapt/result.py
```diff
@@ -27,7 +27,8 @@ class BenchmarkResult:
     stats : Dict[str, Any]
         Measurement data and summary statistics
     tags : Dict[str, Any]
-        Many things. Determines history runs
+        Many things. Must include a ``name`` element; often includes parameters either as
+        separate keys or as a string in a ``params`` key. Determines history runs.
     info : Dict[str, Any]
```
jonkeane, line 32: Thanks for these docs updates

jonkeane (approved):

PR #378, opened 2022-09-23
Better run metadata defaults

alistaire47:
This makes it easier for users to handle run metadata (`run_name`, `run_id`, `run_reason`) correctly (or at least consistently with current usage). Changes:

1. If `run_id` is not already specified elsewhere, sets a single value for all `BenchmarkResult` instances created in the course of running `.transform_results()`
2. If `github` is specified in a result, uses the sha together with `run_reason` (even if it's `None`) as `run_name`.
3. Copies over Austin's client updates to show more info when a request fails
4. Documents a lot of things better with information I've learned.


benchadapt/python/benchadapt/adapters/_adapter.py
```diff
@@ -46,6 +51,7 @@ def __init__(
         self.command = command
         self.result_fields_override = result_fields_override or {}
         self.result_fields_append = result_fields_append or {}
+        self.run_id = uuid.uuid4().hex
```
alistaire47: This could just be generated in `transform_results()` and passed through to `update_result()` if we want it to differ for multiple runs with the same adapter instance

austin3dickey: Maybe this is a better idea

austin3dickey (approved):
Looks good!

benchadapt/python/benchadapt/adapters/_adapter.py
alistaire47: changed

benchadapt/python/benchadapt/adapters/_adapter.py
```diff
@@ -89,7 +90,10 @@ def transform_results(self) -> List[BenchmarkResult]:
         """
         log.info("Transforming results for conbench")
         results = self._transform_results()
-        self.results = [self.update_benchmark_result(res) for res in results]
+        run_id = uuid.uuid4().hex
```
jonkeane, line 95: Will this generate a new runid for each json file that is parsed? (this might be true, I haven't followed it in detail)

Here's an example of this happening in velox: https://github.com/facebookincubator/velox/blob/a5821449193254916293bf509faf7ec66ae6a9be/scripts/veloxbench/veloxbench/implemented_benchmarks.py#L34-L44

For those, each of those json files we want to be in a single run.

alistaire47: No, unless you set things up weirdly so you create a new adapter instance for each file and don't specify the `run_id` elsewhere. That's a bad idea; an adapter instance should always correspond to a run, and the iteration over files should happen in the `_transform_results()` method of the adapter, e.g. [here](https://github.com/conbench/conbench/blob/main/benchadapt/python/benchadapt/adapters/folly.py#L51) for folly.

benchadapt/python/benchadapt/result.py
```diff
@@ -68,6 +76,10 @@ class BenchmarkResult:
     github: Dict[str, Any] = field(default_factory=github_info)
     error: str = None
 
+    def __post_init__(self) -> None:
+        if not self.run_name and self.github.get("commit"):
+            self.run_name = f"{self.run_reason}: {self.github['commit']}"
```
jonkeane, line 106: We might not want to default to this until we clean up the way we identify the main branch for history purposes. IIUC, this would mean that unless someone configures `run_name`, if they are on a PR branch, they would get `commit: ...` as the default, yeah?

alistaire47: They'd get whatever is specified in `run_reason`, which has no default (and therefore could be `None`, which is a bit gross). I don't currently have a way to autodetect what that should be, as adapters could be run in so many ways; populating `run_reason` is still something users will have to do. We could make them specify `run_name` as well, but I suspect most will do it poorly.

benchadapt/python/benchadapt/adapters/_adapter.py
jonkeane: Cool, when we do get to documenting all of this in a more user-facing way, let's make sure to add "create a new adapter instance for each file and don't specify the run_id elsewhere. That's a bad idea" to that

benchadapt/python/benchadapt/result.py
jonkeane: Aah ah ah right. I saw too many "commit" around there and transposed that to the second line, my bad

benchadapt/python/benchadapt/adapters/_adapter.py
alistaire47: I added more to the docs both for the abstract adapter and the result now because it's all fresh in my brain

PR #380, opened 2022-09-28
Configure number of commits for calculating stats

### LINKED ISSUE #379: Make history duration configurable
Right now we have a fixed window of history as the last 100 commits https://github.com/conbench/conbench/blob/2977fcda6af0ccacff88e76777fd333fe8b85bec/conbench/entities/benchmark_result.py#L171 (we should confirm this is actually where that's coming from + that this is actually what's happening. I'm 90% certain this is, but not completely certain!)

Can we make this value be configurable (still default to 100), just like our other configuration values (an env var that we deploy with, cf https://github.com/conbench/conbench/blob/main/conbench/config.py). We have at least one deployment of conbench which would rather use a longer time period even if it means more false negatives.

alistaire47:
Closes #379 by allowing users to specify a `DISTRIBUTION_COMMITS` env var when starting a conbench server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,7 +225,9 @@ Most are self-descriptive, e.g. `DB_PORT`. Exceptions:
 * `DISTRIBUTION_COMMITS`: An integer number of commits to use when calculating
 statistics. The default is 100; larger numbers will lead to more false negatives,
 especially after large changes. We recommend leaving it as the default. Previously
-recorded values will not be recalculated if this value is changed.
+recorded values will not be recalculated if this value is changed. If you would
+like to change previous values, you would need to write a migration of the data
+to recalculate history.
 
 ## Authoring benchmarks
 
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ repository, and the results are hosted on the
 ## Index
 
 * [Contributing](https://github.com/conbench/conbench#contributing)
+* [Configuring the server](https://github.com/conbench/conbench#configuring-the-server)
 * [Authoring benchmarks](https://github.com/conbench/conbench#authoring-benchmarks)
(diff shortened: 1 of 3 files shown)


jonkeane (approved):
Looks good, would you mind adding some docs about this (and especially that going above 100 for now will lead to more false negatives, especially after large changes. We recommend leaving it as the default.) It might also be good to mention that it only changes the data that is recorded when it's set (e.g. if you change it on a deployment all new data will have the change, but old data is not automatically migrated)

alistaire47:
@jonkeane Added some docs; does this align with what you had in mind? Duplicated the description for now.

README.md
```diff
@@ -215,6 +216,17 @@ password in postgres to `postgres`.
...
+## Configuring the server
+
+The conbench server can be configured with various environment variables as
+defined in [config.py](https://github.com/conbench/conbench/blob/main/conbench/config.py).
+Most are self-descriptive, e.g. `DB_PORT`. Exceptions:
+
+* `DISTRIBUTION_COMMITS`: An integer number of commits to use when calculating
+statistics. The default is 100; larger numbers will lead to more false negatives,
+especially after large changes. We recommend leaving it as the default. Previously
+recorded values will not be recalculated if this value is changed.
```
jonkeane:
```suggestion
recorded values will not be recalculated if this value is changed. If you would
like to change previous values, you would need to write a migration of the data
to recalculate history.
```

jonkeane (approved):
Wonderful! Thanks. A small addition to the readme to (very vaguely) wave at where one might do the updating if one wanted to (hopefully no one ever will tho!)

README.md
alistaire47: add this in both places

PR #381, opened 2022-09-29
Specify iterations in adapters

alistaire47:
Quick bugfix; `iterations` always needs to be specified in `stats` for the server right now even though it's redundant with the length of `data`

### DIFF
--- a/benchadapt/python/benchadapt/adapters/folly.py
+++ b/benchadapt/python/benchadapt/adapters/folly.py
@@ -72,6 +72,7 @@ class FollyAdapter(BenchmarkAdapter):
                         "unit": "ns",
                         "times": [result[2]],
                         "time_unit": "ns",
+                        "iterations": 1,
                     },
                     tags={"name": bm_name, "suite": suite, "source": "cpp-micro"},
                     info={},
--- a/benchadapt/python/benchadapt/adapters/gbench.py
+++ b/benchadapt/python/benchadapt/adapters/gbench.py
@@ -218,6 +218,7 @@ class GoogleBenchmarkAdapter(BenchmarkAdapter):
                 }.get(result.unit, result.unit),
                 "times": result.times,
                 "time_unit": result.time_unit,
+                "iterations": len(result.values),
             },
             tags=tags,
             info={},
--- a/benchadapt/python/tests/adapters/test_archery.py
+++ b/benchadapt/python/tests/adapters/test_archery.py
@@ -80,6 +80,7 @@ class TestArcheryAdapter:
             assert result.machine_info is not None
             assert result.stats["data"] == original["values"]
             assert result.stats["times"] == original["times"]
+            assert result.stats["iterations"] == len(original["values"])
 
     def test_run(self, archery_adapter) -> None:
         results = archery_adapter.run()
--- a/benchadapt/python/tests/adapters/test_folly.py
+++ b/benchadapt/python/tests/adapters/test_folly.py
@@ -9,44 +9,44 @@ from benchadapt.result import BenchmarkResult
 folly_jsons = {
     "velox_benchmark_basic_selectivity_vector.json": [
         [
-            "/Users/jkeane/repos/velox/velox/benchmarks/basic/SelectivityVector.cpp",
+            "/Users/floopdorp/repos/velox/velox/benchmarks/basic/SelectivityVector.cpp",
             "sumBaselineAll",
             0,
         ],
         [
-            "/Users/jkeane/repos/velox/velox/benchmarks/basic/SelectivityVector.cpp",
+            "/Users/floopdorp/repos/velox/velox/benchmarks/basic/SelectivityVector.cpp",
             "sumSelectivityAll",
             0,
         ],
         [
-            "/Users/jkeane/repos/velox/velox/benchmarks/basic/SelectivityVector.cpp",
+            "/Users/floopdorp/repos/velox/velox/benchmarks/basic/SelectivityVector.cpp",
             "sumSelectivity99PerCent",
             0,
         ],
         [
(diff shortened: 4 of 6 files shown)


alistaire47:
@jonkeane I might be able to relax the constraint that `iterations` always has to be specified and fill it in server-side with the summary stats. Any sense of whether that's desirable, or is there reason to always specify it?

benchadapt/python/benchadapt/adapters/folly.py
```diff
@@ -72,6 +72,7 @@ def _transform_results(self) -> List[BenchmarkResult]:
                         "unit": "ns",
                         "times": [result[2]],
                         "time_unit": "ns",
+                        "iterations": 1,
```
alistaire47, line 75: I'm not sure this couldn't be higher—does folly have an idea of replications?—but this is implicit in our current processing of the results

jonkeane (commented):
What do we currently do with arrow bench + the velox benchmarks? Is a default of 1 filled in? 1 iteration is ... definitely not right for most of these, but figuring out the right number is hard or impossible. If we already are filling in a default of 1, this is fine let's keep doing it. We might open an issue to treat these kinds of benchmarks slightly differently at some point and fix this, but we shouldn't mess with that now.

I'm surprised we don't have a test that would bump into this already, would it be possible to add one?

benchadapt/python/benchadapt/adapters/folly.py
jonkeane: Yeah, I alluded to this in my other comment: I don't actually think folly returns # of iterations that ran. Though it _does_ at some point have some number of iterations that it runs. 

We probably want at some point to treat these microbenchmarks which run for a set time period and measures how many iterations it gets through slightly differently than our macro benchmarks where each iteration happens regardless of how long it takes. But not today!

alistaire47:
> What do we currently do with arrow bench + the velox benchmarks? Is a default of 1 filled in? 1 iteration is ... definitely not right for most of these, but figuring out the right number is hard or impossible. If we already are filling in a default of 1, this is fine let's keep doing it. We might open an issue to treat these kinds of benchmarks slightly differently at some point and fix this, but we shouldn't mess with that now.

Velox is currently ultimately calling `conbench.record()` which calls a `_stats()` method that [also does `len(data)`](https://github.com/conbench/conbench/blob/main/conbench/runner.py#L364). Arrowbench is ultimately getting passed through that too right now.

For gbench/archery we have [this](https://github.com/conbench/conbench/blob/main/benchadapt/python/benchadapt/adapters/gbench.py#L192-L194) which aggregates separate replications; the current velox implementation doesn't have anything like that.

By "definitely not right" do you mean we should be using the actual number of iterations these do? If so, how should we merge replications? Or are we actually running multiple replications for velox?
 
> I'm surprised we don't have a test that would bump into this already, would it be possible to add one?

Yep, this is in a weird space since it's still just a convention in the JSON, not an actual attribute on an object or something. We could make a `BenchmarkStats` object or something if we wanted, though I'm wary of overcomplicating.

benchadapt/python/benchadapt/adapters/folly.py
alistaire47: `iterations` here is more what archery calls `replications`, not the actual number of iterations run in a period of time (that's in `data` in iterations/s). If we wanted to store raw iterations, we'd need to do it elsewhere, since this field is a single integer, and can't handle a vector of values like you'd get for multiple replications.

jonkeane:
> By "definitely not right" do you mean we should be using the actual number of iterations these do? If so, how should we merge replications? Or are we actually running multiple replications for velox?

IIUC, folly is just like googlebench (and thus archery) where all of the benchmarks that are run are run with as many iterations as you can in some set of time. So they all _do_ have some number of iterations that were run that are used for calculating the statistic that comes out of it, but also IIUC folly doesn't actually even dump that number anywhere. 

This is what I was thinking about when I wanted to make sure that we can send stats separately from the client: I don't think any of the microbenchmark frameworks we use now will give us this information, (but it's totally possible that they already can we "just" don't know how to get to it), but if they did: we could have all the stats data + the actual number of iterations that were run for each microbenchmark from the runner data itself (but never actually see or get the list of iteration times ever. 

Here's an example with R's {bench}:
```
> bench::mark(runif(1000))
# A tibble: 1 × 13
  expression      min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time
  <bch:expr>  <bch:t> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm>
1 runif(1000)  12.1µs 15.6µs    56437.    10.4KB     5.64  9999     1      177ms
# … with 4 more variables: result <list>, memory <list>, time <list>, gc <list>
```

If we could get those details from google bench, folly, whatever, we could pipe them directly into `stats`, `iterations`, etc. 

But this is all more detail than we should dive into right now — if we're hardcoding 1 already let's keep that and put a pin in "Could we actually know more info from these microbenchmarking frameworks and save that?"

alistaire47:
> IIUC, folly is just like googlebench (and thus archery) where all of the benchmarks that are run are run with as many iterations as you can in some set of time. So they all _do_ have some number of iterations that were run that are used for calculating the statistic that comes out of it, but also IIUC folly doesn't actually even dump that number anywhere.

We are recording those values here, just in `data` and normalized by time, since they're meaningless without the time limit. `iterations` here is not that microbenchmark sort of iteration; it's the number of replications, i.e. how many times the number of iterations was recorded, so it's a scalar, not a vector.

jonkeane:
I think we're miscommunicating somewhere here, but it's not important to this change or how conbench currently works, so let's merge this and we can figure out where the disconnect is when | if we ever come back to making conbench understand microbenchmarks versus macrobenchmarks better than it currently does

PR #386, opened 2022-10-03
Add benchrun package and CallableAdapter

alistaire47:
This is a major PR which adds a new {benchrun} Python package that has a [hopefully simple!] framework for assembling Python macrobenchmarks. I've added a pretty comprehensive README which is a good place to start to understand the API.

This PR also adds a `CallableAdapter` to {benchadapt} with which {benchrun} is designed to be used. It is generic, and is designed to work with any Python callable (a function or a class with a `__call__()` method that can be called like one) that returns a list of `BenchmarkResult` instances, avoiding the shelling out other adapters do.

I've tried hard to handle metadata for the user as far as possible, but they still will need to specify some things like `run_reason`, and will want to specify other things like particulars about builds and such. As we migrate {benchmarks} to this, I'll try to assemble a guide of best-practices for what should go where. Eventually we should probably write some integration tests to make sure the whole benchrun -> benchadapt -> conbench server pipeline (among other things) works smoothly, too.


benchrun/python/benchrun/_benchmark.py
```diff
@@ -0,0 +1,263 @@
...
+        Parameters
+        ----------
+        case : dict
+            A dict where keys are parameters and values are scalar arguments for a benchmark
+
+        Returns
+        -------
+        A dict passed to the ``setup_results`` param of ``run()`` when the class is called.
+        """
+        return {}
```
jonkeane: In arrowbench, we have both a setup and a global setup, do you think we'll need a global setup here eventually? We don't have to add it now (and shouldn't if it's not yet needed) but curious your thoughts on that.

benchrun/python/benchrun/_benchmark.py
```diff
@@ -0,0 +1,263 @@
...
+        Code to run in each iteration after timing.
+
+        Parameters
+        ----------
+        case : dict
+            A dict where keys are parameters and values are scalar arguments for a benchmark
+        run_results : dict
+            The results of calling ``run()``. Use for passing data between stages.
+        """
+        pass
```
jonkeane: We might consider calling this something else; for example: presumably this is where validation will happen and it seems weird to call that in a 'teardown' function when it could (rightfully!) result in a failure of the benchmark itself

benchrun/python/benchrun/_benchmark.py
```diff
@@ -0,0 +1,263 @@
...
+    gc_collect : bool
+        Run garbage collection before timing code?
+    gc_disable : bool
+        Disable garbage collection during timing?
+    error_handling : str
+        What should happen if a benchmark errors out? Options: ``"stop"`` (skip future iterations
+        and report only error), ``"break"`` (skip future iterations and report everything so far),
+        ``"continue"`` (run all iterations even if they fail and report everything). As we currently
+        can't report both metrics and errors for the same benchmark, ``"stop"`` and ``"break"`` are
+        currently identical, and ``"continue"`` may run longer, but will report the same thing.
```
jonkeane: This is error handling with in iterations, right? As in, I've got 3 iterations and the 3rd one fails, what to do then, yeah? 

If so, could we add that to the first sentence (as well as it already being there later)?

benchrun/python/benchrun/_benchmark.py
```diff
@@ -0,0 +1,263 @@
...
+    result_fields_append: Dict[str, Any] = None
+    settings = None
+    cache = None
+
+    def __init__(
+        self,
+        iteration: Iteration,
+        case_list: CaseList,
+        result_fields_append: Dict[str, Any] = {},
+        drop_caches: bool = False,
```
jonkeane, line 203: I think we want this to be true? But really we should _also_ have a TODO that does a really investigation into how this impacts the actual data.

benchrun/python/benchrun/case.py
```diff
@@ -0,0 +1,56 @@
...
+    params: Dict[str, list] = None
+    case_list: List[Dict[str, Any]]
+
+    def __init__(self, params: Dict[str, list]) -> None:
+        self.params = params
+
+        crossed_case_list = [
+            dict(zip(self.params.keys(), tup))
+            for tup in itertools.product(*self.params.values())
+        ]
```
jonkeane: This might deserve a comment about what it's doing

jonkeane (commented):
A few comments — this is looking very good. We should also see if @ericdill has any thoughts on what (else) might be needed or helpful to incorporate this into the internal VoltronData-tools based running flow.

benchrun/python/benchrun/_benchmark.py
alistaire47: Hm, so I wrote this so it runs each iteration in a new process, so a global setup/teardown that runs in the main process would be less useful (still good for writing files, etc; less good for importing libraries and setting up contexts) and more confusing. Reading through arrowbench, I'm pretty sure it's running all iterations in a single new process. We could switch to that, if we want? It's maybe less good from a benchmarking perspective, but way simpler from a user perspective.

benchrun/python/benchrun/_benchmark.py
alistaire47: [The current default is false](https://github.com/conbench/conbench/blob/main/conbench/runner.py#L343), which is why I set it this way. We can change that, but it may cause a discontinuity in some benchmarks that weren't explicitly setting it (whenever we get around to porting them).

benchrun/python/benchrun/_benchmark.py
jonkeane: Where does the timing happen? Inside the subprocess?

benchrun/python/benchrun/_benchmark.py
jonkeane: Let's keep this as whatever the current default is and mark a TODO to come back and actually figure out what the right answer is. I think we've cargo-culted ourselves on this one, but that's an empirical question we should dedicate the time to figuring out in the future.

benchrun/python/benchrun/_benchmark.py
alistaire47: Yep, that happens in the subprocess either way

jonkeane: Ok, cool. So then we have to "just" figure out if isolating each iteration is beneficial or detrimental. Ideally we would parameterize this and test — I suspect there's a non-trivial time spend in doing various setup steps in each subprocess, but we should measure that and base our decision on that IMO

alistaire47: It depends on the use-case. If the setup block is "download this dataset", that can be global regardless; if it's "read in this dataset to time writing", yeah, it'd have to happen in each subprocess, whether that's per benchmark or per iteration. Subprocess per-benchmark is definitely faster; the only reason to keep per-iteration is because we think the isolation would result in lower-variance timings. It may prevent some caching/memoisation/JIT effects (maybe?? potentially, but there's no guaranteed effect size here), but doing a big test here would require implementing a lot of benchmarks.

jonkeane: Indeed. It's possible we will want | need both approaches. IMO we shouldn't decide which until we do the test and have data to back up our recommendations. In the mean time, let's flag our uncertainty, pick one, merge it, and create the follow on to do it right in the future

alistaire47: Yeah, actually it might not be all that hard to make two subclasses and allow both. And for the simplicity of programming, subprocess-per-benchmark is probably the default, which means I'll have to write most of both anyway if we want to just do it now.

alistaire47: Alright, in separating out `setup()` and `teardown()` from `before_each()` and `after_each()`, it became a lot easier to just run it exactly like arrowbench, which seems good for continuity. We can reimplement subprocess-per-iteration if we want, but it will be a tad more finicky to program.

I also moved all the `*_results` params to an `env` dict attribute that stages can modify to store/pass state; that felt less messy and more like arrowbench anyway.

benchrun/python/benchrun/_benchmark.py
alistaire47: Done; it's now `setup -> iterations * (before_each -> run -> after_each) -> teardown`

benchadapt/python/tests/adapters/test_callable.py
```diff
@@ -0,0 +1,83 @@
...
+    },
+    "info": {},
+    "machine_info": {
+        "name": "beepboop.local",
+        "os_name": "macOS",
+        "os_version": "12.6",
+        "architecture_name": "arm64",
+        "kernel_name": "21.6.0",
+        "memory_bytes": "17179869184",
+        "cpu_model_name": "Apple M3 Pro",
```
austin3dickey, line 33: this dude is living in the future!

benchrun/python/README.md
```diff
@@ -0,0 +1,149 @@
...
+### `Iteration`
+
+The code to run for a benchmark is contained in a class inheriting from the abstract
+`Iteration` class. At a minimum, users must override the `name` attribute and `run()`
+method (the code to time), but may also override `setup()`, `before_each()`,
+`after_each()` and `teardown()` methods, where `*_each()` runs before/after each
+iteration, and `setup()` and `teardown()` run once before/after all iterations. A
+simple implementation might look like
+
+```python
```
austin3dickey, line 26: Thanks for the examples!

benchrun/python/README.md
```diff
@@ -0,0 +1,149 @@
...
+        self.env = {}
+```
+
+### `CaseList`
+
+An `Iteration`'s methods are parameterized with `case`, a dict where keys are
+parameters for the benchmark, and the values are scalar arguments. Cases are managed
+with an instance of `CaseList`, a class which takes a `params` dict, which is like a
+case dict with the difference that the arguments are lists of valid arguments, not
+scalars. `CaseList` will populate a `case_list` attribute` which contains the grid of
```
austin3dickey:
```suggestion
scalars. `CaseList` will populate a `case_list` attribute which contains the grid of
```

benchrun/python/README.md
```diff
@@ -0,0 +1,149 @@
...
+from benchadapt.adapters import CallableAdapter
+
+my_adapter = CallableAdapter(callable=my_benchmark_list)
+```
+
+Like all adapters, it then has a `run()` method to run all the benchmarks it
+contains (handling generic metadata appropriately for you), a `post_results()`
+method that will send the results to a conbench server, and a `__call__()` method
+that will do both. These are the methods that should be called in whatever CI or
+automated build system will be used for running benchmarks.
```
austin3dickey, line 141: Maybe a quick example of doing this in CI here?

benchrun/python/benchrun/_benchmark.py
```diff
@@ -0,0 +1,280 @@
...
+    """
+    Abstract class defining how to run one iteration of a benchmark.
+
+    Attributes
+    ----------
+
+    name : str
+        A name for the benchmark. Should be specified when inheriting.
+    cache : CacheManager
+        An CacheManager instance for clearing the disk cache when specified.
```
austin3dickey:
```suggestion
        A CacheManager instance for clearing the disk cache when specified.
```

benchrun/python/benchrun/_benchmark.py
```diff
@@ -0,0 +1,280 @@
...
+        """
+        Run iterations and return list of times.
+
+        Parameters
+        ----------
+        case : dict
+            A dict where keys are parameters and values are scalar arguments for a benchmark
+        iterations : int
+            Integer of repetitions to run
+        settings : dict
```
austin3dickey, line 97: Any reason these aren't explicit parameters? It would be nice to see all the settings explicitly

benchrun/python/benchrun/_benchmark.py
```diff
@@ -0,0 +1,280 @@
...
+    case_list: CaseList = None
+    result_fields_append: Dict[str, Any] = None
+    settings = None
+    cache = None
+
+    def __init__(
+        self,
+        iteration: Iteration,
+        case_list: CaseList,
+        result_fields_append: Dict[str, Any] = {},
```
austin3dickey:
```suggestion
        result_fields_append: Dict[str, Any] = None,
```

benchrun/python/benchrun/_benchmark.py
```diff
@@ -0,0 +1,280 @@
...
+            batch_id=batch_id,
+            run_reason=run_reason,
+            stats=res["stats"],
+            error=res["error"],
+            tags={"name": self.iteration.name, **case},
+            # info={},  # TODO: is there common detectable metadata worth putting here?
+            context={"benchmark_language": "Python"},
+        )
+
+        for param in self.result_fields_append:
```
austin3dickey, line 260: Isn't this code already in the BenchmarkAdapter? Why do we have to duplicate it here?

austin3dickey (commented):
Looking great!

benchrun/python/benchrun/_benchmark.py
austin3dickey: (probably that CI README example will help me figure this out)

benchrun/python/benchrun/_benchmark.py
alistaire47: They are explicit on `Benchmark`, which is where they should really be set, but to run this, they have to be passed through from `Benchmark` to `Iteration`, and it's just simpler to pass a dict than everything because with `multiprocessing` I've got to pass all the args through in a tuple:

```python
        proc = mp.Process(
            target=iteration, args=(case, iterations, self.settings, queue)
        )
```
which I kinda hate, and separating them out would exacerbate it. Maybe I could use `NamedTuple` or something? But the tl;dr is that users mostly shouldn't be calling this (though they can); they should use `Benchmark` as an interface instead.

benchrun/python/README.md
alistaire47: I'm not precisely sure what this will look like in CI tbh; presumably for Arrow we'll have to overhaul a bit of arrow-benchmarks-ci. The closest we've come so far was [the Go benchmarks](https://github.com/zeroshade/arrow/blob/go-benchmarks/ci/scripts/go_bench_adapt.py), but that's implementing a new adapter, which doesn't need to happen here. All that really needs to happen is inserting runtime information like `run_reason`. So probably a small Python script which can be called from gh actions or whatnot, but we haven't really figured out best practices for that yet.

benchrun/python/benchrun/_benchmark.py
alistaire47: Yep, and I don't really love the duplication, but the reason is that there will be one `CallableAdapter` instance for all the benchmarks in the `BenchmarkList`, but you may want to set things on some benchmarks but not others, e.g. say a tag for something about the data format for a particular benchmark that's the same for everything now (so it's not necessarily parameterized and therefore a tag) but which may vary in the future.

I could separate it out into a `update_benchmark_result()` util function in `benchadapt` so the same code gets used in both places? Though this isn't huge or likely to change

benchrun/python/benchrun/_benchmark.py
austin3dickey: Yeah, that makes sense. I just read this first before I got to the more user-friendly code below.

benchrun/python/benchrun/_benchmark.py
austin3dickey: yeah maybe we can leave it for now until the packages are more stable

austin3dickey (approved):
nice job!

benchrun/python/README.md
```diff
@@ -0,0 +1,149 @@
...
+{benchrun} is a Python package to run macrobenchmarks, deliberately designed to work
+well with the larger conbench ecosystem.
+
+## Installation
+
+{benchrun} is not [yet] on a package archive like PyPI; you can install from GitHub
+with
+
+```shell
+pip install benchrun@git+https://github.com/conbench/conbench.git@main#subdirectory=benchrun/python
```
ElenaHenderson, line 12: This did not work for me:
```
(base) elena@e benchmarks % pip install benchrun@git+https://github.com/conbench/conbench.git@edward/benchrun#subdirectory=benchrun/python
zsh: no matches found: benchrun@git+https://github.com/conbench/conbench.git@edward/benchrun#subdirectory=benchrun/python
```

alistaire47: Hmm, it's worked for benchadapt in labs/benchmarks: https://github.com/voltrondata-labs/benchmarks/blob/main/requirements.txt#L4 Not sure why that would fail...

alistaire47: wait exactly that does work for me: 

```shell
$ pip install benchrun@git+https://github.com/conbench/conbench.git@edward/benchrun#subdirectory=benchrun/python     
Collecting benchrun@ git+https://github.com/conbench/conbench.git@edward/benchrun#subdirectory=benchrun/python
  Cloning https://github.com/conbench/conbench.git (to revision edward/benchrun) to /private/var/folders/0j/zz6p_mjx2_b727p6xdpm5chc0000gn/T/pip-install-waj5bru7/benchrun_cd08123c40a9404c801db5eadf252c79
  Running command git clone --filter=blob:none --quiet https://github.com/conbench/conbench.git /private/var/folders/0j/zz6p_mjx2_b727p6xdpm5chc0000gn/T/pip-install-waj5bru7/benchrun_cd08123c40a9404c801db5eadf252c79
  Running command git checkout -b edward/benchrun --track origin/edward/benchrun
  Switched to a new branch 'edward/benchrun'
  branch 'edward/benchrun' set up to track 'origin/edward/benchrun'.
  Resolved https://github.com/conbench/conbench.git to commit fc5d2825aedf7f206a3affa1fbc55cfa5281a460
  Preparing metadata (setup.py) ... done
Building wheels for collected packages: benchrun
  Building wheel for benchrun (setup.py) ... done
  Created wheel for benchrun: filename=benchrun-0.0.1-py3-none-any.whl size=15478 sha256=16740270c177f61198904471cffb9a9a0ca78a9acde276b7d726f24b40a5336c
  Stored in directory: /private/var/folders/0j/zz6p_mjx2_b727p6xdpm5chc0000gn/T/pip-ephem-wheel-cache-kuv39y27/wheels/51/99/10/8e53946c25765fba9c37082d375f22d4991383ee3abfa5391d
Successfully built benchrun
Installing collected packages: benchrun
  Attempting uninstall: benchrun
    Found existing installation: benchrun 0.0.1
    Uninstalling benchrun-0.0.1:
      Successfully uninstalled benchrun-0.0.1
Successfully installed benchrun-0.0.1
```
Maybe an ssh issue?

alistaire47: ...or possibly your pip is too old? What I used above: 

```bash
$ pip --version
pip 22.2.2 from /Users/alistaire/mambaforge/envs/arrow-dev/lib/python3.9/site-packages/pip (python 3.9)
```

ElenaHenderson (approved):
I approve this PR since I think we can start with this change and then update stuff as we are migrating benchmarks to use benchadapt and benchrun.

I do wish `BenchmarkAdpater.publish_results` was replaced with `BenchmarkResult.publish` so people could just use `BenchmarkResult` when none of current adapters fit their case and they don't want to invest time into creating a new adapter.

I could not use any adapters for my benchmarks and had to do this to make publish work:
```
class BenchmarkAdapter(benchadapt.adapters.BenchmarkAdapter):
    def __call__(self, *args, **kwargs):
        pass

    def _transform_results(self):
        return self.results


a = BenchmarkAdapter([])
a.results = [result]
a.post_results()
...
```

benchrun/python/README.md
ElenaHenderson: I upgraded pip but same issue. You can ignore my comment. I found another way to install benchadapt and benchrun.

alistaire47:
> I do wish `BenchmarkAdpater.publish_results` was replaced with `BenchmarkResult.publish` so people could just use `BenchmarkResult` when none of current adapters fit their case and they don't want to invest time into creating a new adapter.

Hm, that would require each result carrying around an instance of `GitHubClient`. Maybe we could just put a `.publish_benchmark_result(self, result: BenchmarkResult)` method on `GitHubClient` so users who need to send one result manually could use that?

That does depend on configuring everything correctly on your result directly and not relying on the adapter to get `run_id` and such right for you, though.

jonkeane:
> I do wish BenchmarkAdpater.publish_results was replaced with BenchmarkResult.publish so people could just use BenchmarkResult when none of current adapters fit their case and they don't want to invest time into creating a new adapter.
> 
> I could not use any adapters for my benchmarks and had to do this to make publish work

I wonder if this is a sign that our adapter factoring isn't quite right. Could you say more @ElenaHenderson about what made it hard or impossible to work with? IIRC, our hope was that even folks who needed to write their own runner should use either a standard adapter, or only need to override very small bits of one of the ones that exist, such that we _don't_ need to worry about things like "That does depend on configuring everything correctly on your result directly and not relying on the adapter to get run_id and such right for you, though."

Thoughts? Is there something in benchadapt that we need to improve or change to make that realistic?

.github/workflows/actions.yml
```diff
@@ -30,6 +30,9 @@ jobs:
           pip install -r requirements-build.txt
           pip install .
           pip install coveralls
+          pip install -r benchadapt/python/requirements.txt
+          pip install benchadapt/python
+          pip install benchrun/python
```
austin3dickey: I think the best way to handle this step is to put it all on one line. That way, if a package is specified in multiple places (e.g. a requirements.txt and in this command), pip should put a higher priority on the command-specified one. I also prefer to use `-e` for installing from source in CI, because I've found that omitting that flag can result in the wrong cached package being used in rare occasions. So the end result might look like this:
```
pip install \
  -e . \
  -e benchadapt/python \
  -e benchrun/python \
  -r requirements-test.txt \
  -r requirements-build.txt
```
This assumes you add `coveralls` to `requirements-test.txt`, which I think is a good idea. It's also redundant to specify the requirements files that are already specified in the `setup.py` files, like `requirements-cli.txt` or `benchadapt/python/requirements.txt`.

The only drawback of this approach is if there are conflicting dep versions then pip will fail, but almost none of our deps are version-tagged so I think this should work. It also closer mocks how `pip` will act when a user tries to install from pypi (in which case there'd definitely be a problem if we have conflicts).

.github/workflows/actions.yml
```diff
@@ -55,3 +58,9 @@ jobs:
           coveralls --service=github
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Test benchadapt
+        run: |
+          pytest -vvvv benchadapt/python/tests
+      - name: Test benchrun
+        run: |
+          pytest -vvvv benchrun/python/tests
```
austin3dickey: I lol every time. Technically,
image
but you're free to do whatever you want!

.github/workflows/actions.yml
alistaire47: Oh this makes so much sense, thank you! And yeah, we can leave out the requirements files already picked up by setup.py; not sure why I didn't realize that. A while ago I changed the conbench `setup.py` to at least install its core requirements on pip install, so this that can get less bad now too. I forget if I did the thing so you can do `pip install conbench[dev]` or whatever; I don't think I did? Not sure if people normally do that for testing deps. Seems nice imo.

.github/workflows/actions.yml
alistaire47: yeah we can probably take this back to `-vv`; I think I got this in my brain when I was looking for the debug flag

.github/workflows/actions.yml
alistaire47: 😍 
```
        run: |
          pip install -e .[dev] \
            -e benchadapt/python \
            -e benchrun/python
```
I think it also runs a little faster this way because it doesn't have to do multiple rounds of dependency resolution

ElenaHenderson:
\

> > I do wish BenchmarkAdpater.publish_results was replaced with BenchmarkResult.publish so people could just use BenchmarkResult when none of current adapters fit their case and they don't want to invest time into creating a new adapter.
> > I could not use any adapters for my benchmarks and had to do this to make publish work
> 
> I wonder if this is a sign that our adapter factoring isn't quite right. Could you say more @ElenaHenderson about what made it hard or impossible to work with? IIRC, our hope was that even folks who needed to write their own runner should use either a standard adapter, or only need to override very small bits of one of the ones that exist, such that we _don't_ need to worry about things like "That does depend on configuring everything correctly on your result directly and not relying on the adapter to get run_id and such right for you, though."
> 
> Thoughts? Is there something in benchadapt that we need to improve or change to make that realistic?

 I can not use benchrun or Callable because
- I need to be able to run multiple benchmark cases in one process so I can reuse some expensive setup for each benchmark case
- I need to be able to post results after each query is benchmarked (not once all of them are done)

alistaire47:
> I can not use benchrun or Callable because
> 
>     * I need to be able to run multiple benchmark cases in one process so I can reuse some expensive setup for each benchmark case

This has been addressed in this PR now; you can turn running in a subprocess off by specifying `subprocess=False` when you initialize the `Benchmark`.

>     * I need to be able to post results after each query is benchmarked (not once all of them are done)

This will take more work, so I made https://github.com/conbench/conbench/issues/402

PR #389, opened 2022-10-04
Add entity ERD

alistaire47:
Adds an ERD of the postgres instance in a new README in the `entities` directory. We can put this elsewhere if we don't want it in all installs, but I figured it's not _that_ big and this is as close to where we need it as possible.


jonkeane (approved):
Thanks for this!

Issue #382, opened 2022-09-29
Use the actual git history to situate commits on branches 


jonkeane:
Remove mentions of run name in history + distribution calculations and replace it with something that matches the truth about the git history at the time

Suspicious places:
https://github.com/conbench/conbench/blob/bf68a5b9d52c0fe74dc2e3eedcf72596b6143645/conbench/entities/distribution.py#L102
https://github.com/conbench/conbench/blob/bf68a5b9d52c0fe74dc2e3eedcf72596b6143645/conbench/entities/distribution.py#L156
https://github.com/conbench/conbench/blob/bf68a5b9d52c0fe74dc2e3eedcf72596b6143645/conbench/entities/history.py#L60
https://github.com/conbench/conbench/blob/bf68a5b9d52c0fe74dc2e3eedcf72596b6143645/conbench/entities/run.py#L48
_there might be more_

alistaire47:
@austin3dickey whenever you’re ready for it (Monday?): where things stand:

- code is on [edward/store-branch-info](https://github.com/conbench/conbench/tree/edward/store-branch-info) (sorry, rename it if you like)
- what’s done: 
  - detecting branch when github info gets collected
  - changing the `commit` schema to add `branch` and `fork_point_sha` (I think that's an ok name? `git merge-base --fork-point` is a thing: https://git-scm.com/docs/git-merge-base)
  - a method to get the fork point sha (aka merge base), though it needs a tiny bit of tweaking now that `branch` is stored as `user:branch`. That’s great for forks, but will mess up the conditional that tries to not compare main to main (which errors out).
  - a few tests have been fixed
- left to do:
  - populate records for unbenchmarked commits on the default branch, probably whenever running for a commit on the default branch (which would work nicely for rebase merges)
  - clean up tests. I fixed a tiny bit, but a lot more are broken, and I haven’t looked into why yet. The docker tests (see docs if you haven’t run them; `docker-compose run app pytest -vv conbench/tests`) are passing. I’m not sure if that means the unit tests are failing because they’re hitting an old server, if the docker ones are only passing because I forgot to rebuild an image, or something else. Not sure how much work is here; could be a whole lot or not much.
  - maybe write some more tests if fixing the above doesn’t already get us coverage of new stuff
- out of scope (I think):
  - changing queries. Except maybe we need to for default branch stuff? Not sure, though even if it’s theoretically necessary the old ones will keep working with our squash merges. New comparison queries are def out-of-scope, though; that’s another story.
[...]

PR #356, opened 2022-06-10
Backfilled rows with empty reasons

austin3dickey:
This PR adds a db migration to do a one-time backfill of run reasons using custom logic.

PR #390, opened 2022-10-06
Filter on run reason instead of name prefix

alistaire47:
This attempts to close `#382` (edit: does not, but is useful regardless) by changing filters on run name prefix to filters on run reason, which provided `run_reason` is populated on >100 commits, will not change which commits are selected, but frees up `run_name` to be used for more human-readable purposes.

This might be a fix more for letter than spirit of the issue, but if so hopefully it will be useful to spark conversation about what we want to change.

### DIFF
--- a/conbench/entities/distribution.py
+++ b/conbench/entities/distribution.py
@@ -99,7 +99,7 @@ def get_distribution(benchmark_result, limit):
         .join(Hardware, Hardware.id == Run.hardware_id)
         .join(commits_up, commits_up.c.id == Run.commit_id)
         .filter(
-            Run.name.like("commit: %"),
+            Run.reason == "commit",
             BenchmarkResult.error.is_(None),
             BenchmarkResult.case_id == benchmark_result.case_id,
             BenchmarkResult.context_id == benchmark_result.context_id,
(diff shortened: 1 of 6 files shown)


jonkeane:
> This might be a fix more for letter than spirit of the issue, but if so hopefully it will be useful to spark conversation about what we want to change.

_nods_ yeah, this would move us away slightly from relying on run_name and grepping out of it (which is good), but I was hoping that the way we did that was by replicating more of the git graph in the database (or empowering conbench to look more of it up at run time if that's better).

So something like altering the way that we look up the "parent"[1] here:

https://github.com/conbench/conbench/blob/ebbb7be5f73fa1945fcc070b2f8ee04ae372e5b6/conbench/entities/commit.py#L155-L178

and replace it with something that stores something like the following (I haven't fully worked this out, so this is probably not exhaustive of what we'll need, but also might be too much!):

* actual parent commit sha (regardless of if this is on a branch or not)
* branch name
* if branch is something other than default, the branched-from-commit sha

and then once we have that, we should be able to replace the filter criteria based on run reason | run name with actually matching branches + commit shas (with their parents).


[1] Note the scare quotes: you'll see there that the parent is actually the parent for pull request branches — it's the the branched-from commit on the default branch (or the closest ancestor to it, actually...)

austin3dickey (approved):
IMO we can merge this as-is, because it's a nice improvement on its own. Then we can overhaul these queries once we store more info about branches.

jonkeane:
I'm fine with merging this, though we need to make sure that we migrated run_reason back when we added it? We should confirm that we did on the conbench instances we are running.

austin3dickey:
> we need to make sure that we migrated run_reason back when we added it

Yep we did in https://github.com/conbench/conbench/pull/356 !

PR #391, opened 2022-10-06
Add docs on how to create users

alistaire47:
Just adding a bit to the README about user creation options

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,11 +232,20 @@ to recalculate history.
 
 ## Creating accounts
 
-Follow the login screen's "Sign Up" link, and use the registration key specified
-in the server configuration above.
+By default, conbench has open read access, so a user account is not required to
+view results or read from the API. An account is required if the conbench instance
+is private or to write data to conbench.
 
-To create an additional account, when logged in, go to the gear menu / Users and
-use the "Add User" button to create a new account without the registration key.
+If you do need an account, follow the login screen's "Sign Up" link, and use the
+registration key specified in the server configuration above. If you are a user
+of conbench, you may need to talk to your user administrator to get the
+registration key. SSO can be configured to avoid requiring the registration key.
+
+If you have an account and need to create an additional account (say for a machine
+user of the API) either repeat the process if you have the registration key, or if
+you don't have the registration key (say if your account uses SSO), when logged in,
+go to the gear menu / Users and use the "Add User" button to create a new account
+without the registration key.
 
 ## Authoring benchmarks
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -229,6 +230,14 @@ recorded values will not be recalculated if this value is changed. If you would
 like to change previous values, you would need to write a migration of the data
 to recalculate history.
 
+## Creating accounts
+
+Follow the login screen's "Sign Up" link, and use the registration key specified
+in the server configuration above.
+
+To create an additional account, when logged in, go to the gear menu / Users and
+use the "Add User" button to create a new account without the registration key.
```
jonkeane: Could we also add something like:

If you have set up conbench using an SSO provider, but you're looking to make a user name and password for API access, ...

README.md
```diff
@@ -229,6 +230,14 @@ recorded values will not be recalculated if this value is changed. If you would
 like to change previous values, you would need to write a migration of the data
 to recalculate history.
 
+## Creating accounts
+
+Follow the login screen's "Sign Up" link, and use the registration key specified
+in the server configuration above.
```
jonkeane: Should we mention something here about "if you're a user of conbench, you might need to talk to your conbench administrator to get the registration key"?

README.md
```diff
@@ -229,6 +230,14 @@ recorded values will not be recalculated if this value is changed. If you would
 like to change previous values, you would need to write a migration of the data
 to recalculate history.
 
+## Creating accounts
+
```
jonkeane, line 234: Could we also add in a bit here that describes how _by default_ conbench has open read access, so a user account is not required (even for API access).

But if you have a private conbench instance or you're attempting to write data to conbench, a user account is required.

jonkeane (changes requested):
Thanks for this! A few comments | expansions that we should add.

README.md
alistaire47: Tried again. Better?

README.md
```diff
@@ -229,6 +230,23 @@ recorded values will not be recalculated if this value is changed. If you would
...
+By default, conbench has open read access, so a user account is not required to
+view results or read from the API. An account is required if the conbench instance
```
jonkeane: Maybe "is required only if" (sorry apparently I can't suggest from the mobile interface!)

jonkeane (approved):
Looks good. If we have docs already for setting up with SSO we might link them. If not, could you make an issue to add them (I don't want to bloat this PR, but we really should document that)

alistaire47:
> If we have docs already for setting up with SSO we might link them. If not, could you make an issue to add them (I don't want to bloat this PR, but we really should document that)

https://github.com/conbench/conbench/issues/392

PR #405, opened 2022-10-20
Add new fields to runs and benchmark_results tables and APIs

### LINKED ISSUE #398: Add INFO (json), STARTED_AT (datetime), FINISHED_AT (datetime), ERROR (json) fields to RUNS table and POST /runs api to post runs specific info
I need these fields to keep track of
- runs that could not run at all (e.g., because setup failed) 
- errors preventing a run from executing benchmarks
- when run started
- when run finished

I have been keeping track of this on arrow-bci side but I feel like it is time to add this to Conbench so we can generalize this aspect of running benchmarks.

### LINKED ISSUE #404: Add VALIDATION (JSON) field to benchmark_results table and API
Some benchmarks need validation. We need VALIDATION (JSON) field to keep track of validation and errors if any.

ElenaHenderson:
This PR adds
- these fields to RUNS table and APIs:
    - `finished_timestamp (datetime)`
    - `info (JSON)`
    - `error_info (JSON)`
    - `error_type (str)`
 - `validation (JSON)` to BENCHMARK_RESULTS table and APIs
 - `POST /runs` API endpoint that allows creating runs in conbench before benchmark results are posted
 - `PUT /runs` API endpoint that allows updating these run's fields:
    - `finished_timestamp (datetime)`
    - `info (JSON)`
    - `error_info (JSON)`
    - `error_type (str)`
 
This PR updates UI pages:
- Benchmark page displays Validation section:

Screen Shot 2022-10-20 at 6 09 46 PM

- Run page displays Run section and error_info and info sections if run has values:

Screen Shot 2022-10-20 at 6 08 54 PM
Screen Shot 2022-10-20 at 6 10 18 PM


conbench/entities/run.py
```diff
@@ -150,3 +205,44 @@ def commit_hardware_run_map():
...
+
+
+class _RunFacadeSchemaCreate(marshmallow.Schema):
+    id = marshmallow.fields.String(required=True)
+    name = marshmallow.fields.String(required=False)
+    reason = marshmallow.fields.String(required=False)
+    finished_timestamp = marshmallow.fields.DateTime(required=False)
+    info = marshmallow.fields.Dict(required=False)
+    error_info = marshmallow.fields.Dict(required=False)
+    error_type = marshmallow.fields.String(required=False)
```
alistaire47: What's the intention for how `error_type` is to be used separately from `error_info`? I'm assuming it's supposed to be a low-cardinality value about where it failed, and not an error class (which would go in `error_info`)? Can we document this somewhere? Ideally in descriptions in the swagger, but I'm not sure how hard that is with the way things are set up

conbench/entities/benchmark_result.py
```diff
@@ -98,14 +98,6 @@ def create(data):
         if not case:
             case = Case.create(c)
 
-        # create if not exists
```
austin3dickey, line 101: This feels much cleaner to do this upon run creation

conbench/entities/run.py
austin3dickey: Agreed that it should go in docs somewhere. I think the definitions are in issue https://github.com/conbench/conbench/issues/398

migrations/versions/0363a5c5c9d2_add_fields_to_run_and_benchmark_results.py
```diff
@@ -0,0 +1,41 @@
...
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "0363a5c5c9d2"
+down_revision = "459166107c9d"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
austin3dickey, line 19: Do we want to backfill any of these (maybe error_type)? or perhaps it's not worth the effort

austin3dickey (approved):
Looks great! Small comments

migrations/versions/0363a5c5c9d2_add_fields_to_run_and_benchmark_results.py
ElenaHenderson: I can backfill it when we decide on what categories we want to have. I would avoid doing this for now since we don't know 100%.

conbench/entities/run.py
ElenaHenderson: Documented all new fields in swagger:
Screen Shot 2022-10-21 at 12 07 50 PM

It is not pretty formatting but at least it is documented.

ElenaHenderson: Screen Shot 2022-10-21 at 12 14 00 PM

ElenaHenderson:
@austin3dickey @alistaire47 Thank you for the reviews! I responded to your comments. I will merge this  PR when tests pass. If you still have any concerns after the merge, I will address them in the following PRs.

ElenaHenderson:
Tests and lint passed but something is wrong with sending coverage report, which I don't think an issue with this PR:
```
requests.exceptions.HTTPError: 422 Client Error: Unprocessable Entity for url: https://coveralls.io/api/v1/jobs
```

Issue #402, opened 2022-10-18
Create iterative adapter


alistaire47:
Currently all adapters in benchadapt run all benchmarks, then transform results, then publish them. For sets of benchmarks that are easy to iterate over, it would be nice to run in serial, and run one benchmark, transform its results, and publish it, then move on to the next one. This requires different handling particularly for `run_id`. 

Implementing the new start/finish run endpoints is in scope for these changes, as is integration with benchrun.

PR #406, opened 2022-10-21
Ability to run and process benchmarks iteratively

alistaire47:
This PR changes benchrun and benchadapt to allow benchmarks to be run and processed iteratively, i.e. `for (benchmark in benchmarks) run -> postprocess -> post to conbench` instead of `run all benchmarks -> postprocess all results -> post all results`, closing #402.

Changes:

- benchrun:
    1. Renames `BenchmarkList` to `CallableBenchmarkList` for parallelism with the corresponding adapter.
    2. Adds a new `GeneratorBenchmarkList` class to complement `CallableBenchmarkList`. The core difference is that `GeneratorBenchmarkList`'s `__call__()` method is a generator that iterates over benchmarks and case lists, yielding the `BenchmarkResult` instance from `run_case()`. (By comparison, `CallableBenchmarkList` just calls each benchmark, which runs each case, ultimately returning a list of `BenchmarkResult` instances.)
- benchadapt
    1. Adds a new `GeneratorAdapter` that takes a generator that yields an instance of `BenchmarkResult` in the same way `CallableAdapter` takes a callable that returns a list of instances of `BenchmarkResult`. When called, `GeneratorAdapter` will iterate over the generator, postprocessing each result and posting it to conbench (using existing methods by temporarily pretending the result is the only result) before moving on to the next, which because of the delayed evaluation of generators allows a list of benchmarks to be run one case at a time.
    2. Fixed some typehints

Does not yet implement run-closing calls or validation recording enabled by #398. At least the latter belongs in a separate PR.


benchadapt/python/benchadapt/adapters/generator.py
```diff
@@ -0,0 +1,96 @@
...
+from ..log import log
+from ..result import BenchmarkResult
+from ._adapter import BenchmarkAdapter
+
+
+class GeneratorAdapter(BenchmarkAdapter):
+    """
+    A generic adapter for adapting benchmarks defined in a generator (i.e. a
+    generator or class with an ``__call__()`` method) that yields one
+    `BenchmarkResult` instance at a time. Does not shell out.
```
jonkeane, line 13: > A generic adapter for adapting benchmarks defined in a generator

I wonder if we could add more or different wordings to avoid using "adapter" and "adapting" twice in a row like this?

jonkeane (commented):
I need to take a finer-tooth comb through the code at some point, but wanted to get something up here sooner than that:

We should check and see how this looks + feels when we implement it for real against:

1. a python based running setup (this already is most of the way there with the connection to the runner, but we should write up some examples with real benchmarks and see how that feels — feel free to either borrow arrow benchmarks or make something up in python with a few cases defined by parameters)
2. (which is probably even more important IMHO) what does this feel like to use this from R (or any other non-Python language).  Again this can be a toy benchmark or arrowbench itself, but let's actually see what it feels like to use it + what it would take to implement something with this.

This will help validate that what we're building is lining up with what we will use it for (even better if we can help out others like the work that @ElenaHenderson is doing)

Ultimately, I think we are building this backwards: With the setup in this PR we are interlocking the adapter and the runners in a way that one basically either can only call the list of things to benchmark _from inside_ the adapter or we have to have two separate and parallel ways of calling the list of things to benchmark (one in the adapter and one in the runner). I do truthfully believe this is not setting ourselves up for success, and in many ways is repeating some of the mistakes we made when we wrote voltrondata-labs/benchmarks, which is very very similar in spirit to this approach.

Instead, I'm proposing that we use the adapter like a utility that the runners depend on and shell out to (or however we want to design that API to look like). But the runners are fundamentally (and solely) responsible for the list-of-things to benchmark and writing results out to json for the adapter to suck up + enrich|transform + send to conbench. This separates the areas of concern for each part of our benchmarking stack and gives us more flexibility in connecting these two things together + reusing the adapter in other places.

Some more thoughts on that division of responsibility:

## Benchmark runner (e.g. arrowbench)

What defines how to run the benchmarks + the list of benchmarks that need to be run for a specific run.

 Responsible for: 
* defining cases for each benchmark
* defining what the default set of benchmark (cases) is
* defining how to select specific benchmarks to run (e.g. "only benchmarks with file* in the name" "this specific benchmark case with these parameters")

This should all ideally be in whatever language + frameworks that are responsive to that language. As a benchmark author or someone who wants to add a benchmark this should be stuff that I more-or-less have lots of control over changing


## Adapter

Takes results (either in aggregate, or as they are finished), and sends them to conbench. 
Responsible for:
* filing in or validating run information (run id, run reason, etc.)
* getting the results into a form acceptible to conbench (either through one of the known transformers like folly -> conbench or by accepting well-formed conbench result json
* sending the results to conbench (and dealing with if there are failures or whatever)

Ideally, we write ~one of these (that is, in one language, it might feature a number of different operation modes, but I don't think we should need to write a conbench client in each language that we are benchmarking...) since some of the biggest portion of this is effectively the conbench client. This should also be something that is portable (like datalogistik) so that it can be depended on and re-used easily without hassle by runners 

## The hand off between these
And then the handoff between these is literally a json file that has the details that only the runner knows. The runner then shells out (or again, uses some other API) and tells the adapter to upload that result now.

Some of the pain that we've had with our current setup is explicitly tied to the fact that we currently effectively have an adapter in voltrondata-labs/benchmarks that must be used to call both python and R benchmarks. This means that if we want to add an R benchmark we have to alter the adapter. We also were stymied in doing more targeted filtering because that filtering had to happen at the adapter level (since that's the only thing that is able to trigger the benchmarks when running in CI). This also makes the local versus CI experience of running benchmarks different (or more complicated than it needs to be — this isn't the first use case we need to serve, but it is one that I would like to consider when designing something like this and not forestall).

Finally, if this more tightly-coupled or adapter-wrapping-around-the-runner unblocks something: I'm not opposed to merging it like this and then working on the decoupled, iterative adapter next. But I *do* want us to seriously consider the proposal I'm making and try and code it out and see if it has the properties I think it does (and if it is likely to solve the problems our current setup has, and to figure out if there are any really bad unintended consequences of that, etc.)

alistaire47:
There's lots to chew on here, but I really wish we'd had this conversation earlier, in the design stages before opening #369. If we don't have finalized requirements and [ADRs](https://adr.github.io/) or similar before we start building, we're likely to waste time rewriting later.

As-is, adapters are fully capable of running R benchmarks with cases and filtering and whatnot handled by R code, with the handoff being a shell call and then JSON. Doing so iteratively is possible too, but would require the R to have a way of listing benchmarks to run. All of the running infra would live in R, just like here it lives in benchrun, not benchadapt. The coupling between adapter and runner is deliberately very low—you can adapt essentially anything you can run.

We could write a CLI version that runners call into, but that presents different difficulties:

1. _Maintaining state._ For postprocessing, we need to know what JSON is part of a run. If results all come in one list blob, that's trivial. If we're iterating and making repeated calls to kick off postprocessing and posting, then either the runner side needs to maintain the state and handle `run_id` and such itself, or it pass what has to be more or less `run_id` anyway. There are alternatives like recreating a "start run" and "end run" API like the web API, or creating and passing a new directory where all the JSON for a run will go (though then the runner still has to make a separate call for run finalization). All of this is possible, but it's not trivial to get to work right if you're trying to maintain state on-disk instead of in-memory.
2. _Handling run-time parameters._ Currently benchadapt is built so you can insert run-time parameters like `run_reason` at run time. If configuration goes through the runner, the runner needs to know how to handle these parameters, either by passing them through or by handling them itself. This is doable, but runners will have to be thicker layers rather than just focusing on composing and running benchmarks, and if the inserting is still happening on the CLI side, the handoff gets more complicated.
3. _Existing frameworks._ Existing benchmarking frameworks like archery and folly aren't going to call the CLI, so we still need what is essentially an adapter to a. call it, b. read results, c. transform outputs, and d. post them. That could call a CLI for last tweaks and posting, but all the same code still exists somewhere.
4. _Running in CI._ If calling an adapter always works the same way, I can write a CI implementation guide and template. If each set of benchmarks has its own way of running, I can't—how to pass in `run_reason` and such will always vary.

All of this is possible, but it's trading one set of problems for another which will take work to address and debug. We can plan it out and do it, if that's the direction we decide we want to go. But ultimately all this has little to do with this particular PR, which just enables one possible way of using adapters, and more to do with the big-picture direction.

ElenaHenderson (commented):
The iterative adapter makes sense to me but I still can not take advantage of it for my benchmarks without overriding a lot of code.
The whole benchmark running process is not strait-forward in my case and needs special handling of errors, logging, validation and even case order. 

May I suggest that 
- we only write adapters for benchmarks that use other frameworks and collect results in other formats than conbench needs
- we have a simple `conbench client` for posting results and `conbench utils` for collecting machine info and etc that new benchmarks can take advantage of.

I really think sticking with simple blocks will be more helpful than trying to get users to pick a benchrun class.

jgehrcke:
Closing this for cleanup for now. We can always re-evaluate, re-open, go back to this!

PR #408, opened 2022-10-25
Add a benchmark-level optional info field

### LINKED ISSUE #407: Move info from info.tags into new benchmark_results.info field because info field for benchmark results is unique to each benchmark_results record
The meaning of `INFO.TAGS` field changed since it was added a year ago. Now every benchmark result has its own INFO and there is no INFO that is shared between benchmark results.

Our schema should reflect this change in how `INFO.TAGS` is used .

Another reason for making this change is that users want to post a lot of data into `INFO.TAGS` field but since we index on this field, Postgres can't handle large JSONs being indexed and throws `psycopg2.errors.ProgramLimitExceeded: index row size 4360 exceeds btree version 4 maximum 2704 for index "info_index"` errors.

jonkeane:
Resolves: #407

This field is an open json field and is designed for details that are optional at the benchmark level but that shouldn't make a new case (things like links to logs, links to tracing information, more detailed success | failures).


conbench/entities/benchmark_result.py
```diff
@@ -35,6 +35,9 @@ class BenchmarkResult(Base, EntityMixin):
     data = Nullable(postgresql.ARRAY(s.Numeric), default=[])
     times = Nullable(postgresql.ARRAY(s.Numeric), default=[])
     case = relationship("Case", lazy="joined")
+    # optional info at the benchmark level (i.e. information that isn't a tag that should create a separate case, but information that's good to hold around like links to logs)
```
jonkeane, line 38: This name is not great, but I don't want to write the giant migration that would be required to do the following:

I think we should have at the benchmark level (i.e. that will vary with each benchmark): 
* `case` (which generally come from tags...)
* `optional_info` (or we could call it even more explicitly: `additional_info` or `additional_benchmark_info`)

And then across benchmarks within a run have :
* `context` (just like we have now)
* `context_optional_info` (or `context_additional_info` or whatever to match ^^^) that is our `info` now

This way it's clear(er) that each benchmark has some info about it that matters for performance but is unique to the benchmark (the `case` info), some info that doesn't matter for performance but is unique to the benchmark (the `optional_info`), and then there is info that matters for performance but is shared across benchmarks (`context`) and info that doesn't matter for performance but is shared across benchmarks (`context_optional_info`)

conbench/templates/benchmark-entity.html
```diff
@@ -104,6 +104,17 @@
...
+            <li class="list-group-item active">Additional information</li>
+            {% for k,v in benchmark.optional_info.items() %}
+              <li class="list-group-item" style="overflow-y: auto;">
+                <b>{{ k }}</b>
+                  {% if v is not none %}
+                    <div align="right" style="display:inline-block; float: right;">{{ v }}</div>
+                  {% endif %}
+              </li>
+            {% endfor %}
+          {% endif %}
```
ElenaHenderson, line 53: Did you try to look at this in UI locally?

jonkeane: Not yet, but I need to

conbench/entities/benchmark_result.py
```diff
@@ -236,24 +241,65 @@ class BenchmarkResultSerializer:
 
 
 class _BenchmarkFacadeSchemaCreate(marshmallow.Schema):
-    run_id = marshmallow.fields.String(required=True)
-    run_name = marshmallow.fields.String(required=False)
-    run_reason = marshmallow.fields.String(required=False)
+    run_id = marshmallow.fields.String(
```
jonkeane, line 246: @alistaire47 You might be interested + please suggest additions or corrections here if anything looks wrong

ElenaHenderson (commented):
`record` method in runner.py needs to be updated to take `optional_info`

Once this is done, I will approve the PR.

conbench/entities/benchmark_result.py
alistaire47: Trying to understand what's supposed to go where here. 

Firstly, isn't `context` constant across many runs? Like `benchmark_language` and `compiler_flags` shouldn't change often. So `context_additional_info` doesn't make much sense to me exactly when we're talking about something that can change at the run level. Or am I wrong and is `context_additional_info` supposed to be as invariant as `context`? ...because AFAIK we aren't necessarily using it that way right now.

My understanding:

- `optional_info` is non-history-deciding metadata that can vary per run of a `case` for a given benchmark
- `info` (or whatever; called "Additional Info" in the website) should stay the same across all cases for all benchmarks in a run? —but that's weird, as you can also pass `info` when creating a run, and this (AFAIK) gets stored on `BenchmarkResult`

It seems like the missing level of would then be for info that's constant for all cases of a benchmark, but varies between benchmarks in a run. Maybe `Run.info` should be the one we show under `context` (which should really probably live on the run, too, in theory), and `BenchmarkResult.info` should be constant for all cases of a benchmark, and `optional_info` can vary by case? ...though I'm not sure that's practical given how we don't usually manually create `Run`s (for now).

All that said, I'm not sure how much of metadata we would have to populate each of these. Or are there actually any restrictions on `BenchmarkResult.info` that we couldn't overload the existing field to optionally provide multiple levels of info, nested (e.g. `"info": {"case": {...}, "benchmark": {...}, "run": {...}}`.

As a user, we need to clearly document what kind of metadata should get stuck where, because it's far from obvious from the schema.

alistaire47: I guess there are potentially four levels of info? `context_info`, `run_info`, `benchmark_info`, and `case_info`. Except where they live in the database would be a little weird given we're passing most of these in with a case result.

jonkeane: > BenchmarkResult.info that we couldn't overload the existing field to optionally provide multiple levels of info, nested (e.g. "info": {"case": {...}, "benchmark": {...}, "run": {...}}. 

I like this as a possibility, could you please make an issue for this + pursue it when you have an opportunity as a follow on? That could also be the time when we clean up some of the confusing + inconsistent naming. But I don't want to change too much here yet (since that will (likely) require a decently large migration). This would also be a great time to do a quick + simple design doc like you've suggested elsewhere about this kind of benchmark info.

> I guess there are potentially four levels of info? context_info, run_info, benchmark_info, and case_info. Except where they live in the database would be a little weird given we're passing most of these in with a case result.

Is basically what I've done here for now. It'll unblock us for now and give us a place to store both per-benchmark non-history altering information (`optional_info` here) and non-history altering information that is the same across benchmarks (and like you mention possibly runs) (`info` which gets displayed as "Additional information" in the UI today)

conbench/templates/benchmark-entity.html
```diff
@@ -104,6 +104,17 @@
               {% endif %}
             </li>
           {% endfor %}
+          {% if benchmark.optional_info %}
+            <li class="list-group-item active">Additional information</li>
```
ElenaHenderson: This section needs to be called something else.

We already have `Additional Information` for benchmark.info.

ElenaHenderson (commented):
3 more comments:
1. Version in `setup.py` needs to be bumped.
2. We already have `Additional Information` for `benchmark.info` in `conbench/templates/benchmark-entity.html`. We need different name.
3.  `conbench/tests/populate_local_conbench.py` should populate data for `benchmark.optional_info` so we can view it in UI locally.

conbench/templates/benchmark-entity.html
jonkeane: I was hoping that them being under context and benchmark tags to be enough to differentiate them, but you're right we should name them more expressively. I've changed the new one to "Additional benchmark details" and what used to be called "Additional Information" to "Additional context details"

jonkeane:
> `conbench/tests/populate_local_conbench.py` should populate data for benchmark.optional_info so we can view it in UI locally.

OOOPH this is what I was searching and searching for yesterday and _could not find_ thank you.

conbench/templates/benchmark-entity.html
jonkeane: Ok, I did and it looks fine! Thank you again for explicitly pointing me to the population script — I could not find that yesterday evening when I was struggling to put data there.

jonkeane:
This is ready for another review. I did end up changing the name from `optional_info` to `optional_benchmark_info` so that it's a bit clearer what's going on. We should (eventually) cleanup those names in the API and make it clearer to Ed's point — but not right now since we need this PR sooner than the deprecation cycle + all the updates + migration we would need for making those other names be more evocative + possibly be more well nested in the posting json.

> Version in setup.py needs to be bumped.

done

> We already have Additional Information for benchmark.info in conbench/templates/benchmark-entity.html. We need different name.

done, I ended up changing both so that it's clearer which is associated with the context and which is associated with the benchmark itself

> conbench/tests/populate_local_conbench.py should populate data for benchmark.optional_info so we can view it in UI locally.

done

conbench/entities/benchmark_result.py
```diff
@@ -35,6 +35,9 @@ class BenchmarkResult(Base, EntityMixin):
     data = Nullable(postgresql.ARRAY(s.Numeric), default=[])
     times = Nullable(postgresql.ARRAY(s.Numeric), default=[])
     case = relationship("Case", lazy="joined")
+    # optional info at the benchmark level (i.e. information that isn't a tag that should create a separate case, but information that's good to hold around like links to logs)
+    optional_benchmark_info = Nullable(postgresql.JSONB)
```
ElenaHenderson, line 39: Maybe we call this `execution_info`?
Just want to make sure we are happy with the name since renaming things in Postgres is a procedure.
Ignore me if you don't like this suggestion.

conbench/templates/benchmark-entity.html
```diff
@@ -197,20 +208,20 @@
...
+  {% if plot_history %}
+  $(document).ready(function () {
     Bokeh.embed.embed_item({{ plot_history | safe }});
 });
-{% endif %}
+  {% endif %}
 
 </script>
-{% endblock %}
+{% endblock %}
```
ElenaHenderson: Needs new line

conbench/tests/api/_expected_docs.py
```diff
@@ -1132,7 +1139,7 @@
                         "type": "object",
                     },
                     "error_type": {
-                        "description": "Run's error type. Possible values: none, catastrophic, partial. \n                    None = all attempted benchmarks are good. \n                    Catastrophic =no benchmarks completed successfully. \n                    Partial = some benchmarks completed, some failed",
+                        "description": "Run's error type. Possible values: none, catastrophic, partial.\n                    None = all attempted benchmarks are good.\n                    Catastrophic =no benchmarks completed successfully.\n                    Partial = some benchmarks completed, some failed",
```
ElenaHenderson, line 1167: I don't see descriptions for other fields. 
The way to update this file with new doc strings from marshmallow is to run tests locally (or just one test) which will update this expected doc.

ElenaHenderson (approved):
Approved. A few comments still but they should not block the merge.

conbench/templates/benchmark-entity.html
jonkeane: Ooph, frustrating vscode autosaving here! Thanks

conbench/entities/benchmark_result.py
jonkeane: IMO, `optional_benchmark_info` is a bit more descriptive because it might not be _only_ about the execution environment. It might also include details like tags but that shouldn't separate benchmarks from each other.

conbench/tests/api/_expected_docs.py
jonkeane: Ok, finally got all the pieces in place to run the tests *not* in docker so that I get the updated file along with it. I wonder if it would be possible to have the docker-based tests update the file too?

conbench/entities/benchmark_result.py
```diff
@@ -236,24 +243,65 @@ class BenchmarkResultSerializer:
...
+    run_reason = marshmallow.fields.String(
+        required=False,
+        metadata={
+            "description": "Reason for run (commit, pull request, manual, etc). This should be low cardinality. 'commit' is a special run_reason for commits on the default branch which are used for history"
+        },
+    )
     batch_id = marshmallow.fields.String(required=True)
-    timestamp = marshmallow.fields.DateTime(required=True)
+    timestamp = marshmallow.fields.DateTime(
+        required=True, metadata={"description": "Timestamp the benchmark ran"}
```
austin3dickey, line 264: started? completed? Does it matter?

austin3dickey (approved):
looks great!

conbench/entities/benchmark_result.py
jonkeane: I had the exact same debate, IMO it probably doesn't matter and can be left ambiguous or up to specific implementations to care about. But at least let's leave the docs ambiguous for now

jgehrcke: We removed some of that ambiguity. Related:
https://github.com/conbench/conbench/pull/651#discussion_r1092925503
https://github.com/conbench/conbench/blob/fbf5a6ce898532d44327e3ac961f5abb852843bc/conbench/entities/benchmark_result.py#L507

PR #415, opened 2022-10-27
Adapter for pytest-benchmark

alistaire47:
A rough adapter for [pytest-benchmark](https://github.com/ionelmc/pytest-benchmark/). It has some partially-implemented code to override `machine_info` and `github` with what's in the pytest-benchmark results, but that doesn't provide everything we want. If this is run from the repo and machine the benchmark is run on (i.e. if it's used for benchmarking instead of just uploading existing files), it will work. If it's used to upload existing files, we'll need to make sure everything is getting populated right.


jgehrcke:
Closing this for cleanup for now. We can always re-evaluate, re-open, go back to this!

PR #416, opened 2022-10-27
Accept partial successful benchmarks

### LINKED ISSUE #395: accept all data for a benchmark result even if one iteration is failed
lets say i did 3 iterations of a benchmark and one (or more) failed. Right now, conbench only supports a 'failed' state which means the data for the other iterations that didn't fail aren't available. the ask here is to accept and store all data and let the downstream viewer of said data decide how to interpret the failure state. This is especially important when some benchmark runs can take a long time (e.g., 30 mins per iteration)

- 0/3 failures: success
- 1/3 and 2/3 failures: partial failure, partial success
- 3/3 failures: total failure

jonkeane:
resolves #395 

This works for uploading so far, ~but I need to work more to confirm that things like comparisons will do the correct thing and dis-allow comparing a partial success to something that actually succeeded. Also need to confirm that partial success don't show up in history~ 

This should now work for comparisons as well. The way this is set up, if a partial result (e.g. if `stats.data` has any `None`s in it) and no error is reported, we will add in an error that says it's an incomplete iteration. Thus all of our history + comparisons that check if error exists will all _just work_.

I could imagine a situation where we might want to distinguish between a partial run and an errored run (without needing to inspect `stats.data`) where it might make sense to have a separate property like `completed_iterations: True|False`, but let's deal with that if|when it comes up.

### DIFF
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -194,22 +194,83 @@ s.Index("benchmark_result_context_id_index", BenchmarkResult.context_id)
 
 class BenchmarkResultCreate(marshmallow.Schema):
     data = marshmallow.fields.List(
-        marshmallow.fields.Decimal(allow_none=True), required=True
+        marshmallow.fields.Decimal(allow_none=True),
+        required=True,
+        metadata={
+            "description": "A list of benchmark results (e.g. durations, throughput). This will be used as the main + only metric for regression and improvement. The values should be ordered in the order the iterations were executed (the first element is the first iteration, the second element is the second iteration, etc.). If an iteration did not complete but others did and you want to send partial data, mark each iteration that didn't complete as `null`."
+        },
     )
     times = marshmallow.fields.List(
-        marshmallow.fields.Decimal(allow_none=True), required=True
+        marshmallow.fields.Decimal(allow_none=True),
+        required=True,
+        metadata={
+            "description": "A list of benchmark durations. If `data` is a duration measure, this should be a duplicate of that object. The values should be ordered in the order the iterations were executed (the first element is the first iteration, the second element is the second iteration, etc.). If an iteration did not complete but others did and you want to send partial data, mark each iteration that didn't complete as `null`."
+        },
     )
-    unit = marshmallow.fields.String(required=True)
-    time_unit = marshmallow.fields.String(required=True)
-    iterations = marshmallow.fields.Integer(required=True)
-    min = marshmallow.fields.Decimal(required=False)
-    max = marshmallow.fields.Decimal(required=False)
-    mean = marshmallow.fields.Decimal(required=False)
-    median = marshmallow.fields.Decimal(required=False)
-    stdev = marshmallow.fields.Decimal(required=False)
-    q1 = marshmallow.fields.Decimal(required=False)
-    q3 = marshmallow.fields.Decimal(required=False)
-    iqr = marshmallow.fields.Decimal(required=False)
+    unit = marshmallow.fields.String(
+        required=True,
+        metadata={"description": "The unit of the data object (e.g. seconds, B/s)"},
+    )
+    time_unit = marshmallow.fields.String(
+        required=True,
+        metadata={
+            "description": "The unit of the times object (e.g. seconds, nanoseconds)"
+        },
+    )
+    iterations = marshmallow.fields.Integer(
+        required=True,
+        metadata={
+            "description": "Number of iterations that were executed (should be the length of `data` and `times`)"
+        },
+    )
+    min = marshmallow.fields.Decimal(
+        required=False,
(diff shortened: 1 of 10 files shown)


conbench/entities/benchmark_result.py
```diff
@@ -61,38 +61,66 @@ class BenchmarkResult(Base, EntityMixin):
...
 
+            # calculate any missing stats if data available
+            if data["stats"].get("data"):
+                dat = [to_float(x) for x in benchmark_result_data["data"]]
+
+                # defaults
+                q1 = q3 = mean = median = min = max = stdev = iqr = None
+
+                # calculate stats only if the data is complete
+                complete_iterations = all([x is not None for x in dat])
```
austin3dickey, line 81: Do you think there's value in asserting that if there are Nones in `data`, there are corresponding Nones at the same indexes in `times`, and vice versa?

conbench/entities/benchmark_result.py
```diff
@@ -165,8 +193,12 @@ def create(data):
 
 
 class BenchmarkResultCreate(marshmallow.Schema):
-    data = marshmallow.fields.List(marshmallow.fields.Decimal, required=True)
-    times = marshmallow.fields.List(marshmallow.fields.Decimal, required=True)
+    data = marshmallow.fields.List(
+        marshmallow.fields.Decimal(allow_none=True), required=True
```
austin3dickey: Could you add a brief description of this behavior to the swagger docs?

conbench/tests/populate_local_conbench.py
```diff
@@ -226,6 +262,64 @@ def create_benchmarks_data():
     update_run_with_info(run_id, timestamp)
 
 
+def create_benchmarks_with_history():
```
austin3dickey, line 265: nice, this will be good for my PR

austin3dickey (approved):
Agreed that we can put off "distinguishing partial benchmarks from true errors" until later. It could be nice to treat them differently in alerts, but that's probably not required.

conbench/entities/benchmark_result.py
jonkeane: Yeah, ideally we would validate that those parallel each other — but for now (1) I don't think it actually matters anywhere, (2) there is at least one set of benchmarks that don't send times at all, only data, which would fail or we would need to special case, (3) really is important when we do multi-metrics and would need to be quite a bit different there, (4) it would be good to see this happen in real-life to confirm that our validation is what we want it to be.

For those reasons I think we should punt for now, but very much agreed that it's something to at least considered and made a conscious decisions to punt on for now.

conbench/entities/benchmark_result.py
jonkeane: done

conbench/tests/populate_local_conbench.py
jonkeane: Nice, yeah it took a bit to get it working (and the GH token is sneaky! It would be nice to not need that, though we would need to mock out all of the GH-reliant infrastructure which might defeat some of the purpose for using this as testing)

jonkeane:
Thanks for the comments — I've added those docs. Unless @ElenaHenderson objects, I plan to merge this today

austin3dickey:
Oh, don't forget to bump the version if you'd like to release now.

jonkeane:
> Oh, don't forget to bump the version if you'd like to release now.

Ah, of course will do

conbench/entities/benchmark_result.py
```diff
@@ -165,19 +193,84 @@ def create(data):
...
+        metadata={
+            "description": "The third quartile from `data`, will be calculdated on the server if not present (the preferred method), but can be overridden if sent. Will be marked `null` if any iterations are missing."
+        },
+    )
+    iqr = marshmallow.fields.Decimal(
+        required=False,
+        metadata={
+            "description": "The inter-quartile range from `data`, will be calculdated on the server if not present (the preferred method), but can be overridden if sent. Will be marked `null` if any iterations are missing."
+        },
+    )
```
jgehrcke, line 273: Only seeing now that you very recently added those descriptions @jonkeane -- thanks!

conbench/entities/benchmark_result.py
```diff
@@ -321,11 +414,6 @@ def validate_stats_or_error_field_is_present(self, data, **kwargs):
         if "stats" not in data and "error" not in data:
             raise marshmallow.ValidationError("Either stats or error field is required")
```
jgehrcke, line 415: Oh. Heavily relates to https://github.com/conbench/conbench/issues/813 :)

conbench/entities/benchmark_result.py
jgehrcke: related: https://github.com/conbench/conbench/issues/823

PR #417, opened 2022-10-28
Store branch information on Commits

austin3dickey:
This PR adds `branch` and `fork_point_sha` columns to the Commit table. These are populated when someone posts a new run from a GitHub-enabled repo. The `branch` is obtained on the client side, and the `fork_point_sha` is obtained on the server side by hitting the GitHub API.

This gets us further along on #382 but not all the way. Some things to do in future PRs in order to satisfy that story:

- auto-backfill commit entities for the default branch
- manual backfill branch names on existing commit entities
- change history-related queries across the codebase to use the branch names


conbench/templates/benchmark-entity.html
```diff
@@ -165,6 +165,22 @@
...
+      </li>
+      {% endif %}
+      {% if run.commit.fork_point_sha %}
+      <li class="list-group-item" style="overflow-y: auto;">
+        <b>forked from default at</b>
+        <div align="right" style="display:inline-block; float: right;">
+          {{ run.commit.fork_point_sha }}
+        </div>
+      </li>
+      {% endif %}
```
austin3dickey, line 184: I was back-and-forth on including the fork point SHA, but it was nice for testing so I left it in.

In the future when we auto-backfill all default-branch commits, each Commit entity should have a relationship to its fork-point entity, so it'll be easier to turn this SHA into a nice actual link to the commit.

image

conbench/templates/compare-entity.html
```diff
@@ -175,6 +175,22 @@
               </div>
             </li>
             {% endif %}
+            {% if baseline_run.commit.branch %}
```
austin3dickey, line 178: same comment as above

image

conbench/tests/entities/test_commit.py
```diff
@@ -62,27 +65,32 @@ def test_get_github_commit():
         "author_name": "Diana Clarke",
         "author_login": "dianaclarke",
         "author_avatar": "https://avatars.githubusercontent.com/u/878798?v=4",
+        "branch": branch,
+        "fork_point_sha": sha,
```
austin3dickey, line 69: this is the master branch so the fork point == the sha we care about

benchadapt/python/benchadapt/machine_info.py
```diff
@@ -73,9 +74,14 @@ def python_info():
...
     commit = _exec_command(["git", "rev-parse", "HEAD"])
     repository = _exec_command(["git", "remote", "get-url", "origin"])
+    branch = _exec_command(["git", "branch", "--show-current"])
+    remote = _exec_command(["git", "remote", "get-url", "origin"])
+    fork = re.search("(?<=.com[/:])([^/]*?)(?=/)", remote).group(0)
+
     return {
         "commit": commit,
         "repository": repository.rsplit(".git")[0],
+        "branch": f"{fork}:{branch}",
```
austin3dickey, line 116: As we talked [here](https://github.com/conbench/conbench/pull/369#discussion_r966385000) the goal is eventually to get rid of this duplicated file. But this works for now

conbench/tests/entities/test_commit.py
```diff
@@ -62,27 +65,32 @@ def test_get_github_commit():
...
     # NOTE: This integration test intentionally hits GitHub.
     # TODO: This test will fail once it's no longer one of the most recent 1000
     # commits to the apache/arrow repository.
     repo = "https://github.com/apache/arrow"
     sha = "982023150ccbb06a6f581f6797c017492485b58c"
+    branch = "dianaclarke:ARROW-13266"
     tz = dateutil.tz.tzutc()
     expected = {
-        "parent": "780e95c512d63bbea1e040af0eb44a0bf63c4d72",
+        "parent": "c8668f85a465ea05b2724ec47ff72c4db4d7dfe6",
```
austin3dickey, line 101: This was broken on main. By default we don't `--run-slow` so we weren't testing this.

Note that `c866` is the *true* parent of `9820`, which also happens to be on the same PR branch. `780e` is the fork point of this PR branch.

jonkeane: I haven't run the test myself to see if it literally doesn't run on main, but the behavior before _is_ the behavior I would expect on main, and exactly (one of) the behavior(s) we are correcting on this branch. The change is good IMO, but it might have actually been working as intended on main?

benchadapt/python/benchadapt/machine_info.py
```diff
@@ -73,9 +74,14 @@ def python_info():
 def github_info():
     commit = _exec_command(["git", "rev-parse", "HEAD"])
     repository = _exec_command(["git", "remote", "get-url", "origin"])
+    branch = _exec_command(["git", "branch", "--show-current"])
+    remote = _exec_command(["git", "remote", "get-url", "origin"])
```
jonkeane: This relies on the convention that `origin` is the upstream repo. This is generally true, but AFAIK it is just a convention, and it could be something else entirely.

benchadapt/python/benchadapt/machine_info.py
jonkeane: Could you expand more on this comment? We would be getting rid of this file in deference to what's run on the backend? Or the other way around?

conbench/entities/commit.py
```diff
@@ -21,6 +21,8 @@ class Commit(Base, EntityMixin):
     __tablename__ = "commit"
     id = NotNull(s.String(50), primary_key=True, default=generate_uuid)
     sha = NotNull(s.String(50))
+    branch = Nullable(s.String(100))  # github max chars: 255
```
jonkeane: Out of curiosity what's the 100 here? I assume it corresponds to 255 characters, but how?

conbench/entities/commit.py
```diff
@@ -62,10 +64,12 @@ def create_unknown_context(sha, repository):
         )
 
     @staticmethod
-    def create_github_context(sha, repository, github):
+    def create_github_context(sha, repository: str, github: dict):
```
jonkeane, line 67: Thanks for this cleanup 🙏

conbench/tests/entities/test_commit.py
austin3dickey: Yeah, I ran the test on main and it failed. So `get_github_commit()` has been returning the true git parent for some time. Looking at the code, that makes sense. 

https://github.com/conbench/conbench/blob/ef88d427dd6be11cdd95e17da4e9940927da37e5/conbench/entities/commit.py#L155-L158

Basically, if the true parent is in the last 1000 commits—which it always is!—it will return the true parent as the parent. So this whole time there has been a bug in the behavior we *didn't* want, resulting in the behavior we wanted.

However, to add to the confusion, when a user hits the API for a history, we execute [this other code](https://github.com/conbench/conbench/blob/ef88d427dd6be11cdd95e17da4e9940927da37e5/conbench/entities/distribution.py#L139) to "get the closest parent" which is actually looking for the fork point. So that code is probably what we were thinking is wrong.

conbench/entities/run.py
```diff
@@ -48,16 +48,17 @@ def create(data):
         )
         hardware = hardware_type.upsert(**data.pop(field_name))
 
-        sha, repository = None, None
+        sha, branch, repository = None, None, None
```
jonkeane: I know this doesn't actually matter, but LOL that's a funny git diff that looks very wrong.

conbench/templates/benchmark-entity.html
```diff
@@ -165,6 +165,22 @@
...
+        <div align="right" style="display:inline-block; float: right;">
+          {{ run.commit.branch }}
+        </div>
+      </li>
+      {% endif %}
+      {% if run.commit.fork_point_sha %}
+      <li class="list-group-item" style="overflow-y: auto;">
+        <b>forked from default at</b>
+        <div align="right" style="display:inline-block; float: right;">
+          {{ run.commit.fork_point_sha }}
```
jonkeane, line 181: We might not yet have enough info to do this (especially on past commits), but it would be nice if we could construct URLs to these + ideally also use short hashes for display

jonkeane: This is a bit of an extension of scope too, so if it's too much feel free to make an issue + defer it

conbench/templates/compare-entity.html
```diff
@@ -175,6 +175,22 @@
...
+              <div align="right" style="display:inline-block; float: right;">
+                {{ baseline_run.commit.branch }}
+              </div>
+            </li>
+            {% endif %}
+            {% if baseline_run.commit.fork_point_sha %}
+            <li class="list-group-item" style="overflow-y: auto;">
+              <b>forked from default at</b>
+              <div align="right" style="display:inline-block; float: right;">
+                {{ baseline_run.commit.fork_point_sha }}
```
jonkeane, line 190: Same thing here, re: hyperlinks + short hashes

conbench/templates/compare-entity.html
```diff
@@ -264,6 +280,22 @@
...
+              <div align="right" style="display:inline-block; float: right;">
+                {{ contender_run.commit.branch }}
+              </div>
+            </li>
+            {% endif %}
+            {% if contender_run.commit.fork_point_sha %}
+            <li class="list-group-item" style="overflow-y: auto;">
+              <b>forked from default at</b>
+              <div align="right" style="display:inline-block; float: right;">
+                {{ contender_run.commit.fork_point_sha }}
```
jonkeane, line 295: and one last place

jonkeane (commented):
Thanks for this! It's looking really great

conbench/tests/entities/test_commit.py
jonkeane: Aaaah I see. I'm very glad that we're cleaning up this confusing behavior since it apparently fooled me (and likely the original authors too!) here

benchadapt/python/benchadapt/machine_info.py
austin3dickey: oh right I got it backwards - what I meant is that ideally this logic only lives in the adapter. This PR touches both `conbench/machine_info.py` and `benchadapt/python/benchadapt/machine_info.py`. But we'll be removing `conbench/machine_info.py` eventually.

conbench/tests/entities/test_commit.py
austin3dickey: Me too!! (and I'll actually clean that code up in a subsequent PR; I'm trying to keep this PR to just "add these two fields to the Commit table")

benchadapt/python/benchadapt/machine_info.py
austin3dickey: awesome let me test if there's a more generic way to do this!

conbench/entities/commit.py
austin3dickey: @alistaire47 wrote this—Ed, do you have a reason for the 100 char limit?

conbench/entities/commit.py
austin3dickey: cred to @alistaire47 !

conbench/entities/run.py
austin3dickey: 🤣

conbench/templates/benchmark-entity.html
austin3dickey: Yep see [this](https://github.com/conbench/conbench/pull/417#discussion_r1009559148) comment. I want to link to the commit like we link to the target commit, but I need to set up the backend in a future PR first

conbench/entities/commit.py
alistaire47: I spent a little time trying to figure out how many characters a branch name can be, but it seems like it's system-dependent. Looking more, [it sounds like the _de facto_ GitHub limit is 244 characters](https://stackoverflow.com/questions/24014361/max-length-of-git-branch-name#comment122975637_24014513), though users may run into restrictions before that—various people [try to keep it much shorter](https://medium.com/leantaas-engineering/why-are-we-limiting-git-branch-name-length-to-28-characters-c49cb5f4ff9a).

In practice, I doubt we'll see much over 50, but 244 or 255 are reasonable values because of GitHub. Note that all of that is just the branch name, though; we moved to `{user}:{branch}` to handle forks, so somebody could conceivably have an exceedingly long username as well and blow it out; I'm not sure what GitHub's limit there is.

conbench/entities/commit.py
jonkeane: Thanks @alistaire47 🙏

benchadapt/python/benchadapt/machine_info.py
jonkeane: If I'm reading the backend code correctly that you have here — that grabs the remote from the GitHub API. Maybe we only use that approach (for now?) and then we're immune from how git might be configured wherever this is running

benchadapt/python/benchadapt/machine_info.py
jonkeane: > ideally this logic only lives in the adapter

Is that the ideal state that we want? I definitely agree that having it in one place and not two is good — but I'm not totally sure that the adapter is the right place for this compared to the backend. 

One nice thing about keeping this in the backend is that we have a bit more control over the details of the implementation (and since our comparison flow is also in the backend we can keep them more closely aligned — or change them without needing to go through both the adapter and the backend). It also keeps this logic slightly closer (at least in the same component) to the places that will subsequently use that information (e.g. comparisons). 

What are the pros/cons of putting this in the adapter versus the backend? We also can punt on the specifics decision for now and document some of these questions in an issue and keep the code we have now in both places

conbench/entities/commit.py
jonkeane: I'm getting flashbacks to a prior job where we had a whole string of "surely someone won't make a variable name that's longer than 50 characters!" 😱 

I wonder if 255*2 would be a reasonable enough limit that we won't need to go back and fix it later

benchadapt/python/benchadapt/machine_info.py
austin3dickey: By "backend" you mean the Conbench server, right? Unfortunately I think it's impossible to figure out the GitHub repo name, fork name, branch name, and commit SHA from the server side. Even if the client provided the repo and SHA, it would be fairly expensive or impossible in some cases to figure out the fork and branch using the GitHub API. The client has cheap access to all this information, which is why we've always been figuring it out on the client side.

jonkeane: > By "backend" you mean the Conbench server, right?

Yeah, exactly.

But if the client provides the pull number all of that info is readily available from the pulls endpoint, no? For example: https://api.github.com/repos/apache/arrow/pulls/14560 has `head.label` for fork+branch name (or `head.repo.full_name` + `head.ref` for them separately), as well as the fork_point_sha in `base.sha`. Is that endpoint missing anything that we need?

Since these runs are fundamentally orchestrated from pullrequests, having the PR number should be trivial during the orchestration to run them (I assume we'll still need to send the SHA there too...). And all of the feedback about the outcome of the run will be PR-based anyway with benchalerts, so we'll need the PR number there anyway.

We might _some day_ need to support random-branch benchmarks that aren't associated with PRs, but we absolutely do not need to support them today. And if that lets us avoid things like figuring out how to reliably parse remotes in CI or whatever, IMO that's a win.

alistaire47: I thought we had agreed to go with branch because PR# would have to be manually specified by the user and inserted (i.e. you've got to munge results a lot), whereas branch can be detected

jonkeane: I don't remember that we decided on that level of detail — but do remember distinctly talking about how "PR# would have to be manually specified by the user" isn't quite right for our current use case, like I said above: the current use should always have a PR number easily accessible, and if that makes all the info easy to get from that endpoint we should take advantage of that.

jonkeane: But, regardless of if we did or didn't decide to use one end point or the other: neither of those are themselves requirements — what we care about is the external behavior and then what it takes code-wise to produce that behavior. And my point in this comment is: let's weight the pros and cons of each of these end points and decide which one will make a better interface. Even if we decided in the past to use one, that might not make it the right one now or in the future.

alistaire47: My interest here is to minimize configuration. PR# is known in GHA, so it's another piece of metadata like `run_reason` that needs to be passed through to the results in every implementation. Branch we can automate once for everyone and be done with, and they can override it if they want.

If we care about the PR# we can do it; I'd just rather save some work if we end up with the same thing.

jonkeane: We also don't have an absolute guarantee that when benchmarks are being run, they have a git checkout that can have `git` operations made against (I know of at least one conbench setup that uses a pre-built binary in the benchmark phase, so doesn't have the git repo available). But IIUC benchalerts needs the PR number to operate anyway, so that seems like the lowest-common denominator (additionally, describing the requirement of what to seen is pretty easy to communicate to others implementing this stuff "you need the github pullrequest number" versus "you need the upstream branch + fork formatted in this specific way, you also need to get the fork-merge-sha which is...")

jonkeane: We also don't need to decide this here and now – this PR can go with both of these in place until we decide on which location really is the right on for getting this information in the modal case.

alistaire47: Yeah, if it could be an either `branch` or `pr` needs to be specified (along the lines of `stats`/`error` and `machine_info`/`cluster_info`, that's viable. It's a tad more work to handle and validate, but since we've already got similar logic in the schema it's probably alright

jonkeane: > Yeah, if it could be an either branch or pr needs to be specified (along the lines of stats/error and machine_info/cluster_info, that's viable

Yup, also an option, and that has the benefit of letting us see how they work + which is easier to work with (with the possible downside of being more complex like you mention or having someone somewhere depend on one that we want to deprecate later, but in the abstract at least, those two aren't so bad — we might get into the details and realize it's waaaay more work to maintain both, but I don't think it will be from here)

austin3dickey: You two are commenting too fast for me to keep up :)

So it sounds like with either option, the client must scrape the repo name and commit sha from a local cloned repo. In order to get the repo name, we have to figure out which remote the current branch is tracking (currently on main this is [hardcoded](https://github.com/conbench/conbench/blob/564bc7c0498e06a4ca260e8ce050304a506c06c7/conbench/machine_info.py#L83) to `origin`) and get that remote's github URL. This is doable without assuming anything about the git setup. (Though, note that we've never run into problems with `origin` before.)

If there isn't a local cloned repo, it's up to the runner to supply those two values manually.

The two options define how we get the fork name and branch name. Option 1 is the client collects the fork and branch from the cloned repo in a similar way to collecting the repo and sha. Option 2 is the client provides the PR number and the server figures out the fork and branch from hitting the API.

Pros of 1: in the above process of collecting the repo name and commit sha, we already have the branch and fork handy.
Cons of 1: if there isn't a local cloned repo, the runner configurer must know how to supply the fork and branch.

Pros of 2: the GH PR number should be easily known by the runner configurer. 
Cons of 2: another runner configuration variable. Does not fully support running benchmarks outside of PR builds, or at least we'd need to handle those with more custom logic, including merge-commit and nightly runs. More server-side GitHub API requests.

(Somewhat unrelatedly: right now, `benchalerts` doesn't need to know anything about PR numbers. It just posts alerts on GitHub commits. Maybe someday in the future if we set up alerts to other services, it will need to figure out the PR number. But it could do that by hitting GitHub, not necessarily conbench.)

jonkeane: Personally, I think "So it sounds like with either option, the client must scrape the repo name and commit sha from a local cloned repo" is even a bit too prescriptive even, the repo name is effectively static, so can be in an env var (or dot-conbench file), but sure needs to be set. As for the commit sha: that could come from a local git clone or it could come from asking the software-under-benchmark for it [like we do with pyarrow right now](https://github.com/voltrondata-labs/benchmarks/blob/b1aae88bcc1ca3f9163e2c229d7fafa453b39107/benchmarks/_benchmark.py#L38). IMO the second is a better option if we can do it that way since it guarantees the software being benchmarked is indeed at the revision one is expecting it to be — though there of course will be cases where software isn't configured to provide info like that.

I suspect that option 2 will serve us better in the future because we (as the backend) take on the onus of getting the details of that right at a point when we can relatively easily as opposed to requiring runners or adapters (or folks writing their own versions of both) always to have to get that exactly right.

austin3dickey: Got it, so what I'm hearing now (and I suspect what you've been saying the whole time) is that when it comes to supplying github context, we should prefer more user-supplied configuration over scraping it from the machine setup, as long as the configuration is straightforward.

So with that design principle, here's my suggestion:
1. [optional] We remove the assumption that one conbench server can track multiple repos, and set `GITHUB_REPO_NAME="org/repo"` in the server config file.
2. The recommended way to provide GitHub info is to provide the commit sha and GitHub PR number when the runner POSTs a new run.
   - If we don't do (1), the runner must also provide the repo name in exactly that `org/repo` format.
   - If it's a merge-commit run, the runner provides `"github_pr_number": null` and we'll have to assume the commit sha is on the default branch.
   - We won't be able to handle runs on "special" non-default-branches (like commits to a `dev` branch) unless we add a third configuration flag. When we do want to add support for special branches, we'll probably have to do another overhaul PR with a migration, since `fork_point_sha` might not be as useful anymore.
3. We store the `fork:branch` string, and the fork point sha from the default branch, in the commit table.
4. If a runner is running benchmarks on a locally cloned repo, we could scrape all the necessary information from that repo and skip (2).

In this PR, I won't implement (1) yet. If we skip (4), it will be a breaking change for runners so we'll have to be extra careful about releasing.

jonkeane: Yes, yes I like this. At one step further, it follows a principle that I haven't written down anywhere explicitly (I should) of trying to have the right thing filled in, but if someone really wants or knows what they are doing they can over-ride by sending the information for fields like these. Ideally no one will take the second path (because the first is easier + most likely to be right) but the option is there (which also allows us to over-ride it if we need to before we can get a backend fix in to enable something, for example).

A few comments about pieces: 

For (1) let's see if how helpful it is to make it so any one server has one and only one repo. One place we expected to use this (but haven't yet) is when benchmarking the arrow-rust implementation which is in `apache/arrow-rs`. We don't have to absolutely stick to this if it's super painful to keep sticking to it, but also if it doesn't help much to get rid of this feature, maybe we should keep it around? I'm very on the fence here

> If it's a merge-commit run, the runner provides "github_pr_number": null and we'll have to assume the commit sha is on the default branch.

This sounds right for (almost?) all cases right now.

> We won't be able to handle runs on "special" non-default-branches (like commits to a dev branch) unless we add a third configuration flag. When we do want to add support for special branches, we'll probably have to do another overhaul PR with a migration, since fork_point_sha might not be as useful anymore.

Would it be possible to set this up such that if someone supplies a branch element, we use that? This way someone _could_ start sending random branch names, though the comparison or history might not work with those until we overhaul? But in general we tell folks not to send branch name, and we get the behavior you describe here. If supporting both now is too much, that's totally fine, of course we can punt on being able to send an over-riding branch name for now and only accept commits to the default branch (which is effectively what the current system is restricted to anyway!)

> If a runner is running benchmarks on a locally cloned repo, we could scrape all the necessary information from that repo and skip (2).

Yeah, and this we should also mark as advanced mode + not recommended unless you are very very clear on the details and what it means.

alistaire47: > 4\. If a runner is running benchmarks on a locally cloned repo, we could scrape all the necessary information from that repo and skip (2).

You can't scrape PR# afaik, you have to get it from CI. 

> * If it's a merge-commit run, the runner provides `"github_pr_number": null` and we'll have to assume the commit sha is on the default branch.

This is a little dangerous; that info is probably going to be coming through an env var in CI, and if it's set wrong or not set, it will be null even if it shouldn't be.

austin3dickey: Re: (1), it sounds like we'd really like to support multiple Arrow subprojects on the Arrow conbench (at least), so we can keep all the repo handling around. We'll just have to have runners provide the repo name when posting runs. (unless they're using the auto-scrape method)

>You can't scrape PR# afaik, you have to get it from CI.

Right sorry, I meant that we'd be able to auto-scrape repo, fork, branch, and commit from the local repo, so we don't need the PR number in that case. We're not storing PR number either way, just storing `fork:branch`.

>if someone supplies a branch element, we use that

Yeah this is probably the best way to do special branch handling. I think I'll have to include this behavior anyway in order to get the auto-scrape method to work.

>`"github_pr_number": null` is a little dangerous

True as always. Luckily if a commit is supplied and `github_pr_number` is null, the server will (soon) be doing the default-branch backfill process, and we can validate the new commit is on the default branch during that process.

jonkeane: > > "github_pr_number": null is a little dangerous

> True as always. Luckily if a commit is supplied and github_pr_number is null, the server will (soon) be doing the default-branch backfill process, and we can validate the new commit is on the default branch during that process.

Yeah, and having a nice message there that says "Hey, you sent me a commit but that commit is nowhere on the default branch, something is wrong, are you trying to send a PR branch? If so..." will hopefully reduce confusion and some of the magic that we have now that makes it even more confusing...

benchadapt/python/benchadapt/machine_info.py
```diff
@@ -71,11 +73,48 @@ def python_info():
...
+        warnings.warn(
+            "error in github_info(): probably not in a git repo; returning {}"
+        )
+        return {}
+
+    branches = _exec_command(["git", "branch", "-vv"])
+    if "* (HEAD detached" in branches:
+        remote = "origin"
+        branch = "<DETATCHED BRANCH>"
+        warnings.warn(f"error in github_info(): detached HEAD; returning {branch=}")
```
austin3dickey, line 95: This is the case for most build systems (they run on a detached HEAD that's the hypothetical merge commit) so that's another reason to recommend that most systems should just supply the PR number.

benchadapt/python/benchadapt/machine_info.py
austin3dickey: Okay, this is implemented. If you all have more comments on the implementation let's start a new thread :)

conbench/entities/commit.py
```diff
@@ -197,34 +191,17 @@ def __init__(self):
...
-        # Grabs the last 1000 commits to the default branch. TODO: If the pull
-        # request is old, the parent may not be in the last 1000 commits.
-        branch = self.get_default_branch(name)
-        url = f"{GITHUB}/repos/{name}/commits?sha={branch}&per_page=100"
-        response = self._get_response(url)
-        if response:
-            commits = self._parse_commits(response)
-            if sha in commits:
-                return commits
-            for page in range(2, 11):
```
austin3dickey, line 219: Removing this makes conbench go so much faster, haha

austin3dickey:
Okay this is ready for another review.

benchadapt/python/benchadapt/machine_info.py
jonkeane: Ha, indeed! I forgot about this too

conbench/entities/commit.py
jonkeane: But surely pinging a giant API over and over isn't _that_ slow 😂 😭

conbench/entities/commit.py
```diff
@@ -234,6 +211,41 @@ def get_commit(self, name, sha):
             response = self._get_response(url)
         return self._parse_commit(response) if response else None
 
+    def get_fork_point_sha(self, name: str, sha: str) -> str:
+        """
+        Get the most common ancestor commit between an arbitrary SHA and the default
+        branch.
+
+        Returns ``None`` if sha is not supplied or if GitHub can't find it, otherwise
+        returns the fork point sha, called the "merge base" in git-speak.
```
jonkeane, line 223: Love the detail here

conbench/tests/entities/test_commit.py
```diff
@@ -40,15 +40,28 @@ def test_repository_to_url():
...
+    assert get_github_commit(None, None, None, sha) == {}
 
 
 @pytest.mark.slow
-def test_get_github_commit():
+@pytest.mark.parametrize(
+    "branch",
+    [
+        # as if supplied manually
+        None,
```
jonkeane: I'm not sure I follow the comment here being supplied manually? Wouldn't `None` be like not supplying it?

conbench/tests/entities/test_commit.py
```diff
@@ -62,27 +75,43 @@ def test_get_github_commit():
...
+    [
+        # as if supplied manually
+        (None, 10665),
+        # like github_info() called in forked PR
+        ("dianaclarke:ARROW-13266", None),
+    ],
+)
+def test_get_github_commit_and_fork_point_sha_pull_request(branch, pr_number):
     # NOTE: This integration test intentionally hits GitHub.
     # TODO: This test will fail once it's no longer one of the most recent 1000
```
jonkeane: This TODO is no longer relevant, right?

conbench/entities/run.py
```diff
@@ -210,6 +214,8 @@ def commit_hardware_run_map():
 class GitHubCreate(marshmallow.Schema):
     commit = marshmallow.fields.String(required=True)
     repository = marshmallow.fields.String(required=True)
+    branch = marshmallow.fields.String(required=False, allow_none=True)
+    pr_number = marshmallow.fields.String(required=False, allow_none=True)
```
jonkeane: I know we don't have docs here already, but would you mind either adding them now or making an issue to do so? That would also give us the opportunity to heavily encourage folks to use pr_number only unless they really know what they are doing with the branch

jonkeane (approved):
This is looking great. A few comments, mostly minor

conbench/tests/entities/test_commit.py
austin3dickey: Yeah I can reword this comment. Basically, our recommended way to provide github info is to manually provide the `commit`, `repository`, and optionally `pr_number`. If you go the recommended way, you'll never provide `branch`.

If you go the "advanced" way (scraping with `github_info()`), then `branch` will be the auto-scraped branch string, which is the next test case here.

conbench/entities/commit.py
austin3dickey: thanks - more cred to @alistaire47

conbench/tests/entities/test_commit.py
austin3dickey: good point thanks!

conbench/tests/entities/test_commit.py
jonkeane: That's great, thanks!

conbench/entities/run.py
austin3dickey: Sure! I'm all for more docs. 

Unfortunately these particular docs hide way at the bottom of the swagger page, and I'm assuming no one clicks on them. (scroll down [here](https://conbench.ursa.dev/api/docs/) and click on `BenchmarkCreate` for an example.) So that's why I didn't include them before. Plus hopefully docs for adapters/runners will detail this more. But I can add some advice here just in case!

jonkeane: Yeah, that aspect of the swagger docs is decidedly not great 😞 but having soooome guidance is good and maybe once we've got some more progress we can figure out if we can make swagger better or use something else... But that is *definitely* scope creep for this PR 😂

benchadapt/python/benchadapt/machine_info.py
alistaire47: They'll usually make git info available in env vars though (e.g. [jenkins](https://plugins.jenkins.io/git/#plugin-content-branch-variables)). I wonder if the particular env vars are shared enough across build systems such that we could just check if they're populated? Or make our own as a way for users to pass us the info.

conbench/entities/run.py
alistaire47: They're also available inside each endpoint section if you click over from "Example value" to "Schema". And I, at least, look at these a lot when I'm trying to understand an API (n = 1)

benchadapt/python/benchadapt/machine_info.py
austin3dickey: What I mean is it's the same amount of work—if not more—to have runners get the branch env var and format it `org:branch` than it is to have runners get the PR number env var. So we're actually not saving any configuration if we recommend the "autoscrape from git" method.

It is interesting to think about autoscraping this stuff from the environment, but yeah we'll probably need to deal with at least jenkins, github, buildkite, and circleci, if not more. I think they mostly have different env var names. So that could eventually be more of a pain than it sounds

conbench/entities/commit.py
austin3dickey: I ended up removing `@pytest.mark.slow` from these tests because now they're faster than some other tests, haha

conbench/entities/run.py
austin3dickey: Oh cool!!!!!

benchadapt/python/benchadapt/machine_info.py
alistaire47: Yeah, that would be a lot of work unless everybody already uses the same env var names. 

It may still be worth it to make our own env vars to check, though—otherwise most implementations will need to make them themselves anyway and pass them through to the adapter or whatever is making this JSON.

austin3dickey: Yeah that's a great point. I'll defer that to the adapter work.

PR #418, opened 2022-10-28
Print instead of log what's posted

alistaire47:
Print to stdout so as not to worry about logging across packages. Should solve this better long-term, but for now we need to debug

### DIFF
--- a/benchadapt/python/benchadapt/adapters/_adapter.py
+++ b/benchadapt/python/benchadapt/adapters/_adapter.py
@@ -1,4 +1,5 @@
 import abc
+import datetime
 import subprocess
 import uuid
 from typing import Any, Dict, List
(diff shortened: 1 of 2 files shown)


ElenaHenderson (approved):

benchadapt/python/benchadapt/client.py
```diff
@@ -48,8 +48,17 @@ def get(self, path: str) -> dict:
...
+
+        # log.debug(f"POST {url} {dumps(json)}")
+        print(f"POST {url} {dumps(json)}")
+
+        try:
+            res = self.session.post(url=url, json=json, timeout=self.timeout_s)
+        except requests.exceptions.ReadTimeout as e:
+            print(f"POST timed out: {e.response.content.decode()}. Retrying...")
+            print(f"POST {url} {dumps(json)}")
+            res = self.session.post(url=url, json=json, timeout=self.timeout_s)
```
ElenaHenderson: Can line 60 please be in another try except block so if it fails, we will continue posting other results in the loop?

We should still raise an error though if the any 2nd try failed but after the loop is done.

alistaire47: K, moved all this to the posting loop so we can raise after

ElenaHenderson (approved):
Thank you!

benchadapt/python/benchadapt/adapters/_adapter.py
```diff
@@ -162,10 +164,23 @@ def post_results(self) -> list:
...
         res_list = []
+        error = None
         for result in self.results:
             result_dict = result.to_publishable_dict()
-            res = client.post(path="/benchmarks/", json=result_dict)
+
+            try:
+                res = client.post(path="/benchmarks/", json=result_dict)
+            except requests.exceptions.ReadTimeout as e:
+                print(f"POST timed out: {e.response.content.decode()}. Retrying...")
```
ElenaHenderson: There will be no response when request times out.

ElenaHenderson: You can print `e` and time it happens so it can be tracked in the conbench logs.

alistaire47: Wait if i use `warnings.warn()` will those end up in the logs? That would be less gross than the `print()`s

ElenaHenderson (approved):

benchadapt/python/benchadapt/adapters/_adapter.py
ElenaHenderson: I really don't know and I never know unless I setup logging for the app myself :)

PR #420, opened 2022-10-31
Fix what bokeh 3.0.0 broke

austin3dickey:
This morning, [bokeh 3.0.0](https://github.com/bokeh/bokeh/wiki/Migration-Guides#300) came out and brokeh our plots, which also brokeh our tests. This PR migrates some of the plotting code to use the new 3.0-approved style.

### DIFF
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -60,16 +60,16 @@ def get_display_unit(unit):
 def get_date_format():
     date_format = "%Y-%m-%d"
     return bokeh.models.DatetimeTickFormatter(
-        microseconds=[date_format],
-        milliseconds=[date_format],
-        seconds=[date_format],
-        minsec=[date_format],
-        minutes=[date_format],
-        hourmin=[date_format],
-        hours=[date_format],
-        days=[date_format],
-        months=[date_format],
-        years=[date_format],
+        microseconds=date_format,
+        milliseconds=date_format,
+        seconds=date_format,
+        minsec=date_format,
+        minutes=date_format,
+        hourmin=date_format,
+        hours=date_format,
+        days=date_format,
+        months=date_format,
+        years=date_format,
     )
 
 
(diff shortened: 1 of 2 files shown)


requirements-build.txt
```diff
@@ -1,7 +1,7 @@
 alembic
 apispec
 apispec-webframeworks
-bokeh
+bokeh>=3.0.0
```
austin3dickey, line 4: I think this PR's changes would still work pre-3.0, but it'll be nice to floor it anyway.

austin3dickey:
I still haven't gotten interactive plots to work with the `populate_local_conbench.py` data on my local machine, but those weren't working before either. I'll see if I can mess with it so we know for sure that plots work.

conbench/app/_plots.py
```diff
@@ -236,17 +226,28 @@ def time_series_plot(history, benchmark, run, height=250, width=1000):
...
+        bokeh.models.HoverTool(
+            tooltips=[
+                ("date", "$x{%F}"),
+                ("mean", "@means"),
+                ("commit", "@commits"),
+            ],
+            formatters={"$x": "datetime"},
+            renderers=[hist_line, bench_circle],
+        )
+    )
```
jonkeane, line 250: I'm not going to suggest we do this here, but _at some point_ we should really update these and make them a bit more detailed — I also wonder if it's possible to have hyperlinks here too — I almost always find myself copying info from this onto a URL and doing something when I use them

requirements-build.txt
jonkeane: We'll probably also need to make sure that condaforge picks this up too (cf https://github.com/conda-forge/conbench-feedstock/pull/8)

jonkeane (approved):
Looks good, one comment about a future thing we should improve (def not in this PR) and one about what we'll (probably) need to do downstream before we start deploying this

conbench/app/_plots.py
austin3dickey: Yeah, could be fun. [Looks like](https://discourse.bokeh.org/t/open-link-on-click/99/6) we'd have to delve into a bit of custom javascript to do that. There's not an easy way to do hyperlinks, especially because the hover tool always avoids the cursor

austin3dickey: #421

requirements-build.txt
austin3dickey: I actually just removed the floor. I did a bunch of manual testing and these changes work with really old `bokeh` versions. Also, `bokeh` isn't in that feedstock at all. So I think we can safely skip that step this time

jonkeane: Ah, right, the feedstock is really targetted at distributing the CLI, so that makes sense

austin3dickey:
>I still haven't gotten interactive plots to work with the populate_local_conbench.py data on my local machine

I got it to work! This PR shouldn't break anything

Issue #368, opened 2022-09-06
Runners? Clients? Adapters?


jonkeane:
As we work on modularizing the conbench experience, we've been looking into separating the different modules that conbench uses and needs. We've talked about a number of different utilities in different places, but it's crystalizing in my head a bit that having (better) names for some of these things might help convey and structure our thinking more.

# Conbench ~connector~ adapter
This might be made up of smaller modules that are stored separately but all together, this is the piece of machinary that we can tell people "if you've got a runner, wrap this around it, and you'll get to send stuff to conbench".

Minimally it'll need to: 
1. be able to call a runner of some sort
2. grab the results from the runner (and possibly in some circumstances munge those results to conbench format)
3. add run information (like run name, run reason, etc.)
4. upload those results to conbench

2 and 3 both might need to be extensible such that an implementation could depend on this and then add their own bits of info in there if needed. 

Generally we shouldn't need many of these (and ideally just one). And generally speaking this should be effectively a requirement to use conbench (you _could_ reimplement all of this on your own, but it's a lot of work for not much benefit)

# Benchmark runners
These are the pieces of code that actually run the benchmarks themselves, and output the results. We will likely need one of these for each language (or we should adopt a language's framework like google bench or folly and use that).

These should be abstract enough to be implemented by a bunch of different benchmark suites below, we might even provide template versions of these, though some might combine the benchmark runner with the suite itself if that makes running the benchmarks easier. So they are more optional than the conbench connector above. Using the python | R | C++ | whatever template versions we have will be the easiest way to get going, so long as your runner can be called by the conbench connector, and it spits out results that are either already in conbench's format, or you have extended the connector to munge the results, you're good to go.

# The benchmarks suites themselves
This is the code that is actually tested, the valid cases, setup, etc. This code will always be custom to the things that are being benchmarked, and should be as easily accessible to the developers of the things being benchmarked as possible.

jonkeane:
Also I will say I'm particularly ➖ about the "benchmark suites" name there, since we use suite elsewhere, but needed _something_ to talk about the code that's actually being benchmarked.

jonkeane:
[...]

PR #422, opened 2022-10-31
benchconnect: minimal CLI for posting to conbench

alistaire47:
This is an interpretation of https://github.com/conbench/conbench/issues/368#issuecomment-1297215937 that builds a minimal CLI that currently can post benchmark results from an inline JSON dict, a JSON file, or a dir of JSON files.

This is deliberately minimal to elicit comment about

1. what about the behavior/interface do we like?
2. what do we want to change?
3. what functionality is missing that we want/need?

Obviously the other side that calls this needs to exist, too! But that would live elsewhere.


alistaire47:
What it looks like: 

```bash
$ benchconnect post --json '{"run_id": "1", "run_reason": "test", "error": {"stack_trace": "test"}, "tags": {"name": "benchconnect"}}'
POST http://127.0.0.1:5000/api/login/ {"email": "***", "password": "***"}
POST http://127.0.0.1:5000/api/benchmarks/ {"run_name": "test: 79ccf7169e142da8bcd65a91e805ba4ff35ddba7", "run_id": "1", "batch_id": "6f04660c12b3422286339482b08cddb6", "run_reason": "test", "timestamp": "2022-10-31T21:13:35.250207+00:00", "tags": {"name": "benchconnect"}, "info": {}, "machine_info": {"name": "snork.local", "os_name": "macOS", "os_version": "12.6", "architecture_name": "arm64", "kernel_name": "21.6.0", "memory_bytes": "17179869184", "cpu_model_name": "Apple M1 Pro", "cpu_core_count": "10", "cpu_thread_count": "10", "cpu_l1d_cache_bytes": "65536", "cpu_l1i_cache_bytes": "131072", "cpu_l2_cache_bytes": "4194304", "cpu_l3_cache_bytes": "0", "cpu_frequency_max_hz": "0", "gpu_count": "0", "gpu_product_names": []}, "context": {}, "github": {"commit": "79ccf7169e142da8bcd65a91e805ba4ff35ddba7", "repository": "git@github.com:conbench/conbench"}, "error": {"stack_trace": "test"}}
$
```

or from a file or dir:

```bash
$ mkdir /tmp/results
$ echo '{"run_id": "2", "run_reason": "test", "error": {"stack_trace": "test"}, "tags": {"name": "benchconnect"}}' > /tmp/results/result-1.json
$ echo '{"run_id": "2", "run_reason": "test", "stats": {"data": [1, 2], "unit": "s", "iterations": 2, "times": [], "time_unit": "s"}, "tags": {"name": "benchconnect"}}' > /tmp/results/result-2.json
$
$ benchconnect post --path /tmp/results/result-2.json
POST http://127.0.0.1:5000/api/login/ {"email": "***", "password": "***"}
POST http://127.0.0.1:5000/api/benchmarks/ {"run_name": "test: 79ccf7169e142da8bcd65a91e805ba4ff35ddba7", "run_id": "2", "batch_id": "7c0ba0f7d4b941a6b4518f5be37e35eb", "run_reason": "test", "timestamp": "2022-10-31T21:21:32.079849+00:00", "stats": {"data": [1, 2], "unit": "s", "iterations": 2, "times": [], "time_unit": "s"}, "tags": {"name": "benchconnect"}, "info": {}, "machine_info": {"name": "snork.local", "os_name": "macOS", "os_version": "12.6", "architecture_name": "arm64", "kernel_name": "21.6.0", "memory_bytes": "17179869184", "cpu_model_name": "Apple M1 Pro", "cpu_core_count": "10", "cpu_thread_count": "10", "cpu_l1d_cache_bytes": "65536", "cpu_l1i_cache_bytes": "131072", "cpu_l2_cache_bytes": "4194304", "cpu_l3_cache_bytes": "0", "cpu_frequency_max_hz": "0", "gpu_count": "0", "gpu_product_names": []}, "context": {}, "github": {"commit": "79ccf7169e142da8bcd65a91e805ba4ff35ddba7", "repository": "git@github.com:conbench/conbench"}}
$
$ benchconnect post --path /tmp/results
POST http://127.0.0.1:5000/api/login/ {"email": "***", "password": "***"}
POST http://127.0.0.1:5000/api/benchmarks/ {"run_name": "test: 79ccf7169e142da8bcd65a91e805ba4ff35ddba7", "run_id": "2", "batch_id": "66f6aa5cf95344208eb2280568bee06c", "run_reason": "test", "timestamp": "2022-10-31T21:25:12.618944+00:00", "stats": {"data": [1, 2], "unit": "s", "iterations": 2, "times": [], "time_unit": "s"}, "tags": {"name": "benchconnect"}, "info": {}, "machine_info": {"name": "snork.local", "os_name": "macOS", "os_version": "12.6", "architecture_name": "arm64", "kernel_name": "21.6.0", "memory_bytes": "17179869184", "cpu_model_name": "Apple M1 Pro", "cpu_core_count": "10", "cpu_thread_count": "10", "cpu_l1d_cache_bytes": "65536", "cpu_l1i_cache_bytes": "131072", "cpu_l2_cache_bytes": "4194304", "cpu_l3_cache_bytes": "0", "cpu_frequency_max_hz": "0", "gpu_count": "0", "gpu_product_names": []}, "context": {}, "github": {"commit": "79ccf7169e142da8bcd65a91e805ba4ff35ddba7", "repository": "git@github.com:conbench/conbench"}}
POST http://127.0.0.1:5000/api/benchmarks/ {"run_name": "test: 79ccf7169e142da8bcd65a91e805ba4ff35ddba7", "run_id": "2", "batch_id": "7c476d2c2c9a4200be73e1e6c1f9dd33", "run_reason": "test", "timestamp": "2022-10-31T21:25:12.770746+00:00", "tags": {"name": "benchconnect"}, "info": {}, "machine_info": {"name": "snork.local", "os_name": "macOS", "os_version": "12.6", "architecture_name": "arm64", "kernel_name": "21.6.0", "memory_bytes": "17179869184", "cpu_model_name": "Apple M1 Pro", "cpu_core_count": "10", "cpu_thread_count": "10", "cpu_l1d_cache_bytes": "65536", "cpu_l1i_cache_bytes": "131072", "cpu_l2_cache_bytes": "4194304", "cpu_l3_cache_bytes": "0", "cpu_frequency_max_hz": "0", "gpu_count": "0", "gpu_product_names": []}, "context": {}, "github": {"commit": "79ccf7169e142da8bcd65a91e805ba4ff35ddba7", "repository": "git@github.com:conbench/conbench"}, "error": {"stack_trace": "test"}}
```

jonkeane:
> what about the behavior/interface do we like?

The `post`ing API that's here is great + thanks for the examples. IMO, this is what folks are looking for (e.g. we could call this after each benchmark was run like we've seen in labs/benchmarks + some of our internal benchmark running setups).

> what do we want to change?

I've got a few extentions | additions below, but what's here so far looks great

> what functionality is missing that we want/need?

Two extensions:
Now that we have starting + finishing for runs in conbench, we could have a `benchconnect start-run` or the like that sends information about the run starting + configures any values that need to be configured like run id, hostname, etc.

in `post_results` in this PR, it would be nice to be able to either take the state information from the started run up above, or have it be passed in at that point. For example, the `benchconnect start-run` could use https://github.com/conbench/conbench/blob/main/benchadapt/python/benchadapt/machine_info.py to gather various bits of machine information + store it with the run state. Then fill those values in to the result IFF the result doesn't already have them. This will allow a runner to be able to override the defaults if they have a better understanding of that information.

alistaire47:
> Now that we have starting + finishing for runs in conbench, we could have a `benchconnect start-run` or the like that sends information about the run starting + configures any values that need to be configured like run id, hostname, etc.

Yep, figured. I'm not sure if posting runs is really necessary in most cases since they'll get created when posting benchmarks anyway, but putting would be good practice, and it's not much more work to cover both anyway.

> in `post_results` in this PR, it would be nice to be able to either take the state information from the started run up above, or have it be passed in at that point. For example, the `benchconnect start-run` could use https://github.com/conbench/conbench/blob/main/benchadapt/python/benchadapt/machine_info.py to gather various bits of machine information + store it with the run state. Then fill those values in to the result IFF the result doesn't already have them. This will allow a runner to be able to override the defaults if they have a better understanding of that information.

That's what `BenchmarkResult` already does; I'll make another for run. I'm thinking of making an `augment` method (or something) that fills in the details but doesn't post it, so users can `fill in what they know -> augment -> adjust -> post`.

jonkeane:
> That's what BenchmarkResult already does

As, so since we are using BenchResult in https://github.com/conbench/conbench/pull/422/files#diff-938cb6690a2256ced83319fdb21929e499eb56d1563c785d6109adba0701dcc4R6 this would automatically just work if I gave it JSON that had fields in it that benchadapt would otherwise fill in? 

> I'm thinking of making an augment method (or something) that fills in the details but doesn't post it, so users can fill in what they know -> augment -> adjust -> post.

I like this a lot

alistaire47:
> > That's what BenchmarkResult already does
> 
> As, so since we are using BenchResult in https://github.com/conbench/conbench/pull/422/files#diff-938cb6690a2256ced83319fdb21929e499eb56d1563c785d6109adba0701dcc4R6 this would automatically just work if I gave it JSON that had fields in it that benchadapt would otherwise fill in?

Yep. If `cluster_info` is specified, `machine_info` won't be populated; if neither is specified, `machine_info` will be. Most of the rest is more traditional parameter defaults. There are things that can't be auto-populated, though (like `run_reason`, more obviously `stats`/`error`, and less obviously a `name` tag), so you're not guaranteed to end up with a postable result. Long-term it would be really nice to have a `validate` command too, but we should talk first about how to structure that in such a way that it's easy to keep in sync between client- and server-side code.

jonkeane:
> Long-term it would be really nice to have a validate command too, but we should talk first about how to structure that in such a way that it's easy to keep in sync between client- and server-side code.

💯 on this and 💯 on doing that in a bit and making sure it's in sync with the server (and if we're in a monorepo we might even be able to make it such that they share the same validation bits where that's helpful and useful, yeah?)

austin3dickey:
This sounds like a good plan to me!

alistaire47:
Ok, this is ready for review! ⬆️ draft ⬇️ review

benchadapt/python/benchadapt/run.py
```diff
@@ -0,0 +1,123 @@
...
+    - ``finished_timestamp`` if run is now complete
+    - ``error_type`` (if applicable)
+    - ``error_info`` (if applicable)
+
+    Fields with defaults you may want to override on instantiation:
+
+    - ``machine_info`` if not run on the current machine
+    - ``cluster_info`` if run on a cluster
+    - ``github`` if current git info is unavailable or inaccurate
+    """
```
jonkeane: These can be overridden both at instantiation as well as by results that come in, right? If I'm reading the code below, it does... Do we want to mention that here? I'm not sure myself — I can see arguments for and against

benchconnect/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+A small Python package that creates a CLI for benchmark runners interacting
+with a Conbench server.
+
+This is a light wrapper on top of
+[benchadapt.BenchmarkResult](https://github.com/conbench/conbench/blob/main/benchadapt/python/benchadapt/result.py),
+[benchadapt.BenchmarkRun](https://github.com/conbench/conbench/blob/main/benchadapt/python/benchadapt/run.py), and
+[benchclients.ConbenchClient](https://github.com/conbench/conbench/blob/main/benchclients/python/benchclients/conbench.py).
+If your code is already in Python, it may be simpler to use those classes
+directly.
```
jonkeane, line 11: Love this clarity

benchconnect/benchconnect/_augment.py
```diff
@@ -0,0 +1,97 @@
...
+
+@click.command(
+    help="""
+Fill in missing fields of a benchmark run
+
+Specify either `--json` or `--path`.
+
+JSON will be passed to `benchadapt.BenchmarkRun`, which will fill in default
+values for missing keys and null values. Existing values will not be overridden.
+Please make sure `reason`, etc. are set how you like before posting the results.
```
jonkeane, line 58: Could we make an issue for either validating these on the server or here and add that as a todo. Ideally either the client or the backend would reject wrong results — we have some of that in place already on the backend, but we should add to that + clean it up

benchadapt/python/benchadapt/adapters/_adapter.py
```diff
@@ -5,9 +5,9 @@
 from typing import Any, Dict, List
 
 import requests
+from benchclients.conbench import ConbenchClient
+from benchclients.logging import fatal_and_log, log
 
-from ..client import ConbenchClient
```
austin3dickey, line 9: Do you still need to re-export `benchclient` stuff in `benchadapt/log.py` and `benchadapt/client.py`? I think you can delete those files, no?

benchadapt/python/benchadapt/run.py
```diff
@@ -0,0 +1,123 @@
...
+        Reason for run (e.g. commit, PR, merge, nightly). Should be of low cardinality.
+    info : Dict[str, Any]
+        A schema-less dict of metadata about the run
+    machine_info : Dict[str, Any]
+        For benchmarks run on a single node, information about the machine, e.g. OS,
+        architecture, etc. Auto-populated if ``cluster_info`` not set.
+    cluster_info : Dict[str, Any]
+        For benchmarks run on a cluster, information about the cluster
+    github : Dict[str, Any]
+        Keys: ``repository``, ``commit``. If unspecified, will be auto-populated based on
```
austin3dickey: Could you make the same [changes](https://github.com/conbench/conbench/pull/417/files#diff-c78a2c4b9d0537c37477655277cef51a686c7e4dc23abb4d5b71a497bfc9ac60) to the `github` stuff in this file that I made to the `result.py` file?

You could also consider increasing the scope of this PR to adding recognized github environment variables like [we talked about](https://github.com/conbench/conbench/pull/417/#discussion_r1014259956). If not, maybe make an issue to come back to that discussion later?

benchconnect/README.md
austin3dickey, line 11: nice!

benchconnect/benchconnect/_augment.py
```diff
@@ -0,0 +1,97 @@
...
+Fill in missing fields of a benchmark result
+
+Specify either `--json` or `--path`.
+
+JSON will be passed to `benchadapt.BenchmarkResult`, which will fill in default
+values for missing keys and null values. Existing values will not be overridden.
+Please make sure `run_id`, `run_reason`, `tags["name"]`, etc. are set how you
+like before posting the results.
+
+Augmented JSON will be printed one blob to a line, i.e. newline-delimited.
```
austin3dickey, line 35: nit: maybe add `If --path is a directory,`

benchconnect/benchconnect/_cli.py
```diff
@@ -0,0 +1,61 @@
...
+
+augment.add_command(_augment.result, name="result")
+augment.add_command(_augment.run, name="run")
+
+
+@cli.group(
+    help="""
+Post something to a Conbench API
+
+\b
```
austin3dickey: huh TIL about the backspace character. What is it doing here?

benchconnect/benchconnect/_augment.py
```diff
@@ -0,0 +1,97 @@
...
+        augment_and_print_blob(json=blob, cls=cls)
+
+    elif path and Path(path).resolve().is_file():
+        blob = read_json_file(path=path)
+        augment_and_print_blob(json=blob, cls=cls)
+
+    elif path and Path(path).resolve().is_dir():
+        for file in Path(path).resolve().glob("*.json"):
+            blob = read_json_file(path=file)
+            augment_and_print_blob(json=blob, cls=cls)
```
austin3dickey: What will the workflow look like for "augment and then post a directory of JSONs"? Since `augment` returns ndjson to stdout, how can that be piped to `benchconnect post`? Would you mind putting an example of this somewhere?

benchconnect/benchconnect/_post.py
```diff
@@ -0,0 +1,84 @@
...
+        click.echo("No data to post found!")
+
+
+@click.command(
+    help="""
+Post benchmark result JSON[s] to a Conbench API
+
+Specify either `--json` or `--path`.
+
+For Conbench environment variables, see `benchconnect post --help`.
```
austin3dickey: Wonder if you could import the env var help string instead of pointing to a different help doc here?

benchconnect/benchconnect/_put.py
```diff
@@ -0,0 +1,64 @@
...
+    "Load a JSON file and put the blob to Conbench"
+    with open(Path(path).resolve(), "r") as f:
+        json = load(f)
+
+    put_blob(json=json, endpoint=endpoint, client=client)
+
+
+# This is more abstract than we presently need, but it's easily extensible to
+# new PUT methods we might wrap in the future
+def putter(json: str, path: str, endpoint: str) -> None:
```
austin3dickey: ⛳

benchconnect/README.md
```diff
@@ -0,0 +1,42 @@
...
+or benchmark run (`augment run`) and fills in defaults for unspecified fields.
+This is useful for filling in fields like `machine_info` consistently. Note
+this is stateless, and will not make `run_id` and `suite_id` values correspond
+across results, nor will the resulting JSON necessarily be complete enough to
+post to a Conbench server correctly (it can't fill in things like `run_reason`
+for you); please ensure data is correct before posting.
+
+The `post` and `put` commands authenticate with a Conbench server (see man
+pages for environment variables) and send JSON passed. Since they correspond
+to the API, there are `post result` and `post run` methods, but only `put run`.
```
jonkeane, line 63: It would be nice to have an example in here. I started working on one, and in doing so, wondering if it would make sense to have a slightly higher-level API (basically, abstract the REST bits, giving folks an entry point that are the more-or-less logical operations they are doing)

```
benchconnect run start --run-reason="my fancy run"

# you can send either a path to results
benchconnect result send --path result.json
# or the json itself
benchconnect result send --json="{...}"

benchconnect run finish
```

And further, if you want to see how results will be augmented with default data, you can:

```
benchconnect run start --run-reason="my fancy run" --dry-run
benchconnect result augment --path result.json
```

I've made up a few things in here that would be necessary to get some of this working (e.g. having a `dry-run` option on starting a run that doesn't send to the server, but does setup the state that is needed to persist things like that run reason). I'm also presuming that `benchconnect result send` will automatically augment. The names of these.

On storing state: I know you've brought up multiple times in this PR that this is (currently) stateless. IMO that's making things harder on ourselves than it needs to be (we have to worry about passing things around, or making sure we always get run-reason in the right places). What if we did the simplest possible thing and when we start a run, we save a statefile (working directory is totally fine) that includes these bits? Then we don't have to worry about passing them through the stack.

jonkeane (commented):
This is looking good — I have one relatively major comment about the API here (though all of the work that _is_ here would be used by the higher-level API I'm proposing, there's no throwing anything away — just making the interface closer to what a consumer is thinking about when using this)

benchconnect/tests/test_augment.py
```diff
@@ -0,0 +1,68 @@
...
+minimal_run = {}
+
+
+@pytest.mark.parametrize(
+    "params", [(minimal_result, BenchmarkResult), (minimal_run, BenchmarkRun)]
+)
+def test_augment_blob(params):
+    raw_blob, cls = params
+    augmented_blob = augment_blob(raw_blob, cls=cls)
+    assert augmented_blob.keys() > raw_blob.keys()
```
austin3dickey: nit: parametrize allows multiple inputs, so you could do this to simplify slightly
```suggestion
@pytest.mark.parametrize(
    ["raw_blob", "cls"], [(minimal_result, BenchmarkResult), (minimal_run, BenchmarkRun)]
)
def test_augment_blob(raw_blob, cls):
    augmented_blob = augment_blob(raw_blob, cls=cls)
    assert augmented_blob.keys() > raw_blob.keys()
```

benchconnect/tests/test_augment.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class TestCliAugment:
+    runner = CliRunner()
```
austin3dickey, line 26: This is nice!

austin3dickey (commented):
Looking great! Mostly small stuff

benchadapt/python/benchadapt/adapters/_adapter.py
alistaire47: I know the log is being used [here](https://github.com/voltrondata-labs/benchmarks/blob/main/benchmarks/cpp_micro_benchmarks.py#L8); I'm not sure if the client is by anyone. We should migrate those so we can delete these though.

benchconnect/benchconnect/_augment.py
alistaire47: I mean it will anyway, it'll just be one line the rest of the time

benchconnect/benchconnect/_cli.py
alistaire47: It's a click thing for preserving newlines: https://stackoverflow.com/a/53302580/4497050 I looked a bit for a better way to get `<br>` instead of `<p>`, but this seems like all there is

jonkeane: 🤯 TIL

benchconnect/benchconnect/_augment.py
alistaire47: Hm, it would take a little munging. Maybe methods should take ndjson too?

Originally this was going to be an "all-in-one" situation where it augments and sends in one call, but that seemed opaque and hard to debug. But composability is nice

benchadapt/python/benchadapt/run.py
austin3dickey: I just looked at the server code—if you post a BenchmarkResult with a `run_id` that already exists in the DB, then all these fields will be effectively ignored, because they're only stored on the run:

- machine_info
- cluster_info
- run_name
- run_reason
- github

Is that behavior what you're talking about? (also maybe it would be good to document this behavior somewhere in here if you haven't already)

benchadapt/python/benchadapt/adapters/_adapter.py
austin3dickey: ohhh great call

benchconnect/benchconnect/_augment.py
austin3dickey: Yeah I figured that out eventually—but it took me a while to figure out what this line was saying

benchconnect/README.md
austin3dickey: I do like this api, and it would solve the ndjson problem

benchconnect/tests/test_augment.py
alistaire47: Interesting! But I find it less than obvious that it uses the first list to unpack cases—at a glance I'd guess each of the elements of the latter list are args that will get crossed like if you had >1 `@pytest.mark.parameterize` calls. But it looks like it accepts a tuple for the first arg, which makes it read a little more clearly to me

benchconnect/benchconnect/_augment.py
alistaire47: This led down some weird rabbit holes around stdin and textio, but everything should now accept ndjson piped in (or just passed inline)

benchadapt/python/benchadapt/run.py
alistaire47: Updated, thanks; I wrote this before that PR. 

I didn't add env vars yet; I want a little time to think about the implications that on what implementation looks like beyond the bounds of benchconnect, so I'll put it in another PR. Everything uses the same `machine_info.py` file anyway, so we shouldn't have to change lots of places whenever we get there

benchconnect/benchconnect/_augment.py
alistaire47: Done: https://github.com/conbench/conbench/issues/430 though unfortunately I think this is much easier to say than do

benchadapt/python/benchadapt/run.py
alistaire47: Added some new docs here: https://github.com/conbench/conbench/pull/422/files#diff-c78a2c4b9d0537c37477655277cef51a686c7e4dc23abb4d5b71a497bfc9ac60 which I think is the right place?

benchconnect/README.md
alistaire47: K, added a new stateful workflow, please take a look when you get a minute

benchconnect/tests/test_augment.py
austin3dickey: oh yeah it accepts tuples too, or this (IMO gross) format:
```py
@pytest.mark.parametrize(
    "raw_blob,cls", [(minimal_result, BenchmarkResult), (minimal_run, BenchmarkRun)]
)
```

benchadapt/python/benchadapt/run.py
```diff
@@ -0,0 +1,139 @@
...
+    - ``machine_info`` (generated by default) or ``cluster_info``
+    - ``stats`` or ``error``
+
+    Fields which should generally not be specified directly on instantiation that will
+    be set later for the run:
+
+    - ``name``
+    - ``id``
+    - ``reason``
+    - ``github``
```
austin3dickey, line 59: sorry for the confusion—this should go in the "Fields without comprehensive defaults which should be specified directly" section, with the changes I made in #417

benchadapt/python/benchadapt/run.py
```diff
@@ -0,0 +1,139 @@
...
+
+    name: str = None
+    id: str = field(default_factory=lambda: uuid.uuid4().hex)
+    reason: str = None
+    info: Dict[str, Any] = field(default_factory=dict)
+    machine_info: Dict[str, Any] = field(
+        default_factory=lambda: machine_info(host_name=None)
+    )
+    cluster_info: Dict[str, Any] = None
+    github: Dict[str, Any] = None
```
austin3dickey: Can you make this have a default of `{}` like in the `BenchmarkResult`? Users should have to be intentional about supplying `None` here, opting-in to the git scraping

benchconnect/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Usage
+
+See the man pages:
+
+```shell
+benchconnect --help
+benchconnect augment --help
+benchconnect post --help
+benchconnect put --help
```
austin3dickey, line 50: Reminder to update README with the new workflow (with an example if it's possible)

benchconnect/benchconnect/utils.py
```diff
@@ -0,0 +1,48 @@
+from json import dumps, load, loads
+from pathlib import Path
+from typing import Any, Dict, List
+
+import click
+
+ENV_VAR_HELP = """
```
austin3dickey, line 8: Love it!

benchconnect/benchconnect/utils.py
```diff
@@ -0,0 +1,48 @@
...
+  CONBENCH_URL: The URL of the Conbench server. Required.
+  CONBENCH_EMAIL: The email to use for Conbench login. Only required if the server is private.
+  CONBENCH_PASSWORD: The password to use for Conbench login. Only required if the server is private.
+"""
+
+
+def load_json(json: str, path: str, ndjson: str) -> List[Dict[str, Any]]:
+    "Load JSON from a string, file path, directory path, or stdin"
+
+    stdin = click.get_text_stream("stdin")
```
austin3dickey, line 21: man I love click. This is cool

benchconnect/benchconnect/utils.py
```diff
@@ -0,0 +1,48 @@
...
+    if not ndjson and not stdin.isatty():
+        ndjson = stdin.read()
+
```
austin3dickey, line 24: Would you mind adding a quick check here that not more than 1 option is supplied? Or if so, at least a message about which one you're choosing? I realize that'll probably never happen (and maybe it's technically impossible) but at least you can put that check in one place now

benchconnect/benchconnect/utils.py
```diff
@@ -0,0 +1,48 @@
...
+    if json:
+        return [loads(json)]
+
+    elif ndjson:
+        return [loads(blob) for blob in ndjson.strip().splitlines()]
+
+    elif path and Path(path).resolve().is_file():
```
austin3dickey, line 37: nit: You probably only need either these `.resolve()` calls or the `resolve_path=True` arg in the `--path` click option type, not both

benchconnect/benchconnect/_augment.py
```diff
@@ -0,0 +1,72 @@
...
+"""
+)
+@click.option("--json", default=None, help="A JSON dict of a result to augment")
+@click.option(
+    "--path",
+    default=None,
+    type=click.Path(exists=True, resolve_path=True),
+    help="Path to a JSON file or directory of JSON files containing results to augment",
+)
+@click.argument("ndjson", required=False)
```
austin3dickey, line 45: Is my understanding correct that this meant to be hidden from documentation but it "catches" any stdin piped to the CLI? If so could you add a little comment here like `# catches stdin`? (and to the other lines like this)

benchconnect/benchconnect/_start.py
```diff
@@ -0,0 +1,92 @@
...
+
+This method is the beginning of a workflow for posting a set of results to a
+Conbench API. It takes (directly via a blob or from a path) JSON corresponding
+to the POST /api/runs/ schema. Before posting to Conbench, the data will be
+augmented with `benchadapt.BenchmarkRun`; see `benchmark augment run --help`
+to debug and augment without posting.
+
+When successful, this command will store a statefile called {STATEFILE} in the
+current working directory which will be used to align metadata on subsequent
+data. Accordingly, only on run can be active at a time, and changing the
```
austin3dickey:
```suggestion
data. Accordingly, only one run can be active at a time, and changing the
```

benchconnect/benchconnect/_submit.py
```diff
@@ -0,0 +1,84 @@
...
+from ._start import STATEFILE
+from .utils import ENV_VAR_HELP, load_json
+
+
+def augment_and_post_result(json: dict, client: ConbenchClient) -> None:
+    "Augment a result from the statefile and class, then post it"
+    statefile_path = Path(STATEFILE).resolve()
+
+    if not statefile_path.exists():
+        raise FileNotFoundError(
```
austin3dickey: nit: use `fatal_and_log()`? (there are a few other places in this PR that you could do that instead of `raise` or `assert` if you wish)

benchconnect/benchconnect/_submit.py
```diff
@@ -0,0 +1,84 @@
...
+    for result_key in abstract_result:
+        if result_key in json:
+            assert abstract_result[result_key] == json[result_key], click.BadParameter(
+                "Result metadata does not match run metadata! "
+                f"Key: {result_key} "
+                f"Run: {abstract_result[result_key]} "
+                f"Result: {json[result_key]}"
+            )
+
+        json[result_key] = abstract_result[result_key]
```
austin3dickey, line 38: Most of these keys will be ignored by the server except `run_id` right? I wonder if we could save a tiny bit of time by just storing `run_id` in the statefile and not checking/adding the rest of the run keys every time we post a result. Elena recently [brought up](https://github.com/conbench/conbench/pull/429#discussion_r1017272968) that it's nice to keep the time for posting BenchmarkResults down as much as possible.

Though I realize these are all speedy dict hash operations so it wouldn't save that much time haha.

benchconnect/README.md
austin3dickey: Also, do we still want to expose `augment/post/put`? Seems to me like every usage of this CLI will just use the `start/submit/finish` workflow instead.

austin3dickey (commented):
I like this a lot! A few mostly small things

benchconnect/README.md
alistaire47: > Reminder to update README with the new workflow (with an example if it's possible)

It's there!

> Also, do we still want to expose `augment/post/put`? Seems to me like every usage of this CLI will just use the `start/submit/finish` workflow instead.

Yes, because they're easier to control and necessary for debugging.

benchadapt/python/benchadapt/run.py
alistaire47: oops. this whole section needed reworking; less-tired eyes helped

benchconnect/README.md
austin3dickey: >It's there!

Huh, I don't see it -- can you link to it?

benchconnect/benchconnect/_augment.py
alistaire47: It's not hidden; it shows up in the "Usage" section, but click won't let me specify help for an argument. (Apparently arguments should be "obvious" in bash-land?) It actually _doesn't_ catch piped in stdin; there for whatever reason has to be special handling for that. (I think it may have to do with nested group context? Went down a rabbit hole and this is the simplest, least-bad solution I found.) It does _look_ like what you're piping into, and stdin is handled accordingly. It does allow you to specify ndjson inline, as well; I figured leaving the `--json` flag isn't bad because this requires there to be no newlines within a blob, whereas that can handle it, and the names align with behavior. I'm not averse to documenting the argument more, but I haven't really figured out where to put it.

benchconnect/benchconnect/_submit.py
alistaire47: I thought about what the behavior should be here a bit, and from a least-surprise perspective, it seems like calls should fail if the result metadata conflicts with the run metadata even if it's going to be ignored; you could be posting a result for a different commit or something by accident. There are a few reasons for copying here, none huge:

1. even if the server ignores it, that metadata will get logged, which can be helpful for debugging
2. it's consistent with `augment`
3. it will stop the augmentation from overriding anything it shouldn't

We could just drop it all before posting? But the assertions feel important. Speed here seems exceedingly negligible compared to any network i/o.

benchconnect/README.md
alistaire47: Oh wait, I didn't push 🤦 . Apparently I was very tired yesterday

benchconnect/benchconnect/utils.py
alistaire47: We have to be careful with logging and warning so it goes to stderr, because stdout needs to be only output where there is any. But I think I can make this work

benchconnect/benchconnect/_submit.py
alistaire47: I think I started down a path after discovering that click basically always returns status 0 even if you raise an error, but it turns out that's quite hard to fix, so yeah, that's probably fine

benchconnect/benchconnect/utils.py
alistaire47: The latter handles bash-y things, so I can't drop that. I feel uncomfortable dropping the inline one here because the two are far apart and I'm sure I'm going to pass an unresolved path to this accidentally in the future and not discover it till it explodes months later on somebody else's computer. Maybe I am just scarred.

austin3dickey: ha that's cool!

benchconnect/benchconnect/_augment.py
austin3dickey: oh I see, I missed the `option` vs `argument` distinction in the code. Interesting you can't document it.

austin3dickey (approved):
LGTM!

benchadapt/python/benchadapt/result.py
```diff
@@ -87,6 +87,17 @@ class BenchmarkResult:
...
+    If a result with a new ``run_id`` is posted, a new record for the run will be
+    created. If a run record with that ID already exists, either because of a
+    previous result or the run being posted directly, the following fields will be
+    effectively ignored, as they are only stored on the run:
+
+    - ``run_name``
+    - ``run_reason``
+    - ``github``
+    - ``machine_info``
+    - ``cluster_info
```
jonkeane, line 100: Love the details here

jonkeane (approved):
This looks great — I haven't dug into the implementation details too, too much (let me know if there are any specific spots that you would like me to look at), but the UX|API is really great IMO. I think integrating this into arrowbench (and other iterative-style benchmarks) will be much easier + nicer than needing to define the iterative objects in Python directly.  What do you think?

alistaire47:
> I think integrating this into arrowbench (and other iterative-style benchmarks) will be much easier + nicer than needing to define the iterative objects in Python directly. What do you think?

It depends; there are tradeoffs. 

Easier:

- it doesn't matter at all how you run benchmarks
- everybody can make a system call

Harder:

- pushes JSON management upstream; sources need to produce conbench-valid JSON instead of just something that can be turned into that JSON
- it's not possible to make a generalized way to implement on CI
- python deps
- brittleness to working directory

For code in Python, I think it's easier to either use benchrun, adapters, or the dataclasses and client directly depending on what already exists. For arrowbench specifically, I think it's 6 to one, half-dozen to the other; either way, we've got to clean up the output and write a top-level runner.

One side effect of our unbundling on Arrow specifically that we haven't addressed yet but which will come up when we come to migration is keeping `run_id` consistent across languages/runners. That probably means we'll need to manage the initialization and finalization of the run in CI directly, and pass the `run_id` in to all runners—or if for things using the workflow, have them run from the same working directory and not manage their own run. For adapted things like Archery, you pass through `run_id` to `result_fields_override`. We probably can't easily take full advantage of finalization—passing through and aggregating errors would be a pain. It's possible, but I'm not sure it's worth the effort.

PR #423, opened 2022-11-03
Refactor ConbenchClient into its own package

alistaire47:
Refactors `ConbenchClient` from benchalerts into its own benchclients package. (Not wedded to the name if there are opinions.)

Notable:

- does not migrate the github clients
- exposes `BaseClient` so `benchalerts` can inherit
- renames `log.py` to `logging.py` to disambiguate module/object and exposes `log` at the toplevel
- migrates relevant tests, though this does leave a lot under-tested. We really need integration tests to do this well anyway.
- adds a `put()` method to `BaseClient` because I'm going to need it in #422 anyway
- ~does not migrate benchadapt yet. Probably I'll reexport at least `log` where it is so existing code (of which there is some but not a ton) still works. Can do that here or another PR~ migrates benchadapt, reexporting client in the same place


.github/workflows/actions.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -e .[dev] \
-            -e benchadapt/python \
```
alistaire47, line 35: Removed this for now because its only dep is benchclients, and it tries to install it from the `main` branch, which will fail until this is merged. We can put it back in a future PR, but it only really matters if it needs more dependencies

.github/workflows/actions.yml
```diff
@@ -31,9 +31,10 @@ jobs:
           src: .
       - name: Install dependencies
         run: |
-          pip install -e .[dev] \
-            -e benchadapt/python \
-            -e benchrun/python
+          pip install -e benchadapt/python --no-deps && \
```
alistaire47, line 39: This is janky but necessary for now because it tries to install benchclients from main, which it won't be on till this PR merges. We can revert this right after.

austin3dickey: ah makes sense. We don't have to revert it yet; I'll try to fix it the "right" way when adding integration tests. It's definitely important we don't ever test against the main branch in PRs

benchclients/python/README.md
```diff
@@ -0,0 +1,3 @@
+# benchclients
+
+A lightweight Python package with clients for working with APIs in the Conbench ecosystem
```
austin3dickey, line 3: This one will be only Python, no? Do we need the intermediate `python/` dir?

benchclients/python/benchclients/base.py
```diff
@@ -0,0 +1,75 @@
...
+import requests
+from requests.adapters import HTTPAdapter
+from urllib3.util.retry import Retry
+
+from .logging import log
+
+
+class BaseClient(abc.ABC):
+    """A client to interact with an API.
+    Parameters
```
austin3dickey, line 15: nit: you need a newline above this one, else the RST won't render correctly if we ever render it

benchclients/python/benchclients/base.py
```diff
@@ -0,0 +1,75 @@
...
+        res = self.session.get(url=url, timeout=self.timeout_s)
+        self._maybe_raise(res=res)
+
+        return res.json()
+
+    def post(self, path: str, json: dict) -> Optional[dict]:
+        url = self.base_url + path
+
+        # log.debug(f"POST {url} {dumps(json)}")
+        print(f"POST {url} {dumps(json)}")
```
austin3dickey: Change this back to `log`?

benchclients/python/benchclients/conbench.py
```diff
@@ -0,0 +1,40 @@
...
+
+from requests.adapters import HTTPAdapter
+
+from .base import BaseClient
+from .logging import fatal_and_log
+
+
+class ConbenchClient(BaseClient):
+    """A client to interact with a Conbench server.
+    Parameters
```
austin3dickey, line 13: Same in this docstring

austin3dickey (approved):
looking good! Small stuff

.github/workflows/actions.yml
alistaire47: Yeah, the reason we'd need to revert is if benchadapt adds other deps for some reason. It's deliberately lightweight, so I don't foresee it, but it's possible.

benchclients/python/README.md
alistaire47: Yeah, I was thinking about this and wasn't sure. With things that are CLIs (datalogistik, benchconnect) there's definitely no need for versions in other languages, but for things like this, it's _possible_ that in the future we might add a client in another language. Like if an existing benchmarking framework in, I don't know, Ruby wanted to send their results to conbench but not have deps on various Python stuff, there's a space here to stick a Ruby port that other people could use. To be clear I don't foresee any ports, but directories are cheap and I'm cautious.

PR #427, opened 2022-11-07
urlize some things

### LINKED ISSUE #426: Transform (likely?) URLs into links
When we send extra benchmark information, those sometimes include http-based URLs. It would be nice if our display rendered those as actual clickable links and not just text.

jonkeane:
Resolves #426 

Adds Jinja's `urlize` to some of our free-form fields so that things like https://foo.bar are auto-linked:

image


austin3dickey (approved):
fast, fun, and friendly!

PR #429, opened 2022-11-08
Backfill default branch commits upon Run.create()

austin3dickey:
This PR adds behavior where, upon every creation of a Run (including upon posting a BenchmarkResult without an existing `run_id`), the database will try to backfill any commit entities on the default branch that aren't tracked yet from GitHub. 

This serves the purpose that, in cases where CI fails to post BenchmarkResults from a default-branch commit, Conbench will still know about the commit, so we can display warnings in the app and traverse the commit graph better in the future.

This gets us further on #382 but not all the way.


conbench/api/_examples.py
```diff
@@ -149,9 +149,9 @@ def _api_commit_entity(commit_id, parent_id, links=True):
         "author_name": "Diana Clarke",
         "id": commit_id,
         "message": "ARROW-11771: [Developer][Archery] Move benchmark tests (so CI runs them)",
-        "repository": "https://github.com/apache/arrow",
+        "repository": "https://github.com/org/repo",
```
austin3dickey, line 152: I had to change all these API examples so that the unit tests that use them will know they're in a unit test and not actually hit GitHub.

conbench/entities/_entity.py
```diff
@@ -49,8 +49,14 @@ def search(cls, filters, joins=None, order_by=None):
         return q.filter(*filters).all()
 
     @classmethod
-    def all(cls, limit=None, order_by=None, **kwargs):
+    def all(cls, limit=None, order_by=None, filter_args=None, **kwargs):
+        """Filter using filter_args and/or kwargs. If you just need WHERE x = y, you can
+        use kwargs, e.g. ``all(x=y)``. Else if you need something more complicated, use
+        e.g. ``all(filter_args=[cls.x != y])``.
+        """
```
alistaire47, line 56: Thank you!

conbench/tests/populate_local_conbench.py
```diff
@@ -219,7 +219,7 @@ def create_benchmarks_data():
         "2462492389a8f2ca286c481852c84ba1f0d0eff9",
     ]
 
-    branches = ["apache:master", "austin3dickey:decimal256", None] * 2
```
austin3dickey, line 222: This lie (that these scripted commit shas were on `austin3dickey:decimal256`, not `apache:master`) started to cause this script to fail on this PR. Later when I rework the history queries, I'll write a few true branchy cases in this script so we can test those out in the app better.

conbench/entities/commit.py
```diff
@@ -172,6 +180,62 @@ def get_github_commit(repository: str, pr_number: str, branch: str, sha: str) ->
...
+        repository=repository,
+        order_by=Commit.timestamp.desc(),
+        limit=1,
+    )
+
+    if last_tracked_commit:
+        since = last_tracked_commit[0].timestamp
+    elif os.getenv("DB_HOST") == "localhost":
+        # only for conbench/tests/populate_local_conbench.py
+        since = datetime(2022, 1, 1)
```
alistaire47: There are other cases where the DB will be on localhost, though the majority will still be testing of other sorts. Most things will get caught by the `if` clause here anyway so I'm not that worried, but maybe a log/warning in these other clauses so somebody doesn't end up in a debugging rabbit hole in the future?

alistaire47 (approved):
Looks good!

conbench/entities/commit.py
jonkeane: Agreed, a warning or other kind of messaging would be helpful here

conbench/tests/populate_local_conbench.py
```diff
@@ -275,16 +275,18 @@ def create_benchmarks_data():
 
 
 def create_benchmarks_with_history():
-    # 5 commits in a row in apache/arrow
+    # 7 commits in a row in apache/arrow, one missing
```
jonkeane:
```suggestion
    # 7 commits in a row in apache/arrow, the commented one is missing
```

Just a bit more clarity on the reason for the commenting?

jonkeane (approved):
A few minor comments — this is looking good + really appreciate the smaller isolated chunks where we can

conbench/entities/commit.py
austin3dickey: Great idea

austin3dickey: @ElenaHenderson if `os.getenv("DB_HOST") == "localhost"`, is it a safe assumption that someone is testing on a local computer? If not, is there a better way to check that?

conbench/entities/run.py
```diff
@@ -64,6 +65,7 @@ def create(data):
             )
             if github:
                 commit = Commit.create_github_context(sha, repository, github)
+                backfill_default_branch_commits(repository, commit)
```
ElenaHenderson, line 68: @austin3dickey Did you time how long POST /benchmarks will take with backfill?

This might cause pretty bad performance issues for Apache Arrow benchmark builds.
Some benchmark builds have 2800+ benchmarks and we had to keep POST /benchmarks under 200ms to make posting to conbench reasonable.
If backfill will take 1 second per request, then build will extend by ~45 mins and we tried hard to keep builds like this under 2 hours (which what they are now).

I recommend creating a script and run it as a migration to avoid performance issues with POST /benchmarks. 

Or you can add `Run scripts` to Deploy or Test job and put your script there to backfill.

conbench/entities/commit.py
ElenaHenderson: I don't know about a 100% sure way of checking if somebody is running Conbench locally.

Tests and the app that are run using docker-compose will use DB_HOST = "db" locally.

Anything in production will run on k8s pod and this is how you can check if a process is running in the pod:
`os.getenv("KUBERNETES_SERVICE_HOST") is not None`

People can set DB_HOST to `127.0.0.1` when they run postgres locally too.

ElenaHenderson (commented):
I have a comment on backfilling during POST /benchmarks request. 

I really don't think it is a good idea. We had all sorts of issues when POST /benchmarks was taking more than 200ms and I spent a lot of time refactoring schema and tuning DB to make POST /benchmarks fast to avoid those painful issues.

I recommend just writing a migration for this.

conbench/entities/run.py
jonkeane: We should definitely confirm how long this takes and confirm what it would look like in realistic scenarios. 

I hear Elena's concerns, but personally I'm a bit less concerned because, if I am understanding correctly: 

* For apache/arrow we would almost never actually backfill, because each commit to dev is benchmarked already, and this only backfills if there are missing commits. And, in the (rare) cases when we do hit this, it should be at most a handful of commits not yet benchmarked. And, even then, it should only happen on the first benchmark posted, not on each benchmark after that, no?

* For repos that do merge commits, this could be a considerable number of commits (it will basically always be the number of commits that are on the branch... But even then, it'll be only on the first benchmark posted. But we definitely should investigate what that performance is like (though since we don't really yet support those, IMO if we deferred this investigation that would be OK, not ideal, but ok)

One other possible alternative would be to investigate if we could have this backfilling process be asynchronous, so that it doesn't interfere with the REST call lifecycle?

Finally, I suspect the label "backfill" here _might_ be a bit scary sounding — it's not intended to be like our typical backfill migrations, but more "fill in the gaps of history on the default branch"

conbench/entities/commit.py
austin3dickey: Okay, great. I think I'll keep this as-is then. I don't mind if someone is testing and `DB_HOST` is something other than `localhost`; they'll just have to wait a little longer.

conbench/entities/run.py
austin3dickey: @ElenaHenderson Sorry for the confusion — this "backfill" (really it's more of a "check to see if we need to fill a small gap" like @jonkeane said) only happens when a new Run is created, *not* every time a new BenchmarkResult is created. 

In my understanding, for Arrow benchmarks, we only create a new Run on average every 30 minutes or so. I just timed this locally, and this code will increase the first `POST /benchmarks` time to about 1500 ms, but every subsequent post with the same `run_id` is completely unaffected, sitting around 20 ms. So that's only an increase of about a minute of build time per day.

austin3dickey: Actually, on top of that, this code is only run when a new Run is created on a commit that _doesn't already exist_ in the database. That only happens every 4 hours or so for Arrow.

austin3dickey: example: 
image

ElenaHenderson: @austin3dickey This is good. I completely missed the part that it will run only per RUN creation and not per BENCHMARK_RESULTS creation. I am not worried about this at all then.

ElenaHenderson (approved):
@austin3dickey Thank you for responding to my comment.

@jonkeane If we ever need an async mechanism for conbench, we can add something like celery or something new and better. I kind of wanted this for a while and I know that it will make things with beef up conbench in a good way. Adding celery would take me a few days mostly because I forgot all the important settings that matter. Or can do something else.

jgehrcke:
> One other possible alternative would be to investigate if we could have this backfilling process be asynchronous, so that it doesn't interfere with the REST call lifecycle?

> I have a comment on backfilling during POST /benchmarks request. I really don't think it is a good idea.

I love the pragmatic solution approach here @austin3dickey! It's good that we have what we have. But I also share the thoughts of @ElenaHenderson and @jonkeane. I think a more complete error handling perspective helps finding the right arguments.

We decided that we want to trigger the backfill machinery _during handling a POST HTTP request_ to /api/benchmarks.  

That raises the question: is a successful backfill machinery run a requirement before we can process the user-provided data in the POST request? Or are we just doing a best effort approach here?

In other words: can _accept and insert_ the user-provided data from the POST request (and 2xx-ack it) _even if_ the backfill machinery fails (in whichever way)?

We have to have an opinion about handling all kinds of errors that happen during the backfill machinery, and I don't easily see from code/comments what our opinions were/are. Should the POST to /api/benchmarks fail when we cannot reach the GitHub API? The answer is probably no. Should it fail when the backfill machinery's DB transaction fails for unexpected reasons? The answer seems to be no, too. Inversely, this probably means that we want to ignore _all_ errors happening in the backfill machinery so that we maximize the likelihood for the POST request to actually accept and persist the user-provided data. 

If what I wrote so far makes sense then I think it means that doing the backfill machinery as part of  POST to /api/benchmarks is just _one of many ways_ to trigger that machinery every now and then. And that would be an important insight.

A bit of a harsh way to look at this: the way it's currently built we increased the probability for user-visible failure in  POST requests to /api/benchmarks, and the gain isn't absolutely clear.

A rule of thumb is that it's a great idea to minimize the work done as part of HTTP request handling, and to remove all those sources of transient errors that are not absolutely necessary for generating a success HTTP response. The outcome are less errors seen by HTTP clients (and also a better response latency distribution). I think that is what Jon meant with "doesn't interfere with the REST call lifecycle".

austin3dickey:
Yes, I think this all makes sense. Previously to this PR, we had committed to the principle that "upon POSTing a benchmark that needs a new Commit entity, we'll try our best to find context from GitHub, and if we can't, the POST should still succeed." I tried to follow that principle in this PR, but failed in one or more ways.

With the [history renovation project](https://github.com/conbench/conbench/milestone/1), it's much more important to have detailed commit context in the database than it was before. But it isn't important enough to have the server return 500s when something goes wrong with GitHub context during POST /api/benchmarks. I think correctly doing #441 will help minimize those 500s.

PR #431, opened 2022-11-10
Added commit backfill migration

austin3dickey:
This PR adds a one-time migration script to backfill all default-branch commits on all repos tracked by a Conbench server. This is primarily intended so that the `Commit.fork_point_sha` field is guaranteed to point to another Commit record in the database. (I'm not implementing an actual relationship there yet because the migration might be tricky, but that might be next.)

I tested it out as a dry run with read-only access to the Apache Arrow Conbench DB, and it takes about 4 minutes to run on the 5 repos tracked in that server.

### DIFF
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -314,9 +314,13 @@ class GitHub:
         page = 1
 
         this_page = self._get_response(url + f"&page={page}")
-        if not this_page:
+        if this_page is None:
             print("API request failed")
             return []
+        elif len(this_page) == 0:
+            print("API returned no commits")
+            return []
+
         commits += this_page
 
         while len(this_page) == 100:
--- a/migrations/versions/d3515ecea53d_backfill_all_commits.py
+++ b/migrations/versions/d3515ecea53d_backfill_all_commits.py
@@ -1,4 +1,8 @@
-"""backfill_all_commits
+"""Backfill all commits
+
+Ensures that all default-branch commits are populated correctly in the database, and
+that the database doesn't have non-default-branch commits that lie and say they're on
+the default branch.
 
 Revision ID: d3515ecea53d
 Revises: 480dbbd48927
@@ -54,6 +58,7 @@ def upgrade():
         db_commits_by_sha = {commit.sha: commit for commit in db_commits}
         print(f"Found {len(db_commits)} commits in the database")
 
+        # Go through each default-branch commit and ensure it's correct in the database
         for commit_info in all_commits:
             sha = commit_info["sha"]
             if sha in db_commits_by_sha:
@@ -86,6 +91,23 @@ def upgrade():
                         author_login=commit_info["github"]["author_login"],
                         author_avatar=commit_info["github"]["author_avatar"],
                     )
+                )
+
+        # Ensure there aren't any db commits lying that they're on the default branch
+        default_branch_shas = [commit_info["sha"] for commit_info in all_commits]
+        for db_commit in db_commits:
+            if (
+                db_commit.branch == default_branch
+                and db_commit.sha not in default_branch_shas
+            ):
+                print(
+                    f"Removing branch from sha {db_commit.sha[:7]} because it's not "
+                    "actually on the default branch"
+                )
+                connection.execute(
+                    commit_table.update()
+                    .where(commit_table.c.id == db_commit.id)
+                    .values(branch=None)
                 )
 
 
--- a/migrations/versions/d3515ecea53d_backfill_all_commits.py
+++ b/migrations/versions/d3515ecea53d_backfill_all_commits.py
@@ -1,0 +1,93 @@
+"""backfill_all_commits
+
+Revision ID: d3515ecea53d
+Revises: 480dbbd48927
+Create Date: 2022-11-09 11:33:51.081418
+
+"""
+import datetime
+import uuid
+from typing import List, Tuple
+
+from alembic import op
+from sqlalchemy import MetaData, Table, distinct, select
+
+from conbench.entities.commit import Commit, GitHub, repository_to_name
+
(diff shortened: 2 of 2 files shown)


austin3dickey:
Marked as draft because I want to be sure different CIs are populating `branch/fork_point_sha` correctly before this backfill takes place.

migrations/versions/d3515ecea53d_backfill_all_commits.py
```diff
@@ -0,0 +1,93 @@
...
+
+            else:
+                # insert this commit into the db
+                print(f"Inserting sha {sha[:7]} into the db")
+                connection.execute(
+                    commit_table.insert().values(
+                        id=uuid.uuid4().hex,
+                        sha=sha,
+                        branch=default_branch,
+                        fork_point_sha=sha,
```
alistaire47, line 85: Hm, should `fork_point_sha` be itself for commits to the default branch? I kind of figured it'd be null, which would make it easier to query for available comparisons eventually. Not a strongly-held opinion, though, especially if the gh API is opinionated on the matter

austin3dickey: Yes, I originally started putting the self-sha in `fork_point_sha` instead of None because: 
- the GitHub API returns the self-sha, like you thought
- if a run is posted with a new commit sha, but we can't get info from GitHub about that sha for some reason, we don't fail: we insert a commit without context (so `fork_point_sha` is None in that case). I wanted to distinguish between "commits on the default branch" vs "commits we don't have info about"

So (after this migration script is run, and the databases are cleaned up) we can assume that `commit.sha == commit.fork_point_sha` if and only if the commit is on the default branch.

austin3dickey:
I'm also gonna add one more small chunk of code that ensures any database Commit shas that *aren't* on the default branch don't have the default branch listed under `branch`. That could have happened to a few commits before I fixed the CIs.

austin3dickey:
Okay, this is ready for review. All CI systems that I know of are correctly reporting branch information moving forward. This script will fix the past commits already in the DB.

alistaire47 (approved):

conbench/entities/commit.py
```diff
@@ -314,9 +314,13 @@ def get_commits_to_branch(
         page = 1
 
         this_page = self._get_response(url + f"&page={page}")
-        if not this_page:
+        if this_page is None:
```
jgehrcke, line 317: A question of identity :heart:

migrations/versions/d3515ecea53d_backfill_all_commits.py
```diff
@@ -0,0 +1,115 @@
...
+            continue
+
+        name = repository_to_name(repo)
+        default_branch = github.get_default_branch(name)
+        all_commits = github.get_commits_to_branch(
+            name=name,
+            branch=default_branch,
+            since=datetime.datetime(1970, 1, 1),
+            until=datetime.datetime.now(),
+        )
```
jgehrcke, line 52: It's interesting that we do potentially quite a bit of HTTP interaction with an external service here in the `upgrade()` function. There are probably a number of assumptions here about the robustness of the interaction, and also about rate limiting / quota errors, in relationship to the number of commits that are actually being fetched. Will this raise an exception when one of the underlying HTTP requests fails? Will this fail hard for repos with more than 5000 commits, or will this enter a long retry loop? (5000 requests per hour per api token is one of the quotas that the github api implements)

Some of that is answered by https://github.com/conbench/conbench/blob/13dfb636b3da9d115f38dd3496188c0840389b72/conbench/entities/commit.py#L408 and surroundings.

Just thinking out loud, :shrug: :) no reply expected.

austin3dickey: Nice thoughts! I'm not worried about hitting the rate limit, for a few reasons:

- This script is a database migration, so it's only run once per server. We don't need to worry about this code's performance on future GitHub data, just the data that exists today.
- We're currently lucky enough to be in a position where we're aware of and can maintain most Conbench servers actually running today. Most servers that I know of are only tracking one repo. The exception is https://conbench.ursa.dev/ (the Arrow server) which is tracking 4 repos. Those 4 repos total about 34,000 commits as of this writing.
- The `get_commits_to_branch()` method uses pagination with a [page size](https://github.com/conbench/conbench/blob/3890458e7afaddeda7d52cc5fba471fd3306cbdf/conbench/entities/commit.py#L310) of 100 commits. So this script will only hit the API 340 times!
- When I merge this PR, I'll monitor the deployments on the servers I know about, and fix anything that might pop up.

jgehrcke: Thanks for writing down these thoughts.

And just maybe for clarification: if one of those ~300 HTTP requests fails unexpectedly (e.g. as of a transient error or as of rate limiting) -- then the migration will probably crash, right? And then some outside system (k8s, for example) will restart the entire thing, and the migration will start again, right?

austin3dickey: If it fails on the first page for a repo, `get_commits_to_branch()` should return `[]` and this script will remove all the branches from the commits in the database. This is probably okay because we could try the script again and the branches will be re-populated. I'll keep an eye out for this case.

If it fails on a subsequent page, we'll probably get an error like "can't iadd list and NoneType" and the script will crash. The deployment Buildkite job will fail and will have to be manually re-triggered. Any previous successful transactions will be kept (but for this script that's okay because they'll be skipped upon retry). I'll keep an eye out for this too.

jgehrcke: > get_commits_to_branch() should return []

Oh. I was hoping this function raises an exception :).

PR #433, opened 2022-11-15
Ceilinged SQLAlchemy < 2

austin3dickey:
As we saw in https://github.com/voltrondata-labs/arrow-benchmarks-ci/pull/88 , SQLAlchemy 2.0 is coming out soon, and this repo has at least one thing that will break.

I tried running tests locally with `2.0.0b3`, and got a lot of errors. I'll create an issue to officially migrate and fix those errors, but in the meantime let's ceiling to 2.0 so that we won't be surprised with a breaking change.

### DIFF
--- a/migrations/env.py
+++ b/migrations/env.py
@@ -22,7 +22,13 @@ from conbench.entities import (  # noqa  # isort:skip
 # this is the Alembic Config object, which provides
 # access to the values within the .ini file in use.
 config = context.config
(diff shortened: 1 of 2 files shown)


alistaire47 (approved):

jgehrcke:
Oh, that's a bunch of moving dependencies.
Change looks good.

jonkeane (approved):
Thank you for tracking this down!

Issue #393, opened 2022-10-07
Swagger docs don't give option to poke at live server


ericdill:
I'm currently looking at the Swagger docs for a conbench instance and the only server dropdown choice is localhost
![image](https://user-images.githubusercontent.com/6223243/194589781-dea24743-961e-4ad9-9e6c-1475d1f45d53.png)

I accessed the swagger docs by clicking on 'API Docs' after logging in to conbench, so I'm already auth'd to the server. I'd expect to see the active server as the default option on the above dropdown

![image](https://user-images.githubusercontent.com/6223243/194590212-2380a958-fda6-4ad6-93bc-37d6830464d4.png)

jonkeane:
A public example of swagger docs: https://conbench.ursa.dev/api/docs/#/

Could adding `DB_HOST` and `DB_PORT` to https://github.com/conbench/conbench/blob/fdb30f2934481989e981c49837506c0622d6bb3a/conbench/api/_docs.py#L16 be enough?

jgehrcke:
Hello! Naive question: Is "poke at live server" safe, especially considering mutating API calls with http methods such as POST, DELETE? This would after all change the state in some database.

I suppose the current state (where only localhost:5000 is listed) is tailored for a development environment where a dev server runs locally and where mutating API calls do not need to be carefully crafted (can ignorantly be sent w/o fear of messing up state).

Interaction of the swagger docs with the live server should maybe be limited to read-only API calls? Happy to learn more about the context.

jonkeane:
> I suppose the current state (where only localhost:5000 is listed) is tailored for a development environment where a dev server runs locally and where mutating API calls do not need to be carefully crafted (can ignorantly be sent w/o fear of messing up state).

When I was running this last, I actually had to change this even during local development to go from localhost to 127.0.0.1. So even in that case it would be nice if the swagger docs protected this.

> Hello! Naive question: Is "poke at live server" safe, especially considering mutating API calls with http methods such as POST, DELETE? This would after all change the state in some database.

> Interaction of the swagger docs with the live server should maybe be limited to read-only API calls? Happy to learn more about the context.

Yeah, we might want to disable trying it out (copying Swagger's terms) for post and put on production servers. It would be nice, if we could, to be able to still have the `GET`s or other non-state changing calls work. And ideally, be able to set this so that one could turn this on for a staging server (we don't (yet) have a staging server for arrow, but we do have one internally).

jgehrcke:
[...]

PR #435, opened 2022-11-16
app config: introduce INTENDED_BASE_URL, use for api docs

jgehrcke:
This change is motivated by #393. See commit message for details!

Tested this locally. A few screenshots:

Screenshot from 2022-11-16 10-51-37

'Try it out' feature now works against default dev server (HTTP request was handled by app):
Screenshot from 2022-11-16 10-51-59

This is to show the failure with `localhost` (w/o patch):
Screenshot from 2022-11-16 10-52-35

Review appreciated :)

### DIFF
--- a/conbench/config.py
+++ b/conbench/config.py
@@ -25,6 +25,27 @@ class Config:
     # to recalculate history.
     DISTRIBUTION_COMMITS = int(os.environ.get("DISTRIBUTION_COMMITS", 100))
 
+    # The base URL (scheme, DNS name, path prefix) that regular HTTP clients
+    # are expected to use for reaching the HTTP server exposing the app.
+    # Depends on the deployment and cannot generally be determined by the app
+    # itself (requires human input). The default value is tailored to the Flask
+    # development HTTP server defaults (non-TLS, binds on 127.0.0.1, port
+    # 5000). Can be adjusted via the `--host` and `--port` command line flags
+    # when invoking `flask run`. Three interesting scenarios:
+    # - local dev setup with default listen address: no action needed
+    # - local dev setup with custom listen address: user should set meaningful
+    #   value
+    # - prod/staging setup: user should set meaningful value such as for
+    #   example https://conbench.ursa.dev/
+    #
+    # Currently used for populating the 'Servers' dropdown in the
+    # Swagger/OpenAPI docs website.
+    INTENDED_BASE_URL = os.environ.get(
+        "CONBENCH_INTENDED_BASE_URL", "http://127.0.0.1:5000/"
+    )
+    # Require trailing slash towards tidy URL generation.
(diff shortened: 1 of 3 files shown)


conbench/config.py
```diff
@@ -25,6 +25,27 @@ class Config:
...
+    # 5000). Can be adjusted via the `--host` and `--port` command line flags
+    # when invoking `flask run`. Three interesting scenarios:
+    # - local dev setup with default listen address: no action needed
+    # - local dev setup with custom listen address: user should set meaningful
+    #   value
+    # - prod/staging setup: user should set meaningful value such as for
+    #   example https://conbench.ursa.dev/
+    #
+    # Currently used for populating the 'Servers' dropdown in the
+    # Swagger/OpenAPI docs website.
```
jonkeane, line 42: 💯 I know long comments are frequently looked down on as uncouth, but I love _love_ all the details here for our future selves – thank you!

conbench/config.py
```diff
@@ -25,6 +25,27 @@ class Config:
...
+    INTENDED_BASE_URL = os.environ.get(
+        "CONBENCH_INTENDED_BASE_URL", "http://127.0.0.1:5000/"
+    )
+    # Require trailing slash towards tidy URL generation.
+    assert INTENDED_BASE_URL.endswith("/"), "INTENDED_BASE_URL must have trailing slash"
```
jonkeane: I'm fine leaving this as an assert and making folks do it correctly — but it would be ~the same code to correct it and add the `/` to the end if it didn't have one, yeah?

jonkeane (approved):
Thanks for this!

austin3dickey (commented):
Looks great! Agreed with Jon's review comments.

The build is failing because when running tests, we automatically make changes to the expected OpenAPI documentation, so that reviewers can look over those changes in the PR. Because of https://github.com/conbench/conbench/issues/437 (thanks for opening that issue!), your changes to that file in your container didn't get picked up by git. Could you follow the instructions in the build to fix that?

jgehrcke:
Oh yeah, `test_docs` found
```
E             -  'servers': [{'url': 'http://localhost:5000/'}],
E             ?                              ^^^^^^^^^
E             +  'servers': [{'url': 'http://127.0.0.1:5000/'}],
E             ?  
```
Will add another commit.

conbench/config.py
jgehrcke: You mean something in the lines of
```
    if not INTENDED_BASE_URL.endswith("/"):
        INTENDED_BASE_URL += "/"
```
? Yeah, we can do that!

jgehrcke: pushed a corresponding commit

austin3dickey (approved):

jgehrcke:
I'm going ahead and merge this, let's iterate if needed. Thanks!

PR #436, opened 2022-11-16
debug mode: silence warning, print config

jgehrcke:
Two tiny improvements, see commit messages. Thanks for review!

### DIFF
--- a/.flaskenv
+++ b/.flaskenv
@@ -4,6 +4,7 @@ DB_PASSWORD=postgres
 DB_PORT=5432
 FLASK_APP=conbench
 FLASK_ENV=development
+FLASK_DEBUG=true
 REGISTRATION_KEY=conbench
 SECRET_KEY="Person, woman, man, camera, TV"
 BENCHMARKS_DATA_PUBLIC=true
--- a/conbench/__init__.py
+++ b/conbench/__init__.py
@@ -1,4 +1,5 @@
 import importlib.metadata as importlib_metadata
+import json
 import os
 
 try:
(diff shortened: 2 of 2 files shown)


conbench/__init__.py
```diff
@@ -14,7 +15,15 @@ def create_application(config):
 
     application = f.Flask(__name__)
     application.config.from_object(config)
+
     _init_application(application)
+
+    if application.config["DEBUG"]:
+        c = json.dumps(application.config, sort_keys=True, default=str, indent=2)
```
jonkeane: I don't think I knew that `json.dumps` had such convenience bits as sorting and indents ❤️

.flaskenv
```diff
@@ -4,6 +4,7 @@ DB_PASSWORD=postgres
 DB_PORT=5432
 FLASK_APP=conbench
 FLASK_ENV=development
+FLASK_DEBUG=true
```
jonkeane, line 7: I'm 95% certain this `.flaskenv` is not used in production deployments, so it's all good to hard code debug to true here — but want to double check that you agree that's actually the case here

jonkeane (approved):
Thank you for this clean up! One small question

.flaskenv
jgehrcke: Oh. Yeah great question. If `.flaskenv` is used in production then debug mode was already enabled so far via the `FLASK_ENV=development` setting (it's deprecated, but still honored). That's why I thought that the change won't be a regression. We should certainly double-check, but I like that you are already 95 % confident :).

conbench/__init__.py
jgehrcke: yeah! :) True magic here is the `default=str` which is a meaningful fallback for objects that are not JSON-serializable, such as `datetime.timedelta` objects (needed that here).

jgehrcke:
CI found that I committed before cleaning up:
```
./conbench/__init__.py:22:9: F841 local variable 'c' is assigned to but never used
```

update: added a new commit and squashed, force-pushed right away -- let's re-run CI

austin3dickey (approved):

jgehrcke:
I'm going ahead and merge this, let's iterate if needed. Thanks for the reviews!

Issue #365, opened 2022-08-30
If I post an empty `context` for a run, I get a 500 when I click on the run in the UI


jonkeane:
We should either allow contexts to be empty or we should reject the `POST` if context must have at least one key.

jgehrcke:
Oh. Does an empty context ever have meaning? Otherwise, sounds like the input validation should be tweaked so that each persisted (2xx-acked) context is meaningful and can be meaningfully displayed.

jonkeane:
We've never had a circumstance where we've purposefully wanted an empty context (in other words, there's always been some details about the context of a benchmark run we've wanted to save). But it's possible that some style of running might not need or want to provide that information.

I would be ok either way that we handle this (either requiring it to be non-empty, or changing the display to ignore if it doesn't exist)

jgehrcke:
I played with this a bit.

The current run schema seems to not allow for a `context` property (that's the schema as shown by the API docs):
```
{
  "cluster_info": {
    "info": {},
    "name": "string",
    "optional_info": {}
  },
  "error_info": {},
  "error_type": "string",
  "finished_timestamp": "2022-11-16T12:45:32.192Z",
  "github": {
    "branch": "string",
    "commit": "string",
    "pr_number": 0,
    "repository": "string"
  },
  "id": "string",
  "info": {},
  "machine_info": {
    "architecture_name": "string",
    "cpu_core_count": 0,
    "cpu_frequency_max_hz": 0,
    "cpu_l1d_cache_bytes": 0,
    "cpu_l1i_cache_bytes": 0,
    "cpu_l2_cache_bytes": 0,
    "cpu_l3_cache_bytes": 0,
    "cpu_model_name": "string",
    "cpu_thread_count": 0,
    "gpu_count": 0,
    "gpu_product_names": [
      "string"
    ],
    "kernel_name": "string",
    "memory_bytes": 0,
    "name": "string",
    "os_name": "string",
    "os_version": "string"
  },
  "name": "string",
  "reason": "string"
}
```

jgehrcke:
I built a test to POST a run with a context property. It failed schema validation:
```
    def test_create_run_empty_context(self, client):
        self.authenticate(client)
        payload = _fixtures.VALID_RUN_PAYLOAD.copy()
        payload["context"] = {}
        resp = client.post(self.url, json=payload)
        print(resp.text)
>       assert resp.status == 201
E       AssertionError: assert '400 BAD REQUEST' == 201
[...]

PR #438, opened 2022-11-17
tests: BenchmarkCreate: cover empty context scenario

jgehrcke:
Tiny test coverage extension, motivated by https://github.com/conbench/conbench/issues/365.

### DIFF
--- a/conbench/tests/api/test_benchmarks.py
+++ b/conbench/tests/api/test_benchmarks.py
@@ -869,7 +869,7 @@ class TestBenchmarkPost(_asserts.PostEnforcer):
         resp = client.post(self.url, json=payload)
         assert resp.status_code == 201, f"unexpected response: {resp.text}"
 
-        bechmark_id = resp.json["id"]
+        benchmark_id = resp.json["id"]
         # Confirm that an empty context was created.
         context_url = resp.json["links"]["context"]
         resp = client.get(context_url)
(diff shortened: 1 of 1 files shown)


jonkeane:
This shouldn't block this PR itself, but any idea what's up with the coverage allegedly decreasing (and only for machine_info?) here? I can imagine adding a test like this might not increase line coverage — but decreasing makes me wonder if something is off with our coverage setup?

austin3dickey:
I've noticed that Coveralls has been saying `conbench/machine_info.py`'s coverage has been decreasing on every PR for a long time. It also doesn't display the source when you click through the links, which (IMO) makes it a much less useful tool. I'll create an issue.

conbench/tests/api/test_benchmarks.py
```diff
@@ -843,3 +843,50 @@ def test_neither_stats_nor_error_field_is_present(self, client):
...
+        self.authenticate(client)
+        payload = _fixtures.VALID_PAYLOAD.copy()
+        payload["run_id"] = _uuid()
+        payload["batch_id"] = _uuid()
+
+        payload["context"] = {}
+        resp = client.post(self.url, json=payload)
+        assert resp.status_code == 201, f"unexpected response: {resp.text}"
+
+        bechmark_id = resp.json["id"]
```
austin3dickey: nit:
```suggestion
        benchmark_id = resp.json["id"]
```
(and down below when you use it again)

austin3dickey (approved):
Thanks, looks great!

conbench/tests/api/test_benchmarks.py
jgehrcke: Thanks!

jgehrcke:
Going ahead with this -- thanks for feedback!

PR #440, opened 2022-11-17
truncate posted github messages so they fit

austin3dickey:
This PR truncates the github message in `d3515ecea53d_backfill_all_commits.py` so it always fits in the database.

### DIFF
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -403,7 +403,7 @@ class GitHub:
         return {
             "parent": commit["parents"][0]["sha"] if commit["parents"] else None,
             "date": dateutil.parser.isoparse(commit_author["date"]),
-            "message": commit["commit"]["message"].split("\n")[0],
+            "message": commit["commit"]["message"].split("\n")[0][:240],
             "author_name": commit_author["name"],
             "author_login": author["login"] if author else None,
             "author_avatar": author["avatar_url"] if author else None,
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -231,7 +231,7 @@ def backfill_default_branch_commits(
         try:
             commit = Commit.create_github_context(**commit_info)
             backfilled_commits.append(commit)
-        except s.exc.IntegrityError as e:
+        except Exception as e:
             if "commit_index" in str(e):
                 print(f"Commit {commit_info['sha']} already existed in the database")
             else:
--- a/conbench/tests/entities/test_commit.py
+++ b/conbench/tests/entities/test_commit.py
@@ -66,6 +66,11 @@ def test_get_github_commit_none():
 )
 def test_get_github_commit_and_fork_point_sha(branch):
     # NOTE: This integration test intentionally hits GitHub.
+    if os.getenv("BUILDKITE"):
+        pytest.skip(
+            "Buildkite doesn't have a GITHUB_API_TOKEN so we won't hit the GitHub API"
+        )
+
     repo = "https://github.com/apache/arrow"
     sha = "3decc46119d583df56c7c66c77cf2803441c4458"
     tz = dateutil.tz.tzutc()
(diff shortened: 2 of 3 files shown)


conbench/entities/commit.py
```diff
@@ -403,7 +403,7 @@ def _parse_commit(commit):
         return {
             "parent": commit["parents"][0]["sha"] if commit["parents"] else None,
             "date": dateutil.parser.isoparse(commit_author["date"]),
-            "message": commit["commit"]["message"].split("\n")[0],
+            "message": commit["commit"]["message"].split("\n")[0][:240],
```
jonkeane, line 406: Did we truncate these elsewhere before the recent changes? I imagine we had to have, right?

jonkeane (approved):
Thanks for this, and thanks for the test too. One question (mostly curiosity on my part)

conbench/entities/commit.py
austin3dickey: No! Isn't that crazy? We've just never ingested a commit with a message whose first line is over 250 characters. The backfill exposed one.

ElenaHenderson (approved):

conbench/entities/commit.py
jonkeane: Huh! I wonder if maybe the previous endpoint we were using automatically trimmed them and we didn't realize? 

I find it really surprising that we would only just now bump into it, but it is possible!

conbench/entities/commit.py
```diff
@@ -231,7 +231,7 @@ def backfill_default_branch_commits(
         try:
             commit = Commit.create_github_context(**commit_info)
             backfilled_commits.append(commit)
-        except s.exc.IntegrityError as e:
+        except Exception as e:
```
austin3dickey, line 234: This fixes another issue we were seeing. I thought a `s.exc.IntegrityError` would be raised, but our server was raising a `sqlalchemy.exc.PendingRollbackError`. So this try/except was not catching this expected error. The PendingRollbackError still has `"commit_index" in str(e)` so this change shouldn't cause any false negatives or false positives.

This is pretty hard to test, though.

jonkeane (approved):

conbench/entities/commit.py
austin3dickey: I just queried all commits to apache/arrow, and here's the distribution of length of the first line of commit messages. Only 4 commits ever were over 250 characters!

image

conbench/entities/commit.py
jgehrcke: > This fixes another issue we were seeing.

Would be great to have a super lightweight ticket about that one and to add a code comment in this over-generalized exception handler about _why_ it's over-generalized.

I think that it's also important to pull `backfilled_commits.append(commit)` out of this exception handler (it's just a regular list append, right?) just to make this code more readable.

Even if `if "commit_index" in str(e):` seems to be more or less precise it's always super valuable to log some exception details for future debugging/reasoning, and to be rather neutral when re-writing error messages (there is probably more than one type error message emitted by the stack that might after all contain the sub string `commit_index`)

Example, just to demonstrate briefly:

```python
try:
    commit = Commit.create_github_context(**commit_info)
except Exception as e:
    # Over-generalized exception handler so that this becomes a best effort approach.
    # GitHub commit could not be inserted into database, probably as of unique
    # key constraint. Regardless of the reason, proceed with the next commit.
    # Log the exact reason, however:
    log.info(f"commit {commit_info['sha']} could not be inserted as of error: ...")
    # next loop iteration (process next to-be-backfilled commit)
    continue

log.info(f"inserted commit {commit_info['sha']} to database")
backfilled_commits.append(commit)
```

Another related thought is that we maybe want to move the conflict error handling into the database, and keep interaction with the database at a minimum: one could try to insert all these commits in the _same_ database transaction, ignoring conflicts. Postgres has a feature for that, exposed via SQLAlchemy. I created a tiny issue for tracking that idea: https://github.com/conbench/conbench/issues/441.

austin3dickey: Agreed; I should have done this before I merged! I like #441 better, so I'll hold off on a followup PR until we tackle that issue.

PR #443, opened 2022-11-21
Allow all robots so we can fix incorrect Google search results for conbench

ElenaHenderson:
Googling `conbench` returns this:

image

I disallowed robots about a year ago because they were causing all sorts of load on public conbench.
I am allowing robots again to see if this will fix an issue with google results showing broken link

### DIFF
--- a/conbench/app/robots.py
+++ b/conbench/app/robots.py
@@ -5,7 +5,7 @@ from ..app._endpoint import AppEndpoint
 
 text = """
 User-Agent: *
-Disallow: /
+Allow: /
 """
 
 
--- a/conbench/tests/app/test_robots.py
+++ b/conbench/tests/app/test_robots.py
@@ -1,6 +1,6 @@
 expected_text = """
 User-Agent: *
-Disallow: /
+Allow: /
 """
 
 


austin3dickey (approved):

jgehrcke:
> fix an issue with google results showing broken link

In the screenshot above it's not really obvious to me which link is broken.

I think the question here is if individual Conbench instances should be indexed by Google / search engines at all. If the previous decision was that they should not be indexed, then I concur and the two `http://3.13.114... ` entries above can be looked at as bugs.

The `noindex` concept is used for that:
https://developers.google.com/search/docs/crawling-indexing/block-indexing

note:

>  For the noindex directive to be effective, the page or resource must not be blocked by a robots.txt file, and it has to be otherwise accessible to the crawler. If the page is blocked by a robots.txt file or the crawler can't access the page, the crawler will never see the noindex directive, and the page can still appear in search results, for example if other pages link to it.

ElenaHenderson:
The 3rd link on a screenshot (http://3.13.114.18/) takes a user to a broken conbench deployment. It looks like a load balancer or nginx without any instances. The reason I now want conbench.ursa.dev to be indexed is because I don't want anybody who searches for conbench to think that http://3.13.114.18/ = conbench.ursa.dev

It is really up to whoever runs Conbench service to decide if they want their conbench page to be indexed.
We did not really think about if we want or do not want conbench.ursa.dev to be indexed. We used to have it indexed (without really even thinking about it) and later I disabled it because bots caused some temp performance issues.

jgehrcke:
> It is really up to whoever runs Conbench service to decide if they want their conbench page to be indexed.

Yeah, we can emit `noindex` by default but we can introduce a config parameter for operators to change that.

PR #446, opened 2022-11-21
Always upsert commits with github info

### LINKED ISSUE #441: DB bulk operations (such as commit backfill): consider using PostgreSQL upserts
Motivated by [this loop](https://github.com/conbench/conbench/blob/a797b8a271b7c27b25791e92be2b7eb97af3422a/conbench/entities/commit.py#L228):

```python
    for commit_info in commits_to_try:
    ...
        try:
            commit = Commit.create_github_context(**commit_info)
            backfilled_commits.append(commit)
        except Exception as e:
            if "commit_index" in str(e):
                print(f"Commit {commit_info['sha']} already existed in the database")
            else:
                raise

    return backfilled_commits
```
This issues a database transaction for each loop iteration, and in each iteration we try to handle a key constraint error. This can be done by the database:

> Starting with version 9.5, PostgreSQL allows “upserts” (update or insert) of rows into a table via the ON CONFLICT clause of the INSERT statement. A candidate row will only be inserted if that row does not violate any unique constraints. In the case of a unique constraint violation, a secondary action can occur which can be either “DO UPDATE”, indicating that the data in the target row should be updated, or “DO NOTHING”, which indicates to silently skip this row.

(https://docs.sqlalchemy.org/en/14/dialects/postgresql.html#insert-on-conflict-upsert)

austin3dickey:
Fixes #441. See discussion there.

This PR changes `Commit.create_github_context()` to use an upsert instead of an insert, so there's no chance of the unique constraint on (repository, sha) causing a database transaction rollback. Before this PR, when `Commit.create_github_context()` was used in `backfill_default_branch_commits()`, it was wrapped in a try/except, but the rollbacks could still affect other pending operations on the server.

This PR also wraps the usage of `backfill_default_branch_commits()` inside of `Run.create()` in a try/except, because as discussed, we shouldn't return a 500 if the run was created but commits weren't backfilled.

After this PR, we shouldn't see any 500s due to commit backfills anymore.

### DIFF
--- a/conbench/entities/_entity.py
+++ b/conbench/entities/_entity.py
@@ -1,9 +1,10 @@
 import functools
 import uuid
+from typing import List
 
 import flask as f
 from sqlalchemy import Column, distinct
-from sqlalchemy.dialects.postgresql import insert
+from sqlalchemy.dialects.postgresql import insert as postgresql_insert
 from sqlalchemy.orm import declarative_base
 from sqlalchemy.orm.exc import NoResultFound
 
@@ -93,13 +94,11 @@ class EntityMixin:
         return entity
 
     @classmethod
-    def upsert_do_nothing(cls, data):
-        """Try to insert a row. If there is a conflict, do nothing."""
-        statement = insert(cls).values([data]).on_conflict_do_nothing()
+    def upsert_do_nothing(cls, row_list: List[dict]):
+        """Try to insert rows. If there is a conflict on any row, ignore that row."""
+        statement = postgresql_insert(cls).values(row_list).on_conflict_do_nothing()
         Session.execute(statement)
         Session.commit()
-        # return the row that (now) exists in the database
-        return cls.first(**data)
 
     @classmethod
     def bulk_save_objects(self, bulk):
--- a/conbench/entities/_entity.py
+++ b/conbench/entities/_entity.py
@@ -3,6 +3,7 @@ import uuid
 
 import flask as f
 from sqlalchemy import Column, distinct
+from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.orm import declarative_base
 from sqlalchemy.orm.exc import NoResultFound
 
@@ -90,6 +91,15 @@ class EntityMixin:
         entity = cls(**data)
         entity.save()
         return entity
+
+    @classmethod
+    def upsert_do_nothing(cls, data):
+        """Try to insert a row. If there is a conflict, do nothing."""
+        statement = insert(cls).values([data]).on_conflict_do_nothing()
+        Session.execute(statement)
+        Session.commit()
+        # return the row that (now) exists in the database
+        return cls.first(**data)
 
     @classmethod
     def bulk_save_objects(self, bulk):
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -67,7 +67,7 @@ class Commit(Base, EntityMixin):
 
     @staticmethod
     def create_github_context(sha, repository: str, github: dict):
-        return Commit.create(
+        return Commit.upsert_do_nothing(
             {
                 "sha": sha,
                 "branch": github["branch"],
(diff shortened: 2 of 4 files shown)


jgehrcke:
When I typed up #441 the main idea really was to send all `commits_to_try` in the _same_ upsert/insert statement, i.e. in a single database transaction (as opposed to issuing as many transactions as there are commits, which I think the current patch would still be doing). Sorry if that wasn't clearly worded :).

Relevant SQLAlchemy docs chapter:
https://docs.sqlalchemy.org/en/20/orm/queryguide/dml.html#

I think this code example shows a useful multi-row upsert (for SQLite in this case):
https://docs.sqlalchemy.org/en/20/orm/queryguide/dml.html#orm-upsert-statements

```python
>>> from sqlalchemy.dialects.sqlite import insert as sqlite_upsert
>>> stmt = sqlite_upsert(User).values(
...     [
...         {"name": "spongebob", "fullname": "Spongebob Squarepants"},
...         {"name": "sandy", "fullname": "Sandy Cheeks"},
...         {"name": "patrick", "fullname": "Patrick Star"},
...         {"name": "squidward", "fullname": "Squidward Tentacles"},
...         {"name": "ehkrabs", "fullname": "Eugene H. Krabs"},
...     ]
... )
>>> stmt = stmt.on_conflict_do_update(
...     index_elements=[User.name], set_=dict(fullname=stmt.excluded.fullname)
... )
>>> session.execute(stmt)
```

conbench/entities/commit.py
```diff
@@ -228,14 +228,9 @@ def backfill_default_branch_commits(
     for commit_info in commits_to_try:
         commit_info["github"]["branch"] = default_branch
         commit_info["github"]["fork_point_sha"] = commit_info["sha"]
-        try:
-            commit = Commit.create_github_context(**commit_info)
+        commit = Commit.create_github_context(**commit_info)
```
jgehrcke: About the removal of the exception handler:

While the row conflict would here now be eliminated as a reason for this statement to raise an exception -- the individual database transaction can of course still fail with a whole range of transient errors. Without exception handler in the loop body that means that an individual transient error may prematurely crash the `for commit_info in commits_to_try` loop.

A good way to address this is to issue 1 instead of N database transactions, as suggested by my higher level review comment on this pull request! :)

austin3dickey: I'm not too worried about prematurely crashing this loop, since the backfill will be attempted again later, but I agree that one transaction lessens the chance of something going wrong and also probably speeds up this function a little. I'll change it now!

conbench/tests/entities/test_commit.py
```diff
@@ -161,10 +181,11 @@ def test_backfill_default_branch_commits():
     )
 
     # this should backfill all 335 default-branch commits before that one
-    backfilled_commits = backfill_default_branch_commits(repository, commit_1)
```
austin3dickey, line 164: The changes to this test are because `backfill_default_branch_commits()` no longer returns anything, because we can't tell which commits were inserted and which were ignored during the upsert transaction.

conbench/entities/commit.py
austin3dickey: Okay, I changed to a single transaction. Let me know what you think!

jgehrcke: Thanks!

conbench/tests/entities/test_commit.py
jgehrcke: > because we can't tell which commits were inserted and which were ignored during the upsert transaction.

For the record: if we ever need this, then I believe the sqlalchemy docs show how to get a result from this single transaction, showing which rows were newly inserted.

conbench/entities/_entity.py
```diff
@@ -91,6 +93,13 @@ def create(cls, data):
         entity.save()
         return entity
 
+    @classmethod
+    def upsert_do_nothing(cls, row_list: List[dict]):
```
jgehrcke, line 97: Lovely, thanks for the type annotation here :heart:.

jgehrcke (approved):
:rocket: cool! Let's go ahead with this.

conbench/tests/entities/test_commit.py
austin3dickey: I briefly looked in the docs yesterday, but couldn't quite find an example of what I wanted, especially for `on_conflict_do_nothing()` (as opposed to `on_conflict_do_update()`). However I'm sure it's possible to use `.returning()` in some way that it returns only the commits that were inserted. We'd just have to look into it a little more.

PR #451, opened 2022-11-23
Introduce stdlib-based logging infrastructure

jgehrcke:
Baby steps towards using Python's stdlib `logging` for more insights and debugging while feature development.

I have split this out of another branch that I am developing in locally.

Hope that doesn't look intimidating. Approach is based on years of experience, code is similar to what we did [before](https://github.com/dcos/bouncer/blob/f2116912adc43dafbb4b96668ebca17b3f924603/bouncer/logging.py).

The high-level config interface for now has three parameters:
```
    LOG_LEVEL_STDERR = os.environ.get("CONBENCH_LOG_LEVEL_STDERR", "INFO")
    LOG_LEVEL_FILE = None
    LOG_LEVEL_SQLALCHEMY = "WARNING"
```

That means that by default no log file is created, the stderr log level is INFO. The latter can be overridden with the environment variable `CONBENCH_LOG_LEVEL_STDERR`. I override that to DEBUG in my local test environment.

### DIFF
--- a/conbench/__init__.py
+++ b/conbench/__init__.py
@@ -1,6 +1,9 @@
 import importlib.metadata as importlib_metadata
 import json
+import logging
 import os
+
+import conbench.logger
 
 try:
     __version__ = importlib_metadata.version(__name__)
@@ -10,20 +13,31 @@ except Exception:
 del importlib_metadata
 
 
+# Pre-configure logging before reading user-given configuration.
+conbench.logger.setup(level_stderr="DEBUG", level_file=None)
+log = logging.getLogger(__name__)
+
+
 def create_application(config):
     import flask as f
 
-    application = f.Flask(__name__)
-    application.config.from_object(config)
+    app = f.Flask(__name__)
+    app.config.from_object(config)
 
-    _init_application(application)
+    _init_application(app)
 
-    if application.config["DEBUG"]:
-        print(
-            f"* Flask config:\n{json.dumps(application.config, sort_keys=True, default=str, indent=2)}"
-        )
+    # Re-configure logging using user-given configuration details.
+    log.debug("re-configure logging")
+    conbench.logger.setup(
+        level_stderr=app.config["LOG_LEVEL_STDERR"],
+        level_file=app.config["LOG_LEVEL_FILE"],
+    )
 
-    return application
+    log.debug(
+        "flask app config:\n%s",
+        json.dumps(app.config, sort_keys=True, default=str, indent=2),
+    )
+    return app
 
 
 def _init_application(application):
--- a/conbench/logger.py
+++ b/conbench/logger.py
@@ -1,0 +1,94 @@
+"""
+Set up Python's logging infrastructure for the web application.
+
+Acquire root logger instance and populate it with handlers, applying a certain
+log format. This takes effect for Conbench code as well as integrated
+libraries, such as SQLAlchemy
+
+A package or module within Conbench should obtain and use a logger instance in
+the 'conbench' sub-tree of the logging namespace hierarchy, via e.g.
+
+    log = logging.getLogger('conbench.xxx')
+"""
+
+
+import logging
+import logging.handlers
+
+
+def setup(
+    level_stderr="ERROR",
+    level_file="DEBUG",
+    logfilepath="conbench.log",
+    level_sqlalchemy="WARNING",
+):
+    """Set up root logger for the Conbench application.
+
+    Args:
+        level_stderr: `None` or a valid log level string.
+        level_file: `None` or a valid log level string.
+        logfilepath: Path to where a log file should be opened
+            in append mode. Takes effect only when `level_file`
+            defines a logging level.
+
+    Valid log level strings:
+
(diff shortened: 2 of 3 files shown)


jgehrcke:
Subsequent PRs that I would love to open right now make use of this change -- I am super open to changing things after review, but would for now love to go ahead and merge this now that CI went green. [EAFP](https://en.wikipedia.org/wiki/EAFP) *cough*.

conbench/logger.py
```diff
@@ -0,0 +1,94 @@
...
+    if not handlers:
+        # Mute root logger (prevent logging.lastResort from taking effect).
+        rootlog.addHandler(logging.NullHandler())
+
+    sa_logger = logging.getLogger("sqlalchemy.engine")
+    sa_logger.setLevel(getattr(logging, level_sqlalchemy))
+
+    # Do not show log msgs like
+    # [urllib3.connectionpool] DEBUG: Starting new HTTP connection (1): ....
+    logging.getLogger("urllib3.connectionpool").setLevel("INFO")
```
jonkeane, line 94: This is fantastic, thank you!

jonkeane (commented):
> That means that by default no log file is created, the stderr log level is INFO

In this default case, will the logging be sent to stdout (or stderr) then? Mostly I want to make sure we will still be capturing some diagnostics with datadog like we are from instances currently

conbench/logger.py
```diff
@@ -0,0 +1,94 @@
...
+Set up Python's logging infrastructure for the web application.
+
+Acquire root logger instance and populate it with handlers, applying a certain
+log format. This takes effect for Conbench code as well as integrated
+libraries, such as SQLAlchemy
+
+A package or module within Conbench should obtain and use a logger instance in
+the 'conbench' sub-tree of the logging namespace hierarchy, via e.g.
+
+    log = logging.getLogger('conbench.xxx')
```
austin3dickey, line 11: Is this equivalent to `log = logging.getLogger(__name__)` or are there exceptions? If it is equivalent, maybe we should recommend using `__name__` here so it's more future-proof if we rename/reorganize the files.

conbench/__init__.py
```diff
@@ -10,20 +13,31 @@
...
 
-    if application.config["DEBUG"]:
-        print(
-            f"* Flask config:\n{json.dumps(application.config, sort_keys=True, default=str, indent=2)}"
-        )
+    # Re-configure logging using user-given configuration details.
+    log.debug("re-configure logging")
+    conbench.logger.setup(
+        level_stderr=app.config["LOG_LEVEL_STDERR"],
+        level_file=app.config["LOG_LEVEL_FILE"],
```
austin3dickey, line 33: I think you're missing `level_sqlalchemy=app.config["LOG_LEVEL_SQLALCHEMY"]`, no?

conbench/logger.py
```diff
@@ -0,0 +1,94 @@
...
+"""
+
+
+import logging
+import logging.handlers
+
+
+def setup(
+    level_stderr="ERROR",
```
austin3dickey, line 20: Right now we have defaults in 3 different places (for example, `level_stderr="ERROR"` here, `level_stderr="DEBUG"` in `__init__.py`, and `LOG_LEVEL_STDERR="INFO"` in the app config). Would you be open to removing the defaults from this internal function and explicitly passing them through in the other two files, so that it's a little clearer which level will be used at what times? Then we don't have to think as hard about "which default level got picked up in this case"

conbench/__init__.py
```diff
@@ -1,7 +1,10 @@
 import importlib.metadata as importlib_metadata
```
austin3dickey, line 1: You mentioned

>I override that to DEBUG in my local test environment.

Would you mind adding a quick example to the README, in the "run tests" section, about how you do that? It's probably a good practice for other devs as well!

austin3dickey (commented):
Just a few small comments for a potential followup PR:

conbench/logger.py
jgehrcke: There might be exceptions when `__name__` becomes too long :). I like the idea of recommending `__name__`. Will do. (but after all this is not a strict rule -- logs should be useful and what we consider useful in this context might change over time).

conbench/__init__.py
jgehrcke: Yes! Thank you!

conbench/logger.py
jgehrcke: >  Would you be open to removing the defaults from this internal function and explicitly passing them through in the other two files, so that it's a little clearer which level will be used at what times?

Yes, totally open. I agree. I also found that this smells a bit in terms of readability/transparency. I was basically just copy/pasting code around.

conbench/__init__.py
jgehrcke: Great idea! 

I will keep observing my dev flow for another while though before I feel comfortable updating the README -- right now I run tests with docker-compose, and I do the logging level change by adding
```
      CONBENCH_LOG_LEVEL_STDERR: DEBUG
```
to the app/environment section in  my local checkout of `docker-compose.yml`.

Issue #449, opened 2022-11-22
ci: docker-compose: rethink startup dependencies via condition service_healthy


jgehrcke:
The `condition: service_healthy` directives in our `docker-compose.yml` cannot be relied upon.

I tried adding a new service:

```yaml
  dex:
    image: dexidp/dex:v2.35.3
    healthcheck:
      test: "curl -f http://127.0.0.1:5556"
      interval: 10s
      timeout: 5s
      retries: 20
      start_period: 2s
```

I also added this to `app`:

```
    depends_on:
      dex:
        condition: service_healthy
```

A subsequent `docker-compose run app` resulted in:

```
...
 ---> 6c3750049d90
Successfully built 6c3750049d90
Successfully tagged conbench_app:latest

ERROR: for app  Service "dex" is missing a healthcheck configuration
ERROR: Encountered errors while bringing up the project.
```

This was a WTF moment because clearly(?) the `dex` service has a healthcheck configuration. And here we go:

https://github.com/docker/compose/issues/7973 discussed that

> This error doesn't happen with docker-compose version 1.26.2. Similar bug has existed before in https://github.com/docker/for-mac/issues/1341 Maybe this is a regression from the Merge 2.x and 3.x compose formats and align with COMPOSE_SPEC schema change? condition: service_healthy is going away with the version 3 but should work until support for 2.x versions is removed completely I believe.

but there was no further input (issue closed by stale bot).

I noticed that there is 
https://docs.docker.com/compose/compose-file/compose-file-v2/
vs
https://docs.docker.com/compose/compose-file/compose-file-v3/

and that `condition: service_healthy` is only mentioned in the v2 documentation. Equipped with that I searched the web further and indeed found that [Version 3 no longer supports the condition form of depends_on.](https://github.com/docker/compose/issues/8154#issuecomment-817760624).

We are hard-coding version 3 [here](https://github.com/conbench/conbench/blob/093e61385f9716b801ec03a773cfe281c43ce4c8/docker-compose.yml#L2). 

That is, the error message is highly misleading. It's not that `dex` does not define a healthcheck. It's that `condition: service_healthy` does not work with the v3 schema.

I love the sentiment of [this](https://stackoverflow.com/a/47714157) SO answer:

[...]

PR #452, opened 2022-11-23
db create_all: apply retrying strategy for test env

jgehrcke:
As described in https://github.com/conbench/conbench/issues/449 the docker-compose-managed 'readiness' checks are difficult to see through. Currently I run tests locally via

```
docker-compose down && \
   docker-compose build app && \
   docker-compose run app pytest -k TestXXXX -s conbench/tests
```

Rarely, that results in Postgres not yet accepting connections when the first `create_all()` is issued, resulting in a `sqlalchemy.exc.OperationalError` crashing the test suite.

Instead of figuring out how to make the container startup order/timing more robust in the context of docker-compose, I am embracing the principle where we try to auto-heal transient issues pragmatically, in the spirit of [this](https://stackoverflow.com/a/47714157).

https://github.com/jd/tenacity is the successor of https://github.com/rholder/retrying.

Example for log msgs when auto-healing this worked, with the DEBUG log level:
```
[221123-13:13:01.785] [1] [conbench.db] DEBUG: Starting call to 'conbench.db.create_all', this is the 1st time calling it.
[221123-13:13:01.786] [1] [conbench.db] INFO: result after attempt 1 for <function create_all at 0x7f2b94df0af0>: (psycopg2.OperationalError) connection to server at "db" (172.29.0.3), port 5432 failed: Connection refused
	Is the server running on that host and accepting TCP/IP connections?

(Background on this error at: https://sqlalche.me/e/14/e3q8)
[221123-13:13:02.787] [1] [conbench.db] DEBUG: Starting call to 'conbench.db.create_all', this is the 2nd time calling it.
```

### DIFF
--- a/conbench/db.py
+++ b/conbench/db.py
@@ -1,12 +1,9 @@
 import logging
 
-
 import sqlalchemy.exc
+import tenacity
 from sqlalchemy import create_engine
 from sqlalchemy.orm import scoped_session, sessionmaker
-
-import tenacity
-
 
 engine = None
 session_maker = sessionmaker(future=True)
--- a/conbench/db.py
+++ b/conbench/db.py
@@ -1,9 +1,19 @@
+import logging
+
+
+import sqlalchemy.exc
 from sqlalchemy import create_engine
 from sqlalchemy.orm import scoped_session, sessionmaker
+
+import tenacity
+
 
 engine = None
 session_maker = sessionmaker(future=True)
 Session = scoped_session(session_maker)
+
+
+log = logging.getLogger(__name__)
 
 
 def configure_engine(url):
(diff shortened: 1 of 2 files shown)


jonkeane (approved):
This is really great, thanks!

We do use retry in one other location: https://github.com/conbench/conbench/blob/main/conbench/util.py#L11 where it might make sense to transition to tenacity though that code is likely to go away, @alistaire47 is working on a project to make abstract connectors and adapters that are separate packages from conbench itself in https://github.com/conbench/conbench/tree/main/benchconnect and https://github.com/conbench/conbench/tree/main/benchadapt so maybe we leave that code like it is since it's (effectively) deprecated

jgehrcke:
Thank you for taking a look, Jon! I am going ahead and merge this for easier PR management. Super open to changing everything about this if more review comments come in.

PR #454, opened 2022-11-23
config: introduce CONBENCH_OIDC_ISSUER_URL, oidc code cleanup

jgehrcke:
What we did so far with "Google SSO" was actually a standards-compliant OpenID Connect (OIDC) single sign-on flow. That's cool.

OIDC is a standard on top of OAuth2. There's a plethora of identity providers (Saas, etc) that can serve as an OIDC provider.

An OIDC provider is identified by its issuer URL, or just the issuer. In case of Google, the issuer is `https://accounts.google.com`

So -- an OpenID Connect (OIDC) issuer URL precisely identifies an OIDC identity provider. Make this configurable, for testing and more flexible deployments.

This PR tries to explicitly support the legacy/current config scenario where the environment shows a `GOOGLE_CLIENT_ID` (and secret) -- which has so far implied that in the corresponding deployment we want to use `https://accounts.google.com` as issuer. 

This PR also touches on a few variable names etc to make the code less gooogly and more openidy, and to help towards better readability.

### DIFF
--- a/conbench/api/_google.py
+++ b/conbench/api/_google.py
@@ -1,44 +1,72 @@
 import json
+import logging
 import os
+import time
 
 import flask as f
 import requests
 
+from ..config import Config
 
-def get_google_config():
+log = logging.getLogger(__name__)
+
+# Rely on INTENDED_BASE_URL to have a trailing slash.
+OIDC_CALLBACK_URL = Config.INTENDED_BASE_URL + "api/google/callback"
+
+
+def get_oidc_config():
     client_id = os.environ.get("GOOGLE_CLIENT_ID", None)
     client_secret = os.environ.get("GOOGLE_CLIENT_SECRET", None)
-    discovery_url = "https://accounts.google.com/.well-known/openid-configuration"
+    discovery_url = Config.OIDC_ISSUER_URL + "/.well-known/openid-configuration"
     return discovery_url, client_id, client_secret
 
 
-def get_google_client():
+def get_oidc_client():
     from oauthlib.oauth2 import WebApplicationClient
 
-    discovery_url, client_id, _ = get_google_config()
-    google = requests.get(discovery_url).json()
+    discovery_url, client_id, _ = get_oidc_config()
+
+    # Pragmatic healing for transient errors. Better: cache OP config across
+    # login requests. Also, if we want to retry here in the future: consider
+    # using tenacity.
+    for attempt in range(4):
+        try:
+            oidc_provider_config = requests.get(discovery_url).json()
+            break
+        except requests.exceptions.RequestException as exc:
+            log.info("err getting OP config (attempt %s): %s -- retry", attempt, exc)
+            time.sleep(2)
+
     client = WebApplicationClient(client_id)
-    return client, google
+    return client, oidc_provider_config
 
 
 def auth_google_user():
-    client, google = get_google_client()
-    redirect_uri = f.url_for("api.callback", _external=True, _scheme="https")
+    client, oidc_provider_config = get_oidc_client()
+
+    # http://127.0.0.1:5000/api/google/callback
+    # redirect_uri = f.url_for("api.callback", _external=True, _scheme="https")
+
+    # If either redirect URL or the authorization endpoint (at the OP) do not
+    # use the HTTPS scheme then the function below is expected to throw
+    # `InsecureTransportError`. For testing, this can be changed by setting
+    # the environment variable OAUTHLIB_INSECURE_TRANSPORT.
+
     return client.prepare_request_uri(
-        google["authorization_endpoint"],
-        redirect_uri=redirect_uri,
+        oidc_provider_config["authorization_endpoint"],
+        redirect_uri=OIDC_CALLBACK_URL,
         scope=["openid", "email", "profile"],
     )
 
 
 def get_google_user():
-    client, google = get_google_client()
-    _, client_id, client_secret = get_google_config()
+    client, oidc_provider_config = get_oidc_client()
+    _, client_id, client_secret = get_oidc_config()
 
     token_url, headers, body = client.prepare_token_request(
-        google["token_endpoint"],
-        authorization_response=f.request.url.replace("http://", "https://"),
-        redirect_url=f.request.base_url.replace("http://", "https://"),
+        oidc_provider_config["token_endpoint"],
+        authorization_response=f.request.url,  # .replace("http://", "https://"),
+        redirect_url=f.request.base_url,  # replace("http://", "https://"),
         code=f.request.args.get("code"),
     )
     token_response = requests.post(
@@ -49,7 +77,7 @@ def get_google_user():
     )
     client.parse_request_body_response(json.dumps(token_response.json()))
 
-    uri, headers, body = client.add_token(google["userinfo_endpoint"])
+    uri, headers, body = client.add_token(oidc_provider_config["userinfo_endpoint"])
     return requests.get(
         uri,
         headers=headers,
--- a/conbench/config.py
+++ b/conbench/config.py
@@ -53,6 +53,22 @@ class Config:
     LOG_LEVEL_FILE = None
     LOG_LEVEL_SQLALCHEMY = "WARNING"
 
+    # If `OIDC_ISSUER_URL` is after all `None`: disable OpenID Connect (OIDC)
+    # single sign-on. If this is not `None` then it must be a valid OIDC issuer
+    # notation (that is, a URL).
+    OIDC_ISSUER_URL = os.environ.get("CONBENCH_OIDC_ISSUER_URL", None)
+    if OIDC_ISSUER_URL is not None:
+        assert OIDC_ISSUER_URL.startswith("http")
+        # Remove all trailing slashes.
+        OIDC_ISSUER_URL = OIDC_ISSUER_URL.rstrip("/")
+    else:
+        # legacy config support: when CONBENCH_OIDC_ISSUER_URL is set in the
+        # environment it takes precedence. If it is not set and if
+        # GOOGLE_CLIENT_ID is set in the environment then set issuer to
+        # "https://accounts.google.com"
+        if "GOOGLE_CLIENT_ID" in os.environ:
+            OIDC_ISSUER_URL = "https://accounts.google.com"
+
 
 class TestConfig(Config):
     DB_NAME = os.environ.get("DB_NAME", f"{APPLICATION_NAME.lower()}_test")
--- a/conbench/api/auth.py
+++ b/conbench/api/auth.py
@@ -92,7 +92,9 @@ class CallbackAPI(ApiEndpoint):
         try:
             google_user = _google.get_google_user()
         except Exception as e:
-            self.abort_400_bad_request(f"OpenID Connect single sign-on flow failed: {e}.")
+            self.abort_400_bad_request(
+                f"OpenID Connect single sign-on flow failed: {e}."
+            )
 
         email = google_user["email"]
         given = google_user["given_name"]
--- a/conbench/api/auth.py
+++ b/conbench/api/auth.py
@@ -92,7 +92,7 @@ class CallbackAPI(ApiEndpoint):
         try:
             google_user = _google.get_google_user()
         except Exception as e:
-            self.abort_400_bad_request(f"Google SSO failed {e}.")
+            self.abort_400_bad_request(f"OpenID Connect single sign-on flow failed: {e}.")
 
         email = google_user["email"]
         given = google_user["given_name"]


conbench/api/_google.py
```diff
@@ -1,44 +1,72 @@
...
 import flask as f
 import requests
 
+from ..config import Config
 
-def get_google_config():
+log = logging.getLogger(__name__)
+
+# Rely on INTENDED_BASE_URL to have a trailing slash.
+OIDC_CALLBACK_URL = Config.INTENDED_BASE_URL + "api/google/callback"
```
jgehrcke, line 14: Here we make use of https://github.com/conbench/conbench/blob/d4030f80bbfba0484104d01db3124c75206ba2aa/conbench/config.py#L50

conbench/api/_google.py
```diff
@@ -1,44 +1,72 @@
...
+
+
+def get_oidc_config():
     client_id = os.environ.get("GOOGLE_CLIENT_ID", None)
     client_secret = os.environ.get("GOOGLE_CLIENT_SECRET", None)
-    discovery_url = "https://accounts.google.com/.well-known/openid-configuration"
+    discovery_url = Config.OIDC_ISSUER_URL + "/.well-known/openid-configuration"
```
jgehrcke, line 20: I think this change shows pretty well what this PR is mostly about.

conbench/api/_google.py
```diff
@@ -1,44 +1,72 @@
...
+        try:
+            oidc_provider_config = requests.get(discovery_url).json()
+            break
+        except requests.exceptions.RequestException as exc:
+            log.info("err getting OP config (attempt %s): %s -- retry", attempt, exc)
+            time.sleep(2)
+
     client = WebApplicationClient(client_id)
-    return client, google
+    return client, oidc_provider_config
```
jgehrcke, line 41: And maybe for clarity and for serving curiosity: the provider config is a standardized JSON document served at `http....well-known/openid-configuration`, example:

```json
{
 "issuer": "https://accounts.google.com",
 "authorization_endpoint": "https://accounts.google.com/o/oauth2/v2/auth",
 "device_authorization_endpoint": "https://oauth2.googleapis.com/device/code",
 "token_endpoint": "https://oauth2.googleapis.com/token",
 "userinfo_endpoint": "https://openidconnect.googleapis.com/v1/userinfo",
 "revocation_endpoint": "https://oauth2.googleapis.com/revoke",
 "jwks_uri": "https://www.googleapis.com/oauth2/v3/certs",
 "response_types_supported": [
  "code",
  "token",
  "id_token",
  "code token",
  "code id_token",
  "token id_token",
  "code token id_token",
  "none"
 ],
 "subject_types_supported": [
  "public"
 ],
 "id_token_signing_alg_values_supported": [
  "RS256"
 ],
 "scopes_supported": [
  "openid",
  "email",
  "profile"
 ],
 "token_endpoint_auth_methods_supported": [
  "client_secret_post",
  "client_secret_basic"
 ],
 "claims_supported": [
  "aud",
  "email",
  "email_verified",
  "exp",
  "family_name",
  "given_name",
  "iat",
  "iss",
  "locale",
  "name",
  "picture",
  "sub"
 ],
 "code_challenge_methods_supported": [
  "plain",
  "S256"
 ],
 "grant_types_supported": [
  "authorization_code",
  "refresh_token",
  "urn:ietf:params:oauth:grant-type:device_code",
  "urn:ietf:params:oauth:grant-type:jwt-bearer"
 ]
}
```

This fully specifies the feature set and behavior of the OpenID Connect provider.

conbench/api/_google.py
```diff
@@ -1,44 +1,72 @@
...
 
 
 def auth_google_user():
-    client, google = get_google_client()
-    redirect_uri = f.url_for("api.callback", _external=True, _scheme="https")
```
jgehrcke, line 26: For clarity, instead of

```python
redirect_uri = f.url_for("api.callback", _external=True, _scheme="https")
```
we now do
```python
Config.INTENDED_BASE_URL + "api/google/callback"
```
at the top of this module.

jgehrcke: The `api/google` part of that URL maybe should at some point change into something less googly for cosmetic reasons.

conbench/api/_google.py
```diff
@@ -1,44 +1,72 @@
...
+    client, oidc_provider_config = get_oidc_client()
+    _, client_id, client_secret = get_oidc_config()
 
     token_url, headers, body = client.prepare_token_request(
-        google["token_endpoint"],
-        authorization_response=f.request.url.replace("http://", "https://"),
-        redirect_url=f.request.base_url.replace("http://", "https://"),
+        oidc_provider_config["token_endpoint"],
+        authorization_response=f.request.url,  # .replace("http://", "https://"),
+        redirect_url=f.request.base_url,  # replace("http://", "https://"),
```
jgehrcke, line 69: Those schema rewrites were a way to work around the `InsecureTransportError` explained in a code comment above.

A tidier way to deal with this error is the `OAUTHLIB_INSECURE_TRANSPORT` environment variable, as also described in said code comment.

The rewrites, I believe, were only necessary when the current state was initially developed, i.e. with Conbench running locally, using Google's OIDC provider.

conbench/api/_google.py
jonkeane: This _is_ a fun minimal change to go from single source -> OIDC I wish all abstraction PRs were as simple as this! 🎉

conbench/api/_google.py
```diff
@@ -1,44 +1,72 @@
...
+def get_oidc_client():
     from oauthlib.oauth2 import WebApplicationClient
 
-    discovery_url, client_id, _ = get_google_config()
-    google = requests.get(discovery_url).json()
+    discovery_url, client_id, _ = get_oidc_config()
+
+    # Pragmatic healing for transient errors. Better: cache OP config across
+    # login requests. Also, if we want to retry here in the future: consider
+    # using tenacity.
```
jonkeane, line 31: This is super minor, and I don't really mind either way since this doesn't sound quite like a thing we will definitely do in the future, but I like calling TODOs out with `# TODO:` since many IDEs flag those | makes them easier to search.  Up to you if you think using tenacity is really a to-do or if it's more a "if we have issues with this code, we might consider"

conbench/api/_google.py
```diff
@@ -1,44 +1,72 @@
...
+    return client, oidc_provider_config
 
 
 def auth_google_user():
-    client, google = get_google_client()
-    redirect_uri = f.url_for("api.callback", _external=True, _scheme="https")
+    client, oidc_provider_config = get_oidc_client()
+
+    # http://127.0.0.1:5000/api/google/callback
+    # redirect_uri = f.url_for("api.callback", _external=True, _scheme="https")
```
jonkeane, line 48: I can't quite tell if this is a comment or old code that we should remove?

conbench/api/_google.py
jonkeane: _nods_ cool, we should be able to remove them totally then, yeah?

jonkeane (commented):
Thanks for this, I especially appreciate all the comments (both in code and on the PR), makes reviewing a breeze!

For deployed instances, the URL | entity at https://github.com/conbench/conbench/blob/d4030f80bbfba0484104d01db3124c75206ba2aa/conbench/templates/login.html#L15 will stay the same (until we add a new provider), yeah?

conbench/config.py
```diff
@@ -53,6 +53,22 @@ class Config:
...
     LOG_LEVEL_SQLALCHEMY = "WARNING"
 
+    # If `OIDC_ISSUER_URL` is after all `None`: disable OpenID Connect (OIDC)
+    # single sign-on. If this is not `None` then it must be a valid OIDC issuer
+    # notation (that is, a URL).
+    OIDC_ISSUER_URL = os.environ.get("CONBENCH_OIDC_ISSUER_URL", None)
+    if OIDC_ISSUER_URL is not None:
+        assert OIDC_ISSUER_URL.startswith("http")
+        # Remove all trailing slashes.
+        OIDC_ISSUER_URL = OIDC_ISSUER_URL.rstrip("/")
```
jgehrcke, line 63: Oh, and in `Config.OIDC_ISSUER_URL + "/.well-known/openid-configuration"` we rely on the OIDC issuer URL to _not_ have trailing slashes.

I think that issuer URLs are usually provided w/o trailing slash (for example, see above's JSON doc representing Google's OIDC provider). But just in case someone adds a trailing slash we remove it here again.

conbench/api/_google.py
jgehrcke: Yes! Will do in one of the next PRs.

conbench/api/_google.py
jgehrcke: It's meant to be a comment, but so far it isn't really. Right! Will make this a useful comment in one of the next patches.

conbench/api/_google.py
jgehrcke: >  I like calling TODOs out with # TODO

Good to know. Me, too! :)  In this case indeed I just wanted to quickly dump my thoughts, it's basically a "think about this again" todo item and because it's so unspecific and not so important I found a TODO label a little too fat.

jgehrcke:
> `<a href="{{ url_for('api.google') }}">Google Login</a> ` ... will stay the same (until we add a new provider), yeah?

Yes, this is independent of the identity provider. It is the provider-agnostic entry point into the OIDC flow -- the HTTP response generated by this endpoint (served by the conbench app) is a redirection response to the identity provider.

As commented somewhere in here it makes sense to rename this endpoint so that it's less googly :) (for purely cosmetic/style reasons).

jgehrcke:
Thanks for the lovely feedback here. I will go ahead and merge this, so that we can transition to the more interesting part of this series of patches -- introducing Dex! :)

jgehrcke:
Technical retrospective.

This patch broke the OIDC-based login flow against Google in those Conbench instances that had it enabled. Users that were logged in were not affected. Functionality other than login was not affected.

A user reported that problem (thanks @jihoonson). Error screenshot:



That is, the critical error detail is `redirect_uri_mismatch`.

(note: the identity provider is configured with the _expected_ redirect URL, by an operator. it rejects any oncoming so-called 'authorization request' when the redirect URL that is part of that request does not match the URL that it has been configured with, therefore the mismatch).

How did that happen? 

With this PR, the OIDC callback URL started relying on `Config.INTENDED_BASE_URL` which defaults to `http://127.0.0.1:5000`. The affected deployments did not set this new configuration variable and therefore started generating 

```
http://127.0.0.1:5000/api/google/callback
```
instead of e.g.
```
https://conbench.ursa.dev/api/google/callback
```

How did callback URL generation work previously? It relied on this line of magic:

```
redirect_uri = f.url_for("api.callback", _external=True, _scheme="https")
```

[Here](https://tedboy.github.io/flask/_modules/flask/helpers.html#url_for), that is documented via

```
    :param _external: if set to ``True``, an absolute URL is generated. Server
      address can be changed via ``SERVER_NAME`` configuration variable which
      defaults to `localhost`.
```

However, we do not set Flask's `SERVER_NAME` configuration variable in any of our deployments and yet the generated value was not using `localhost`. So, the documentation I just quoted does not seem to be entirely correct. There seems to be undocumented magic behavior that makes Flask _try to_ derive a meaningful IP address or DNS name when `SERVER_NAME` is not set -- and in our case, this magic behavior clearly worked in the past, and we relied on that.

I will look into that further, it's really interesting to understand why this worked in the first place at all.

jgehrcke:
In the latest [docs for `app.url_for()`](
https://github.com/pallets/flask/blob/cc66213e579d6b35d9951c21b685d0078f373c44/src/flask/app.py#L1917) we get to the more interesting behavioral specification, emphasis mine:

> When not in an active request, URLs will be external by default, but this requires setting `SERVER_NAME` so Flask knows what domain to use. `APPLICATION_ROOT` and `PREFERRED_URL_SCHEME` should also be configured as needed. **This config is only used when not in an active request.**

In other words, when `url_for()` is run in the context of an active request, the scheme and host information are taken from the incoming request, as sent by the HTTP client (or HTTP proxy).

I wanted to see what the code really does. Flask/Werkzeug use the concept of a [URL adapter](https://github.com/pallets/flask/blob/cc66213e579d6b35d9951c21b685d0078f373c44/src/flask/app.py#L1998) which is specific to the request context. The URL adapter is after all implementing the logic for building URLs.



New adapters are generated here:

https://github.com/pallets/flask/blob/cc66213e579d6b35d9951c21b685d0078f373c44/src/flask/app.py#L2212

This calls out to Werkzeug:
```
            return self.url_map.bind_to_environ(
                request.environ,
                server_name=self.config["SERVER_NAME"],
                subdomain=subdomain,
            )
```

And in Werkzeug's [bind_to_environ()](https://github.com/pallets/werkzeug/blob/8b3bdc812b1b17dddc68028150df386ff2d3d4d8/src/werkzeug/routing/map.py#L246) we see a critical piece of documentation:

>  Werkzeug will use `SERVER_NAME` and `SERVER_PORT` (or `HTTP_HOST` if provided)

the code then proceeds with 
```
wsgi_server_name = get_host(env).lower()
```
and `get_host()` [is documented with](https://github.com/pallets/werkzeug/blob/8b3bdc812b1b17dddc68028150df386ff2d3d4d8/src/werkzeug/wsgi.py#L92)

>     The ``Host`` header is preferred, then ``SERVER_NAME`` if it's not set.

"if it's not set" refers to the HOST header. Notably, an HTTP request must set the HOST header for it to be standards-compliant. From https://www.rfc-editor.org/rfc/rfc7230#section-5.4:

```
   A server MUST respond with a 400 (Bad Request) status code to any
   HTTP/1.1 request message that lacks a Host header field 
```

That is, when one uses Flask's `url_for(..., _external=True, ...)` then the base URL is built using information from the incoming request, in particular (as speculated yesterday via Slack) from the HOST header field. The scheme used by client is also known, communicated between software components as part of the WSGI spec.

That is it seems that _even with_  Flask's `SERVER_NAME` set the HOST header set by the incoming request takes precedence.

jgehrcke:
After all, that `url_for()` behavior seems to be rather well documented and stable in code (maybe not at the outside). Equipped with that knowledge I feel good about building a more backwards-compatible version of this PR which falls back to the 'old' method of dynamically generating the `redirect_uri` flow parameter as long as `CONBENCH_INTENDED_BASE_URL` is not set, or does not even consider the new config parameter in the first place.

conbench/api/_google.py
jgehrcke: Done in https://github.com/conbench/conbench/pull/457.

conbench/api/_google.py
jgehrcke: Taken care of in https://github.com/conbench/conbench/pull/457.

Issue #439, opened 2022-11-17
Coveralls is not performing optimally


austin3dickey:
As discussed in https://github.com/conbench/conbench/pull/438#issuecomment-1318823589, Coveralls has a few issues:
- `conbench/machine_info.py`'s coverage has been falsely "decreasing" on every PR for a long time
- it doesn't display the source when you click through the links
- it treats the `conbench/tests/` directory as source code because it's under `conbench/`, which can artificially inflate the coverage percent

See if we can solve these problems, or else find a new solution.

jgehrcke:
Updated thinking: maybe we want to remove coveralls for now, indeed. The overhead right now is I think not worth it.

austin3dickey:
Shall we repurpose this issue for that? I'm hoping to tackle it in the next month or two. Maybe we can find a good alternative.

jgehrcke:
Also, publishing the coverage report to the external API is a source for transient issues. Seen for example here: https://github.com/conbench/conbench/pull/558#issuecomment-1361688365

Issue #319, opened 2022-03-30
ci: rely only on GitHub Actions for per-PR CI (remove Buildkite pipeline from PRs)


ElenaHenderson:
@dianaclarke Were GHA tests enabled because Buildkite pipelines are not public?

dianaclarke:
I wanted the tests to run independent of buildkite, yes. My preference would be to leave them as GitHub Actions, but it's totally up to you :)

ElenaHenderson:
@dianaclarke Thank you for the quick response. I am leaving this as it is for now until we have more good reasons.

jgehrcke:
>  My preference would be to leave them as GitHub Actions,

I also quite like GHA for the integration with the PR workflow. A downside of BK is when external contributors chime in -- we've always had some workflow struggles around that (builds not triggering, and more).

The big advantage of BK is when CI is really resource intense -- with BK, one can fully take control of the underlying infrastructure. To me it seems like the current conbench CI is lightweight enough and kind of perfectly suited for running on GHA.

jgehrcke:
I am in favor of removing one of both rather sooner than later. So that we learn what we're potentially missing out on. I'd say in the current state of the system it really makes sense to remove the BK CI from PRs.

assignUser:
Just for completeness sake/another vote for GHA:

> The big advantage of BK is when CI is really resource intense -- with BK, one can fully take control of the underlying infrastructure. 

GHA has two different solutions for this now: self-hosted runners and [pay-as-you-go larger, github-hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners)

austin3dickey:
+1 on removing BK. Now that we have uncoupled the `conbench-test` BK pipeline from the `conbench-deploy` one, it seems there's not as much a reason to keep the BK one around.

jonkeane:
I'm also a +1 on simplifying | unifying on GHA. One additional thing that's nice about GHA (I'm sure we could make buildkite do this, somehow, but AFAIK it doesn't out of the box): we get actions on forks for external contributors

ElenaHenderson:
I am good if we disable BK test pipelines as long as:
1. GHA tests migrations the same way BK does.
2. We auto-deploy only after tests pass although since we no longer auto-deploy into Production, it might be ok. 
3. We runs tests agains main branch before deploying anything in Prod.

austin3dickey:
[...]

PR #455, opened 2022-11-23
tests: cover oidc flow against dex

jgehrcke:
This introduces [Dex](https://dexidp.io/docs/) as a test dependency. I have had quite a bit of success in the past testing single sign-on and related features using Dex, we even put it into one of our products. It's pretty mature and (compared to other identity providers like Shibboleth or Keycloak etc) rather simple. The docker image is small, the configuration interface is rather neat. 

Using that dependency I have added a test which exercises the complete OIDC flow from the user agent's point of view. That kind of test is super valuable, will certainly help preventing regressions in subsequent feature work.

### DIFF
--- a/conbench/api/auth.py
+++ b/conbench/api/auth.py
@@ -97,8 +97,8 @@ class CallbackAPI(ApiEndpoint):
             )
 
         email = google_user["email"]
-        given = google_user["given_name"]
-        family = google_user["family_name"]
+        given = google_user.get("given_name", "")
+        family = google_user.get("family_name", "")
         user = User.first(email=email)
         if user is None:
             data = {
--- a/conbench/config.py
+++ b/conbench/config.py
@@ -81,6 +81,7 @@ class Config:
         assert OIDC_CLIENT_ID is not None
         assert OIDC_CLIENT_SECRET is not None
 
+
 class TestConfig(Config):
     DB_NAME = os.environ.get("DB_NAME", f"{APPLICATION_NAME.lower()}_test")
     SQLALCHEMY_DATABASE_URI = (
--- a/conbench/config.py
+++ b/conbench/config.py
@@ -69,6 +69,17 @@ class Config:
         if "GOOGLE_CLIENT_ID" in os.environ:
             OIDC_ISSUER_URL = "https://accounts.google.com"
 
+    # Note that GOOGLE_CLIENT_ID/GOOGLE_CLIENT_SECRET are the legacy env vars
+    # previously used for enabling OIDC SSO. Keep using these env vars for now.
+    OIDC_CLIENT_ID = os.environ.get("GOOGLE_CLIENT_ID", None)
+    OIDC_CLIENT_SECRET = os.environ.get("GOOGLE_CLIENT_SECRET", None)
+
+    # Require client ID and client secret to be set if issuer is configured.
+    # Those three parameters are all required for an OIDC authorization code
+    # flow.
+    if OIDC_ISSUER_URL is not None:
+        assert OIDC_CLIENT_ID is not None
+        assert OIDC_CLIENT_SECRET is not None
 
 class TestConfig(Config):
     DB_NAME = os.environ.get("DB_NAME", f"{APPLICATION_NAME.lower()}_test")
--- a/conbench/tests/app/test_auth.py
+++ b/conbench/tests/app/test_auth.py
@@ -146,11 +146,19 @@ class TestLoginOIDC(_asserts.AppEndpointTest):
 
     def test_oidc_flow_against_dex(self, client):
 
-        # r = client.get("/login/", follow_redirects=True)
-        # TODO: parse this from the HTML doc
-        # initiate_flow_url = "http://127.0.0.1:5000/api/google/"
-        # follow_redirects=True)
-        r0 = client.get("/api/google/")
+        # TODO: parse this 'initiate flow URL' from the HTML login page, i.e.
+        # from the button/ link that people would actually click. If that is a
+        # _relative_ href then construct an absolute URL from it (see below).
+
+        # As long as `client` is the Flask test client we do not use an actual
+        # HTTP client/server interaction. The URL here must be specified in
+        # absolute terms because the request handler uses scheme, host, port to
+        # dynamically generate the OIDC callback URL. We could set any scheme
+        # and host, but we want to match the allow-listed callback URL that Dex
+        # is aware of. That is is currently set to
+        # http://127.0.0.1:5000/api/google/callback -- see
+        # containders/dex/config.yml
+        r0 = client.get("http://127.0.0.1:5000/api/google/")
 
         # `r0` is meant to be a redirect response, redirecting to the identity
         # provider. Extract the full URL we've been redirected to. The URL
--- a/conbench/tests/app/test_auth.py
+++ b/conbench/tests/app/test_auth.py
@@ -1,5 +1,5 @@
 import logging
-from urllib.parse import urljoin, parse_qs
+from urllib.parse import parse_qs, urljoin
 
 import requests
 from bs4 import BeautifulSoup
--- a/conbench/tests/app/test_auth.py
+++ b/conbench/tests/app/test_auth.py
@@ -1,5 +1,5 @@
 import logging
-from urllib.parse import urljoin
+from urllib.parse import urljoin, parse_qs
 
 import requests
 from bs4 import BeautifulSoup
@@ -143,6 +143,27 @@ class TestLoginOIDC(_asserts.AppEndpointTest):
     def test_login_page_shows_sso_link(self, client):
         r = client.get("/login/", follow_redirects=True)
         assert "Google Login" in r.text
+
+    def test_dynamic_callback_url(self, client):
+        """This test is mainly for demonstration purposes how the dynamic
+        OIDC callback URL construction works. One important bit that it covers
+        though is that all three parts of the base url are dynamically
+        constructed: scheme, host, port.
+
+        OIDC callback URL construction does not need to stay dynamic. If we
+        decide to make it static (based on INTENDED_BASE_URL) then this test
+        here can be removed.
+        """
+        r0 = client.get("http://foobar:7000/api/google/")
+        authorization_request_url = r0.headers["location"]
+        data = parse_qs(authorization_request_url)
+        assert "redirect_uri" in data
+        assert "foobar:7000" in data["redirect_uri"][0]
+
+        # Now check that scheme is dynamic, too.
+        data = parse_qs(client.get("https://foo/api/google/").headers["location"])
+        assert "redirect_uri" in data
+        assert "https://foo" in data["redirect_uri"][0]
 
     def test_oidc_flow_against_dex(self, client):
 
--- a/conbench/tests/app/test_auth.py
+++ b/conbench/tests/app/test_auth.py
@@ -166,7 +166,7 @@ class TestLoginOIDC(_asserts.AppEndpointTest):
         # we expect it to send a login page in the response. That login
         # page shows an email/password login form. Extract form submission
         # endpoint URL from login page. It's a relative URL.
-        assert r1.status_code == 200
+        assert r1.status_code == 200, f"bad response: {r1.text}"
         rel_login_post_url = parse_login_page(r1.text)
         log.info("rel_login_post_url: %s", rel_login_post_url)
 
--- a/conbench/tests/app/test_auth.py
+++ b/conbench/tests/app/test_auth.py
@@ -1,5 +1,13 @@
+import logging
+from urllib.parse import urljoin
+
+import requests
+from bs4 import BeautifulSoup
+
 from ...config import TestConfig
 from ...tests.app import _asserts
+
+log = logging.getLogger(__name__)
 
 
 class TestRegister(_asserts.AppEndpointTest):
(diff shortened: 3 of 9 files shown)


conbench/api/auth.py
```diff
@@ -97,8 +97,8 @@ def get(self):
             )
 
         email = google_user["email"]
-        given = google_user["given_name"]
-        family = google_user["family_name"]
+        given = google_user.get("given_name", "")
+        family = google_user.get("family_name", "")
```
jgehrcke, line 101: If an OIDC provider does not yield given_name and family_name as part of the userinfo endpoint then we still want the login flow to succeed. We got their email address which is enough use as user ID. The set of properties returned by the userinfo endpoint by an OIDC provider is _not_ standardized, i.e. for compatibility it's important to expect little.

conbench/tests/app/test_auth.py
```diff
@@ -129,3 +137,124 @@ def test_logout_unauthenticated(self, client):
...
         response = client.get("/users/", follow_redirects=True)
         self.assert_login_page(response)
+
+
+class TestLoginOIDC(_asserts.AppEndpointTest):
+    def test_login_page_shows_sso_link(self, client):
+        r = client.get("/login/", follow_redirects=True)
+        assert "Google Login" in r.text
+
+    def test_oidc_flow_against_dex(self, client):
```
jgehrcke, line 168: This piece of code exercises the entire OIDC flow from the user agent's (UA) point of view.

What is the UA? For clarity, there are three parties:

- service provider/relying party (RP, conbench)
- identity provider/ OIDC provider (OP, e.g. Dex)
- user agent (UA, the HTTP client/browser that the human interacts with)

conbench/tests/app/test_auth.py
```diff
@@ -129,3 +137,124 @@ def test_logout_unauthenticated(self, client):
...
+
+        # r = client.get("/login/", follow_redirects=True)
+        # TODO: parse this from the HTML doc
+        # initiate_flow_url = "http://127.0.0.1:5000/api/google/"
+        # follow_redirects=True)
+        r0 = client.get("/api/google/")
```
jgehrcke: This relates to @jonkeane's point around [here](https://github.com/conbench/conbench/pull/454#issuecomment-1325181808).

conbench/tests/containers/dex/config.yml
```diff
@@ -0,0 +1,48 @@
...
+  # config:
+  #   file: /dex-conbench-tests.db
+
+web:
+  http: 0.0.0.0:5556
+  # https: 0.0.0.0:8900
+  # tlsCert: /dexcerts/server.crt
+  # tlsKey: /dexcerts/server.key
+
+# Also support the implicit flow (tested) and hybrid flows (untested)
```
jgehrcke: I need to update this comment, does not apply here

docker-compose.yml
```diff
@@ -6,20 +6,38 @@ services:
...
       REGISTRATION_KEY: "code"
       SECRET_KEY: "Person, woman, man, camera, TV"
-    depends_on:
-      db:
-        condition: service_healthy
+      OAUTHLIB_INSECURE_TRANSPORT: 1
+      GOOGLE_CLIENT_ID: conbench-test-client
+      GOOGLE_CLIENT_SECRET: AnotherStaticSecret
+      CONBENCH_LOG_LEVEL_STDERR: DEBUG
+      CONBENCH_OIDC_ISSUER_URL: http://dex:5556/dex-for-conbench
```
jgehrcke, line 36: Here we see how we make use of the new config variables introduced in the previous PRs.

Here we also see that we probably want to rename `GOOGLE_CLIENT_ID` / `GOOGLE_CLIENT_SECRET` into something like `OIDC_CLIENT_ID/...`. (while still supporting GOOGLE_CLIENT_ID for our legacy environments).

requirements-build.txt
```diff
@@ -12,6 +12,7 @@ flask-swagger-ui
 Flask-WTF
 gitpython
 gunicorn
+lxml
```
jgehrcke: Oh, I think this should also go into `requirements-test.txt` -- commit upcoming.

jgehrcke:
Looking into test failures right now. Got the main test passing locally, but it looks like there are side effects on other tests that need to be addressed. Commits upcoming.

jgehrcke:
One test failure:
```
    discovery_url = Config.OIDC_ISSUER_URL + "/.well-known/openid-configuration"
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```
So, `Config.OIDC_ISSUER_URL` was None in this test run. Which makes me wonder why the docker-compose environment change (`CONBENCH_OIDC_ISSUER_URL: http://dex:5556/dex-for-conbench`) isn't enough.

docker-compose.yml
```diff
@@ -6,20 +6,38 @@ services:
...
+      FLASK_DEBUG: "true"
       REGISTRATION_KEY: "code"
       SECRET_KEY: "Person, woman, man, camera, TV"
-    depends_on:
-      db:
-        condition: service_healthy
+      OAUTHLIB_INSECURE_TRANSPORT: 1
+      GOOGLE_CLIENT_ID: conbench-test-client
+      GOOGLE_CLIENT_SECRET: AnotherStaticSecret
+      CONBENCH_LOG_LEVEL_STDERR: DEBUG
```
jgehrcke: CC @austin3dickey that's actually how I after all set the log level to DEBUG in my case. Either I remove this here again before merging this PR, or we leave this in here for everyone. Are we sure `docker-compose.yml` is only ever used for dev/CI purposes?

jgehrcke:
OK `buildkite/conbench-test`  went green, but `build / build (pull_request)` is red.

I'd love some input on 
```
      - name: Run tests
        run: |
          coverage run --source conbench -m pytest -vv --durations=20 conbench/tests/
```
[here](https://github.com/conbench/conbench/blob/5eb4325b3fef675cf62a613eb0b3dba904fb8cf5/.github/workflows/actions.yml#L63).

I kind of assumed that the docker-compose way of running tests is currently the way to run tests. Now after having added that new containerized test dependency I found that in that GH actions workflow we do not use docker-compose, but the the GH actions method for setting up container dependencies.

The docker-compose-provided DNS names are rather deeply entangled into the way the test is currently built.

That is, I think I'd love for us to also use the docker-compose way of doing things in the GH actions workflow.

Hmmm. Test runner fragmentation is always an exciting topic :) Need to think about this a little more.

austin3dickey:
@jgehrcke See https://github.com/conbench/conbench/issues/319 for a (very very brief) past discussion about GHA. We could take this conversation there.

jgehrcke:
> @jgehrcke See https://github.com/conbench/conbench/issues/319 for a (very very brief) past discussion about GHA. We could take this conversation there.

That's super related, indeed. I believe that we can get to a state where we only use one of both. Maybe GHA is better in terms of external contributors.

In need for a pragmatic solution now. 

Is the output from `coverage run --source conbench -m pytest -vv --durations=20 conbench/tests/` used for generating some coverage report like [this](https://github.com/conbench/conbench/pull/454#issuecomment-1325148465)?

jgehrcke:
Locally I have just confirmed that this works
```
docker-compose run app coverage run --source conbench -m pytest -vv --durations=20 conbench/tests/
```

That is, it's rather likely that it's more or less easy to change the GHA workflow file to use the docker-compose approach.

austin3dickey:
> Is the output from [GHA] used for generating some coverage report

Yes, this output powers Coveralls. Personally, I think coveralls has issues -- see #439 -- so I'd be okay if you added docker-compose to GHA and had to disable Coveralls for now (if that's necessary).

jgehrcke:
current problem: as part of running `docker-compose run app alembic upgrade head` the lot output suggests that `create_all()` is called:
```
[221123-16:47:46.328] [1] [conbench.db] DEBUG: Starting call to 'conbench.db.create_all', this is the 1st time calling it.
```
and likewise the subsequent migration fails with
```
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.DuplicateTable) relation "case" already exists
[SQL: 
CREATE TABLE "case" (
	id VARCHAR(50) NOT NULL, 
	name TEXT NOT NULL, 
	tags JSONB NOT NULL, 
	PRIMARY KEY (id)
)
]
(Background on this error at: https://sqlalche.me/e/14/f405)
```

jgehrcke:
Sigh. The syntax doumented [here](https://docs.docker.com/compose/environment-variables/#pass-environment-variables-to-containers) does not easily seem to work:
```
web:
  environment:
    - DEBUG
```
when mixed with entries that set a value.

jgehrcke:
Why the force-push? This PR is now based on the not-yet-merged https://github.com/conbench/conbench/pull/457, i.e. contains the commits from there, too. I want to get this one here green before landing #457.

jgehrcke:
> Overall coverage decreased (-94.7%) to 0.0%

OK that did not work. :-). Deleting coverall's comment for now.

Edit: I was wondering how `coveralls` discovers the coverage data persisted to disk by Ned's `coverage`.

Turns out `coveralls` includes `coverage` as a library and gets data like [this](https://github.com/TheKevJames/coveralls-python/blob/b6809185cdb6eaaf3b25daa0131f678f3e382dab/coveralls/api.py#L398):
```
        workman = coverage.coverage(config_file=config_file)
        workman.load()
        workman.get_data()
```
I think the best idea to not mess with all that magic really is to run `coverage` and then `coveralls` in the _same_ environment, right after one another -- w/o trying to manually transfer the state that needs to be communicated from coverage to coveralls.

Edit2:

When running `coveralls` in a container it still needs access to all these env vars:
```
GITHUB_ACTIONS
GITHUB_REF
GITHUB_SHA
GITHUB_HEAD_REF
GITHUB_REPOSITORY
GITHUB_RUN_ID
GITHUB_TOKEN
```
(documented here: https://coveralls-python.readthedocs.io/en/latest/usage/tox.html#github-actions)

It's a bit annoying to pass all these along. I think a good, different way would be to use https://github.com/marketplace/actions/coveralls-github-action. It's clearly documented with 

> The action's step needs to run after your test suite has outputted an LCOV file. 

I can work with that and make `coverage` output that specific report format, as documented at https://coverage.readthedocs.io/en/6.5.0/cmd.html#cmd-lcov.

jgehrcke:
On https://coveralls.io/github/conbench/conbench it looks like the coveralls dance works properly again, it shows a report corresponding to commit [ee5117e](https://github.com/conbench/conbench/pull/455/commits/ee5117e7eef1dc982170ad011618fe433879b8ed):

image

(the decrease in coverage is expected because in that run I disabled a lot of the test suite to get a faster result).

jgehrcke:
I split out the GitHub Actions work into https://github.com/conbench/conbench/pull/459 and landed that first.

conbench/config.py
```diff
@@ -53,6 +53,37 @@ class Config:
...
+
+    # Require client ID and client secret to be set if issuer is configured.
+    # Those three parameters are all required for an OIDC authorization code
+    # flow.
+    if OIDC_ISSUER_URL is not None:
+        assert OIDC_CLIENT_ID is not None
+        assert OIDC_CLIENT_SECRET is not None
+
+    # dirty debug
+    print(os.environ)
```
jgehrcke: still needs to be removed

jgehrcke:
Merged https://github.com/conbench/conbench/pull/457, and then rebased this here on current head of main, force-pushed.

conbench/tests/app/test_auth.py
```diff
@@ -129,3 +137,153 @@ def test_logout_unauthenticated(self, client):
...
+        r0 = client.get("http://foobar:7000/api/google/")
+        authorization_request_url = r0.headers["location"]
+        data = parse_qs(authorization_request_url)
+        assert "redirect_uri" in data
+        assert "foobar:7000" in data["redirect_uri"][0]
+
+        # Now check that scheme is dynamic, too.
+        data = parse_qs(client.get("https://foo/api/google/").headers["location"])
+        assert "redirect_uri" in data
+        assert "https://foo" in data["redirect_uri"][0]
```
jgehrcke, line 166: Those tiny tests above are new to confirm the dynamic nature of callback URL generation (the past/current behavior 'locked in' to tests).

jgehrcke:
I am rather positive and confident with the current state of this -- would love go ahead and develop the 'target URL' support based on this. Squash-merging now.

conbench/tests/containers/dex/config.yml
```diff
@@ -0,0 +1,51 @@
+# issuer: http://127.0.0.1:5556/dex-for-conbench
+# The DNS name `dex` is provided by e.g. docker-compose
+issuer: http://dex:5556/dex-for-conbench
+# will expose http://127.0.0.1:5556/dex/.well-known/openid-configuration
+
+storage:
+  type: sqlite3
+  # config:
+  #   file: /dex-conbench-tests.db
```
austin3dickey, line 9: Do you still need these comments in this file?

austin3dickey (commented):
Looks fantastic!

conbench/tests/containers/dex/config.yml
jgehrcke: Personally I like when some of those stay around so that one can quickly change things w/o having to consult docs too much, but I am also happy with doing a proper cleanup :).

austin3dickey: Yes, agreed! We can keep these around for that purpose. I'm just not quite sure after reading these comments what sorts of features they might enable or why we might want to do that in the future.

PR #459, opened 2022-11-24
github actions: invoke test runner with docker-compose approach

jgehrcke:
As part of https://github.com/conbench/conbench/pull/455 I spent a couple of hours trying to change the way we run tests in GitHub Actions. I have now split out this work into this separate PR.

Instead of running `pytest` directly in the GHA container the goal was to use `docker-compose` to run `pytest` _in_ the `app` container as defined by `docker-compose.yml`. That is to make use of the dependency management that docker-compose provides, which is for now a good thing to have when introducing Dex as a new (containerized) test dependency that needs to be reachable under a predictable DNS name.

As part of making this change the biggest challenge was to not break the coverage data workflow. I ran into the same wall multiple times and almost gave up (the commit history shows the difficulties, I suppose), but after all I think the coverage data flow is now the same as before.

What's new in terms of coverage report information flow:

- We let the pytest test runner in the `app` container write the coverage data to a particular location in that container file system.
- We then invoke `coverage lcov` in the same container to write the coverage data in the so-called LCOV file format to a file that is accessible from the 'host', under `$(pwd)/_conbench-coverage-dir/coverage.lcov`.
- We use the GitHub action called `coverallsapp/github-action` (seems to be rather popular) actually submit the report -- with the input parameter titled `path-to-lcov`, pointing to said LCOV file.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -37,12 +37,14 @@ jobs:
       - name: Start PostgreSQL DB
         run: |
           docker-compose run --detach db
+      # Run `alembic upgrade head` via docker-compose so that it is within
+      # the network that the DB is also in, reachable via DNS name `db`
       - name: Migrations
         run: |
-          alembic upgrade head
+          docker-compose run app alembic upgrade head
         env:
           DB_USERNAME: "postgres"
-          DB_HOST: "localhost"
+          DB_HOST: "db"
           DB_PASSWORD: "postgres"
           DB_NAME: "conbench_test"
           DB_PORT: "5432"
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -55,7 +55,8 @@ jobs:
           docker-compose down && \
           docker-compose build app && \
           docker-compose run app coverage run \
-            --data-file=/etc/conbench-coverage-datafile --source conbench -m pytest -vv -s --durations=20 conbench/tests/
+            --data-file=/etc/conbench-coverage-dir/.coverage --source conbench \
+              -m pytest -vv -s -k test_dynamic_callback_url --durations=20 conbench/tests/
         env:
           DB_USERNAME: postgres
           GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -40,13 +40,15 @@ jobs:
       - name: Lint (isort)
         run: |
           isort --check .
-      - name: Run tests, generate coverage report
+      - name: Run tests, generate coverage data
         run: |
           docker-compose down && \
           docker-compose build app && \
           docker-compose run app coverage run \
             --source conbench \
-              -m pytest -vv -s -k test_dynamic_callback_url --durations=20 conbench/tests/ \ &&
+              -m pytest -vv -s -k test_dynamic_callback_url --durations=20 conbench/tests/
+      - name: Convert coverage data to LCOV
+        run: |
           docker-compose run app coverage lcov -o /etc/conbench-coverage-dir/coverage.lcov
       - name: Publish coverage
         uses: coverallsapp/github-action@master
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -55,7 +55,7 @@ jobs:
           docker-compose down && \
           docker-compose build app && \
           docker-compose run app coverage run \
-            --data-file=/etc/conbench-coverage-datafile --source conbench -m pytest -vv -s --durations=20 conbench/tests/ && \
+            --data-file=/etc/conbench-coverage-datafile --source conbench -m pytest -vv -s --durations=20 conbench/tests/
         env:
           DB_USERNAME: postgres
           GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/actions.yml
```diff
@@ -34,6 +24,36 @@ jobs:
         with:
           options: --check --diff
           src: .
+      - name: Install flake8, isort
+        run: pip install flake8 isort
+      - name: Lint (flake8)
+        run: flake8
+      - name: Lint (isort)
+        run: isort --check .
```
jgehrcke, line 32: I took liberty to tweak this part a bit so that the flake8 and isort results come back faster than before. (the previous `Install dependencies` step took quite a while and did set up way more things than needed for `flake8` and `isort`)

jgehrcke: We may want to pin flake8 and isort versions in the future (didn't do so far via requirements-test.txt).

.github/workflows/actions.yml
```diff
@@ -34,6 +24,36 @@ jobs:
...
+      - name: Start PostgreSQL DB
+        run: docker-compose run --detach db
```
jgehrcke, line 34: After this step, the docker-compose-managed PostgreSQL container is left running, available for subsequent steps. Effectively replacing
```
    services:
      postgres:
        image: postgres:11
```

.github/workflows/actions.yml
```diff
@@ -34,6 +24,36 @@ jobs:
...
+      # Run `alembic upgrade head` via docker-compose so that it is within
+      # the network that the DB is also in, reachable via DNS name `db`
+      - name: Migrations
+        run: docker-compose run -e CREATE_ALL_TABLES app alembic upgrade head
+        env:
+          CREATE_ALL_TABLES: false
```
jgehrcke, line 40: This is to prevent the app to create tables during initialization.

.github/workflows/actions.yml
```diff
@@ -34,6 +24,36 @@ jobs:
...
+      - name: Run tests, generate coverage data
+        run: |
+          docker-compose down && \
+          docker-compose build app && \
+          docker-compose run -e COVERAGE_FILE=/etc/conbench-coverage-dir/.coverage app \
+            coverage run --source conbench \
+              -m pytest -vv -s --durations=20 conbench/tests/
```
jgehrcke, line 47: At the very heart this leaves the test runner invocation unchanged:

```
pytest -vv -s --durations=20 conbench/tests/
```

Before, that was `coverage run --source conbench -m pytest -vv --durations=20 conbench/tests/` -- i.e., the same.

.github/workflows/actions.yml
```diff
@@ -34,6 +24,36 @@ jobs:
...
+      - name: Convert coverage data to LCOV
+        run: |
+          docker-compose run -e COVERAGE_FILE=/etc/conbench-coverage-dir/.coverage app \
+            coverage lcov -o /etc/conbench-coverage-dir/coverage.lcov
+      - name: Publish coverage
+        uses: coverallsapp/github-action@master
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          path-to-lcov: _conbench-coverage-dir/coverage.lcov
```
jgehrcke, line 56: See PR description for some background for this.

docker-compose.yml
```diff
@@ -6,17 +6,27 @@ services:
...
+      # network and is both reachable by other containers on that network, and
+      # discoverable by them at a hostname identical to the container name."
+      # That is why the DNS name `db` works.
       DB_HOST: "db"
       DB_PASSWORD: "postgres"
       DB_NAME: "postgres"
       DB_PORT: "5432"
       FLASK_APP: "conbench"
       FLASK_ENV: "development"
+      FLASK_DEBUG: "true"
```
jgehrcke, line 25: That's to get rid of a warning msg. Also see https://github.com/conbench/conbench/pull/436.

docker-compose.yml
```diff
@@ -6,17 +6,27 @@ services:
...
       REGISTRATION_KEY: "code"
       SECRET_KEY: "Person, woman, man, camera, TV"
+      # Default log level is INFO, change to DEBUG as needed.
+      # CONBENCH_LOG_LEVEL_STDERR: DEBUG
```
jgehrcke, line 29: CC @austin3dickey

jgehrcke:
Here we can see that the lint result is now available after ~10 seconds (just a tiny side effect of this PR, but valuable):
image

I am going ahead with squash-merging this so that I can start rebasing (and conflict-resolving https://github.com/conbench/conbench/pull/455.

.github/workflows/actions.yml
austin3dickey: This is great! It ensures there's no configuration in this yml file that might become out-of-date with the docker-compose file

docker-compose.yml
austin3dickey: Great! This would be nice to add to the README. Right now, it seems the README isn't recommending the best way to run tests locally:

https://github.com/conbench/conbench/blob/c086edf17a451002d278fe10b837c5a06edd0db8/README.md?plain=1#L150-L152

It would be cool to add something like `-e CONBENCH_LOG_LEVEL_STDERR=DEBUG` to this, and probably the `down` and `build` commands too.

jgehrcke: > it seems the README isn't recommending the best way to run tests locally:

Yes, let's change that! https://github.com/conbench/conbench/pull/468

Issue #411, opened 2022-10-27
Redirect to desired url after login


mariusvniekerk:
When going to a conbench url from a link and not being logged in already you are not redirected to the desired url upon completing the login flow.

jgehrcke:
Quick thoughts: the URL path of entry needs to be passed along all parties contributing to the login flow (three: user, api server, identity provider)

- the server component emitting the 401 response (not authenticated) will redirect to the login URL, adding a query parameter containing the entry path
- in the OAuth2 flow initiation (in the api server) that entry path must be encoded in a state-holding parameter.
- the login callback http request handler must then decode the entry path from the bits it got back from the identity provider, and can then emit an http redirect response

code section to encode the entry path:
https://github.com/conbench/conbench/blob/fdb30f2934481989e981c49837506c0622d6bb3a/conbench/api/_google.py#L26

code section to reconstruct the entry path and to redirect the client to:
https://github.com/conbench/conbench/blob/fdb30f2934481989e981c49837506c0622d6bb3a/conbench/api/auth.py#L81

jgehrcke:
This now works for SSO (not regular login), and only for these views: 
- batches
- benchmarks
- compare
- index
- runs

Closing this for now.

Follow-up work to do, if desired (let's review the need):

- make this work for username/password login
- make this work for other views, too.

PR #462, opened 2022-11-25
OIDC: carry target URL across flow

jgehrcke:
Context: https://github.com/conbench/conbench/issues/411

This patch introduces the concept of a target URL, which is the URL that the user actually wanted to visit before they were required to log in. 

Information flow:
- If the login page is GET with a specific query parameter (`?target=<target-url>`), then the HTML template renders the SSO button/link on the login page so that it _also_ carries a query parameter with the value `<target-url>`.
- If the SSO flow is initiated with that very query parameter, then the information is retained across the flow, i.e. pushed back and forth as part of the OIDC flow in the so-called `state` parameter. 

### Discussion: (mis)usage of the `state` parameter in the so-called authorization code flow?

The security implications of the `state` parameter in the context of OpenID Connect can be quite severe, and vary depending on the exact conditions and flow type used. There is a plethora of confusing content on the Internet, and so I always like to go back to the specification.

The specification for the so-called authorization code flow lives here: https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth

In the authorization code flow the so-called access token and ID token are never seen by the user agent, but only by the relying party (conbench backend).

In the spec for that flow we see that the `state` parameter is "only" `RECOMMENDED`, not required:

> RECOMMENDED. Opaque value used to maintain state between the request and the callback. Typically, Cross-Site Request Forgery (CSRF, XSRF) mitigation is done by cryptographically binding the value of this parameter with a browser cookie.

Whether or not it's a good place to store a redirect URL, _instead of_ "cryptographically binding the value of this parameter with a browser cookie"?

From https://auth0.com/docs/secure/attack-protection/state-parameters#csrf-attacks:

>  You can use the state parameter to encode an application state that will put the user where they were before the authentication process started. 

from https://stackoverflow.com/questions/26132066/what-is-the-purpose-of-the-state-parameter-in-oauth-authorization-request:

> The developers of ckanext-oauth2 use the state parameter also to store info about the previously visited page, to redirect the user back there after login, e.g.: {"came_from": "/dashboard"}

From https://www.oauth.com/oauth2-servers/server-side-apps/authorization-code/:

> The state parameter serves two functions. When the user is redirected back to your app, whatever value you include as the state will also be included in the redirect. This gives your app a chance to persist data between the user being directed to the authorization server and back again, such as using the state parameter as a session key. This may be used to indicate what action in the app to perform after authorization is complete, for example, indicating which of your app’s pages to redirect to after authorization.

> The state parameter also serves as a CSRF protection mechanism if it contains a random value per request. When the user is redirected back to your app, double check that the state value matches what you set it to originally.

from https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow:

> A value included in the request that is also returned in the token response. It can be a string of any content that you wish. A randomly generated unique value is typically used for preventing cross-site request forgery attacks. The value can also encode information about the user's state in the app before the authentication request occurred. For instance, it could encode the page or view they were on.

---------------

If we don't use the `state` parameter for security purposes -- why not simply persist a cookie storing the 'target URL' -- it would be sent _with_ the final callback? Well, reliably linking cookie state and login flow would need some basic state management (such as manual cookie deletion). The `state` parameter provides after all a _simple_ way to carry information across the flow -- *non-securely* (if carried around w/o a cryptographic signature).

### DIFF
--- a/conbench/api/_google.py
+++ b/conbench/api/_google.py
@@ -1,3 +1,4 @@
+import base64
 import json
 import logging
 import time
@@ -90,24 +91,27 @@ def gen_oidc_authz_req_url(user_came_from_url: str) -> str:
     log.debug("Initiate OIDC SSO flow. redirect_uri: %s", abs_oidc_callback_url)
     log.debug("user_came_from_url: %s", user_came_from_url)
 
-    camefrom_encoded = ""
-    try:
-        camefrom_encoded = urllib.parse.quote(user_came_from_url)
-    except Exception as exc:
-        # Continue with the login flow w/o carrying the target URL around.
-        # NOTE: maybe emit a 400 Bad Request response instead, showing err
-        # detail.
-        log.info("drop target URL: encoding failed with: %s", exc)
+    state = encode_target_url(user_came_from_url)
+    if not state:
+        # In case `encode_target_url()` returned a zero-length string pass
+        # state=None into `prepare_request_uri()` below, resulting in oauthlib
+        # to generate random state -- although there does not seem to be a
+        # security gain (there is no validation at the end of the flow) that
+        # enhances compatiblity (and also resembles legacy behavior).
+        state = None
 
     url_to_redirect_user_to = client.prepare_request_uri(
         oidc_provider_config["authorization_endpoint"],
         redirect_uri=abs_oidc_callback_url,
-        # The `openid` scope renders this OAuth2 flow to be an OpenIDConnect
-        # (OIDC) flow.
+        # The `openid` scope is an essential ingredient to make this OAuth2
+        # flow be an OpenID Connect (OIDC) flow.
         scope=["openid", "email", "profile"],
-        # Additional parameter to carry across the flow. Usually security
-        # purpose. TODO: combine with non-guessable state.
-        state="RANDOM" + camefrom_encoded,
+        # Additional parameter to carry across the flow. Usually this parameter
+        # has a security purpose. For the time being we do not use it for that,
+        # but we use it for communicating `camefrom_encoded` across the flow.
+        # Discussion can be found in
+        # https://github.com/conbench/conbench/pull/462.
+        state=state,
     )
 
     return url_to_redirect_user_to
@@ -155,6 +159,12 @@ def conclude_oidc_flow():
         # Continue with the login flow w/o using the target URL around.
         log.info("drop target URL: decoding failed with: %s", exc)
         log.debug("camefrom_encoded: %s", camefrom_encoded)
+    # Parse encoded target URL from state.
+    user_came_from_url = ""
+    if "state" in authorization_response:
+        user_came_from_url = decode_target_url_from_oidc_state(
+            authorization_response["state"]
+        )
 
     log.info("user_came_from_url: %s", user_came_from_url)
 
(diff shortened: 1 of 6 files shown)


conbench/api/_google.py
```diff
@@ -88,34 +86,179 @@ def auth_google_user():
...
+    userinfo_url, headers, body = client.add_token(
+        oidc_provider_config["userinfo_endpoint"]
+    )
+
+    userinfo = requests.get(
+        userinfo_url,
         headers=headers,
         data=body,
     ).json()
+
```
jgehrcke, line 215: this needs another exception handler, involves an HTTP request and JSON-deserializing the response body.

jgehrcke:
The test suite in this PR shows the functionality added: 

If we visit `login?target=https://foobar.com` then the SSO login link carries this information. If we then press the SSO login button and let ourselves get carried across the login flow -- then the flow is after all concluded with a redirect to https://foobar.com.

The test suite shows that this works for URLs in general, including relative URLs.

With this patch said feature will not be used (unless users _manually_ render their login page with the `target` query parameter :hankey:).

Follow-up work is required to introduce the desired UX: upon an `authentication required` error and the corresponding redirect to the login page -- _that's_ where we will use that feature. That work will be done in subsequent patches.

conbench/tests/app/test_auth.py
```diff
@@ -141,11 +142,37 @@ def test_logout_unauthenticated(self, client):
...
+            "https://rofl.com",
+            "https://foo.bar?x=y",
+            # Test a literal %20 and a literal %2F to be carried across the
+            # flow. This is a good way to find flaws in the
+            # URL-encoding-decoding information flow (the goal is that these
+            # character sequences are communicated verbatim, i.e. end up being
+            # emitted as-is in the final redirect URL).
+            "https://foo.bar/path/?x=%20yz",
+            "https://foo.bar/path/?x=y%2Fz",
+        ],
```
jgehrcke, line 173: Reviewers, maybe start here, following this `target_url` parameter.

conbench/app/_endpoint.py
```diff
@@ -11,8 +11,43 @@ def as_bool(x):
...
+        processing and redirect to the login page if needed. Maybe build a
+        simple decorator instead with a name that relates to access control
+        (such as `enforce_access_control()`). Decorated views enforce access
+        control (with the business logic as in _this_ function here),
+        non-decorated views do not enforce. If access is denied then the
+        redirect to the login page should be emitted from that decorator -- the
+        redirect will contain view-specific parameters (target URL).
+        """
+
+        # TODO(JP): the environment variable readout should be managed via the
```
austin3dickey, line 35: I like this idea; want to do it now?

austin3dickey (approved):
This looks good to me, though I admit I don't know much about the OIDC flow and can't provide suggestions there. Thanks for all the documentation and comments. I think the usage of `state` sounds fine.

conbench/app/_endpoint.py
jgehrcke: Thanks for having a look! That's part of my attempt to not do toooo much in a single pull request, so I will leave this for the follow-up work on access control:

> Follow-up work is required to introduce the desired UX: upon an authentication required error and the corresponding redirect to the login page -- that's where we will use that feature. That work will be done in subsequent patches.

jgehrcke:
Going to land this now, to make it easier to manage upcoming patches.

PR #457, opened 2022-11-24
config: introduce CONBENCH_OIDC_ISSUER_URL 

jgehrcke:
Like #454, but with backwards compatibility (without reliance on CONBENCH_INTENDED_BASE_URL being set to a meaningful value).

PR #465, opened 2022-11-29
OIDC: fix #464: hard-code scheme again (do not use WSGI info)

### LINKED ISSUE #464: oidc: `redirect_uri_mismatch`: WSGI environment does not reflect client URL scheme
In one of our staging deployments we found that OpenID Connect against Google fails with a `redirect_uri_mismatch`. The mismatch is as of the redirect URL scheme being set to `http`, whereas `https` is expected.

With the [last patch](https://github.com/conbench/conbench/pull/457) we started relying on Flask's `url_for()` to be able to properly derive the URL scheme used by the client.

In https://github.com/conbench/conbench/pull/454#issuecomment-1326338524 I wrote

> The scheme used by client is also known, communicated between software components as part of the WSGI spec.

How does that work? [Here is](https://github.com/pallets/werkzeug/blob/8b3bdc812b1b17dddc68028150df386ff2d3d4d8/src/werkzeug/routing/map.py#L295) where Werkzeug/Flask reads the scheme that the client used from the WSGI properties of the incoming HTTP request. The WSGI server (in our case gunicorn) derives the scheme from the HTTP header `X-FORWARDED-PROTO` (or similar, see https://docs.gunicorn.org/en/stable/settings.html#secure-scheme-headers) if set.

It seems like in case of our staging environment the `X-FORWARDED-PROTO` header is either not arriving at gunicorn or gunicorn ignores its value. I just found that gunicorn may not respect it for security reasons because of the `forwarded_allow_ips` setting: https://docs.gunicorn.org/en/stable/settings.html#forwarded-allow-ips (as pointed out in https://stackoverflow.com/a/48707678/145400). In that context, it suddenly important whether the reverse proxy and gunicorn run on separate hosts or not.

jgehrcke:
See #464.

The approach I opted for here required changing the default value of `Config.INTENDED_BASE_URL`, and also results in `f.url_for("api.callback", _external=True, https=True)` _not_ being covered by the test suite. Those are two reasons for why I initially decided against this approach when working on https://github.com/conbench/conbench/pull/457.

However, #464 has shown that we really need to keep the behavior of `f.url_for("api.callback", _external=True, https=True)` to keep legacy environments working.

This is now the third iteration. The confusion, complexity, error-proneness around this will decrease once we require `INTENDED_BASE_URL` to be set by Conbench operators.

### DIFF
--- a/conbench/api/_docs.py
+++ b/conbench/api/_docs.py
@@ -5,6 +5,17 @@ import apispec_webframeworks.flask
 from ..api import _examples as ex
 from ..config import Config
 
+
+# `api_server_url` is used for populating the 'Servers' dropdown in the
+# Swagger/OpenAPI docs website. The default value is tailored to the Flask
+# development HTTP server defaults (non-TLS, binds on 127.0.0.1, port 5000).
+# Can be adjusted via the `--host` and `--port` command line flags when
+# invoking `flask run`.
+api_server_url = "http://127.0.0.1:5000/"
+if Config.INTENDED_BASE_URL is not None:
+    api_server_url = Config.INTENDED_BASE_URL
+
+
 spec = apispec.APISpec(
     title=Config.APPLICATION_NAME,
     version="1.0.0",
@@ -13,7 +24,7 @@ spec = apispec.APISpec(
         apispec_webframeworks.flask.FlaskPlugin(),
         apispec.ext.marshmallow.MarshmallowPlugin(),
     ],
-    servers=[{"url": Config.INTENDED_BASE_URL}],
+    servers=[{"url": api_server_url}],
 )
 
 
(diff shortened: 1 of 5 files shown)


jgehrcke:
This is unlikely to worsen things over the current HEAD of main -- so I am going ahead and merge this. Will then rebase the work in progress in https://github.com/conbench/conbench/pull/462 on top of that.

jgehrcke:
>  we really need to keep the behavior of `f.url_for("api.callback", _external=True, https=True)` to keep legacy environments working.

Yes, it's just that I had written that from memory whereas the actual line of code in question looked like this:

https://github.com/conbench/conbench/blob/6d8bfe182ea6b1043dd99a79918bc72732ed783e/conbench/api/_google.py#L26

So, instead of `_scheme="https"` my brain produced `https=True`, and somehow I did not double-check. That feels slightly embarrassing. As pointed out above, this code is not covered by the test suite.

Thanks @austin3dickey for uncovering this via a staging environment!

Issue #437, opened 2022-11-16
devprod: docker-compose tests do not pick up code changes


jgehrcke:
`docker-compose run app pytest -vv conbench/tests` does not reliably pick up code changes to test/app files. Code gets into this test runner via a Docker image build:

https://github.com/conbench/conbench/blob/353f1f9cf7dceb42cd287d8a0035b1ef9259dba5/Dockerfile#L9

docker-compose does not seem to execute that image build upon every `run`. Makes sense. If it were to re-execute the Docker image layer cache would see the update and invalidate the cache layer. Therefore, a quick workaround is to run `docker-compose build app`, i.e. I currently run
```
docker-compose build app && \
  docker-compose run app pytest -k test_create_benchmark_empty_context -vv conbench/tests
```
to quickly iterate on a test. That works, it's however a little inefficient because for example it executes `RUN pip install .` for every test runner invocation.

Better approach would be to mount the code from the host into the container (downside, as always with privileged containers, would be root-created `__pycache__` etc).

Action item: find dockerized test invocation method that is efficient and picks up code changes; document that in the README.

austin3dickey:
For clarity, are you suggesting we mount the code only during local development/testing, and keep the current "ADD" setup for CI and CD?

jgehrcke:
I think that is what I mean, yes :)

jgehrcke:
@sayantikabanik also runs into this challenge when using the containerized workflow. It's important to prioritize addressing this issue here, so that local changes are automatically picked up by the flask app.

Typical approach: mount local checkout into container.

points to consider:
- this must work w/o mounting the checkout into the container (for CI, and for having a stable reference).
- disadvantage of mounting local checkout into container: files written by container processes will for starters be root-owned: 
https://github.com/conbench/conbench/pull/468#discussion_r1036083675
- challenge of docker-compose: mount only conditionally.

jgehrcke:
This has been addressed with the `make run-app-dev` command. 

Also see
https://github.com/conbench/conbench/issues/503
https://github.com/conbench/conbench/pull/546

PR #468, opened 2022-11-30
README: document current recommended test invocation method

jgehrcke:
Between individual developers and CI it's valuable to settle on a common method to run tests, and to document that method.

As of today, CI in Buildkite as well as in GitHub Actions uses the docker-compose approach. This is also useful to use locally, at least that is what I have been running tests with in the last couple of days.

The purpose of this patch is to document this method, and to enumerate a couple of useful configuration switches.

The command shown in this README patch run locally on my machine against current HEAD of `main`:

```bash
$ docker-compose down && docker-compose build app && \
    docker-compose run app \
    pytest -vv conbench/tests
Stopping conbench_dex_1 ... done
[...]
Removing conbench_db_1                 ... done
Removing network conbench_default
Building app
Sending build context to Docker daemon  7.628MB
Step 1/8 : FROM python:3.8
 ---> 928a35165ae5
Step 2/8 : COPY requirements-build.txt /tmp/
 ---> Using cache
 ---> 6ad8d4d47682
[...]
Step 8/8 : RUN pip install .
 ---> Running in e479d9ad2cbe
Processing /app
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Requirement already satisfied: click in /usr/local/lib/python3.8/site-packages (from conbench==1.62.0) (8.1.3)
[...]
Successfully installed conbench-1.62.0
[...]
Successfully built 8c9b160f5a0a
Successfully tagged conbench_app:latest
Creating network "conbench_default" with the default driver
Creating conbench_db_1  ... done
Creating conbench_dex_1 ... done
Creating conbench_app_run ... done
======================================================= test session starts ========================================================
platform linux -- Python 3.8.15, pytest-7.2.0, pluggy-1.0.0 -- /usr/local/bin/python
cachedir: .pytest_cache
rootdir: /app
collected 380 items                                                                                                                

conbench/tests/test_version.py::test_version PASSED                                                                          [  0%]
conbench/tests/api/test_auth.py::TestLoginPost::test_extra_field <- conbench/tests/api/_asserts.py PASSED                    [  0%]
conbench/tests/api/test_auth.py::TestLoginPost::test_cannot_set_id <- conbench/tests/api/_asserts.py PASSED                  [  0%]
conbench/tests/api/test_auth.py::TestLoginPost::test_required_fields <- conbench/tests/api/_asserts.py PASSED                [  1%]
conbench/tests/api/test_auth.py::TestLoginPost::test_empty_fields <- conbench/tests/api/_asserts.py PASSED                   [  1%]

[...]

conbench/tests/entities/test_history.py::test_history PASSED                                                                 [ 99%]
conbench/tests/entities/test_user.py::test_user_repr PASSED                                                                  [ 99%]
conbench/tests/entities/test_user.py::test_user_str PASSED                                                                   [100%]

============================================ 378 passed, 2 skipped in 114.62s (0:01:54) ============================================
```

My environment: docker-compose version 1.29.2, Linux 6.0.9-300.fc37.x86_64

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,13 +143,23 @@ password in postgres to `postgres`.
 
 
 ### Run tests
-    (conbench) $ cd ~/workspace/conbench/
-    (conbench) $ pytest -vv conbench/tests/
 
+The canonical test invocation method for now is via `docker-compose` (manages test dependencies automatically, and is also used in CI):
 
-### Run tests with docker
-    (conbench) $ cd ~/workspace/conbench/
-    (conbench) $ docker-compose run app pytest -vv conbench/tests/
+```
+$ docker-compose down && docker-compose build app && \
+    docker-compose run app \
+    pytest -vv conbench/tests
+```
+
+This command attempts to stop and remove previously spawned test runner containers, and it rebuilds the `app` container image prior to running tests to pick up code changes in the local checkout.
+
+Useful command line arguments for local development (can be combined as desired):
+
+* `... pytest -k test_login`: run only string-matching tests
+* `... pytest -x`: exit upon first error
+* `... pytest -s`: do not swallow log output during run
+* `... run -e CONBENCH_LOG_LEVEL_STDERR=DEBUG app ...`
 
 
 ### Format code


README.md
```diff
@@ -143,13 +143,23 @@ password in postgres to `postgres`.
 
 
 ### Run tests
-    (conbench) $ cd ~/workspace/conbench/
-    (conbench) $ pytest -vv conbench/tests/
```
jgehrcke, line 147: Notably, this method is not easily expected to succeed because it does not automatically manage Dex.

README.md
```diff
@@ -143,13 +143,23 @@ password in postgres to `postgres`.
...
+```
+
+This command attempts to stop and remove previously spawned test runner containers, and it rebuilds the `app` container image prior to running tests to pick up code changes in the local checkout.
+
+Useful command line arguments for local development (can be combined as desired):
+
+* `... pytest -k test_login`: run only string-matching tests
+* `... pytest -x`: exit upon first error
+* `... pytest -s`: do not swallow log output during run
+* `... run -e CONBENCH_LOG_LEVEL_STDERR=DEBUG app ...`
```
jonkeane, line 162: Love this documentation!

README.md
```diff
@@ -143,13 +143,23 @@ password in postgres to `postgres`.
...
 
+The canonical test invocation method for now is via `docker-compose` (manages test dependencies automatically, and is also used in CI):
 
-### Run tests with docker
-    (conbench) $ cd ~/workspace/conbench/
-    (conbench) $ docker-compose run app pytest -vv conbench/tests/
+```
+$ docker-compose down && docker-compose build app && \
+    docker-compose run app \
+    pytest -vv conbench/tests
```
jonkeane, line 152: Do we need both build + run here? I thought run without build would do the building if it needed to?

jonkeane: And, is it possible (and advisable) that we mount the local directory so that for tests like the docs tests that will automatically generate the new version (with a warning), will do that + it is persisted in the filesystem?

jgehrcke: > Do we need both build + run here?

Great question. Yes, the additional `build` is critical to pick up local changes to files.

`run` does not rebuild by itself. I ran into that and found that this is actually expected behavior, documented by docker-compose.

Related: https://github.com/conbench/conbench/issues/437

jgehrcke: > that we mount the local directory so that for tests like the docs tests that will automatically generate the new version

Mounting a checkout into a test container has typically been the most efficient way I have seen in the past to make a containerized test runner reflect the local checkout on the host (typically faster than a Docker image ADD or COPY).

However, test runner output files then often end up written as `root` onto the host -- that really can be an annoyance. Addressing that can then again increase complexity in an unpleasant way ("do not run processes in container as root", relates to https://docs.docker.com/engine/security/userns-remap/).

Sounds like we need some docs-related output emitted by the test runner every now and then? I'll have to look at the specifics.

jonkeane: Ooph, yeah I forgot about the root user thing. In that case, I say let's ship this as is, and we can follow up to see if the docs-related changes can be separated out, or excepted for this case, etc.

Another thought might be hooking up a comment github action that adds the updated docs to the branch when run. That way we can see the failure in CI, then say "@githubbot, please fix up the docs, I did mean to change those" and it would commit the updated expectations? Regardless, IMO that's out of scope for this PR

austin3dickey:
I like this -- one open question is, does [this workflow](https://github.com/conbench/conbench#to-populate-local-conbench-with-sample-runs-and-benchmarks) of populating a local conbench with fake data and being able to interact with the local webapp still work with docker? If it does, but the commands change, could you change those in this PR too?

README.md
jgehrcke: > Another thought might be hooking up a comment github action that adds the updated docs to the branch when run. That way we can see the failure in CI, then say "@githubbot, please fix up the docs, I did mean to change those" and it would commit the updated expectations? Regardless, IMO that's out of scope for this PR

I will have to see what the technical goals are in detail, and then it's easier for me to think about all that.

Maybe this automatic-docs-generation step that we talk about here can be split out of the test suite and run in its own CI step. But that's just me thinking out loud w/o fully understanding what we even talk about :).

In any case, the "current recommended approach to run tests" can of course change again very soon if we feel like  there's a good reason! Personally, I am not the biggest fan of deeply relying on docker-compose.

jgehrcke:
> I like this -- one open question is, does [this workflow](https://github.com/conbench/conbench#to-populate-local-conbench-with-sample-runs-and-benchmarks) [...] still work with docker?

Good question. 

One way I think about this, to keep things simple and structured: if it's not part of the test suite, it does not need to work in the context of this new/current way of running the test suite. That is, if someone needs to use this workflow today: I assume they can just act as before (run conbench straight on the host, not in a container). Yes, they would not be able to easily connect Conbech to Dex (the magic that docker-compose provides). But they probably also don't need that.

I have the feeling that "the recommended way of running tests" is going to fluctuate in the near term. Let's land this documentation change now as it adds value. Let's think about other README updates separately.

PR #463, opened 2022-11-28
Metadata updated for ClusterCreate API

sayantikabanik:
resolves #450

PR #471, opened 2022-11-30
Pass CI=true to GHA tests

austin3dickey:
In [this PR](https://github.com/conbench/conbench/pull/463#issuecomment-1331464712) we found that CI was not failing when the expected OpenAPI docs did not match the actual generated docs. 

This is because #459 changed the tests to run inside docker-compose, so the `CI=true` environment variable was not set anymore -- so the test assumed we were running locally, manually changed `_expected_docs.py`, and succeeded.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -42,7 +42,10 @@ jobs:
         run: |
           docker-compose down && \
           docker-compose build app && \
-          docker-compose run -e COVERAGE_FILE=/etc/conbench-coverage-dir/.coverage app \
+          docker-compose run \
+            -e COVERAGE_FILE=/etc/conbench-coverage-dir/.coverage \
+            -e CI=true \
+            app \
             coverage run --source conbench \
               -m pytest -vv -s --durations=20 conbench/tests/
       - name: Convert coverage data to LCOV
--- a/conbench/entities/hardware.py
+++ b/conbench/entities/hardware.py
@@ -190,13 +190,7 @@ class MachineCreate(marshmallow.Schema):
 class ClusterCreate(marshmallow.Schema):
     name = marshmallow.fields.String(required=True)
     info = marshmallow.fields.Dict(required=True)
-    optional_info = marshmallow.fields.Dict(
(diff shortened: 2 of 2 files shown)


austin3dickey:
^ I made a quick change to the docs and confirmed that CI does fail. (That change was reverted.)

jonkeane (approved):
Thanks for this!

jgehrcke (commented):
:rocket: 

Thanks!

This is a reminder that maybe I should have compared the number of tests reported as passing before/after the switch of GHA to using docker-compose.

jgehrcke:
I have not looked into this at all, but one interesting question that I might answer myself in the future: why did we in the past decide not to execute these tests on our machines? Too slow?

austin3dickey:
Oh, this environment variable doesn't enable/disable tests. It changes the behavior of `test_docs()` depending on whether it's running locally or in CI. That test compares the OpenAPI docs coming from the endpoint to `_expected_docs.py`. If they are different:

- if locally, it will automatically update and format `_expected_docs.py` for you, so you can commit that file. Then the test succeeds.
- if in CI, it will fail.

This relates to @jonkeane 's [recent comment](https://github.com/conbench/conbench/pull/468#discussion_r1036258970) that maybe this isn't the optimal way to do this kind of test. The point of the test is to ensure that PR reviewers can see the proposed changes to the OpenAPI docs. But the current workflow leads to some friction sometimes.

jgehrcke:
> Oh, this environment variable doesn't enable/disable tests. 

Thank you for clarifying!

As far as I understand this magic may result in scenarios where tests pass for me locally, but then fail when executed in pull request CI. That's a bit of a hassle from a developer productivity point of view, I think I'd love for these test failures to happen locally if possible.

> if locally, it will automatically update and format _expected_docs.py for you, so you can commit that file. Then the test succeeds.

Not quite sure if that's now a bug or not but I get a test failure in PR CI where I did not get one locally. Will have in inspect this; created https://github.com/conbench/conbench/issues/496.

PR #467, opened 2022-11-30
short commit `sha` appended to commit message [conbench ui table]

sayantikabanik:
resolves #445

PR #479, opened 2022-12-01
UI updates : conbench index page

### LINKED ISSUE #478: Use the commit message as a link to run in the index.html
And then maybe add a small github icon for the link to the actual github commit?

sayantikabanik:
resolves #478 

~Logo details:~
~- GitHub logo updated (source - official GitHub site `github.com/logos`)~
~- size fixed to static 30*30, consistent with the author_avatar~

Updated details : https://github.com/conbench/conbench/pull/479#issuecomment-1361055908


jgehrcke:
Cool! Notably, this somehow overlaps with https://github.com/conbench/conbench/pull/467/files. When I see the short commit hash being used in a URL, I assume that the link takes me to the commit.

About the icon: what if the commit is not on GitHub but on a different platform like GitLab (I don't know what I am talking about)? If this shows a GitHub icon in each row then it would not really add information and maybe make things a little busy, visually. It's fine if you ignore my input for now, I assume that @jonkeane and you have probably talked about that face to face :).

austin3dickey:
>what if the commit is not on GitHub but on a different platform like GitLab

I believe we don't support other platforms at this time. Commit/repo/PR information is submitted via the API in the "github" key, which hits the GitHub API to gather more information. Adding support for other platforms would mean lots of changes around that workflow.

austin3dickey:
@sayantikabanik Could you please upload a screenshot of what this looks like? I'm having a bit of trouble visualizing it.

jonkeane:
Before I saw this PR I commented on t[he issue itself](https://github.com/conbench/conbench/issues/478#issuecomment-1333975875) (sorry!), but the operative part:

What if we did something like this, which would be a slight scope extension of re-ordering, but while we're here it might be nice):

| Run | Commit Message | Hardware | Reason | Commit SHA | Author | Repository | 
| --- | --- | --- | --- | --- | --- | --- |  
| [2022-12-01 15:19](https://conbench.ursa.dev/runs/d1be8f6c6273429294ef52f0ab6ae816/) |[GH-14787: [Java] [Doc] Update table.rst (#14794)](https://conbench.ursa.dev/runs/d1be8f6c6273429294ef52f0ab6ae816/) | voltron-pavilion | commit | [1d9f778 :github icon:](https://github.com/apache/arrow/commit/1d9f7781acf5e5b3679ac2d2aee9e837435ac87e) | Larry White | [apache/arrow](https://github.com/apache/arrow) | 

Or maybe we condense the run (timestamp) with the commit message into one field? And|or maybe we swap the repo and the author around here too?

sayantikabanik:
> @sayantikabanik Could you please upload a screenshot of what this looks like? I'm having a bit of trouble visualizing it.

- Icon here points to the commit on GitHub
- Commit message links to the respective run(s)

Screenshot 2022-12-02 at 9 22 59 AM

sayantikabanik:
> Or maybe we condense the run (timestamp) with the commit message into one field? And|or maybe we swap the repo and the author around here too?

another thought that I had was, combining  run (timestamp) with reason. Might also add a little extra detail about "when an accident occur?", "when was the PR created?" something on similar lines, if this is making sense. 

Commit(message) currently appears busy, adding any extra details might make it visually heavy to digest.

jonkeane:
Agreed that the commit message might be a little bit (too) busy. Want to try out a few different options with it and without it and put up some screen shots for us to decide between?

sayantikabanik:
> Agreed that the commit message might be a little bit (too) busy. Want to try out a few different options with it and without it and put up some screen shots for us to decide between?

Updated two variations here - https://github.com/conbench/conbench/issues/478#issuecomment-1334976039

conbench/templates/index.html
```diff
@@ -25,16 +25,16 @@ <h3>Language-independent Continuous Benchmarking (CB) Framework</h3>
 <div class="row">
     <div class="col-md-12">
         <table id="runs" class="table table-striped table-bordered table-hover">
-          <caption>Runs</div>
+          <caption>Runs
```
jgehrcke, line 28: Oh. How did that even work/render before? Reminds of many years ago when I was manually writing quite a bit of HTML, always involving https://validator.w3.org/ before being confident about validity.

sayantikabanik:
Ready for review. 

Changes done:
- New short commit hash col added (linked to commit URL)
- URL stripped from commit message 
- New css class added
- few JS changes to accommodate the updates

Updated UI:
Screenshot 2022-12-21 at 2 42 31 PM

conbench/templates/index.html
```diff
@@ -25,16 +25,16 @@ <h3>Language-independent Continuous Benchmarking (CB) Framework</h3>
...
           </caption>
             <thead>
                 <tr>
                     <th scope="col">Run</th>
                     <th scope="col">Author</th>
                     <th scope="col">Repository</th>
-                    <th scope="col">Commit</th>
-                    <th scope="col">Hash</th>
                     <th scope="col">Hardware</th>
+                    <th scope="col">Commit hash</th>
```
jgehrcke, line 36: I'd say let's just call this `Commit`. Then the header is as wide as the column content, and it's still precise.

jgehrcke: I'd say let's just call this `Commit`. Then the header is as wide as the column content, and it's still precise.

(in the screenshot the `Commit hash` column header is the only one that consumes two lines of text right now)

conbench/templates/index.html
sayantikabanik: Code editors at rescue :D

conbench/templates/index.html
```diff
@@ -25,16 +25,16 @@ <h3>Language-independent Continuous Benchmarking (CB) Framework</h3>
...
+                    <th scope="col">Commit</th>
```
jgehrcke, line 36: And I think this column header should then be `Commit message` :).

conbench/static/app.css
```diff
@@ -109,3 +109,10 @@ table.dataTable td {
   white-space: nowrap;
   overflow: hidden;
 }
+
+.table-entry {
```
jgehrcke, line 113: Why did we need to add this?

conbench/static/app.css
```diff
@@ -109,3 +109,10 @@ table.dataTable td {
...
+  width: -moz-fit-content;
+  width: fit-content;
```
jgehrcke, line 115: Why these two properties?

conbench/static/app.css
```diff
@@ -109,3 +109,10 @@ table.dataTable td {
...
+  padding: 3px;
+  margin-bottom: 1em;
```
jgehrcke: How did we decide between using padding vs. margin? 

When I look at the screenshot then I think for just a hand full of rows this looks ok/nice, but when displaying many entries (like, 100) then the additional spacing might maybe be wasteful?

We probably want central vertical alignment in each cell (same padding to top and bottom, and no margin).

conbench/templates/index.html
```diff
@@ -98,20 +113,12 @@ <h3>Language-independent Continuous Benchmarking (CB) Framework</h3>
...
 <script type="text/javascript">
-  var table = $('#runs').dataTable( {
-      "responsive": true,
+  var table = $('#runs').DataTable( {
       "order": [[0, 'desc']],
-      "columnDefs": [
-          {
-              "targets": [4],
-              "visible": false,
-              "searchable": true,
```
jgehrcke, line 108: Why can this safely go? For example, what did `"searchable": true,` mean and why is it ok to remove this? :)

jgehrcke (commented):
This looks great, thanks for the screenshot! :rocket: 

Also thanks for the four bullet points. I have left an idea about column headers, and two questions about the CSS and JS changes.

conbench/templates/index.html
sayantikabanik: Previously I was running into a issue described here :https://github.com/conbench/conbench/issues/478#issuecomment-1340697361

I was ignoring the JS bit and wasn't able to find the reason. This points to the col "Hash" to be silent for some reason (which honestly I don't have a solid reasoning to)

jgehrcke: Ah, this `"targets": [4],` means that the entire option block only targeted one specific column?  OK!

I think this patch is fine then even without further understanding.

Even if this breaks someone's flow. We have to go through this kind of pain to understand what users want, need.

conbench/static/app.css
sayantikabanik: Hit and trial mostly

- minimum relative margin (Eg: the commit col was touching the cell boundary, seemed little off)

conbench/static/app.css
sayantikabanik: The table currently is defined with class `col-md-12` (https://getbootstrap.com/docs/4.1/layout/grid/) equal size for all the cols.
This helps to overwrite those/ or we can assign different width for each.

conbench/templates/index.html
sayantikabanik: Sounds good

conbench/static/app.css
sayantikabanik: Added my thoughts below, short reasoning: overwriting the existing table class

conbench/static/app.css
jgehrcke: > equal size for all the cols.

The way I understand the current/previous setup is that `col-md-12` practically defines minimal width, and the columns could get wider with content. For example, current conbench.ursa.dev has wide columns, wider than 1/12th of total width.

This should not block this patch at all, just thinking out loud.

conbench/templates/index.html
```diff
@@ -54,28 +54,43 @@ <h3>Language-independent Continuous Benchmarking (CB) Framework</h3>
...
-                     {% if run.reason %}
-                       <td><div class="ellipsis-175">{{ run.reason }}</div></td>
-                     {% else %}
-                       <td></td>
-                     {% endif %}
+                    <td>
+                         {% if run.commit.display_message %}
+                            <div class="table-entry">{{ run.commit.display_message }}</div>
+                         {% else %}
+                            <div class="table-entry">{{ run.commit.sha }}</div>
```
austin3dickey: Maybe if there is no display message this should be just a blank string instead of the 40-character sha?

conbench/templates/index.html
```diff
@@ -54,28 +54,43 @@ <h3>Language-independent Continuous Benchmarking (CB) Framework</h3>
...
+                         </div>
+                    </td>
 
-                     {% if run.commit.display_message %}
-                       <td><a href="{{ run.commit.url }}">{{ run.commit.sha[:7] }} {{ run.commit.display_message }}</a></td>
-                     {% else %}
-                       <td><a href="{{ run.commit.url }}">{{ run.commit.sha }}</a></td>
-                     {% endif %}
+                    <td>
+                         <div class="table-entry">{{ run.hardware.name }}</div>
```
austin3dickey, line 72: You may still need the ellipsis property here; I think hardware names could be long (see the [arrow conbench](https://conbench.ursa.dev/) for example)

austin3dickey (approved):
Looks really good!

austin3dickey:
oh - you can delete the GitHub logo file now right?

jgehrcke:
@sayantikabanik if you rebase this onto current head of main and re-run CI -- then CI will create UI screenshots for us :)

sayantikabanik:
Do the screenshots appear on the PR (over here)? 💫

conbench/templates/index.html
sayantikabanik: I was thinking on similar lines, not sure if the users might be expecting the hash instead?

conbench/static/app.css
sayantikabanik: True that thanks:)
Inside the cells previously we were using `ellipsis-175` css class
That's where the width got restricted

conbench/templates/index.html
sayantikabanik: Blank made more sense (to me) I added the change. We can reverse if necessary

conbench/templates/index.html
sayantikabanik: The "table-entry" class has the "fit-width" property, that should cover this use case. 

To validate: screenshots from CSS editors (picked a random long hardware name)

Using "table-entry" class
Screenshot 2022-12-22 at 11 26 05 AM

Using "ellipsis-175" class
Screenshot 2022-12-22 at 11 27 41 AM

sayantikabanik:
Thanks @jgehrcke @austin3dickey for the feedback :)

Adding this recent screenshot below (just for quick reference)

Screenshot 2022-12-22 at 11 40 55 AM

sayantikabanik:
Will keep this open for some more time until Jan (post we return from holidays), since it's dev/user facing change. Open to any additional feedback.

conbench/templates/index.html
jgehrcke: > I think hardware names could be long (see the [arrow conbench](https://conbench.ursa.dev/) for example)

Hm. May there be a misunderstanding here? I believe that Austin's point was that we do want this column to have a max width, and not have it consume the full width of very long hardware names.

jgehrcke:
Thanks for the exchange of opinions on this, and for the work, and for the patience @sayantikabanik!

I think we should land this now. I am a tiny bit lost in the details here to be honest. However, I trust that it's highly likely an overall improvement! In case it is _not_, then landing this now and seeing the downsides is the most efficient way to learn (and I trust that the downsides would be of cosmetic nature, and not really break anyone's workflow). And then we can quickly iterate based on new insights, in January.

The strongest reason to land this now is I think of psychological nature:  this is / should have been a rather small-scope topic and it can quickly become frustrating to see something like this take so long between feedback cycles and multiple people being involved. I hope it feels good to see this landed before the holidays :christmas_tree:, and if this raises feedback then we hopefully will have it already readily available early Jan.

jgehrcke:
before/after froms cb-staging:
**Before**:
Screenshot from 2022-12-23 11-43-07

**After**:
image

I like that change.

sayantikabanik:
Is there a cb-staging URL which I can access?
(Got it, thanks)

austin3dickey:
Right now, Voltron Data is hosting a staging Conbench server internally, which automatically gets deployed upon merges to `main` here. We hope to make a public staging server at some point in the future.

Issue #480, opened 2022-12-01
oidc: staging environment against google shows `OAuth 2 MUST utilize https`


jgehrcke:
Seen yesterday when @austin3dickey and I deployed from (at that time) head of main to a staging env. Upon initiating the flow:
```
{"code": 400, "description": {"_errors": ["OpenID Connect single sign-on flow failed: (insecure_transport) OAuth 2 MUST utilize https.."]}, "name": "Bad Request"}
```

jgehrcke:
Debugging, with the help of logs, code, etc.

What fails is the conclusion of the flow, i.e. the error response shown above is to a GET request to `/api/google/callback?code=4%2F0Afg....`.

jgehrcke:
Corresponding access log line showing the 400 response, redacted:
```
172.31.9.214 - - [01/Dec/2022:13:30:34 +0000] "GET /api/google/callback?code=4%2F0AfgeXvu[...]ompt=none HTTP/1.1" 400 162 "https[...]/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/107.0.0.0 Safari/537.36"
```

jgehrcke:
While CI is running on the patch, my attempt to explain the problem:

This oauthlib method call raised an exception: https://github.com/conbench/conbench/blob/175439dfbcfc18c2c51bb4e1b43635c65207e02e/conbench/api/_google.py#L153

As of https://github.com/conbench/conbench/issues/464 even in 'production' when actual user agents reach the Conbench HTTP API via TLS, i.e. with the URL scheme `https://`, the WSGI environment might tell the app that the access happened via `http://`, so that `f.request.url` starts with the 'wrong' scheme.

Security-wise, we do not need oauthlib to do _that_ kind of validation. One way to disable is: always replace `http://` with `https://`. That is what we did in the legacy version. However, that approach was not viable anymore once we started testing the flow locally.

In the patch (#480) I am now taking the approach where we distinguish between test environment and all other environments.

jgehrcke:
This has been fixed, and also the next problem of the same kind was fixed by #499. Closing!

PR #481, opened 2022-12-01
oidc: address issue #480

jgehrcke:
The legacy approach included this scheme replacement magic ([code](https://github.com/conbench/conbench/blob/6d8bfe182ea6b1043dd99a79918bc72732ed783e/conbench/api/_google.py#L38)):
```
    token_url, headers, body = client.prepare_token_request(
        google["token_endpoint"],
        authorization_response=f.request.url.replace("http://", "https://"),
        redirect_url=f.request.base_url.replace("http://", "https://"),
        code=f.request.args.get("code"),
    )
```

Must do this now conditionally only "in production", and not in the testing environment. Not doing this case distinction led to #480.

### DIFF
--- a/conbench/api/_google.py
+++ b/conbench/api/_google.py
@@ -1,3 +1,4 @@
+import os
 import base64
 import logging
 import time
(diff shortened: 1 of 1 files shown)


austin3dickey (approved):
Seems good to me. Sounds like we should prioritize making INTENDED_BASE_URL required soon! 🙂

jgehrcke:
>  Sounds like we should prioritize making INTENDED_BASE_URL required soon! slightly_smiling_face

Yes :see_no_evil: 

Thanks for having a look!

jgehrcke:
Let's give this a try. With the current rate of (bad) luck I kind of believe this is not yet the end of the story of unbreaking legacy environments, but we will find out the hard way.

PR #489, opened 2022-12-02
app startup: log sanitized config on INFO level

jgehrcke:
For now, log two configuration objects of relevance: the app's global config object, and the derived Flask config object. (we seem to mix usage both, and for debugging it's important to have both).

How this looks on my machine:
```
[221202-13:18:32.499] [1] [conbench] DEBUG: Flask app config object:
{
  "APPLICATION_NAME": "Conbench",
  "APPLICATION_ROOT": "/",
  "BOOTSTRAP_CDN_FORCE_SSL": false,
  "BOOTSTRAP_LOCAL_SUBDOMAIN": null,
  "BOOTSTRAP_QUERYSTRING_REVVING": true,
  "BOOTSTRAP_SERVE_LOCAL": false,
  "BOOTSTRAP_USE_MINIFIED": true,
  "CREATE_ALL_TABLES": true,
  "DB_HOST": "db",
  "DB_NAME": "postgres",
  "DB_PASSWORD": "*******res",
  "DB_PORT": "5432",
  "DB_USERNAME": "postgres",
  "DEBUG": true,
  "DISTRIBUTION_COMMITS": 100,
  "ENV": "development",
  "EXPLAIN_TEMPLATE_LOADING": false,
  "INTENDED_BASE_URL": "http://127.0.0.1:5000/",
  "JSONIFY_MIMETYPE": null,
  "JSONIFY_PRETTYPRINT_REGULAR": null,
  "JSON_AS_ASCII": null,
  "JSON_SORT_KEYS": null,
  "LOG_LEVEL_FILE": null,
  "LOG_LEVEL_SQLALCHEMY": "WARNING",
  "LOG_LEVEL_STDERR": "DEBUG",
  "MAX_CONTENT_LENGTH": null,
  "MAX_COOKIE_SIZE": 4093,
  "OIDC_CLIENT_ID": "conbench-test-client",
  "OIDC_CLIENT_SECRET": "*******ret",
  "OIDC_ISSUER_URL": "http://dex:5556/dex-for-conbench",
  "PERMANENT_SESSION_LIFETIME": "31 days, 0:00:00",
  "PREFERRED_URL_SCHEME": "http",
  "PROPAGATE_EXCEPTIONS": null,
  "REGISTRATION_KEY": "*******ode",
  "SECRET_KEY": "******* TV",
  "SEND_FILE_MAX_AGE_DEFAULT": null,
  "SERVER_NAME": null,
  "SESSION_COOKIE_DOMAIN": null,
[...]

PR #492, opened 2022-12-02
config: repair TESTING detection

jgehrcke:
I was just browsing through Conbench instance logs and did not see the config being logged on INFO level. I expected that after https://github.com/conbench/conbench/pull/489. Found the oversight.


jgehrcke:
I went ahead with merging to see the outcome of this patch in conbench.ursa.dev after auto-deploy.

PR #494, opened 2022-12-05
config: make INTENDED_BASE_URL required

jgehrcke:
A few more changes on this PR.

For example, this changes the `Config` object being an instantiated class object instead of just being a class constructor. See commit messages for more detail.

### DIFF
--- a/conbench/config.py
+++ b/conbench/config.py
@@ -73,12 +73,6 @@ class ConfigClass:
     def __init__(self):
         self.OIDC_ISSUER_URL = self._get_oidc_issuer_url_from_env_or_exit()
 
-class TestConfig(Config):
-    DB_NAME = os.environ.get("DB_NAME", f"{APPLICATION_NAME.lower()}_test")
-    SQLALCHEMY_DATABASE_URI = (
-        f"postgresql://{Config.DB_USERNAME}:{Config.DB_PASSWORD}"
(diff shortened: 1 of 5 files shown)


conbench/config.py
```diff
@@ -87,10 +50,83 @@ class Config:
...
+        if oiu is not None:
+            assert self.OIDC_CLIENT_ID is not None
+            assert self.OIDC_CLIENT_SECRET is not None
+
+        return oiu
+
+
+Config = ConfigClass()
+# for legacy code that imports via `from ..config import TestConfig`
+TestConfig = Config
```
jgehrcke, line 131: Towards more pythonic naming, in the future the `Config` object should be renamed, reflecting that it's a config singleton object, globally available, ideally immutable after app startup. All capital letters to indicate a constant. I have chosen the old name `Config` here to not break all those modules that import this object.

docker-compose.yml
```diff
@@ -12,7 +12,7 @@ services:
       - dex
       - db
     environment:
-      APPLICATION_NAME: "Conbench"
+      APPLICATION_NAME: "conbench-for-pytest"
```
jgehrcke, line 15: This change gives a bit more explicit test coverage.

jgehrcke:
Merging this to see the effect on the staging environment.

In a separate patch we can then simplify the business logic to rely on this config var being set to a meaningful value.

PR #495, opened 2022-12-05
buildkite: disable conbench-deploy trigger

jgehrcke:
Quoting the commit msg:
```
Effectively, this disables auto-deploy
to conbench.ursa.dev upon a merge-to-main.

In more specific terms, this disables
triggering the pipeline

  apache-arrow/conbench-deploy

at the end of pipeline

  apache-arrow/conbench-test

(when the branch name is `main`).

What's not visible in this commit is
that we add a new build trigger for
another buildkite pipeline that will
deploy after merge-to-main into a
more appropriate staging environment.
```

### DIFF
--- a/.buildkite/conbench-test/pipeline.yml
+++ b/.buildkite/conbench-test/pipeline.yml
@@ -26,10 +26,3 @@ steps:
     command: "exit 0"
     concurrency: 1
     concurrency_group: "conbench-deploy"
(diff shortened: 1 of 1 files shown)


jgehrcke:
Landing this. The new deploy-to-staging webhook is in place. Let's see if it works.

PR #497, opened 2022-12-05
oidc: rely on INTENDED_BASE_URL to be set to a meaningful value

jgehrcke:
I like to do things step-by-step and now that https://github.com/conbench/conbench/pull/494 landed we can remove some of the daunting complexity.

### DIFF
--- a/conbench/api/_google.py
+++ b/conbench/api/_google.py
@@ -10,6 +10,9 @@ from ..config import Config
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


jgehrcke:
That's now expected to unbreak SSO in the staging enviroment. Let's see!

jgehrcke:
> That's now expected to unbreak SSO in the staging enviroment. Let's see!

There still was a breakage in the staging environment. With better log output it was now easy to spot and address.

Follow-up patch here: https://github.com/conbench/conbench/pull/499

Update: #499 worked. Good. Case closed, kindof.

PR #499, opened 2022-12-05
oidc: introduce parse_state_from_authorization_response(), fix state parser

jgehrcke:
The business logic in the callback endpoint became spaghetti code. Factor this part of the work out in its own function.

The two input parameters are now documented nicely, also via type annotations. At the heart this patch includes a change from `parse_request_uri_response(f.request.url)` to `parse_request_uri_response(cur_request_url_abs)` which addresses an oauthlib `insecure_transport` error (see code comments for a better explanation)

### DIFF
--- a/conbench/api/_google.py
+++ b/conbench/api/_google.py
@@ -151,6 +151,9 @@ def conclude_oidc_flow():
     # confirm that (it cannot reliably know).
     cur_request_url_abs = f.request.url.replace("http://", "https://")
 
+    # Parse target URL from state parameter in authorization response. If it's
+    # not an empty string then treat this as the URL the user actually wanted
+    # to visit before going into the login flow.
     user_came_from_url = parse_state_from_authorization_response(
         client, cur_request_url_abs
     )
@@ -207,20 +210,23 @@ def conclude_oidc_flow():
 
 def parse_state_from_authorization_response(c: WebApplicationClient, url: str) -> str:
     """
+    Parse target URL from state parameter in authorization response.
+
+    Raise exception or return the target URL as a string (may be of length 0).
+
     url: absolute URL that the user agent used to get here (modulo scheme, it's
     always expected to be HTTPS, see considerations above).
 
     Extract authorization response structure from incoming URL. Response is
-    expected to have retained the `state` parameter which we're using to store
-    the URL the user actually wanted to visit before going into the login flow.
-
-    This is expected to raise an exception or return a string object (which may
-    be of length 0).
+    expected to have retained the `state` parameter. Extract that, and then
+    decode
     """
     try:
         # This performs standards-compliant parsing of the so-called
         # authorization response and can therefore raise various oauthlib
-        # exceptions.
+        # exceptions. `url` is expected to always start with https:// which is
+        # why this is not expected raise an exception related to insecure
+        # transport.
         authorization_resp = c.parse_request_uri_response(url)
     except Exception as exc:
         log.info("parse_request_uri_response() failed: %s", exc)
--- a/conbench/api/_google.py
+++ b/conbench/api/_google.py
@@ -4,11 +4,9 @@ import time
 
 import flask as f
 import requests
-
 from oauthlib.oauth2 import WebApplicationClient
 
 from ..config import Config
-
 
 log = logging.getLogger(__name__)
 
--- a/conbench/api/_google.py
+++ b/conbench/api/_google.py
@@ -5,7 +5,10 @@ import time
 import flask as f
 import requests
 
+from oauthlib.oauth2 import WebApplicationClient
+
 from ..config import Config
+
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


conbench/api/_google.py
```diff
@@ -4,6 +4,7 @@
 
 import flask as f
 import requests
+from oauthlib.oauth2 import WebApplicationClient
```
jgehrcke, line 7: I have pulled this up so that I could use that in a type annotation.

conbench/api/_google.py
```diff
@@ -219,6 +206,42 @@ def conclude_oidc_flow():
     return user_came_from_url, userinfo
 
 
+def parse_state_from_authorization_response(c: WebApplicationClient, url: str) -> str:
```
austin3dickey, line 209: small nit: would you mind calling this `client`? Single-character variables are a little hard to read, IMO

austin3dickey (approved):
looks great! fingers crossed!

jgehrcke:
Merging, and keeping an eye on logs of staging environment.

jgehrcke:
This was automatically deployed to our staging environment where I could confirm that Google SSO works again.

conbench/api/_google.py
jgehrcke: Huh! :) On the one hand I can be happy to change this because I can totally choose to not care in this case.

But because it's a calm morning and I just meditated I feel like trying to put my opinion into words. Maybe for myself more than for others, it's an interesting exercise. Feel free to skip reading.

I think I do not generally agree with

> Single-character variables are a little hard to read

There certainly are cases when single-character variable names are dangerous and a bit difficult to deal with, especially when they are re-used and when their definition is rather distant from their usage. They can also be tiring, I know what you mean I think.

However, there are also other cases where they are even likable when used well, and can even make code more readable and look beautiful because they allow for focusing on the verbs more than on the subjects, grammatically speaking (if that makes sense). 

In many pieces of code there is a variable that takes front stage and represents a really important object within a rather local context. And this important entity is often very locally represented by this one character of text. Some examples that we often come across: `f` for file, `r` for reader, `w.` for writer, `s` for server, `c` for client, well :) Or `s` for stream, `r` for response, and so on. Oftentimes we use those in a Python context manager, but other times these one-character variables are also the arguments to _short_ functions.

Here, I think the reading distance between `c: WebApplicationClient` and `c.parse_request_uri_response(url)` is pretty short. That is, "what is `c`?" is quickly answered. More importantly, even `client` would not be able to convey the complexity of this kind of object anyway. If the argument was called `client` then the immediate question "what kind of `client` is this, exactly?" would come up, and the type annotation would again help a bit. That's also why -- mainly for readability -- I did use the type annotation here.

Maybe for completeness, two examples for (IMO) readable code:

```python
    # Create gipc pipe and expose the read end as `r` and the write end as `w`.
    with gipc.pipe() as (r, w):

        # Start child process for receiving messages. It inherits the standard
        # streams of this process and prints the received messages to stdout.
        p = gipc.start_process(target=child_process, args=(r, ))

        # Start greenlet (in the current, the parent, process). It periodically
        # sends a message to the child process through the pipe, via gipc's IPC.
        g = gevent.spawn(_writegreenlet, w)

        # Keep the pipe alive, and let the two entities communicate as long as
        # the shutdown procedure has not been invoked.
        while not shutdown:
            gevent.sleep(0.01)

        # Once we're here the shutdown procedure has been invoked. Terminate the
        # message sender greenlet. `kill()` always returns None; never raises an
        # exception.
        g.kill(block=True)

        print('Write greenlet terminated. Send SIGTERM to child process')
        p.terminate()

        # Wait for child process to terminate, reap it (read exit code).
        p.join()
        print('Child process terminated. Exit code: %s' % (p.exitcode, ))
```
(from [here](https://github.com/jgehrcke/gipc/blob/4f5195c1c830e4e8684c6b558fdad5dd89c7930f/examples/infinite_send_to_child.py#L77))


And then maybe this here from the Go stdlib docs:
```golang
s := &http.Server{
	Addr:           ":8080",
	Handler:        myHandler,
	ReadTimeout:    10 * time.Second,
	WriteTimeout:   10 * time.Second,
	MaxHeaderBytes: 1 << 20,
}
log.Fatal(s.ListenAndServe())
```

I should stop typing now. Hope this doesn't waste anyone's time. :)

austin3dickey: haha I really appreciate the thorough response! It's much more clear to me that this is a stylistic opinion that depends on diverse backgrounds and experiences reading text. Different people have different opinions of beauty.

My background: On previous teams I've been on, we tried to stay away from single-character arguments to functions, because we had lots of poor experience trying to understand unreadable code with non-intuitive or ill-defined variable names. So it's just something I've become accustomed to when reading code. For example, I still think those code blocks you attached could be improved with a few more characters. 😉  But I agree that I use them all the time for context managers. And my comment was just a `nit` so I love hearing your opinions :)

jgehrcke: <3

PR #500, opened 2022-12-06
Remove Buildkite from per-PR CI

jgehrcke:
See #319.


jgehrcke:
I disabled the webhook for `buildkite/conbench-test` -- I think that's now an exceptional case where this status will stay yellow here, and not disappear.

jgehrcke:
Note to self: at the heart of the Kubernetes database migration job we run `alembic upgrade head` directly:

https://github.com/conbench/conbench/blob/ddf496c77a7624fe3087c82fe53e7ec4e78b134c/migration-job.yml#L24

That speaks in favor of removing the docker-compose `migration` service and instead run this directly in the `app` service.

jgehrcke:
Going ahead with this.

PR #505, opened 2022-12-06
tests: speedup: empty tables instead of drop_all/create_all

jgehrcke:
commit msg:
```
This is based on prior experience.
In my local environment that speeds up the test
suite duration from ~120 seconds to ~80 seconds.

In GHA the test suite takes considerably longer
(~300 seconds), and the same relative speedup
would be be a noticable wall time win.
```

### DIFF
--- a/conbench/db.py
+++ b/conbench/db.py
@@ -1,6 +1,5 @@
 import functools
 import logging
-import time
 
 import sqlalchemy.exc
 import tenacity
--- a/conbench/db.py
+++ b/conbench/db.py
@@ -1,9 +1,13 @@
+import functools
 import logging
+import time
 
 import sqlalchemy.exc
 import tenacity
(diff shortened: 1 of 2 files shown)


jgehrcke:
`3m 43s` (now) vs. `4m 55s` (in the previous PR), for the `make tests` step. That's good.

A significant portion of the time in those ~4 minutes is spent building the container images which we can and should also cache between builds.

jgehrcke:
It's unclear to me what else needs to be done to get rid of this stale `buildkite/conbench-test` status report:

image

(the webhook is deleted from GH, and the corresponding pipeline in BK is set to `Disable all GitHub activity` -- from here, we need to do research).

jgehrcke:
Going ahead with this. As always, post-merge discussion very very welcome.

Issue #473, opened 2022-11-30
Change z-score calculations to incorporate `begins_distribution_change`


austin3dickey:
See comments below for full discussion. Current plan: when calculating the z-score of a BenchmarkResult `B` on commit `C`,
- for the distribution mean, use the mean of the means of all BenchmarkResults that:
  - share the same case/context/hardware/repo as `B`
  - are on the default branch
  - are in the commit ancestry of `B`
  - starting with either the earliest BenchmarkResult that matches these criteria, or the latest matching BenchmarkResult where `begins_distribution_change` is True, whichever is later
  - ending with the latest matching BenchmarkResult before `C`
- for the distribution standard deviation, calculate the sample standard deviation of the differences between the mean of each BenchmarkResult from the rolling distribution mean (which is calculated the same way as above)
  - This always starts at the earliest BenchmarkResult that matches the criteria, not accounting for `begins_distribution_change`
- use the mean of `B` for the z-score calculation

jonkeane:
A few options we discussed in #383, but with more summary and polish here:

We could (and IMHO should) treat the step-change markers we have from #472 and #574 as known-mean shifts in the distribution. 

We currently use two measures of history: the mean of the measure (for tracking central tendency) and the standard deviation. If we centered all of our measures around the means from their segments before calculating the standard deviation for the whole set, we will have our measure of variability that is independent from the mean shifts. IIRC, @alistaire47 called this the standard deviation of the residuals (where the model is the mean of the segment). I'm happy to throw some code together that shows this off (in the style of what's in #383), but need to do that later.

We also should consider adding an exponential weighting so that past history has less (or effectively no) influence on recent measures. But IMO we should put that as a separate ticket and not try and tackle it here.

I also think even with just this scaling, we probably are at the point that we want (or possibly even need) to pull these calculations out of being executed exclusively in the database.

Finally, we should consider what this looks like in the UI + how we message to folks about this. I think improvements to the plots (some of #477) to show the area of uncertainty and if bots fall outside of it is absolutely critical. After that, we should describe why we're using history for this. And finally, ideally we would also have a robust and detailed document describing the approach + math for folks that want to dive that deeply

alistaire47:
To summarize my thoughts on how to do this: 

1. Calculate the means (or median or whatever) of each segment (separated by step changes)
2. Subtract off (1) from the raw data to get residuals across segments
[...]

Issue #442, opened 2022-11-18
consider replacing print() with logging infra, add request context


jgehrcke:
It's a good idea to not mix stdout/stderr for the app's log output because then the relative timing between messages sometimes becomes unclear (when querying logs from a log aggregator).

It also makes sense to add a bit of an HTTP request ID to log messages so that it becomes unambiguously clear which log messages correspond to which HTTP request. This is necessary for debugging stuff when two or more HTTP requests are ever handled concurrently (not sure if our gunicorn deployment method even allows for that or not).

We can cheaply get both by using Python's logging scaffold from the stdlib.

jgehrcke:
First steps here: https://github.com/conbench/conbench/pull/451

PR #507, opened 2022-12-06
Reworked queries to use git history

### LINKED ISSUE #382: Use the actual git history to situate commits on branches

austin3dickey:
Finally resolves #382.

This PR reworks the SQLAlchemy queries used throughout the codebase that search for BenchmarkResults, in order to utilize the new git information we store on Commits. Highlights:

- A new `Commit.ancestor_commit_query` property that lets you easily access a Commit's ordered ancestors
- Removes all the brittle filters of `run.reason == "commit"` in favor of smarter ways of filtering commits:
  - A Distribution now uses commits from the commit's true ancestry (including PR commits)
  - z-scores now use commits from the default branch in the commit's true ancestry
  - Histories (powering plots) now use commits from the default branch
  - Default baseline run is from the last commit with a matching result in the commit ancestry, which could be a PR commit
- When getting BenchmarkResults from the API, now z-scores can be set in more instances
- Fixed a bug where histories could return BenchmarkResults from other repos

I should have split this into smaller PRs, but there were many interconnected pieces that I'd need to constantly be changing, so I'm presenting this as one big change. Apologies for the large review.


conbench/api/_examples.py
```diff
@@ -159,7 +159,7 @@ def _api_commit_entity(commit_id, parent_id, links=True):
             "self": "http://localhost/api/commits/%s/" % commit_id,
         },
         "branch": "some_user_or_org:some_branch",
-        "fork_point_sha": "some_fork_point_sha",
+        "fork_point_sha": "02addad336ba19a654f9c857ede546331be7b631",
```
austin3dickey, line 162: This PR is relying on the [assumption](https://github.com/conbench/conbench/pull/431#discussion_r1021908358) that a commit's sha == its (non-None) fork_point_sha if and only if it's on the default branch. I had to update our fake commits to reflect that.

conbench/api/benchmarks.py
```diff
@@ -105,30 +105,22 @@ def get(self):
                 filters=[Case.name == name_arg],
                 joins=[Case],
             )
-            # TODO: cannot currently compute z_score on an arbitrary
-            # list of benchmark_results - assumes same machine/sha/repository.
```
austin3dickey, line 109: Fixing this TODO was a nice side effect of this PR.

conbench/api/_examples.py
jgehrcke: Quick note about terminology. The Internet often uses "sha" instead of "hash", whereas "hash" really is the better word.

I trust we all know this, but "sha" is actually SHA which is a rather well-known abbreviation for a very specific set of _algorithms_.  That is, "a sha" is not "a hash", because the hash is the output of the algorithm. Well well. :)

jgehrcke: That is, in code I kind of find `_hash` better than `_sha` :nerd_face:  I'd totally understand if the idea of renaming sounds like a huge nono, especially because of database schema etc.

conbench/entities/commit.py
```diff
@@ -36,6 +42,69 @@ class Commit(Base, EntityMixin):
...
+        if self.sha == self.fork_point_sha:
+            return self
+        else:
+            return Commit.first(sha=self.fork_point_sha, repository=self.repository)
+
+    @property
+    def ancestor_commit_query(self) -> Query:
+        """Return a query that returns the IDs and timestamps of all Commits in the
+        direct ancestry of this commit, ordered starting with this commit, backwards in
+        lineage.
```
jgehrcke: > backwards in lineage

I trust that this is precise.

Can we make this a little more vivid for me, maybe use git terminology or maybe just tree or graph terminology, may an ASCII drawing?

Does this walk the tree upwards, taking the shortest path towards the default branch, and then it walks along the default branch?

In a git repo each commit is guaranteed to have at most _one_ parent (right?), so what I just said is probably obvious, and implied by "backwards in lineage".

If I were to have a local checkout of a repository, can we say which `git` command we would run to end up with the same ordered list of hashes?

For example, one can run this command in the conbench repo to find the ancestry of commit `e5d5257`, it will output at most 100 commit hashes, in order of history:
```
$ git log --pretty=%P -n 100 e5d5257
```

Which brings me to a question: do we (intend to) store the (partial) history of a git repo in our database, i.e. the relationship between commits?

Thought: for looking up complex history, there might be the interesting debate of doing local `git ...` operations (managed by Conbench) vs. doing complex database queries. The former might make things easier in terms of staying up-to-date with history changes in the remote end.

Feel free to ignore, this is me thinking out loud while trying to become familiar with the topic.

conbench/entities/distribution.py
```diff
@@ -108,109 +111,160 @@ def get_distribution(benchmark_result, limit):
...
 
-def get_closest_parent(run):
-    commit = run.commit
-    if commit.timestamp is None:
-        return None
+def get_closest_ancestor(
+    benchmark_result: "BenchmarkResult", branch: Optional[str] = None
+) -> Optional[Commit]:
+    """Given a BenchmarkResult, return the most recent ancestor commit on the given
+    branch that also has a BenchmarkResult in the database with the same
```
jgehrcke: Great docstring, thanks!

conbench/entities/distribution.py
```diff
@@ -108,109 +111,160 @@ def get_distribution(benchmark_result, limit):
...
+    hardware/case/context. If branch is not given (default), search all branches.
```
jgehrcke: > If branch is not given (default), search all branches.

I am not really clear what the starting point of this search is. Is this guaranteed to be a single commit, implicitly defined by `benchmark_result`?

Is the following statement true? Walk the git commit tree upwards and find the first commit that has a benchmark annotation.

How far back does this go at most, up to the origin? If not: maybe "walk at most N steps"?

If my mental model is correct, then in the process of walking backwards we might or might not visit different branches -- do we really care? It seems like "If branch is not given (default)" we do not care.

But if `branch` is given: what really is the behavior then?

It's cool to see all this -- and it makes me wonder how appropriate our database schema / DB choice is for searches like that, and how working with `git` directly (or tooling from the `git` ecosystem that directly operates on git repos in the local file system might help.

conbench/entities/distribution.py
```diff
@@ -108,109 +111,160 @@ def get_distribution(benchmark_result, limit):
...
+    # ancestor commit to each BenchmarkResult regardless of branch, which isn't ideal
+    # (see docstring) but it's good enough.
+    a_default_commit = benchmark_results[0].run.commit.get_fork_point_commit()
+    default_branch = a_default_commit.branch if a_default_commit else None
 
-    distributions = Session.query(*cols).filter(*where).all()
-    lookup = {f"{d.case_id}-{d.context_id}": d for d in distributions}
+    # Cache Distribution query results by commit/hardware/case/context, in case multiple
+    # input benchmark_results (e.g. from the same Run) share those attributes
+    cached_distributions = {}
```
jgehrcke, line 175: The stupid saying is: there's only two hard problems in ...
- off-by one
- naming
- cache invalidation -- haha

So. When/how is this cache invalidated? I was `Ctrl + F` this patch to find the word "invalidation" but didn't find it :)

jgehrcke (approved):
@austin3dickey this is great work! Thanks for what's obviously a huge effort :rocket:. Great code comments, pretty clearly readable. However, this is complicated matter and despite the code clarity, I cannot properly review this in a time-efficient manner w/o talking to you or sitting in front of the screen with you.

To make my review useful I'd like to focus first on behavior/correctness of code (and not on style etc), but for doing that I understand too little about the actual goals so far.

I can see that there are various really important discussion points here in problem/solution space, and also about architectural choices.

Let's sit together and you can explain all this to me, I would love for us to do that! However, we do not need to wait with merging this PR (we can, though).

conbench/api/_examples.py
austin3dickey: Haha, yes I thought the same thing when I onboarded onto this codebase. But "sha" is EVERYWHERE (not only in this repo, but also in many other repos that use Conbench)

conbench/entities/commit.py
alistaire47: Some thoughts related to JP's, some not:

At first I was unsure of the name; it sounded like it was to get a particular ancestor commit, when really it's getting the whole history. Coming back after reading the code, I get the intent better and mind less, but maybe `commit_ancestry_query` or `commit_history_query` would be more obvious?

As there's no limit on this query, it will return everything previous on `fork_point_commit.branch`, which in the case of a busy repo like Arrow can be a lot (I see 34,298 commits with branch `apache:master` in postgres). Depending on how this is used, that's fine! But maybe a comment or note in the docstring of warning?

For the ordering bit, even if they're ordered, I tend to assume SQL results are unordered in downstream processing. Is there value to adding an index column so order is easily maintained downstream? Otherwise I worry somebody will not take the care you have here and sort by timestamp because it makes sense at first.

Per JP's comments: 

> Thought: for looking up complex history, there might be the interesting debate of doing local `git ...` operations (managed by Conbench) vs. doing complex database queries. The former might make things easier in terms of staying up-to-date with history changes in the remote end.

this does feel like basically `git log`, and we do have lots of other bits that similarly feel like other git commands. It does seem like an interesting idea to use git server-side directly instead of reimplementing functionality. Probably we wouldn't want to download files; [it looks like maybe that's possible](https://stackoverflow.com/questions/59276829/how-to-git-clone-without-downloading-the-file-objects)? But it does seem right now the work that would entail outweighs the benefits, but maybe that balance will shift.

conbench/entities/distribution.py
```diff
@@ -1,15 +1,24 @@
...
 from ..entities._entity import Base, EntityMixin, NotNull, Nullable, generate_uuid
-from ..entities.commit import Commit
+from ..entities.commit import CantFindAncestorCommitsError, Commit
 from ..entities.hardware import Hardware
 from ..entities.run import Run
 
+if TYPE_CHECKING:
+    from ..entities.benchmark_result import BenchmarkResult
+
+log = logging.getLogger(__name__)
```
alistaire47, line 20: I know this repo has mostly had print logging previously and there's desire to do more structured logging; does it make sense to create a single logger all files could use, or do we want each file to be separate like this? If so it should probably be in a separate PR; I think I saw an issue for logging improvements already.

conbench/entities/distribution.py
```diff
@@ -50,26 +59,21 @@ class Distribution(Base, EntityMixin):
...
+    """
     from ..entities.benchmark_result import BenchmarkResult
 
-    commits_up = (
-        get_commits_up(benchmark_result.run.commit, limit)
-        .subquery()
-        .alias("commits_up")
-    )
+    commit: Commit = benchmark_result.run.commit
+    ancestor_commits = commit.ancestor_commit_query.limit(commit_limit).subquery()
```
alistaire47: Ah, here's the `limit`. I guess this part is here because parameterizing a property is weird. I don't love separating the parts of the query, but it does make sense in context.

conbench/entities/distribution.py
```diff
@@ -108,109 +111,160 @@ def get_distribution(benchmark_result, limit):
...
+) -> Optional[Commit]:
+    """Given a BenchmarkResult, return the most recent ancestor commit on the given
+    branch that also has a BenchmarkResult in the database with the same
+    hardware/case/context. If branch is not given (default), search all branches.
 
-    hardware_entities = Hardware.all(hash=run.hardware.hash)
-    hardware_ids = set([m.id for m in hardware_entities])
+    Return None if one isn't found. Should not raise anything.
+    """
+    from ..entities.benchmark_result import BenchmarkResult
```
alistaire47: Why import this here instead of at the file level? Am I missing something?

conbench/entities/commit.py
austin3dickey: For JP's comments: Good ideas; I just added a comparison to `git log` and an ASCII graph to the docstring.

>do we (intend to) store the (partial) history of a git repo in our database, i.e. the relationship between commits?

Yes! This was the solution we chose to solve https://github.com/conbench/conbench/milestone/1 . (Click "show more" in the description of that milestone to see more context.)

>doing local `git ...` operations (managed by Conbench)

This is a good idea, but it can't deal with git history rewrites well, which happen often in the repos we track. These queries are complex but are able to fully track the temporal relationships between different BenchmarkResults that Conbench stores, even after history rewrites.

conbench/entities/distribution.py
alistaire47: Yeah, the branch handling here seems weird to me in two ways, operating under the assumption that the goal here is to get the closest ancestor with results:

1. If `branch` is populated, it can only effectively be the current branch or the ancestor branch (probably the default branch); otherwise because it starts from `ancestor_commit_query`, there will be no results on other branches (which would be weird anyway). I'm not sure any of this could handle multiple ancestor branches (e.g. a longer-running feature-branch or some dev branch shenanigans), but that's probably out-of-scope for now anyway.
2. If `branch` is not populated, this docstring makes it sound like it's going to search any branch and get the last commit with a result (which might be a very odd comparison), but since it's using `ancestor_commit_query` still, I don't think that's true—it would only use those from the current or ancestor (i.e. default) branch. Which makes sense.

So if I understand correctly, if `branch=None` it will get the parent (or next ancestor with results); if `branch=<default/ancestor branch>` it will get the fork point (or next ancestor with results), and no other value for `branch` will return anything useful. Maybe just a flag for ancestor branch instead of taking any branch?

tl;dr I think the code is sensible, but the framing is confusing

conbench/entities/distribution.py
```diff
@@ -108,109 +111,160 @@ def get_distribution(benchmark_result, limit):
...
+        query = query.filter(ancestor_commits.c.ancestor_branch == branch)
 
-    parent = None
-    if result:
-        commit_id = result[1]
-        parent = Commit.get(commit_id)
+    # see Commit.ancestor_commit_query() comments for why we order this way
+    closest_benchmark_result = query.order_by(
+        "branch_order", ancestor_commits.c.ancestor_timestamp.desc()
+    ).first()
```
alistaire47: ah yes, this is exactly the case that I was thinking of when suggesting an index column. It's fine without it if we always remember to sort this way when messing with these results in the future; an index just makes it more obvious that maybe you shouldn't sort by timestamp

conbench/entities/distribution.py
austin3dickey: For JP's comments:

>Is this guaranteed to be a single commit, implicitly defined by benchmark_result?

Yes, perhaps I should clarify on the docstring: a BenchmarkResult always belongs to one Run which belongs to one Commit. This function attempts to find the closest ancestor commit to the given BenchmarkResult's commit, so that it can look up the distribution's mean and standard deviation at that time, so that we can calculate a z-score.

>Is the following statement true? Walk the git commit tree upwards and find the first commit that has a benchmark annotation.

I'd say "Walk the git commit tree backwards and find the first commit that has a recorded BenchmarkResult that shares the given BenchmarkResult's hardware, case, and context." We've decided in the past that BenchmarkResults with different combos of hardware/case/context are incomparable.

>How far back does this go at most, up to the origin?

Yes, up to the origin. Imposing a commit limit on this query wouldn't improve performance.

>If my mental model is correct, then in the process of walking backwards we might or might not visit different branches

Yes, and this is okay. If branch is not given, we'll return the first matching commit. If the branch is given, we'll return the first matching commit on the given branch. The default behavior of `set_z_scores()` uses this argument to filter to the default branch, so that we only compare the BenchmarkResult to results on the default branch.

conbench/entities/distribution.py
austin3dickey: In this case, the cache isn't referenced anymore after exiting this function scope (so it's eventually garbage-collected). So the cache won't be invalidated during a call to `set_z_scores()`. 

I don't think we need to worry about one of these query results changing during one call to `set_z_scores()`, because:

- that only would happen when BenchmarkResults are being POSTed at the exact same time they're being GETed (probably rare)
- even if it does happen, in most cases the z-score might only be changed by a few decimal places (since theoretically, the distribution mean/standard deviation don't have big changes too often)

conbench/entities/distribution.py
```diff
@@ -108,109 +111,160 @@ def get_distribution(benchmark_result, limit):
...
-    ]
-    if len(benchmark_results) == 1:
-        where.extend(
-            [
-                Distribution.case_id == first.case_id,
-                Distribution.context_id == first.context_id,
-            ]
-        )
+    For the start commit of the distribution, uses whatever commit_limit was used when
+    writing to the Distribution table.
```
alistaire47, line 168: As we get further down the change-detection road, this is going to change. At a glance I was thinking it might make sense to structure this so that if we want to change this to "since commit _x_ which was a mean-shift", but really even as we get there pragmatically we'll probably still want to keep a cap on the max commits to go back (if nothing has changed with some very-stable code, there's no point in going back 10 years anyway). So future changes would be additive, I think? Does that ring roughly true?

conbench/entities/distribution.py
```diff
@@ -108,109 +111,160 @@ def get_distribution(benchmark_result, limit):
...
+
+        if (
+            benchmark_result.mean is not None
+            and distribution
+            and distribution.mean_mean is not None
+            and distribution.mean_sd  # is positive
+        ):
+            benchmark_result.z_score = (
+                benchmark_result.mean - distribution.mean_mean
+            ) / distribution.mean_sd
```
alistaire47: 1. thank you for `d` -> `distribution`; much more readable
2. I see this is the way it already was, so this is fine, but maybe we should talk about whether using `benchmark_result.mean` is the best choice. I see a few considerations:
  a. Median would be less influenced by outlier iterations. But [arguably] outliers matter here, so mean is more apt. Median plus separate calculations for spread would be more robust, if we want to go there.
  b. I think there's value here to using the z-scores of the raw iteration times instead of the means; the more means we take, the more the central limit theorem will mute interesting signal. The nice part is then handling spread becomes easier; we just return all the data instead of worrying about summary stats of summary stats of summary stats.

Not sure that's all right, but I do think this is an area worth more thought. If others agree (?), let's open an issue.

conbench/entities/history.py
```diff
@@ -32,7 +34,14 @@ class HistorySerializer:
...
 
-def get_history(case_id, context_id, hardware_hash):
+def get_history(case_id, context_id, hardware_hash, repo) -> List[tuple]:
+    """Given a case/context/hardware/repo, return all non-errored BenchmarkResults
+    (past, present, and future) on the default branch that match those criteria, along
+    with information about the stats of the distribution as of each BenchmarkResult.
+    Order is not guaranteed.
+
+    Primarily used to power the blue line in the timeseries plots.
+    """
```
alistaire47, line 44: Thank you for the docstring! I think (?) this also powers the `/history/` API as well, which is a rational place for anybody doing their own data investigations into their metrics to start.

alistaire47: Hmm, so this _only_ gets results on the default branch. That makes sense for the blue line, but it would be very handy to have a version of this that would return `ancestor_commit_query` joined with results

conbench/entities/run.py
```diff
@@ -80,57 +86,79 @@ def create(data):
...
+        - is in this Run's commit ancestry
+        - shares this Run's reason (but see note)
+        - has *any* BenchmarkResults that share the hardware/case/context of *any* of
+            this Run's BenchmarkResults
+
+        Note: if there are no matches for those criteria, we search for ANY run_reason
+        on the default branch. This helps runs on the first commits of pull requests.
+
+        Returns None if there are no matches.
+        Returns a random Run if there are multiple matches.
```
alistaire47: er...this seems worrying. I guess this is the case where there are multiple runs on a machine for one commit? In that case, it may be worth prioritizing runs where nothing failed, given one reason for that situation would be something external caused a run to fail, so it got rerun. Though if we're not actually doing that there's less reason to filter, I guess. A warning log or something seems prudent, though.

alistaire47 (approved):
This is a lot! And good. A few thoughts, some of which may deserve issues.

Apologies if any comments end up out-of-order; I'm pretty sure more have come in before I finalized my review.

conbench/entities/commit.py
austin3dickey: For Ed's comments:

>maybe commit_ancestry_query or commit_history_query would be more obvious?

Yeah! Since `history` is used in other contexts elsewhere, maybe I'll change to `commit_ancestry_query`

>But maybe a comment or note in the docstring of warning?

Will do. This was intentional because downstream queries that use this as a subquery can choose whether to impose their own limit or not.

>Is there value to adding an index column so order is easily maintained downstream?

Ah, this is a much better idea than copying/pasting the rationale everywhere. 🙂

conbench/entities/distribution.py
austin3dickey: For Ed's comments:

> this docstring makes it sound like it's going to search any branch

Yeah, I can clarify this docstring. Originally I had two separate functions—`get_closest_ancestor()` and `get_closest_ancestor_default_branch()`—but they were so similar that I decided to combine them. 

>Maybe just a flag for ancestor branch instead of taking any branch?

Since the default branch name isn't standard, we need to figure it out once and pass it to this function.

---

Okay, I just had another thought when writing this. I think this function is redundant with `Run.get_closest_ancestor()`. We could just do `benchmark_result.run.get_closest_ancestor().commit`, as long as I can figure out how to subset that to default branch commits. I'll mess around with this.

conbench/entities/distribution.py
austin3dickey: This setup was proposed in #451; when each file is separate, I do like how it looks in our log archive. I think we're planning to replace all `print()`s in #442.

conbench/entities/distribution.py
austin3dickey: Yeah, there are a lot of uses of `commit.ancestor_commit_query` in this PR, and this is the only place that needs a limit. I kinda like how it's a method of `Query` so you can modify the `ancestor_commit_query` after the fact, without having to change the internals of that property.

conbench/entities/distribution.py
austin3dickey: There would be a circular import due to

https://github.com/conbench/conbench/blob/614e947a1b95525ba48c57ded6559ca8f9de8b82/conbench/entities/benchmark_result.py#L22

I tried reversing the direction of this functional import, but it caused a lot of issues with SQLAlchemy (I think it could be magically using import order to determine how to delete tables?) so I left the direction as it was before this PR. I decided it wasn't worth the time to investigate.

conbench/entities/distribution.py
austin3dickey: yes yes I love this!!

conbench/entities/distribution.py
austin3dickey: I was sorta excited to get rid of the limit in #473, but I suppose it's a good point that there's no point going back 10 years. Though LIMIT in a subquery doesn't typically increase performance, as far as I'm aware. Either way, I'm thinking I'll revisit when working on that issue.

conbench/entities/distribution.py
austin3dickey: When calculating `mean_mean`, couldn't we do a weighted-by-N mean of means in order to get the same result as taking the mean of the raw times? And if so, maybe we could do that with SD too (I'm rusty on stats)? The only trouble with using raw data is that Distribution is (I think) intended as a performance enhancer by calculating things quickly upon write, not read

conbench/entities/history.py
austin3dickey: I was assuming no one except the webapp used `/api/history` so thanks for proving me wrong 🙂 maybe you could make another issue (related to #384?) that adds a way to return the non-default commits too? I'm sticking with default branch here since that was the behavior before

conbench/entities/run.py
austin3dickey: hm, yeah it would have to be 2 runs with the same commit, same reason, same hardware, sharing at least 1 case/context (same as one on the given Run). Oh, I could tiebreak with latest Run timestamp/finished_timestamp?

alistaire47: Yeah, I was thinking about timestamps, but unfortunately it seems like the two cases I can think of for this happening suggest opposite heuristics: 

- if something broke and another build was necessary, later is better
- if someone is kicking off builds by hand in a way they shouldn't, earlier is better

Regardless, deterministic results are good, so maybe any choice is better than none

conbench/entities/history.py
alistaire47: My thoughts on what an ideal history endpoint might look like haven't fully coalesced yet, though I do think there's more we can do. Curious for @boshek's thoughts on the matter as well in terms of what's useful

conbench/entities/distribution.py
alistaire47: Yeah, weighting would be good if the number of iterations per run can vary. (It probably _usually_ doesn't, but in some cases may.)

But I think there's more to try here in how we're aggregating; something like the mean of the absolute value of the z-scores of each iteration (instead of the z-score of a mean of each iteration) would mute spread information less (albeit at the cost of the sign). A knob to play with later, anyway, if/when there's time to pull lots of data and experiment.

conbench/entities/run.py
jgehrcke: >  deterministic results are good

I agree that there should be one definite result here.

In the longer run, maybe all matches should be used in the information flow. What, fundamentally, is a good reason to pick just one data point if all data points are 'good' (meaningful) ones?

conbench/entities/commit.py
austin3dickey: Okay, I reworked `commit_ancestry_query` to return a `commit_order` column that downstream queries may take advantage of.

conbench/entities/distribution.py
austin3dickey: Okay I deduplicated some code and am now using `Run.get_baseline_run()` in `set_z_scores()`. Got rid of this confusing `get_closest_ancestor()` function entirely.

conbench/entities/run.py
austin3dickey: Cool, I went with tiebreaking by choosing the latest Run.timestamp.

>In the longer run, maybe all matches should be used in the information flow.

Yeah. Right now this powers the `compare ->` button at the top right of a page like [this](https://conbench.ursa.dev/runs/0bc0e39524c14349bbb0556e63799e13/). That compares a run to exactly one other run. But with #384 coming up, in the future there could be a way in the webapp to pick the baseline run out of a list of "recommended" baseline runs. I think we could accommodate this edge case there.

austin3dickey:
I addressed all the feedback, thanks all!

I just tried to get a quick, rough estimate of performance by making the database calls associated with hitting https://conbench.ursa.dev/api/compare/commits/3ce5e3446e69d039f47ca2c5ecd0f5b4bd1fc459...16d40c844a98129299224d871b65f258d5ec8741/ . It took a *lot* longer with this PR's changes. I'm going to spend some time figuring out how I can cut down on that time.

jgehrcke:
> It took a lot longer with this PR's changes. I'm going to spend some time figuring out how I can cut down on that time.

OK, cool. Good to keep an eye on perf. What would the fallout be if we were to merge this as-is? Would some of our deployments start breaking visibly?

Hope you don't need to go too crazy trying to squeeze out perf improvements with the current approach. I think some slow-down is OK, initially.

The way I look at this big patch: it's an extremely valuable milestone towards finding and locking in the 'intended user-facing behavior' (modulo performance, probably). Once we are confident that that's what users want API-wise etc (but faster!) we can look into radically changing implementation towards making things a lot lot lot faster (orders of magnitude, hopefully), but then against tests etc. 

As expressed in some of the discussion above, I think there might be good reasons to potentially iterate on very high-level architecture decisions. It's the right thing to not try to answer all the architecture questions before landing this, but we should also consider that things might (have to?) change a lot:
- what's done as part of request processing and what's done out-of-sync
- which role the relational database will after all play vs. using smart caching, and potentially other crazy but simple ideas  (like using an in-mem graph database, and/or using an abstraction on top of an actual fs-based git repo, ... ?)
- the way we represent and organize/manage git repositories within conbench

austin3dickey:
>What would the fallout be if we were to merge this as-is?

Good question. I set up a little test script locally that connects directly to the Arrow Conbench DB (with read-only creds) and runs the same python code that runs when you hit https://conbench.ursa.dev/api/compare/commits/3ce5e3446e69d039f47ca2c5ecd0f5b4bd1fc459...16d40c844a98129299224d871b65f258d5ec8741/ . On the `main` branch this takes about 3 seconds. On my branch this takes 1-2 minutes. They return the same JSON.

Of course this isn't necessarily indicative of actual performance, because of network I/O over my slowish internet. For example, curling that URI from my computer only takes 0.7 seconds, not 3. But the order of magnitude change on this branch worries me, because "normal" requests could start to take longer than the current 30-second worker timeout, causing failures in our deployments. So I'm going to see if there's any low-hanging fruit that can speed things up.

> we should also consider that things might (have to?) change a lot

Agreed, these are all good ideas! Eventually it's probably worth just proof-of-concept-ing out a few of these ideas and seeing where they break and where they perform better.

I'd even be willing to call this PR a proof-of-concept still. If I can't get performance to be stable, we might have to abandon this idea entirely, or at least not merge it until we can explore additional ideas. That's always disappointing but it's much better than never having tried!

austin3dickey:
I sped up the `get_baseline_run()` query in the latest commit. There were two main issues:

1. We were searching for a baseline run on ALL commits in the ancestry of the given run's commit, and that just takes a while in a database as big as Arrow's. I was able to get a huge speedup by only searching the last 20 commits of the ancestry during the commit subquery. I think that number is a good balance between
    - performance (lower the better)
    - chance of a false negative (higher the better) -- from my experience, it would be rare for a certain case/context/hardware to go more than 20 commits without reporting any results, and if it does, the impact on the user (can't find a baseline run) is low
2. Previously I was doing `Session.query(Run)`. Because of this, SQLAlchemy was adding 3 more denormalizing left joins to the query because it thought I needed more columns than I did. Now I'm doing `Session.query(BenchmarkResult.run_id)` and then (after getting only 1 run_id) doing `Run.get(<the run_id>)`. This skips those left joins in the big query and speeds everything up.

Now, my local test script is running in 1.7 seconds, which is great because the `main` branch's performance on the same test script is at 2.8 seconds. So now that method is sped up in this PR. This could help address #549 .

austin3dickey:
Okay, I also performance-tested `set_z_scores()` with all the ways to hit `/api/benchmarks/?X=Y`, and realized querying by `name` will take too long no matter what, so I stopped setting z-scores in that case (which is how we had it before this PR anyway). I also started caching baseline run in `set_z_scores()` to speed up the other ways to hit that endpoint. All other performance tests I tried were successful.

I'm all done working on this for now. I don't want to merge before the holidays, so I'll leave this open until the first week of January. Please feel free to take another look in the meantime!

jgehrcke:
> I'm all done working on this for now.

Haha I can imagine this feels like a rabbit hole :rabbit2: :rabbit:. You're really doing great work here, the performance analysis seems to have yielded super productive insights! I'd be happy to land this sooner than later but I totally respect that you want to leave this for January.

conbench/entities/commit.py
```diff
@@ -39,6 +45,100 @@ class Commit(Base, EntityMixin):
...
+        Ordering by commit_order.desc(), the following commits would return the
+        following ordered ancestors:
+
+        A  :  A
+        B  :  B, A
+        C  :  C, B, A
+        D  :  D, B, A
+        E  :  E, C, B, A
+        F  :  F, D, B, A
+        C2 :  C2, F, B, D, A
```
jonkeane: typo?

```suggestion
        C2 :  C2, F, D, B, A
```

conbench/entities/distribution.py
```diff
@@ -108,109 +115,122 @@ def get_distribution(benchmark_result, limit):
...
+    # Cache Distribution query results by commit/hardware, in case multiple input
+    # benchmark_results (e.g. from the same Run) share those attributes
+    cached_distributions = {}
+    # Also cache respective baseline runs by each Run's ID. Note that two
+    # benchmark_results from the same Run might not have the exact same baseline run,
+    # because one baseline run might be missing the case/context of the other
+    # benchmark_result. In those cases, we might fail to populate the z-score of the
+    # second benchmark result. But that's too rare and small-impact to justify not
+    # caching for performance reasons.
+    cached_baseline_runs = {}
```
jonkeane, line 182: Mostly confirming my understanding of the code below (but I think it critically touches on this comment, so I'm commenting here): 

For each run (well technically it's a list of benchmarks, but IIUC, those are a result of getting the list of benchmarks form the run, so effectively a run, yeah?), we cache the first baseline run that matches and then use that as the baseline for all subsequent comparisons. Is that right? 

For now I think that's totally fine (and it mimics what our current behavior is), but we might in the future want to loosen this up so that each benchmark finds it's own baseline rather than relying on going through runs. Doing this would unlock the ability to do runs that only have subsets of benchmarks in them (e.g. only benchmark the python code in arrow when we update pyarrow or cpp files) without needing to make sure that whenever a run has one benchmark of a type it has all of them. But we can (and should!) defer that for the future — just want to flag that as a thing we will want to loosen up in the future

conbench/entities/distribution.py
```diff
@@ -108,109 +115,122 @@ def get_distribution(benchmark_result, limit):
...
-        Distribution.mean_sd,
-    ]
+        if not baseline_run:
+            log.debug("No baseline run; setting benchmark_result.z_score = None")
+            continue
 
-    distributions = Session.query(*cols).filter(*where).all()
-    lookup = {f"{d.case_id}-{d.context_id}": d for d in distributions}
+        commit = baseline_run.commit_id
+        hardware = baseline_run.hardware.hash
```
jonkeane: This might be only an issue while reviewing this code (given the large scope of changes), but I wonder if we should name these `baseline_commit` and `baseline_hardware` to make sure it's clear that that's what they are later on.

When reviewing this, at first, I was worried that the line below in 211 was actually using the distribution starting from the contender and not the baseline and was super confused until I came back to find this

conbench/entities/distribution.py
```diff
@@ -108,109 +115,122 @@ def get_distribution(benchmark_result, limit):
...
+            for distribution in distributions
+            if distribution.case_id == benchmark_result.case_id
+            and distribution.context_id == benchmark_result.context_id
+        ]
+
+        if (
+            benchmark_result.mean is not None
+            and matching_distributions
+            and matching_distributions[0].mean_mean is not None
+            and matching_distributions[0].mean_sd  # is positive
```
jonkeane: This is a statement about the sd, right? And not a test of positivity?

```suggestion
            and matching_distributions[0].mean_sd  # is always positive
```

conbench/entities/history.py
jonkeane, line 44: Love these doc strings! We don't need it now, but in a follow on we might extend this to _also_ include errored BenchmarkResults and display them with a 💣 or something in the plot itself

conbench/entities/run.py
```diff
@@ -101,57 +103,93 @@ def create(data):
         run.save()
         return run
 
-    def get_baseline_run(self):
+    def get_baseline_run(
+        self,
+        commit_limit: int = 20,
```
jonkeane, line 108: We can leave this as is for now, but a few ideas that might make this restriction not so bad (none of these are mutually exclusive, so we might end up doing all of them):

* Could we have the code below iterate through in chunks of 20 until it finds history? This way the bast majority of cases (where within 20 commits, there's _something_ that could be a baseline) is fast, but if that's not found we degrade gracefully (e.g. find something even if it takes 5x as long going back 100 commits) 

* Have you profiled to see what makes this query take so long? Maybe it's possible to pass a subset of the data to the ancestor query (only the benchmarks we're interested in) that would make walking back like this faster

* Should we make this tuneable (like we did for the history distribution look back window)? In case there are repos where large gaps happen + they don't mind the slow performance?

conbench/entities/run.py
```diff
@@ -101,57 +103,93 @@ def create(data):
...
+        - has a BenchmarkResult with the given case_id/context_id, if those are given
+        - if they aren't given, has a BenchmarkResult with the case_id/context_id of
+          *any* of this Run's BenchmarkResults
+
+        Returns None if there are no matches. This could be a false negative if
+        ``commit_limit`` is too low (though note that the query takes longer with a
+        higher ``commit_limit``).
+
+        If there are multiple matches, prefer a Run with the same reason as this Run,
+        and then find the latest commit, finally tiebreaking by latest Run.timestamp.
```
jonkeane, line 127: Could you expand a bit more about the reasoning behind matching run reasons? I suspect you're right to do this, but I'm curious what you're imagining might happen here that this would be disambiguating?

jonkeane (approved):
This looks really really great. Like others have said above — this is a huge amount of work + effort.

None of my comments are blocking (and frankly most of them are more about future extensions and refinements we might make...) so I'm happy to merge this whenever you're comfortable to.

One more thing that might be interesting to show off is how this lets us have more flexibility. Do we have any examples in our current conbench(es) that show off the new flexibility this provides us? Maybe we need to wait until we deal with step changes?

conbench/entities/commit.py
austin3dickey: oh wow thank you!!

conbench/entities/distribution.py
austin3dickey: Yes, this is correct. One clarification to your "IIUC" parenthetical: a person or the webapp could hit `/api/benchmarks` for the last 500 benchmarks regardless of run, or `/api/benchmarks?name=X` which also disregards run, or `/api/benchmarks?batch_id=X` which could theoretically return benchmarks from multiple runs (since there's currently no technical restriction that one batch belongs to one run). That's why we're handling potentially-multiple runs in this function.

I agree that loosening this would be nice, so that each benchmark finds its own baseline. That's actually how I had originally written this function before I changed it in my latest commit. The problem was it was taking a really long time to `get_baseline_run()` for every single benchmark. There's probably a smart middle ground where we do less-aggressive caching.

conbench/entities/distribution.py
austin3dickey: sure!

conbench/entities/distribution.py
austin3dickey: What I meant was we need to make sure both that it's not None and not 0, and I was using python truthiness to make both of those checks at once. Since you asked the question, it's probably too magic :) I can change it to be more explicit

conbench/entities/history.py
austin3dickey: Oh yeah that could be fun!

conbench/entities/distribution.py
austin3dickey: Made #567

conbench/entities/history.py
austin3dickey: Made #568

conbench/entities/run.py
austin3dickey: Good ideas! Some thoughts on them:
- Re: iteration, I suspect it's more common for a run to be the "true" first run in history (e.g. the first with a certain hardware or new set of case/contexts) than for it to have a baseline that's more than 20 commits away. But I don't have the database stats to back this up. 🙂 either situation is probably pretty rare! But yeah maybe we could exponentially iterate at most 3 times so that "true" first runs don't get stuck. We also mustn't take more than 30 seconds or we'll return an Internal Server Error.
- Yes, I profiled the python code before I introduced the `commit_limit`, and 99%+ of the time is spent waiting for the database to execute the query. So it's definitely a query optimization problem. I used EXPLAIN ANALYZE to look at the different parts of the execution, and while I'm fairly inexperienced at doing that, I could tell that most of the time was spent evaluating whether runs on very old commits matched the criteria or not. That's why `commit_limit` worked so well.
- Yep tunability is a great idea.

austin3dickey: Made #569 and #570.

conbench/entities/run.py
austin3dickey: This was inspired by the `conbench/tests/api/test_runs.py:: test_closest_commit_different_machines_should_omit_test_runs` test. In that test, we set up this run history (newer runs first). Each run is on a different commit:
```
run ID | machine | run reason
-------+---------+-----------
4      | m1      | commit
3      | m2      | commit
2      | m1      | testing
1      | m1      | commit
```

And that test has always ensured that the baseline of run 4 is run 1, not runs 2 or 3. I know the intention of the test was to exclude "testing" run reasons, but it also told me that we care more about preferring the same run reason than preferring a newer commit. 

This also means that if you select a "nightly" run, you'll get a "nightly" run as the baseline, even if there's a "commit" run on a commit in between. I thought that was nice, clean behavior. If you think it will lead to more confusion though, we could change it.

jonkeane: > This also means that if you select a "nightly" run, you'll get a "nightly" run as the baseline, even if there's a "commit" run on a commit in between. I thought that was nice, clean behavior. If you think it will lead to more confusion though, we could change it.

I think this is totally fine to do for now (and I believe there are actually even no cases where this would happen right now) — but I do suspect if the future we actually will want to open this up a bit and allow for nightlies that have the same benchmarks as other types of runs (like a commit run) to be allowed to be considered as part of the history. There are some cases where this is more important than others (e.g. having a nightly run that has more robust history data because it also includes commit-to-default branch is one good example where having more flexibility is good. On the other hand, looking at per-commit benchmarks to detect a regression would be less useful to allow interspersed different run reasons).

I don't think we should change this PR or adjust this now, but we might (need to) do so in the future

austin3dickey: >to be allowed to be considered as part of the history

To clarify, this "prefer run reason" behavior only affects what happens when you're finding a baseline run (like clicking on the "compare ->" arrow in the webapp, or hitting the compare API). We still include ALL run reasons in the blue line in the timeseries plots (aka the history api), and in the baseline distribution when calculating z-scores.

austin3dickey:
>Do we have any examples in our current conbench(es) that show off the new flexibility this provides us? 

In my head here are the main benefits of the PR:
1. When benchmarks don't run on a commit (e.g. CI failure?), subsequent runs will be able to "skip over" that commit and find the previous baseline run that actually ran.
    - corollary: PR benchmarks branching off of a commit that didn't get a run will still be able to find a baseline.
    - corollary: The issue we've seen where the "compare ->" button is (unintentionally) missing should be fixed now.
2. Baseline runs will be correctly found for rebased branches, since we no longer rely on commit timestamp.
3. Run name no longer needs to follow the weird brittle "commit:" regex on the default branch.
4. And, like you mentioned, this PR sets us up for future work, which will allow users to find different baseline runs if they want (like the fork point or previous commit or head of the default branch), and to better manage "modes" of distributions so we don't evaluate a benchmark against stale historic data.

I don't have links to examples of these, but future work should result in nice links. :)

Issue #496, opened 2022-12-05
rethink: magic of setting CI=true and _expected_docs


jgehrcke:
I just ran into a case where a test failed in CI which did not fail locally:

```
=========================== short test summary info ============================
FAILED conbench/tests/api/test_docs.py::TestDocs::test_docs - RuntimeError: The 'CI' env var was set so we're assuming this test was run in CI. However, /app/conbench/tests/api/_expected_docs.py was not updated with the latest docs changes. Run `pytest conbench/tests/api/test_docs.py` locally to fix the file automatically.
============= 1 failed, 383 passed, 2 skipped in 209.92s (0:03:29) =============
```

With developer productivity in mind let's see if we can synchronize this again. Maybe this was even a regression introduced by moving to the docker-compose-based method to invoke tests locally.

austin3dickey:
That's right, I forgot about this. I typically run my local tests outside of docker-compose, so this hasn't been a problem for me. But because of #437, if you run local tests inside of docker-compose, the file doesn't get changed locally and so it can't be pushed.

With the current setup, the property that's most important to me is that PR reviewers can see the diff of how the OpenAPI docs might change along with a given code change. I don't really care how we get to that property, and agree that the current way has friction. 

I am intrigued by @jonkeane's [recent idea](https://github.com/conbench/conbench/pull/468#discussion_r1036258970) that we could trigger a bot to automatically update the PR with the expected doc changes. I like this opt-in method a little better than having the bot do it without being asked, because there's always a chance that a developer could cause a doc change that was unintended, and we wouldn't want that to slip through.

Other options are to keep the current setup and mount the repo to the local docker-compose container so that local changes are picked up, or get rid of the `_expected_docs.py` file altogether and have a bot otherwise attach the doc diff to the PR (maybe a comment or a GitHub Check or a build artifact similar to the webapp screenshots we're discussing in #483), or something else.

jgehrcke:
Ignoring bot magic for now, trying to get a better understanding of this actual challenge. I am rather confident that I have seen this type of challenge before (still need to confirm).

The most important principle I think is to make the local test suite have a signal that is strong: if it passes, it's very likely that CI passes. 

In that spirit, I think it's the right solution to make this test error out if `_expected_docs.py` does not match the output given by `/api/docs.json`. That should error out in any test environment (locally and in CI).

However, if it _does_ error out the error message should give a really useful hint for a developer for how to rebuild `_expected_docs.py`.

[...]

PR #509, opened 2022-12-06
tests_docs: do not look at CI var

jgehrcke:
This is acting on my thinking in #496 -- would appreciate your review @austin3dickey -- thanks!

### DIFF
--- a/conbench/tests/api/test_docs.py
+++ b/conbench/tests/api/test_docs.py
@@ -1,10 +1,12 @@
 import ast
+import logging
 import os
-import subprocess
 
 from openapi_spec_validator import validate_spec
 
 from ...tests.api import _asserts
+
+log = logging.getLogger(__name__)
 
 this_dir = os.path.abspath(os.path.dirname(__file__))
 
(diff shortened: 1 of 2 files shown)


conbench/tests/api/test_docs.py
```diff
@@ -1,38 +1,43 @@
...
+            # Hoping that this shows a useful diff.
+            log.info("caught assertion error: %s", exc)
+            raise Exception(
+                "/api/docs.json and _expected_docs.py are out of sync. "
+                "Run `make rebuild-expected-api-docs` to regenerate "
+                "_expected_docs.py and then review the diff manually."
+            )
+
+        # Maybe this is where we need a Python object structure, and not just
+        # a JSON document.
```
austin3dickey, line 42: I believe `validate_spec()` is intended as a cheap way to ensure the docs will render correctly using swagger. I'm not sure if ser/de to JSON will break anything here?

conbench/tests/api/test_docs.py
```diff
@@ -1,38 +1,43 @@
...
             expected_docs = ast.literal_eval(f.read())
 
         try:
+            # Note(JP): it seems that the hope here is that `assert_200_ok`
+            # performs a deep object equality inspection, relying on the
+            # `assert r.json == expected` equality check in the implementaion
+            # of `assert_200_ok()`. But how deep is that, really? Typically,
+            # the safest way to make sure that all details are after all equal
+            # is to perform stable string serialization. What would we lose if
+            # we were to compare JSON documents here?
```
austin3dickey, line 30: All I know is that `assert_200_ok()` is used in dozens of our tests, and never has had a false positive or negative that I'm aware of. So I'm fine with it. But you're welcome to try to replace its internals with a JSON serializer!

austin3dickey (approved):
looks great!

conbench/tests/api/test_docs.py
jgehrcke: It makes complete sense that this function wants to have a Python object structure as input. But that could simply be `validate_spec(resp.json)`.

conbench/tests/api/test_docs.py
jgehrcke: :)
The false negatives are hard to see of course.

I mean, Python has rather deep deep deep comparison when basic data types are used (lists, tuples, dicts), and therefore it is likely that the outest equality operator after all means "compare _all_ the details". But there is no language guarantee.

jgehrcke:
Thanks for the review, @austin3dickey.

For the record, I am wondering what we'd lose if were to do:

```python
resp = client.get("/api/docs.json")
assert resp.data== open("expected_docs.json"), "rb").read()
validate_spec(resp.json)
```

That confirms that `docs.json` represents a valid OpenAPI spec, and that directly confirms that the JSON byte sequence in the response to matches the expected byte sequence byte-by-byte.

The JSON doc should for this be in stable sort order and ideally be pretty-printed to make diffs readable. But I think that's already the case.

I am not really sure what the complexity in `ast.literal_eval(f.read())` and `black` is good for in this context, but I trust that at some point there was value or maybe there _is_, and I don't see it.

Moving on from here, for now :)

PR #517, opened 2022-12-07
makefile: remove CI=true

jgehrcke:
See #496.

This removes `-e CI=true` and also changes indentation.


jgehrcke:
I have done `± grep -nR 'env' | grep CI` and similar to try to make sure that no part of the codebase has a look at an environment variable called `CI` anymore.

austin3dickey (approved):

PR #523, opened 2022-12-08
access control: app: use authorize_or_terminate decorator

jgehrcke:
This is a followup from my previous patch where I left this comment: https://github.com/conbench/conbench/blob/285dc13d459c9ea5ac1a37612236669dc2320a1a/conbench/app/_endpoint.py#L63


jgehrcke:
Merging, next patch that builds on top of this is in the pipeline.

PR #524, opened 2022-12-08
makefile, readme: small tweaks/fixes

jgehrcke:
Thanks @austin3dickey for feedback!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ tests:
 lint:
 	flake8
 	isort .
-	black --diff .
+	black .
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -23,7 +23,7 @@ tests:
 lint:
 	flake8
 	isort .
-	black --diff .
+	black .
```
jgehrcke, line 26: note to self: that's something we may want to test in CI, too (that `make lint` indeed performs expected modifications).

Issue #528, opened 2022-12-09
Any duration measurement: use monotonic time source: time.monotonic(), time.perf_counter()


jgehrcke:
```
± grep -nR 'time.time()'
benchrun/python/benchrun/_benchmark.py:144:            start_time = time.time()
benchrun/python/benchrun/_benchmark.py:146:            end_time = time.time()
conbench/runner.py:362:            iteration_start = time.time()
conbench/runner.py:364:            times.append(time.time() - iteration_start)
conbench/util.py:45:            start = time.time()
conbench/util.py:47:            print("Time to POST", url, time.time() - start)
```

Consider this type of code:
```python
import time

t0 = time.time()
do_something()
duration = time.time() - t0
```
This is asking the operating system twice "hey, which wall time do you think we have right now?".

`time.time()` is constantly affected by system time adjustments. These are typically smeared out across longer time periods (and therefore they _should be_ insignificant) but sometimes there might be singularities. In any case, `duration` can easily become affected by how exactly system time updates over the course of time. The impact may be larger in home-grown environments than in a cloud provider's environment of course.

Tiny bit of background on how ntpd operates: https://docs.ntpsec.org/latest/ntpd.html

> The maximum slew rate possible is limited to 500 parts-per-million (PPM) as a consequence of the correctness principles on which the NTP protocol and algorithm design are based. As a result, the local clock can take a long time to converge to an acceptable offset, about 2,000 s for each second the clock is outside the acceptable range. During this interval, the local clock will not be consistent with any other network clock and the system cannot be used for distributed applications that require correctly synchronized network time.

A much better tool for the task at hand (measuring a duration) are monotonic time sources that the CPU provides. High resolution, and they are meant for measuring time differences. A monotonic time source cannot tell tell what time it is (absolute value is meaningless) but the difference between two samples is meaningful.

:hourglass_flowing_sand: 

Better:
```python
import time

t0 = time.monotonic()
do_something()
duration = time.monotonic() - t0
```

Or even https://docs.python.org/3/library/time.html#time.perf_counter -- depending on the order of magnitude of the expected duration.

I came across this really important distinction when I followed PEPs in detail like 10 years ago. I did read https://peps.python.org/pep-0418/ with great interest. It's a great read. It also mentions `time.perf_counter(): benchmarking, most precise clock for short period`.

PR #529, opened 2022-12-09
duration measurements: use monotonic time source

jgehrcke:
See #528.

I used `grep` to find all those places in this repo where we take two `time.time()` samples for calculating a duration.


alistaire47 (approved):
Looks good, thanks!

austin3dickey (approved):
Nice catch!

PR #537, opened 2022-12-13
redact SQLALCHEMY_DATABASE_URI

austin3dickey:
This URI has a sensitive password inside it.


jgehrcke (approved):
:heart: Great catch, thanks for watching out for this!

jgehrcke:
follow-up thought: it would be nice to in the future log some parts of this URL; those that may be considered non-sensitive. Might help debugging things.

Issue #400, opened 2022-10-17
Add tag keys to the default display name


austin3dickey:
Currently, if the "name" tag isn't set, the display name of a BenchmarkResult is a [concatenation](https://github.com/conbench/conbench/blob/main/conbench/hacks.py#L22) of the values of most tags. It would be nice to include the keys of the tags in this concatenation as well.

jgehrcke:
How would I test this, where does the display name show?

https://github.com/conbench/conbench/blob/a797b8a271b7c27b25791e92be2b7eb97af3422a/conbench/hacks.py#L1 is a little too magical to rationalize the extent of this ticket based on code only. Let's maybe find an example where we demonstrate the current state and then show how the future state should look like.

What is the meaning of case here btw? Case as in 'scenario'?

austin3dickey:
Sorry for the confusing language in the code; as usual, these naming conventions are under-documented. In this ticket, a "case" corresponds to one BenchmarkResult in the database.

Here's a specific example. We have a [workflow](https://github.com/conbench/benchalerts) to be run in CI, asking a Conbench server for any possible regressions on a given commit, and posting a summary of those regressions as a GitHub Check back to the GitHub commit. A commit might have multiple benchmark runs, and each run might have multiple "cases" (or BenchmarkResults). In the GitHub Check Markdown summary, I want to display those cases with nice display names.

[Here's what the markdown summary looks like today.](https://github.com/conbench/benchalerts/blob/5658826e2c17f72637945308f804d311d0b0b7b4/tests/unit_tests/expected_md/summary_regressions.md) Notice the "snappy, nyctaxi_sample, csv, arrow" strings (which were taken directly from the GET /api/compare/runs/ endpoint). Those would be more informative if they looked like "compression=snappy, dataset=nyctaxi_sample, ..." etc.

Another place these names display is on the compare page in the webapp. For example, https://conbench.ursa.dev/compare/runs/c86780aab02a471b9c8223854a47295c...2e4b0516b7fc4b02b5900bf097065c19/

PR #538, opened 2022-12-13
conbench cli: show result timing overview

jgehrcke:
I work on building a new benchmark. While doing so I want to be able to easily compare the timing information for various cases/scenarios run by a single `conbench` cli invocation. For example, when I do `conbench dataset-serialize --all true --iterations=6 nyctaxi_multi_parquet_s3` then when the entire process completes, I was looking for timing summary across all the cases/scenarios run. So that it's easier for me to actually iterate on/develop the benchmark method.

With the pragmatic code in this PR, I got for example the following output, which helped me make decisions:

```
$ conbench dataset-serialize --all true --iterations=6 nyctaxi_multi_parquet_s3

[...]

[221213-19:50:00.354] [283214] [conbench.cli] INFO: result timing overview:

format: parquet, selectivity: 1pc
    --> min: 0.469 s    mean±SE: (0.49 ± 0.01) s
format: arrow, selectivity: 1pc
    --> min: 0.0468 s   mean±SE: (0.0502 ± 0.0009) s
format: ipc, selectivity: 1pc
    --> min: 0.0479 s   mean±SE: (0.054 ± 0.002) s
format: feather, selectivity: 1pc
    --> min: 0.0581 s   mean±SE: (0.061 ± 0.001) s
format: csv, selectivity: 1pc
    --> min: 0.410 s    mean±SE: (0.435 ± 0.009) s
format: parquet, selectivity: 10pc
    --> min: 2.22 s     mean±SE: (2.26 ± 0.01) s
format: arrow, selectivity: 10pc
    --> min: 0.154 s    mean±SE: (0.162 ± 0.002) s
format: ipc, selectivity: 10pc
    --> min: 0.189 s    mean±SE: (0.22 ± 0.02) s
format: feather, selectivity: 10pc
    --> min: 0.171 s    mean±SE: (0.20 ± 0.01) s
format: csv, selectivity: 10pc
    --> min: 3.78 s     mean±SE: (3.94 ± 0.05) s
```

### DIFF
--- a/conbench/cli.py
+++ b/conbench/cli.py
@@ -2,14 +2,12 @@ import json
 import logging
 import math
 
-
 import click
 import sigfig
 
 from . import __version__
 from .runner import LIST, REGISTRY
 from .util import register_benchmarks
-
 
 log = logging.getLogger(__name__)
 
--- a/conbench/cli.py
+++ b/conbench/cli.py
@@ -1,10 +1,17 @@
 import json
+import logging
+import math
+
 
 import click
+import sigfig
 
 from . import __version__
 from .runner import LIST, REGISTRY
 from .util import register_benchmarks
+
+
+log = logging.getLogger(__name__)
 
 register_benchmarks()
 BENCHMARKS = {}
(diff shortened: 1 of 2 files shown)


jgehrcke:
I understand that we might phase out this CLI soon, and it's OK to never merge this PR. If this is however useful for others for now, maybe it makes sense to pull it in. Happy to receive review/questions :).

requirements-cli.txt
```diff
@@ -5,3 +5,4 @@ py-cpuinfo
 pytest>=7.0.0
 PyYAML
 requests
+sigfig
```
austin3dickey, line 8: Adding a new dependency to the CLI will require some changes to https://github.com/conda-forge/conbench-feedstock too, but I'm not sure of the exact process. @ElenaHenderson might know. This is an underdocumented aspect of the CLI right now.

Either way, I think bumping the version in setup.py is a good idea.

conbench/cli.py
```diff
@@ -81,6 +86,61 @@ def _to_cli_name(name):
...
+    timestrings = []
+
+    for r in results:
+        tags = r["tags"]
+
+        # That represents the case/senario. Note that `tags` has been enriched
+        # with key/value pairs beyond the case -- filter these out again.
+        tagstring = ", ".join(
+            f"{k}: {tags[k]}" for k in sorted(tags.keys()) if k not in common_tag_keys
+        )
```
austin3dickey, line 112: We have similar behavior [in hacks.py](https://github.com/conbench/conbench/blob/main/conbench/hacks.py#L1) (which itself has inspired issues like #400). Would you consider consolidating that code? I like this strategy of filtering out common tags instead of a hardcoded list of keys (since technically tags is a freeform dict) but I understand it will take some thought to apply that strategy to what exists in hacks.py today.

austin3dickey (commented):
This is nice! I'm assuming the test suite covers most of these branches?

IMO we can merge this now (pending some comments below) and think about how to put this code in the new runners later.

requirements-cli.txt
jgehrcke: Interesting. https://github.com/conda-forge/conbench-feedstock/pull/8 is a template maybe. Huh, all that is quite a bit of process! :)

conbench/cli.py
jgehrcke: > We have similar behavior [in hacks.py](https://github.com/conbench/conbench/blob/main/conbench/hacks.py#L1) (which itself has inspired issues like https://github.com/conbench/conbench/issues/400).

lol, now I understand what this was/is about :-) and thank you for the thorough response on #400 -- I had missed this!

jgehrcke: >  I like this strategy of filtering out common tags instead of a hardcoded list of keys

Cool, thanks for this thought. Then I suppose I now 'know' (roughly) how to consolidate these two.... :shrug: 

I will see about that.

conbench/cli.py
```diff
@@ -81,6 +86,61 @@ def _to_cli_name(name):
...
+
+        # Assume that stdev=0 for now means there is only one sample.
+        if r["stats"]["stdev"] == 0:
```
alistaire47, line 115: This will throw a keyerror if a case failed because `stats` will not be populated (though `error` will). Well, mostly; I know we were working towards accepting both keys for cases where some iterations failed and some didn't, though I forget where we are in that change—I think Jon merged a server change, but I'm not sure if this CLI can actually handle that situation yet. 

In any case, we should probably write this to handle either of `stats` or `error` being missing without erroring out.

conbench/cli.py
```diff
@@ -81,6 +86,61 @@ def _to_cli_name(name):
...
+            timestring = f"{sigfig.round(r['stats']['data'][0], 2)} s"
+        else:
+            # Assume that there are at least three samples. Standard deviation
+            # is known, calc standard error of the mean:
+            mean = r["stats"]["mean"]
+            stdev = r["stats"]["stdev"]
+            min = r["stats"]["min"]
```
alistaire47, line 122: There was a fairly recent change so that summary stats like these don't need to be supplied by the user, and the server will do so from `data` before inserting the database record. (This is nice because it ensures consistency in calculation regardless of what is actually running benchmarks.) I think this runner is still calculating them client-side so this will work, but if we port this elsewhere, we can't depend on these existing.

alistaire47 (commented):
This is nice, thank you! A couple comments about generalization we'll need to address to port this to other runners.

jgehrcke:
Closing this for cleanup -- the ideas in here are valuable, and we can always go back to this.

PR #540, opened 2022-12-15
Migration script to clean commits

austin3dickey:
This migration script will run one time per Conbench server. It will clean up the commits in some servers (including Arrow) whose GitHub PAT recently expired, leading to commits without detailed information. It will also ensure that the branch corresponds to the fork_point_sha when possible.

### DIFF
--- a/migrations/versions/4d8d67396f79_clean_commits.py
+++ b/migrations/versions/4d8d67396f79_clean_commits.py
@@ -1,0 +1,124 @@
+"""clean commits
+
+We'd like to make the assumption that a Commit's sha == its fork_point_sha if and only
+if it's on the default branch. Due to a few bugs, this wasn't always the case. We have
+fixed those bugs, and now would like to clean up the historic commits so they follow
+this rule.
+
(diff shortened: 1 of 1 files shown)


austin3dickey:
There are still dozens of open PRs in the Arrow and Velox repos whose branches fork from commits from before we started reporting PR numbers to Conbench correctly. If those authors don't rebase, but they add new commits, and run PR benchmarks, there's still the possibility of commit branches being reported incorrectly.

I think that's okay. I'll go ahead and merge this now. If we need to clean the databases again in the future, I'm sure we can figure out a way to run this specific migration again on those databases.

That does mean that if server code ever needs to query "commits on the default branch," it's probably generally a better pattern to filter to `Commit.sha == Commit.fork_point_sha` than `Commit.branch == default_branch`. We can trust `Commit.sha == Commit.fork_point_sha` to be a more accurate indicator because it comes straight from GitHub, vs the branch, which is reported by the benchmark runner (and we've seen that can make mistakes).

Issue #531, opened 2022-12-13
populating local conbench with data


sayantikabanik:
The current docker workflow worked flawlessly, was able to run the app.
While trying to update the db with data, running into below error:

Script I am using for data population - `conbench/tests/populate_local_conbench.py`

```shell
requests.exceptions.ConnectionError: HTTPConnectionPool(host='127.0.0.1', port=5000): Max retries exceeded with url: /api/register/ (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x1012007f0>: Failed to establish a new connection: [Errno 61] Connection refused'))
``` 

A thought: could we integrate the above under makefile?

jgehrcke:
Oh hey! Yeah that's a great suggestion that this flow should be supported by a Make target, and that this should be documented in the README.

So, you did `make run-app`, and then were you able to open `http://127.0.0.1:5000/` in the browser, seeing the Conbench UI? If that's the case then `python -m conbench.tests.populate_local_conbench` executed on your machine should also be able to reach the API via `(host='127.0.0.1', port=5000)`.

sayantikabanik:
I tried again, previously my daemon unexpectedly stopped (now rectified)
this time ran into 02 different issues

```shell
requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
```

After adding a header was able to connect to the API
`session.headers.update({'User-Agent': 'custom_agent'})`

```shell
{"code": 400, "description": {"email": ["Email address already in use."]}, "name": "Bad Request"}
```
Flagged as exception.

Happy to raise a PR, if this sounds like a good fix/probable issue

jgehrcke:
Hey!

> `requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))`

Why do we think that this (mainly) was triggered by a bad user agent header? In CI and other environments this works without `session.headers.update({'User-Agent': 'custom_agent'})`.

> `{"code": 400, "description": {"email": ["Email address already in use."]}, "name": "Bad Request"}`

Sounds like this method was run against a database that was already partly populated.

We should set some goalds for `populate_local_conbench.py` -- should this exit indicating when executed repeatedly or should it exit non-zero when it sees that it's _not_ running against a fresh/empty database state?

I'd vote for the latter I think.

Maybe we need another helper that says "wipe DB state". Then the user can do "wipe-db, populate-db" in succession.

[...]

PR #544, opened 2022-12-19
devprod: add /api/wipe-db endpoint

jgehrcke:
This is motivated by https://github.com/conbench/conbench/issues/531. During local development it may be convenient to have an API endpoint that one can hit with the browser and/or curl. Tested this locally. Also added a test case.

README update follows.

CC @sayantikabanik

### DIFF
--- a/conbench/api/index.py
+++ b/conbench/api/index.py
@@ -18,6 +18,13 @@ log = logging.getLogger(__name__)
 
 @api.route("/docs.json")
 def docs():
+    d = spec.to_dict()
+
+    # In TESTING mode there is a special endpoint that we do not want to mess
+    # with test_docs.
+    if "/api/wipe-db" in d["paths"]:
+        del d["/api/wipe-db"]
+
     return f.jsonify(spec.to_dict())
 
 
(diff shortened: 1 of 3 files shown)


jgehrcke:
Note that `empty_db_tables()` is tested implicitly between all tests, i.e. I only built a pragmatic test that confirms existence of the endpoint, not a test that confirms that this indeed wipes DB state.

jgehrcke:
This is part of a series of smaller devprod PRs. Merging.

conbench/api/index.py
```diff
@@ -9,11 +10,21 @@
...
 @api.route("/docs.json")
 def docs():
+    d = spec.to_dict()
+
+    # In TESTING mode there is a special endpoint that we do not want to mess
+    # with test_docs.
+    if "/api/wipe-db" in d["paths"]:
+        del d["paths"]["/api/wipe-db"]
+
     return f.jsonify(spec.to_dict())
```
austin3dickey, line 28: You wanted to `return f.jsonify(d)`, no?

jgehrcke: Yes. Thank you for your pair of eyes!!

Question is: why did the tests pass nevertheless? :D Looking now.

PR #546, opened 2022-12-19
devprod: mount local checkout into container

jgehrcke:
This introduces a new Makefile target which invokes the docker-compose machinery in a special way, using an extension config on top of the normal config.

That extension config has instructions to mount the local checkout dir (the root of the Conbench repo) into the Conbench container image, at `/app`.

This patch also overrides the gunicorn command/settings.

I reduced the number of workers from 5 to 2 (not just 1, so that during local dev one can also see some concurrency issues that would otherwise happen in prod).

Found one concurrency issue: `db.create_all()` was not guarded against concurrent execution. This did not happen so far because gunicorn was started with `--preload`. I removed `--preload` to make the reload-upon-file-change more complete (init code changes would otherwise _not_ lead to auto-reload).

Why not always mount local checkout into /app when doing `make app-run`? Maybe that would have been easier and also possible. But then, `make app-run` is not really testing the contents of the container image defined by Dockerfile
(which is the image that's being used in prod).  I feel like mounting the local checkout into the container image is special enough of a use case to get its own Makefile target. So far I thought of `make run-app` to be mainly for users who 'just' want to try out the app locally. 

Like `make app-run`, this target is not to be tested in CI trivially because it launches a long-running process. Thinking about adding a commit to test this in CI, but for today I think merging this w/o test is good enough. Testing locally that this works.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -75,5 +75,5 @@ jobs:
             export CONBENCH_BASE_URL=http://$(docker compose port app 5000) && \
             make db-populate && make db-populate
           make teardown-app
-
+          make test-run-app-dev
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,11 @@ COPY requirements-test.txt /tmp/
 RUN pip install -r /tmp/requirements-build.txt
 RUN pip install -r /tmp/requirements-test.txt
 
+# This Dockerfile currently defines the image used for production environments.
+# It also contains all test dependencies and most CI dependencies because it's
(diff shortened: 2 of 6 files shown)


jgehrcke:
> Like make app-run, this target is not to be tested in CI trivially because it launches a long-running process. Thinking about adding a commit to test this in CI, but for today I think merging this w/o test is good enough. Testing locally that this works.

update: I have just added a commit which should cover most of this in CI.

jgehrcke:
The CI log output isn't great to reveal what happens, but indeed CI exercised `make test-run-app-dev`, and things went green. Merging!

PR #547, opened 2022-12-19
makefile: add teardown-app target

jgehrcke:
Rename, to make this name more expressive and correct.

Will be part of the 'public interface', i.e. should have mention in README.


jgehrcke:
documentation for this: https://github.com/conbench/conbench/pull/548

Issue #556, opened 2022-12-21
Finding all commits to the master branch of apache/arrow: may take longer than WORKER TIMEOUT


jgehrcke:
```
conbench-app-1 | 127.0.0.1 - - [21/Dec/2022:13:25:10 +0000] "GET /api/ping/ HTTP/1.1" 200 110 "-" "curl/7.74.0"
conbench-app-1 | [2022-12-21 13:25:13 +0000] [1] [CRITICAL] WORKER TIMEOUT (pid:12)
conbench-app-1 | Finding all commits to the master branch of apache/arrow between 1970-01-01T00:00:00Z and 2022-03-03T14:16:51Z
conbench-app-1 | [2022-12-21 13:25:13 +0000] [12] [INFO] Worker exiting (pid: 12)
```
Seen in CI in https://github.com/conbench/conbench/issues/555#issuecomment-1361312925.

jgehrcke:
Confirmed that in CI (yes, I have used it a lot today) this can fail with:
```
conbench/tests/entities/test_commit.py::test_get_github_commit_and_fork_point_sha[None] {'message': 'API rate limit exceeded for installation ID 19416693.', 'documentation_url': 'https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting'}
FAILED
conbench/tests/entities/test_commit.py::test_get_github_commit_and_fork_point_sha[apache:master] {'message': 'API rate limit exceeded for installation ID 19416693.', 'documentation_url': 'https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting'}
FAILED
conbench/tests/entities/test_commit.py::test_get_github_commit_and_fork_point_sha_pull_request[None-10665] {'message': 'API rate limit exceeded for installation ID 19416693.', 'documentation_url': 'https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting'}
FAILED
conbench/tests/entities/test_commit.py::test_get_github_commit_and_fork_point_sha_pull_request[dianaclarke:ARROW-13266-None] {'message': 'API rate limit exceeded for installation ID 19416693.', 'documentation_url': 'https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting'}
FAILED
[...]

PR #558, opened 2022-12-21
get_github_commit(): clarify interface, expect rate limiting errs in tests

jgehrcke:
Motivated by https://github.com/conbench/conbench/issues/556 where CI suffered from rate limiting errors during GitHub HTTP API usage.

At the high level this patch

- changes and clarifies the interface of `get_github_commit()`
- adjusts error handling code in `run.py` to use that interface
- adjusts to expect rate limiting errs, and to not fail in this case

Update: analogue treatment for `backfill_default_branch_commits()`: expect errors.

If we decide that this approach is too relaxed for _testing_ I think the clarification of interfaces and the slightly improved error handling is still valuable.

### DIFF
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -1,6 +1,6 @@
 import functools
-import logging
 import json
+import logging
 import os
 from datetime import datetime
 from typing import List
@@ -18,7 +18,6 @@ from ..entities._entity import (
     Nullable,
     generate_uuid,
 )
-
 
 log = logging.getLogger(__name__)
 
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -185,10 +185,6 @@ def get_github_commit(repository: str, pr_number: str, branch: str, sha: str) ->
     # GitHub HTTP API failed, e.g. with a 4xx rate limiting response.
     commit = github.get_commit(name, sha)
 
-    # Note(JP): I think with current err handling this cannot happen anymore.
-    if commit is None:
-        return {}
-
     if branch:
         commit["branch"] = branch
     elif pr_number:
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -1,4 +1,5 @@
 import functools
+import logging
 import json
 import os
 from datetime import datetime
@@ -17,6 +18,9 @@ from ..entities._entity import (
     Nullable,
     generate_uuid,
 )
+
+
+log = logging.getLogger(__name__)
 
 
 class Commit(Base, EntityMixin):
@@ -157,12 +161,32 @@ def repository_to_url(repository):
 
 
 def get_github_commit(repository: str, pr_number: str, branch: str, sha: str) -> dict:
+    """
+    This function interacts with the GitHub HTTP API. Exceptions related to API
+    interaction errors are not handled here (and should be expected and handled
+    in the caller). Expected error sources are among the following :
+
+    - transient issues on DNS or TCP level
+    - transient HTTP errors such as 5xx
+    - transient but permanent-ish HTTP errors such as rate limiting
+    - HTTP authentication/authorization errors
+    - permanent error responses
+
+    If this is free of bugs then all expected exceptions should derive from
+    requests.exceptions.RequestException.
+
+    """
     if not repository or not sha:
         return {}
 
     github = GitHub()
     name = repository_to_name(repository)
+
+    # `github.get_commit()` below may raise an exception if the GitHub
+    # GitHub HTTP API failed, e.g. with a 4xx rate limiting response.
     commit = github.get_commit(name, sha)
+
+    # Note(JP): I think with current err handling this cannot happen anymore.
     if commit is None:
         return {}
 
@@ -287,7 +311,12 @@ class GitHub:
             response = self.test_commit(sha)
         else:
             url = f"{GITHUB}/repos/{name}/commits/{sha}"
+            # _get_response() may raise an exception, for example if the GH
+            # HTTP API returned a non-2xx HTTP response (e.g. in case of rate
+            # limiting).
             response = self._get_response(url)
+
+        # Note(JP): I think now `response` can never be just `None`.
         return self._parse_commit(response) if response else None
 
     def get_commits_to_branch(
@@ -413,9 +442,27 @@ class GitHub:
             "author_avatar": author["avatar_url"] if author else None,
         }
 
-    def _get_response(self, url):
+    def _get_response(self, url) -> dict:
+        """
+        Return deserialized JSON-structure or raise an exception.
+        """
+        # This can raise exceptions corresponding to transient issues related
+        # to DNS, TCP, HTTP during sending request, while waiting for response,
+        # or while receiving the response.
         response = self.session.get(url) if self.session else requests.get(url)
+
+        # An HTTP response was received.
         if response.status_code != 200:
-            print(response.json())
-            return None
+            # This is here to log details an make debugging easier.
+            log.info(
+                "got unexpected HTTP response with code %s: %s",
+                response.status_code,
+                response.text,
+            )
+
+        # Raise an exception for all known HTTP error response types.
+        response.raise_for_status()
+
+        # This may raise an exception if JSON-deserialization fails. If JSON
+        # deser succeeds then this is known to be a dict at the outest level.
         return response.json()
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -306,17 +306,16 @@ class GitHub:
         return f"{org}:{branch}"
 
     def get_commit(self, name, sha):
+        # Pragmatic method for testing.
         if sha in self.test_commits:
-            response = self.test_commit(sha)
-        else:
-            url = f"{GITHUB}/repos/{name}/commits/{sha}"
-            # _get_response() may raise an exception, for example if the GH
-            # HTTP API returned a non-2xx HTTP response (e.g. in case of rate
-            # limiting).
-            response = self._get_response(url)
+            return self._parse_commit(self._mocked_get_response(sha))
 
-        # Note(JP): I think now `response` can never be just `None`.
-        return self._parse_commit(response) if response else None
+        url = f"{GITHUB}/repos/{name}/commits/{sha}"
+
+        # _get_response() may raise an exception, for example if the GH
+        # HTTP API returned a non-2xx HTTP response (e.g. in case of rate
+        # limiting).
+        return self._parse_commit(self._get_response(url))
 
     def get_commits_to_branch(
         self, name: str, branch: str, since: datetime, until: datetime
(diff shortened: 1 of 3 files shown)


conbench/entities/commit.py
```diff
@@ -413,9 +441,27 @@ def _parse_commit(commit):
...
         }
 
-    def _get_response(self, url):
+    def _get_response(self, url) -> dict:
+        """
+        Return deserialized JSON-structure or raise an exception.
+        """
+        # This can raise exceptions corresponding to transient issues related
+        # to DNS, TCP, HTTP during sending request, while waiting for response,
+        # or while receiving the response.
```
jgehrcke, line 463: timeout errors, of course (upon TCP `connect()`, between two bytes `recv()`d, or for complete request/response cycle.

jgehrcke:
CI failed with
```
E            'author_avatar': 'https://avatars.githubusercontent.com/u/878798?v=4',
E            'author_login': 'dianaclarke',
E            'author_name': 'Diana Clarke',
E         -  'branch': 'dianaclarke:ARROW-13266',
E         +  'branch': 'apache:master',
E            'date': datetime.datetime(2021, 7, 6, 21, 51, 48, tzinfo=tzutc()),
E            'fork_point_sha': '780e95c512d63bbea1e040af0eb44a0bf63c4d72',
E            'message': 'ARROW-13266: [JS] Improve benchmark names',
E            'parent': 'c8668f85a465ea05b2724ec47ff72c4db4d7dfe6',
E           }
```
Looking, then adding commit. Probably small oversight while copy/pasting code.

conbench/tests/entities/test_commit.py
```diff
@@ -135,9 +147,15 @@ def test_get_github_commit_and_fork_point_sha_pull_request(branch, pr_number):
         "branch": "dianaclarke:ARROW-13266",
         "fork_point_sha": "780e95c512d63bbea1e040af0eb44a0bf63c4d72",
     }
-    assert (
-        get_github_commit(repo, branch=branch, sha=sha, pr_number=pr_number) == expected
-    )
+    try:
+        result = get_github_commit(repo, branch=branch, sha=sha, pr_number=None)
```
jgehrcke: Yes, here: `pr_number=pr_number` turned to be `pr_number=None` :fire: -- Fixing this :fire_engine:

conbench/entities/commit.py
```diff
@@ -413,9 +441,27 @@ def _parse_commit(commit):
...
+                "got unexpected HTTP response with code %s: %s",
+                response.status_code,
+                response.text,
+            )
+
+        # Raise an exception for all known HTTP error response types.
+        response.raise_for_status()
+
+        # This may raise an exception if JSON-deserialization fails. If JSON
+        # deser succeeds then this is known to be a dict at the outest level.
```
jgehrcke, line 479: Might raise `ValueError` if JSON-deserialization fails, i.e. that particular exception is not from the `requests.exceptions` hierarchy. Have not tested though.

austin3dickey: Yeah, or in particular if the response has no content I think `.json()` will fail. But I don't think any of the endpoints we hit have that behavior.

conbench/entities/commit.py
```diff
@@ -157,12 +160,32 @@ def repository_to_url(repository):
...
         return {}
 
     github = GitHub()
     name = repository_to_name(repository)
+
+    # `github.get_commit()` below may raise an exception if the GitHub
+    # GitHub HTTP API failed, e.g. with a 4xx rate limiting response.
     commit = github.get_commit(name, sha)
+
+    # Note(JP): I think with current err handling this cannot happen anymore.
```
austin3dickey: Agreed! I think you can delete this line

conbench/tests/entities/test_commit.py
```diff
@@ -103,7 +107,15 @@ def test_get_github_commit_and_fork_point_sha(branch):
         # this is the master branch, so the fork point sha == the commit sha
         "fork_point_sha": sha,
     }
-    assert get_github_commit(repo, branch=branch, sha=sha, pr_number=None) == expected
+    try:
+        result = get_github_commit(repo, branch=branch, sha=sha, pr_number=None)
+    except Exception as exc:
+        log.info("exc during get_github_commit(): %s", exc)
+        if "API rate limit" in str(exc):
+            pytest.skip("GitHub API rate limit seen, skip test")
```
austin3dickey, line 126: Yep I really like this.

austin3dickey (approved):

conbench/entities/commit.py
jgehrcke: Yes, right.

To summarize: once we're here and JSON deserialization fails this means we got a 2xx response from the GitHub HTTP/JSON API _without_ JSON body in the response. Which _should not_ happen. But since it's not in control of this program here, it's part of the possible outcomes of this function, in the strictest sense.

conbench/entities/commit.py
jgehrcke: Cool. Thanks for double-checking. Added a commit!

conbench/entities/run.py
```diff
@@ -59,12 +63,24 @@ def create(data):
...
+            gh_commit_dict = None
+            try:
+                # get_github_commit() may raise all those exceptions that can
+                # happen during an HTTP request cycle.
+                gh_commit_dict = get_github_commit(
+                    repository=repository, pr_number=pr_number, branch=branch, sha=sha
+                )
+            except Exception as exc:
+                log.info(
+                    "treat as unknown commit: error during get_github_commit(): %s", exc
```
jgehrcke, line 79: Austin, is this "treat as unknown commit" meaningful/correct?

austin3dickey: Yes, that should be the case here!

jgehrcke:
Nice. Works already quite well, test suite revealed another test that now needs more exception handling. Useful log output during test suite:
```
------------------------------ Captured log call -------------------------------
INFO     conbench.entities.commit:commit.py:452 got unexpected HTTP response with code 403: {"message":"API rate limit exceeded for installation ID 19416693.","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

edit: yes, `backfill_default_branch_commits()` is now expected to raise HTTP-related exceptions as well. Adding commit(s).

conbench/tests/entities/test_commit.py
```diff
@@ -135,9 +159,16 @@ def test_get_github_commit_and_fork_point_sha_pull_request(branch, pr_number):
...
     }
-    assert (
-        get_github_commit(repo, branch=branch, sha=sha, pr_number=pr_number) == expected
-    )
+    try:
+        result = get_github_commit(repo, branch=branch, sha=sha, pr_number=pr_number)
+    except Exception as exc:
+        log.info("exc during get_github_commit(): %s", exc)
+        if "403 Client Error" in str(exc):
+            pytest.skip("GitHub API rate limit seen, skip test")
```
jgehrcke, line 167: Of course this bakes in the assumption that `"403 Client Error"` corresponds to the API rate limit error(s).

I am rather certain that the `exc` object holds a reference to the HTTP error response object in case `exc` derives from `request.exceptions.HTTPError`. That is, we could _also_ inspect the error response body here and do string-matching on `API rate limit`, if we ever feel like that's necessary.

jgehrcke:
`ci / testsuite` failed with a Coveralls-related API error:
```
  with:
    github-token: ***
    path-to-lcov: _conbench-coverage-dir/coverage.lcov
    coveralls-endpoint: https://coveralls.io
  env:
    GITHUB_API_TOKEN: ***
Using lcov file: _conbench-coverage-dir/coverage.lcov
Error: Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
```
Looks like a transient one, maybe also related to rate limiting: https://github.com/lemurheavy/coveralls-public/issues/750

Can ignore.

conbench/tests/entities/test_commit.py
```diff
@@ -18,6 +19,20 @@
 this_dir = os.path.abspath(os.path.dirname(__file__))
 
 
+log = logging.getLogger(__name__)
+
+
+def backfill_default_branch_commits_ign_rate_limit(*args, **kwargs):
```
jgehrcke, line 25: I should have added a comment here. This is supposed to be a super pragmatic wrapper around `backfill_default_branch_commits()` that behaves like the original _except_ for rate limiting errors.

conbench/entities/commit.py
```diff
@@ -223,7 +246,7 @@ def backfill_default_branch_commits(
     )
     commits_to_try = commits[1:-1]  # since/until are inclusive; we want exclusive
 
-    print(f"Backfilling {len(commits_to_try)} commit(s)")
+    log.info(f"Backfilling {len(commits_to_try)} commit(s)")
```
austin3dickey, line 249: nice, thanks for this

conbench/entities/commit.py
```diff
@@ -287,7 +310,12 @@ def get_commit(self, name, sha):
             response = self.test_commit(sha)
         else:
             url = f"{GITHUB}/repos/{name}/commits/{sha}"
+            # _get_response() may raise an exception, for example if the GH
+            # HTTP API returned a non-2xx HTTP response (e.g. in case of rate
+            # limiting).
             response = self._get_response(url)
+
+        # Note(JP): I think now `response` can never be just `None`.
         return self._parse_commit(response) if response else None
```
austin3dickey: I think `self._get_response(url)` can't be None now, but `self.test_commit(sha)` can be None due to this line:

https://github.com/conbench/conbench/blob/c029a3d9ca5b3099f87b3983b8f420e46f3d8dd8/conbench/entities/commit.py#L419-L420

Wonder if we should change that line to `raise requests.exceptions.RequestException` to more correctly mock the new behavior? Then this line can just be `return self._parse_commit(response)`.

austin3dickey (approved):
Looks good! One small thing.

conbench/entities/commit.py
jgehrcke: Yes, this needs more work. Thanks.

Trying to make sense of this.

The relationship between test_commit() and parse_commit() and get_commit() is a little mind-bending without code comments. :)

jgehrcke: > `raise requests.exceptions.RequestException`

Good idea to raise an exception, but it's IMO for now better to not raise an exception that otherwise only `requests` is supposed to raise. 

I have added a commit with a bit of commentary and also with
```
raise Exception("_mocked_get_response(): simulate _get_response() error")
```
which makes use of the fact that so far we're indeed using overgeneralized exception handlers.

conbench/entities/commit.py
```diff
@@ -415,25 +414,40 @@ def session(self):
             session.headers = {"Authorization": f"Bearer {token}"}
         return session
 
-    def test_commit(self, sha):
+    def _mocked_get_response(self, sha) -> dict:
```
jgehrcke, line 417: I did this quick rename from `test_commit()` to `_mocked_get_response()` to hopefully make the name a tiny bit better reflect the purpose of this method.

conbench/entities/commit.py
```diff
@@ -415,25 +414,40 @@ def session(self):
...
+        """
+        Note(JP): this function performed magic before and I am trying to write
+        a docstring now. Maybe: load commit information from disk, if
+        available. Otherwise, if the commit hash `sha` contains the magic words
+        'unknown' or 'testing' then pretend as if fetching these from the
+        GitHub HTTP API failed, and raise an exception simimar to _get_response
```
jgehrcke, line 423: simimar -> similar. well :) next time.

conbench/entities/commit.py
```diff
@@ -415,25 +414,40 @@ def session(self):
...
+        would do.
+        """
+
         if "unknown" in sha or "testing" in sha:
-            return None
-        fixture = f"../tests/entities/{self.test_shas[sha]}"
```
jgehrcke, line 394: Removed the abstract word 'fixture' from this code. There is a relative file path (now `path`), and a file object (now `f`).

jgehrcke:
Thanks for the review(s), Austin! Merging :ship:

conbench/tests/entities/test_commit.py
jgehrcke: This back-fired: https://github.com/conbench/conbench/issues/637#issue-1562612211

Did not remember this. Test suite passed because we skipped this test because of a rate limiting error. However, the patch was bad and in absence of rate limiting errors the test started "working again" in `main` (after merge), and then failed, exposing the bug.

Issue #472, opened 2022-11-30
Track step changes on BenchmarkResults


austin3dickey:
- Add a boolean column to BenchmarkResult called `is_step_change` that tracks whether this BenchmarkResult is sufficiently "different" from previous BenchmarkResults with the same hardware/case/context
- Add a way for humans using the webapp to mark a BenchmarkResult as a step change or not
- Mark step changes on the plots

jgehrcke:
Interesting discussion! Maybe my comment from https://github.com/conbench/conbench/pull/575#issuecomment-1375491063 should better have been placed here. :) Well.

Issue #383, opened 2022-09-29
Don't blow out history when there's a large change


jonkeane:
We rely on distributions of past runs to have confidence a change is really a regression or improvement, but our current setup will incorporate those changes into the distribution making the variance of non-alerted changes too large.

There are two stages here:
* identify different segments of stability
* merge together the distributions some how from the different segments. Naively, it's "just" scalling the distributions and calculating SDs, or taking the mean of the SD. But more likely we would actually want to treat this a sort of weighted calculation where the more recent segments are given a greater weight than further away segments.

Simulation of this:
```r
library(dplyr)
library(slider)
library(ggplot2)

fake_data_1 <- data.frame(
  time = rnorm(1000, mean = 5.28, sd = 0.01),
  true_regression = FALSE
  )

fake_data_2 <- data.frame(
  time = rnorm(10, mean = 11.2, sd = 0.01),
  true_regression = TRUE
)

fake_data_3 <- data.frame(
  time = rnorm(120, mean = 5.28, sd = 0.01),
  true_regression = FALSE
)

data <- bind_rows(fake_data_1, fake_data_2, fake_data_3)

# our current strategy
data_proc <- data %>%
  mutate(
    num = row_number(),
    sd = slider::slide_dbl(time, sd, .before = 100, .after = 0),
    mean = slider::slide_dbl(time, mean, .before = 100, .after = 0),
    z_score = (time - lag(mean)) / sd
  )

ggplot(data_proc) +
  aes(x = num, y = time) +
  geom_line() +
  geom_linerange(mapping = aes(ymin = mean - sd * 5, ymax = mean + sd * 5), alpha = 0.2, color = "blue") +
  ggtitle("times plotted over iterations", subtitle = "the grey band is mean +/- sd*5 (or our z-score threshold)")


# a slightly different way
special_sd <- function(x, y) {
  x_true <- x[y]
  x_false <- x[!y]

  mean(c(sd(x_true), sd(x_false)), na.rm = TRUE)
}

special_mean <- function(x, y) {
  x_true <- x[y]
  x_false <- x[!y]

  mean(x_false, na.rm = TRUE)
}

data_proc <- data %>%
  mutate(
    num = row_number(),
    sd = slider::slide2_dbl(time, true_regression, special_sd, .before = 100, .after = 0),
    mean = slider::slide2_dbl(time, true_regression, special_mean, .before = 100, .after = 0),
    z_score = (time - lag(mean)) / sd
  )

ggplot(data_proc) +
[...]

PR #574, opened 2023-01-05
Add `is_step_change` column to BenchmarkResult

austin3dickey:
This PR adds a boolean column to BenchmarkResult that tracks whether the result is "sufficiently different" from the previous commit (for the same hardware/case/context). It also adds a PUT endpoint to `/api/benchmarks` that allows users to update this field.

As of this PR, there's no way to update the field via the webapp (that's coming next) or automatically (that's further down the road).

Addresses some of #472 but not all.


.github/workflows/actions.yml
```diff
@@ -39,6 +39,8 @@ jobs:
             coverage lcov -o /etc/conbench-coverage-dir/coverage.lcov
       - name: Publish coverage
         uses: coverallsapp/github-action@master
+        # sometimes the coveralls API fails transiently, and we don't want that to affect this build
+        continue-on-error: true
```
austin3dickey, line 43: I kept running into failures here so I added this to this PR. I could split it out if you wish.

alistaire47 (approved):
The implementation looks good, though I do have one conceptual question: Is it worthwhile to store a richer value than a boolean here? It might later be useful to at least know sign info (i.e. is this a positive or negative step?), and we could equally well store positive or negative z-scores for changes and nulls for non-steps. Any choice depends on where we want to go with this in the future, which I don't think we've fully defined yet, but I want us to be conscious of the choice we're making here in terms of specificity. And if we are already, great!

austin3dickey:
>Is it worthwhile to store a richer value than a boolean here?

My general plan (see the conversation starting [here](https://github.com/conbench/conbench/issues/383#issuecomment-1330896435)) is to use these step changes as simply "mode" boundaries, so that when calculating statistics about a distribution of results we don't include data from beyond the boundaries. For that use, it doesn't really matter whether the change was an improvement or regression. 

I'm against storing actual z-scores on the BenchmarkResults since handling out-of-order data would get complex and we're slowly moving away from storing rolling calculations in the database (see #577). 

Another option, since this is to be used as a mode delimiter, is to store some sort of "mode_id" on the BenchmarkResults instead, so that users could theoretically connect two modes that aren't next to each other. But managing mode_ids seems tricky as well, because each newly created BenchmarkResult would have to look up and use its baseline's mode_id, and adding a mode boundary in historic data would require a lot of database updates. I think the `is_step_change` solution is simpler. Though I'm always open to a rename if it's clearer.

conbench/entities/benchmark_result.py
```diff
@@ -56,9 +56,12 @@ class BenchmarkResult(Base, EntityMixin):
     iqr = Nullable(s.Numeric, check("iqr>=0"))
     error = Nullable(postgresql.JSONB)
     validation = Nullable(postgresql.JSONB)
+    is_step_change = NotNull(s.Boolean, default=False)
 
     @staticmethod
     def create(data):
+        # note: is_step_change is ignored for BenchmarkResult.create();
+        # it should only be set via BenchmarkResult.update()
```
jonkeane: We might want to think about this a bit more: I very much could imagine knowing _when we send a benchmark result_  that we expect a step change, and so should not alert for things that have that happen. Though I would say we almost certainly shouldn't build the infrastructure to make this possible through out our runners, etc right now (we've got bigger things we need to address first, and this is probably a very niche idea), but it's not impossible.

But, if we are going to _not_ support that (generally, throughout our stack), IMO we should error (or at the very least warn!) if someone sends that key (I usually assume if I can `GET` a field, I can `POST` it too)

conbench/api/benchmarks.py
```diff
@@ -50,6 +51,33 @@ def get(self, benchmark_id):
...
+            content:
+                application/json:
+                    schema: BenchmarkUpdate
+        tags:
+          - Benchmarks
+        """
+        benchmark_result = self._get(benchmark_id)
+        data = self.validate_benchmark(self.schema.update)
+        benchmark_result.update(data)
+        set_z_scores([benchmark_result])
```
jonkeane, line 78: This PR does not yet exclude changes from the z-score calculation at all, right? If so, this line probably shouldn't be here since at this point it's at best re-calculating the same numbers as were calculated the first time (or if we've got bugs or we have an odd order of execution, it's changing the z-score numbers which could be incredibly confusing if you have a github alert saying there's some big regression, but then you click through and all is fine).

jonkeane (changes requested):
Looking good — one major question (confirming that we're waiting until #473 to change the z-score calculation + a suggestion for that as well) and a few smaller suggestions

conbench/entities/benchmark_result.py
austin3dickey: Yeah, this is a good call. I'll just enable being able to POST `is_step_change`.

conbench/api/benchmarks.py
austin3dickey: Yep, it's correct that I haven't changed any calculations affecting z-score in this PR! But here's why I think this line is still okay:
- This line doesn't actually change anything in the database or the webapp; it just adds a `"z_score": X` field to the API response of the PUT request.
- We haven't changed any z-score calculations yet, so when you GET /api/benchmarks/<id> you'll always get a z-score that ignores any step changes.
- I think it's best to have the PUT response return the same schema as the GET response, and further I think it's better to have the z-score field return the same thing too (instead of nullifying out z-score for PUT).

conbench/entities/benchmark_result.py
austin3dickey: okay done!

conbench/api/benchmarks.py
jonkeane: Aaah got it. This isn't something we should do here, but perhaps `set_z_scores` should have a more informative name `add_z_scores` `calculate_z_scores`? that make it a bit more obvious it's not persisting these anywhere

austin3dickey: Yeah I totally agree!! It's a bad name. Since it'll go away with #577 maybe we just ignore it for now?

Issue #476, opened 2022-11-30
Add behavior that automatically tries to determine if a BenchmarkResult is a step change


austin3dickey:
Probably just start with the same behavior as finding regressions/improvements

Issue #474, opened 2022-11-30
Manually exclude BenchmarkResults from a Distribution


austin3dickey:
- Add a boolean column to BenchmarkResult called `exclude_from_distribution` that tracks whether this BenchmarkResult should not be included in distribution-type queries (histories, z-scores, etc)
- Add a way for humans using the webapp to mark a BenchmarkResult to exclude or not (encourage people not to use this until some time has passed and they're sure)
- Mark exclusions on the plots

jgehrcke:
From a UX point of view I think one way to expose this would be "annotate as outlier". The annotation should allow for submitting a reason via a free text field. One consequence of outlier annotation then can be "exclude from distribution".

jgehrcke:
I still think that being able to label one specific data point as "outlier" (not considered in _any_ analysis) is important and powerful, and largely orthogonal to what we have (the nice ability to label a distribution change, where still each data point is used, but the rolling window is reset).

jgehrcke:
Had the thought that we may want to emphasize this entire aspect of Conbench, by adding to the mission/premise something like "Conbench helps you annotate and curate raw benchmark result data".

jgehrcke:
Still in my previous mental model of https://github.com/conbench/conbench/issues/171#issuecomment-1375505842:

> i) annotate expected, explainable, permanent change (reset statistical aggregate calculation at a certain point in time)
> ii) annotate an outlier (exclude a data point from statistical aggregate calculation)

We have done (i).

Now we still need to be able to annotate an outlier (exclude _one_ result from the aggregate).

Example scenario A: 
![image](https://user-images.githubusercontent.com/265630/217194319-830a0314-a6d3-4c3b-a5c8-f76d395b86cf.png)

Example scenario B:
![image](https://user-images.githubusercontent.com/265630/217197065-a3650802-1edd-4b93-9b11-9ac8cfdccbcf.png)


@jonkeane do you agree that in both of these scenarios the rolling window analysis can benefit from being able to exclude the _one_ offending data point? I think in some meetings you had the opinion that we wouldn't need (i) _and_ (ii), and I disagreed, potentially not actually understanding what your point was :). I was waiting for specific examples to continue this discussion. The two screenshots above are decent examples I think.

jonkeane:
[...]

PR #575, opened 2023-01-05
Enable toggling step changes in webapp

### LINKED ISSUE #472: Track step changes on BenchmarkResults

austin3dickey:
Fixes #472.

This PR enables users to toggle whether BenchmarkResults are step changes in the webapp. It also displays step changes on the plots as vertical lines.

Sorry for the poor-quality GIF, but here's the workflow in action. If logged in, you can toggle with the checkbox, and then the line is persisted on all plots of the same history.

recording

### DIFF
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -6,7 +6,7 @@ from typing import Optional
 
 import bokeh.plotting
 import dateutil
-from bokeh.models import Spacer
+from bokeh.models import Spacer, Span
 
 from ..hacks import sorted_data
 from ..units import formatter_for_unit
@@ -401,6 +401,32 @@ def time_series_plot(history, benchmark, run, height=380, width=1100):
             legend_label="current benchmark (min)",
             name="benchmark",
         )
+
+    # visually separate out distribution changes
+    dist_change_in_legend = False
+    for result in history:
+        if result["change_annotations"].get("begins_distribution_change", False):
+            p.add_layout(
+                Span(
+                    location=dateutil.parser.isoparse(result["timestamp"]),
+                    dimension="height",
+                    line_color="purple",
+                    line_dash="dashed",
+                    line_alpha=0.5,
+                )
+            )
+
+            if not dist_change_in_legend:
+                # hack: add a dummy line so it appears on the legend
+                p.line(
+                    [dateutil.parser.isoparse(result["timestamp"])] * 2,
+                    [result["mean"]] * 2,
+                    legend_label="distribution change",
+                    line_color="purple",
+                    line_dash="dashed",
+                    line_alpha=0.5,
+                )
+                dist_change_in_legend = True
 
     hover_renderers = [
         scatter_mean_over_time,
--- a/conbench/app/benchmarks.py
+++ b/conbench/app/benchmarks.py
@@ -11,6 +11,16 @@ from ..app._util import augment, display_message, display_time
 from ..config import Config
 
 
+class UpdateForm(flask_wtf.FlaskForm):
+    toggle_distribution_change = w.SubmitField(
+        render_kw={
+            "title": 'A distribution change means a sufficiently "different" '
+            "distribution began, such that data from before shouldn't be statistically "
+            "compared to data after."
+        }
+    )
+
+
 class DeleteForm(flask_wtf.FlaskForm):
     delete = w.SubmitField("Delete")
 
@@ -128,12 +138,24 @@ class RunMixin:
 
 
 class Benchmark(AppEndpoint, BenchmarkMixin, RunMixin, TimeSeriesPlotMixin):
-    def page(self, benchmark, run, form):
-        if not flask_login.current_user.is_authenticated:
-            delattr(form, "delete")
-
+    def page(self, benchmark, run, delete_form, update_form):
         if benchmark is None:
             return self.redirect("app.index")
+
+        update_button_color = "default"
+        if flask_login.current_user.is_authenticated:
+            if benchmark["change_annotations"].get("begins_distribution_change", False):
+                update_form.toggle_distribution_change.label.text = (
+                    "Unmark this as the first result of a distribution change"
+                )
+                update_button_color = "info"
+            else:
+                update_form.toggle_distribution_change.label.text = (
+                    "Mark this as the first result of a distribution change"
+                )
+        else:
+            delattr(delete_form, "delete")
+            delattr(update_form, "toggle_distribution_change")
 
         return self.render_template(
             "benchmark-entity.html",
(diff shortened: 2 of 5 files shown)


austin3dickey:
Build will fail until #574 is merged.

EDIT: it's merged! Should work now

conbench/app/_plots.py
```diff
@@ -237,6 +238,28 @@ def time_series_plot(history, benchmark, run, height=250, width=1000):
...
+    for step_change in step_changes:
+        p.add_layout(step_change)
+    if step_changes:
+        # hack: add a dummy line so it appears on the legend
+        one_result = [result for result in history if result["is_step_change"]][0]
+        p.line(
+            [dateutil.parser.isoparse(one_result["timestamp"])] * 2,
+            [one_result["mean"]] * 2,
+            legend_label="Step change",
+            line_color="black",
```
alistaire47: I think black is visually confusing here, since it can look like part of the line itself (i.e. a very fast vertical wiggle, which it kind of is, but not like that). Red comes to mind, but that's more appropriate for regressions than improvements; if we're not going to separate them, we should probably use a more neutral color like blue or purple or orange or something. Or we could color based on direction if we track it (which it seems like we should). We may need special handling for spikes that immediately revert then though.

Opacity would help too; maybe 50-70% so it's more of an event than the distribution ribbon, but less of a datapoint than the line.

conbench/app/_plots.py
```diff
@@ -237,6 +238,28 @@ def time_series_plot(history, benchmark, run, height=250, width=1000):
...
+    ]
+    for step_change in step_changes:
+        p.add_layout(step_change)
+    if step_changes:
+        # hack: add a dummy line so it appears on the legend
+        one_result = [result for result in history if result["is_step_change"]][0]
+        p.line(
+            [dateutil.parser.isoparse(one_result["timestamp"])] * 2,
+            [one_result["mean"]] * 2,
+            legend_label="Step change",
```
alistaire47: This really deserves a fuller explanation somewhere explaining the definition here in terms of z-scores and what the threshold is, etc. Maybe in a hover if we can?

alistaire47 (approved):
I like the direction here—if we're tracking regressions and improvements, we should show them! A couple thoughts on visual impact.

conbench/app/_plots.py
austin3dickey: These are great ideas. Maybe a dotted line would work well.

conbench/app/_plots.py
austin3dickey: I'm sorta leaving this vague on purpose for now -- I think we need to collect more data before we can make a solid recommendation on how to use the `step_change` field. We can definitely come back and add a better definition once we do, e.g., #476.

jonkeane: While I agree that gathering data is really great, we still will need to nudge folks in the right direction in order for them to use it at all (such that we would have any substantive points labelled as step-changes).

We can be vague about the details if we think that's helpful but something along the lines of "Mark this as a large shift in the data, which should be treated separately from the data before it for the purposes of variability measurement"

conbench/app/_plots.py
jonkeane: Agreed here, that the black might be a bit stark. I suspect we will need to treat the two cases we are currently imagining slightly differently (especially later in the history after more runs are added to the plots):

* if it's a one-off sudden regression or improvement that goes back to the mean right after  
    here, we probably want this point to be lowered opacity to show that it's being ignored 
    
* if it's a step-change and the benchmarks are now at this new higher or lower number  
    here, we probably want to highlight where this began, and what all is included in determining variability here.

austin3dickey: Yes, agreed. I'm planning to handle the case of "spikes" in #474 . Those will be rendered differently on the plots. 🙂 

I'll attach a screenshot of the new line style.

conbench/app/_plots.py
austin3dickey: That makes sense to me! I'll update the description by the checkbox

conbench/app/benchmarks.py
```diff
@@ -11,6 +11,12 @@
 from ..config import Config
 
 
+class UpdateForm(flask_wtf.FlaskForm):
+    is_step_change = w.BooleanField(
+        "Is this result a step change?", render_kw={"onchange": "this.form.submit()"}
+    )
```
jonkeane: You might get a second notification about this, I thought I sent this already, but I can't find it on github (it might have been eaten by browser cache?). Aaanyway:

This is relatively minor, and I'm not sure if it's a flask thing or a bootstrap thing (or something else): Is there anyway we could make this a toggle button instead of a checkbox? I fear that checkboxes look more like todo lists (e.g. something that should generally be set) whereas for this it should be relatively rare that it's toggled.

austin3dickey: oh sure let me mess around with it!

conbench/app/_plots.py
austin3dickey: Here's the new line style (and the new button below the plot) 
Screenshot 2023-01-06 at 14 38 02

conbench/app/_plots.py
austin3dickey: What do you think of this as hover-text on the button?
Screenshot 2023-01-06 at 14 37 49

conbench/app/benchmarks.py
austin3dickey: Changed to a button; see the screenshots I just posted

austin3dickey:
If you want to play with this: clone the branch, run `make run-app-dev`, then in a separate terminal run `make db-populate`, then open http://127.0.0.1:5000/ in your browser! (Thanks @jgehrcke for the nice `make` commands!)

jgehrcke:
Ho ho :christmas_tree:  I am not sure if I can productively provide feedback here, because I'd like to talk about concept and purpose, whereas the change and thought process and implementation is already so advanced that I might be too late for that.

Some quick and potentially ignorant thoughts are around the UX and how this 'step change' concept might be entangled with other concepts. Sorry if some of that is answered or discussed here in this ticket or elsewhere.

What's the primary purpose of the annotation? I see two possibilities:

1) Is this meant to annotate an outlier, with the effect of simply excluding a data point from rolling time window analysis?

2)  Is this meant to annotate an expected, significant change (in either direction)? Does this then have the effect to 'reset' rolling time window analysis (forget everything before this point in time, start building aggregates from here on again)? 

I want to say that 'step change' is not a term I am used to. Where is that an established term (happy to learn)? I have worked a lot with step functions in continuous space, but with discrete data points all changes look like a 'step change' :-) :shrug:. I think the term 'step change' points to (2) more than (1), but I'd hope we can find a nicer/more intuitive user-facing term, based on the consequences of the annotation. Based on what I've seen, I'd say: this is about being able to annotate change, significant change, expected to be permanent. Certainly reviewed change, hopefully (but not necessarily) explainable change -- with the goal to reset aspects about statistical analysis.

If it's (2) and the expected change is permanent, it could be lovely to enrich the annotation with a free text field where users should explain the reason for the change. This can be powerful also when improving performance over time: get an auto-generated plot that shows how performance improved over time and also _why_.

austin3dickey:
>What's the primary purpose of the annotation? I see two possibilities

You're correct that there are 2 possibilities here; this PR addresses (2) (#472) and there will be a future PR to address (1) (#474).

I just made #580 and #581 to address your two ideas!

austin3dickey:
I just opened #582 which will change some of the code in this webapp PR. I'll rebase this onto that (partly to see if the new approach is "easy enough" to develop with)

conbench/app/_plots.py
austin3dickey: Update: 
Screenshot 2023-01-10 at 16 07 42

conbench/app/_plots.py
austin3dickey: Update: 
Screenshot 2023-01-10 at 16 08 12

austin3dickey:
It wasn't too hard to rebase onto #582! Check out the code again and see if you agree.

conbench/app/_plots.py
alistaire47: Maybe pedantic, sorry, but I don't love "permanent" as a descriptor because inevitably there will be more changes in the future. "Durable" maybe?

Regardless, do we want users to not click this button for spikes, or quick deviations for a few commits that revert? If they're because of issues that resolve, clicking seems good in terms of getting them healthy distributions (and therefore future alerting). If we want to keep outliers separate that's ok, but I'm not convinced we need to here—though this is definitely the sort of thing it'd be useful to collect free text about as we figure out how to evolve our handling here.

jonkeane: > If we want to keep outliers separate that's ok, but I'm not convinced we need to here—though this is definitely the sort of thing it'd be useful to collect free text about as we figure out how to evolve our handling here.

Yeah agreed, I would also strongly advocate for bundling "purposeful change that does indeed persist" with "this outlier is a blip that I can't explain which goes down immediately after" together _for now_. We'll follow up with making that distinction later when we have more real info if any of this is helpful + we have real data we can work with.

Maybe we could call this just "a change in distribution" instead of qualifying it. It's a bit ambiguous, but we might want to be purposefully so for now (and we have the hover over where we can explain that in a bit more detail.

conbench/app/benchmarks.py
jonkeane: Would it be possible to have some styling on that button to make it clear when it's marked and when it's not? This also might already exist and I'm only seeing screenshots of un-set buttons so far?

conbench/app/_plots.py
austin3dickey: Yep, this makes sense, and ties into the fact that we can use the new `change_annotations` for anything. Let's try it.

conbench/app/benchmarks.py
austin3dickey: How's this? The possible button colors are [here](https://getbootstrap.com/docs/3.3/components/#btn-dropdowns-single)
Screenshot 2023-01-12 at 14 19 37
Screenshot 2023-01-12 at 14 19 46

conbench/app/_plots.py
austin3dickey: Okay, changed everything to "distribution change."

austin3dickey:
I merged #582 and rebased; all tests pass so I'll merge this. Thanks for all the feedback! We can iterate more in future issues if you have more feedback.

jgehrcke:
@austin3dickey thanks for all the work on this, and it's great that we merged this and related PRs!

Now I am a tiny little lost and do not know what the consequences are as of today when I click that button: do we reset the rolling window analysis, or do we only show the vertical line?

I would like when the precise list of consequences is documented / specified in the Conbench UI. Could be something in the lines of "as of today, clicking this button results in a vertical line being shown in the plot above. In the future, we plan to ...".

austin3dickey:
@jgehrcke Good call; I forgot to document this. Right now it just adds a vertical line. I will document the full behavior in #473.

jgehrcke:
> will document the full behavior in https://github.com/conbench/conbench/issues/473.

Have created https://github.com/conbench/conbench/issues/691 for this! :)

Issue #475, opened 2022-11-30
benchalerts: Allow wildcard suppression of alerts


austin3dickey:
## Current plan
When setting up a benchalerts script to run in CI, you may provide string wildcard patterns for various fields like benchmark name, hardware name, tags, etc. If an alert for a specific benchmark result matches one of those string patterns, it will be suppressed.

## Old plan

Ability to mark a Distribution (i.e. a certain case/context/hardware/repo) as "never alert" by storing that info in the DB.
- Add a boolean column to Distribution called no_alerts that tracks whether this Distribution is "weird" enough to justify never alerting on regressions, etc
- Add a way for humans using the webapp to mark a Distribution as worthy of alerts or not
- Never say there was a regression/improvement if no_alerts is True

jonkeane:
I wonder if instead of distributions here, we add it at the benchmarkresult level (for now?)

austin3dickey:
Do you mean: mark a BenchmarkResult as "never alert on any results sharing this case/context/hardware/repo" or "never alert for this result in particular"?

jonkeane:
Part of doing this task will be figuring out if either of those is useful to users and if either of those is feasible and then weighing that to figure out if + when we should do it

austin3dickey:
This might actually be more suited to a `benchalerts` configuration rather than something that permanently lives in the database. I imagine something like this: when setting up a `benchalerts` script to run in CI, you may provide string wildcard patterns for various fields like benchmark name, hardware name, tags, etc. If an alert for a specific benchmark result matches one of those string patterns, it will be suppressed.

jonkeane:
> This might actually be more suited to a benchalerts configuration rather than something that permanently lives in the database. I imagine something like this: when setting up a benchalerts script to run in CI, you may provide string wildcard patterns for various fields like benchmark name, hardware name, tags, etc. If an alert for a specific benchmark result matches one of those string patterns, it will be suppressed.

Yup, totally. We have two levels | areas where we flag these (one in benchalerts and one in the web UI). I'm more than happy to try out some things in benchalerts, see how they feel even if they don't (yet) bubble up into the UI (e.g. having benchalerts _not_ alert for a regression, but the UI still shows it as such) for a little bit while we play around with what works. Eventually we will want to unify the behaviors (which of course is easiest if the code is unified, but that's not the only way!), but getting it in front of people + getting their reactions + iterating on what's useful is the most important part of this, so doing it in one place first, iterating, etc. is the right move

[...]

PR #582, opened 2023-01-09
Change `is_step_change` to `change_annotations`

### LINKED ISSUE #580: Consider renaming `is_step_change`
>I want to say that 'step change' is not a term I am used to. Where is that an established term (happy to learn)? I have worked a lot with step functions in continuous space, but with discrete data points all changes look like a 'step change' :-) :shrug:. I think the term 'step change' points to (2) more than (1), but I'd hope we can find a nicer/more intuitive user-facing term, based on the consequences of the annotation. Based on what I've seen, I'd say: this is about being able to annotate change, significant change, expected to be permanent. Certainly reviewed change, hopefully (but not necessarily) explainable change -- with the goal to reset aspects about statistical analysis.

_Originally posted by @jgehrcke in https://github.com/conbench/conbench/issues/575#issuecomment-1375491063_

### LINKED ISSUE #581: Add JSON column to BenchmarkResult for "why" the result was a step change or outlier
>If the expected change is permanent, it could be lovely to enrich the annotation with a free text field where users should explain the reason for the change. This can be powerful also when improving performance over time: get an auto-generated plot that shows how performance improved over time and also _why_.

_Originally posted by @jgehrcke in https://github.com/conbench/conbench/issues/575#issuecomment-1375491063_

austin3dickey:
Fixes #581 and #580. 
Helps with #472, #473, #474, and #475.

This PR introduces `change_annotations`, a new free-form JSONB column on BenchmarkResult that is intended to store (post-analysis) information about how that BenchmarkResult stands among its distribution of results. In the future, it will store whether that result is the beginning of a step change/discontinuity/outlier/etc, as well as reasons for why it was annotated that way (user-given or bot-given).

This new functionality supersedes the `is_step_change` column so I removed that.


conbench/tests/api/_expected_docs.py
```diff
@@ -1075,9 +1075,9 @@
             },
             "BenchmarkUpdate": {
                 "properties": {
-                    "is_step_change": {
-                        "description": "Is this result the first result of a sufficiently 'different' distribution than the result on the previous commit (for the same hardware/case/context)? That is, when evaluating whether future results are regressions or improvements, should we treat data from before this result as incomparable?",
-                        "type": "boolean",
+                    "change_annotations": {
+                        "description": "Post-analysis annotations about this BenchmarkResult that\ngive details about whether it represents a change, outlier, etc. in the overall\ndistribution of BenchmarkResults.\n\n\nThis endpoint will only update the user-specified keys, and leave the rest alone. To\ndelete an existing key, set the value to null.\n",
```
jonkeane, line 1103: We should add details about why one might set this field (and we should also be explicit that — for now — all details inside of it are entirely ignored). The prior boolean had "That is, when evaluating whether future results are regressions or improvements, should we treat data from before this result as incomparable?" which is IMO a decent description + vague enough for us to get to the end of this project without needing to come back and change it but still be meaningful to folks using conbench to have some idea of what to do with this.

jonkeane (commented):
The code here looks ok to me and I'm fine to merge this and move on if it's helpful — but I'm a bit worried that this is more code than is necessary for us to get to our end goals for the history project so that we can start validating + exploring what kinds of discontinuities folks report + how to deal with those differently (if we even need to). 

The # of lines isn't huge, but there's a non-trivial maintenance burden for exposing more complications like this further up the stack (e.g. in #575 we'll need to validate json or add in some hardcoded json etc.) IMHO, we would be better served to leave this be as a boolean for now and get the history improvements in front of folks rather than try to figure out what the whole set of ways we might see deviations in the future + plan out + document + validate all of those (which could be connected to the spirit of as-much-as-needed-as-little-as-possible. 

But I won't block this from merging if you want to, let's do it (decisively) now and move on to the meat of this project.

conbench/tests/api/_expected_docs.py
austin3dickey: Agreed - I will do this in #575 when I write code that needs a specific element of the JSON.

austin3dickey:
>I'm a bit worried that this is more code than is necessary

My current hypothesis is that documenting and using the JSON field `change_annotations.begins_permanent_change` will only be a little bit harder than documenting and using `is_step_change`, and allows more flexibility for change later (no pun intended). I think that benefit is worth it. If I can't do #575 "easily," I'll reject that hypothesis, so maybe I won't merge this PR until that PR is rebased onto this one.

conbench/entities/benchmark_result.py
```diff
@@ -56,7 +56,7 @@ class BenchmarkResult(Base, EntityMixin):
     iqr = Nullable(s.Numeric, check("iqr>=0"))
     error = Nullable(postgresql.JSONB)
     validation = Nullable(postgresql.JSONB)
-    is_step_change = NotNull(s.Boolean, default=False)
+    change_annotations = NotNull(postgresql.JSONB, server_default="{}")
```
alistaire47: Is it more useful to make it nullable or have a default of `{}`? Not sure which is simpler to query

jonkeane:
> If I can't do https://github.com/conbench/conbench/pull/575 "easily," I'll reject that hypothesis, so maybe I won't merge this PR until that PR is rebased onto this one.

Love this!

alistaire47 (approved):

conbench/entities/benchmark_result.py
austin3dickey: I'm not sure -- I did some basic googling earlier today and couldn't quickly find an answer. To me it "feels like" the {} is better because my Python code can always use `result.change_annotations.get("asd")` instead of something like `(result.change_annotations or {}).get("asd")`

austin3dickey: After a lot of discussion with @ElenaHenderson , I made this nullable. It's just simpler to do the migration that way.

PR #584, opened 2023-01-10
Webapp: dropdown search enabled for individual columns 

### LINKED ISSUE #564: search not yeilding results for certain inputs
*Below I am filtering using the `lang`/language column*

### Bit more insight:
The input "R" didn't fetch any results while other inputs (Python, JavaScript etc) yielded the desired outcome. The instant thought that comes to mind is "maybe single char search is buggy" or maybe not (more digging could help)

https://conbench.ursa.dev/runs/9292568e6e4b4c1882d0564bf8ebc3e7/?search=python

Screenshot 2022-12-23 at 3 40 34 PM

https://conbench.ursa.dev/runs/9292568e6e4b4c1882d0564bf8ebc3e7/?search=R
Screenshot 2022-12-23 at 3 40 26 PM

sayantikabanik:
resolves #564 

#### Details

- This PR enables search for specific columns for the DataTable
- Add-on to the existing search functionality
- New JS method `specific_column_search` under `app.html` which can be used across the templates to enable search for targeted cols

```javascript
// example usage
// requires 02 input: table_id & col indexes in array format
specific_column_search('#runs', [3,6]);
```

https://user-images.githubusercontent.com/17350312/211490986-dbe1c1dc-f5d8-4540-b793-624a7ec43a5b.mov

### DIFF
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -155,17 +155,28 @@ function specific_column_search(data_table_id, column_indexes) {
         var table = $(data_table_id).DataTable();
         table.columns(column_indexes).every(function() {
             var that = this;
+            var reset = ""
             var mySelectList = $("<select></select>")
                 .appendTo(this.header())
                 .on("change", function() {
                     that.search($(this).val()).draw();
                 });
-
+            mySelectList.append(
+                        $(`<option value="${reset}">${reset}</option>`)
+                    );
             this.cache("search").unique().sort().each(function(param) {
-                mySelectList.append(
-                    $(`<option value="${param}">${param}</option>`)
-                );
+                if (param!==""){
+                    mySelectList.append(
+                        $(`<option value="${param}">${param}</option>`)
+                    );
+                }
+                else{
+                    mySelectList.append(
+                        $(`<option value="${param}">null</option>`)
+                    );
+                }
             });
+
         });
     });
 }
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -138,25 +138,28 @@ function enable_search_query_string(data_table) {
     });
 
 }
+/*
+INFO:
+    Adds search ability for specific columns in the DataTable,
+    Creates a *table* instance and update the header with the dropdown menu
+    List of unique values based on the column data fields are added
+    Output is yield based on the selected option from the dropdown menu
 
-// requires two inputs: table_id & column_indexes
-function specific_column_search(data_table, column_indexes){
+INPUTS:
+ string: data_table_id
+ array: column_indexes
+ */
+function specific_column_search(data_table_id, column_indexes){
     $(document).ready(function() {
-        var table =  $(data_table).DataTable();
-        table.columns(column_indexes)
-		.every(function () {
+        var table =  $(data_table_id).DataTable();
+        table.columns(column_indexes).every(function () {
             var that = this;
 			var mySelectList = $("<select></select>")
 			.appendTo(this.header())
-			.on("change", function () {
-				that.search($(this).val()).draw();
+			.on("change", function () {that.search($(this).val()).draw();
 			});
 
-        this
-			.cache("search")
-            .unique()
-			.sort()
-			.each(function (param) {
+        this.cache("search").unique().sort().each(function (param) {
 				mySelectList.append(
 				$(`<option value="${param}">${param}</option>`)
 				);
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -173,9 +173,8 @@ function specific_column_search(data_table_id, column_indexes) {
                     mySelectList.append(
                         $(`<option value="${param}">null</option>`)
                     );
-                }
+                }           
             });
-
         });
     });
 }
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -162,15 +162,14 @@ function specific_column_search(data_table_id, column_indexes) {
                     that.search($(this).val()).draw();
                 });
             mySelectList.append(
-                        $(`<option value="${reset}">${reset}</option>`)
-                    );
+                $(`<option value="${reset}">${reset}</option>`)
+            );
             this.cache("search").unique().sort().each(function(param) {
-                if (param!==""){
+                if (param !== "") {
                     mySelectList.append(
                         $(`<option value="${param}">${param}</option>`)
                     );
-                }
-                else{
+                } else {
                     mySelectList.append(
                         $(`<option value="${param}">null</option>`)
                     );
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -121,7 +121,7 @@ $(document).ready(function() {
 });
 
 function enable_search_query_string(data_table) {
-    $(data_table).on( 'search.dt', function () {
+    $(data_table).on('search.dt', function () {
         var current_search_value = $('.dataTables_filter input').val();
         var new_url = window.location.protocol + "//" + window.location.host + window.location.pathname
         if (current_search_value) {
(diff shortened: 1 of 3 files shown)


sayantikabanik:
Single char search still feels buggy (example: lang=R), came across couple of similar questions on forum. Though not much activity there

https://www.datatables.net/forums/discussion/18977/filter-not-working-for-single-charcter

conbench/templates/app.html
```diff
@@ -136,8 +136,36 @@ <h4 class="modal-title"><span class="glyphicon glyphicon-trash submenu"></span>&
             table.search("{{ search_value }}").draw();
         {% endif %}
     });
+
 }
 
+// requires two inputs: table_id & column_indexes
+function specific_column_search(data_table, column_indexes){
```
jgehrcke: Ah, I see that the `unique()` down there is really important :)

Could you maybe add a code comment specifying the behavior of this function?

Something like "identify the set of values in ...., and then ..."

conbench/templates/index.html
```diff
@@ -119,8 +119,8 @@ <h3>Language-independent Continuous Benchmarking (CB) Framework</h3>
           "searchPlaceholder": "commit hash works too",
       }
 });
-
 enable_search_query_string('#runs');
+specific_column_search('#runs', [3,6]);
```
jgehrcke: Could you add a code comment for what this does? Something like: "in the Runs table, add a setpicker to the headers of columns three and six, which are ..." (I don't know if I have understood this!)

conbench/templates/app.html
```diff
@@ -136,8 +136,36 @@ <h4 class="modal-title"><span class="glyphicon glyphicon-trash submenu"></span>&
...
+			.cache("search")
+            .unique()
+			.sort()
+			.each(function (param) {
+				mySelectList.append(
+				$(`<option value="${param}">${param}</option>`)
+				);
+			});
+		});
+	});
```
austin3dickey: I know we don't have a linter set up for our HTML files (sorry!), but would you mind cleaning up the indentation a little for this function? Looks like the rest of this file uses 4-space tabs.

conbench/templates/index.html
austin3dickey: Unfortunately the [Arrow conbench instance](https://conbench.ursa.dev/) has many more hardware names than our repo's dummy data -- I just counted 68 in the last 1000 runs:
```
>>> from benchclients import ConbenchClient
>>> c = ConbenchClient()
>>> runs = c.get('/runs')
>>> len(set(run['hardware']['name'] for run in runs))
68
```
I think this will be too many for a dropdown. Instead of hardcoding which columns to apply the dropdown to, I wonder if it would be better to automatically apply it to any column that has less than, for example, 10 unique values? That might be tricky to code but could be worth it for the UX.

Alternatively, you could save that for a future PR, and just get rid of the dropdown for the Hardware column. It will probably still be useful for Reason and benchmark Language.

conbench/templates/app.html
```diff
@@ -136,8 +136,36 @@ <h4 class="modal-title"><span class="glyphicon glyphicon-trash submenu"></span>&
...
 }
 
+// requires two inputs: table_id & column_indexes
+function specific_column_search(data_table, column_indexes){
+    $(document).ready(function() {
+        var table =  $(data_table).DataTable();
+        table.columns(column_indexes)
+		.every(function () {
+            var that = this;
+			var mySelectList = $("<select></select>")
```
austin3dickey: Looking at the dummy data, I can see that the Reason column has a "blank" choice, which might not have been intentional because some Reasons are actually blank in the dummy data. But I actually really like that behavior because it allows users to "deselect" the filter. Could you try to add a blank to every option list on purpose?

image

austin3dickey (commented):
Nice, I like this feature!

conbench/templates/index.html
jonkeane: Or maybe in circumstances when there are many unique values, we switch over to a free-text search. We don't have to do that in this PR, but it would be nice to add that for columns that have many unique values IMHO

alistaire47: > Unfortunately the [Arrow conbench instance](https://conbench.ursa.dev/) has many more hardware names than our repo's dummy data -- I just counted 68 in the last 1000 runs

This is my fault and needs cleaning up; I fixed the C++ microbenchmark hardware name issue, but I think the Go ones are still problematic, so I haven't cleaned up the db yet. The real number should be [7-8](https://github.com/voltrondata-labs/arrow-benchmarks-ci/blob/main/config.py#L65) (Go is still separate for now).

jonkeane:
I really love this feature! Thank you for pushing it forward. IMO, we can merge this as is (pending any change requests above). 

But seeing this made me realize: it would be _really nice_ if we could have drop down searches for low cardinality columns, and then free-form text search for all other columns. We can (and should!) do that as a follow on unless you think it's easier to tack on here.

sayantikabanik:
> But seeing this made me realize: it would be _really nice_ if we could have drop down searches for low cardinality columns, and then free-form text search for all other columns. We can (and should!) do that as a follow on unless you think it's easier to tack on here.

The existing *Search* bar is still functional and works well for free text searches. Probably we could utilize that?

conbench/templates/app.html
sayantikabanik: I shall add a condition to identify empty/null value and then a separate reset feature. 
Else it will turn little confusing

sayantikabanik:
Tweaked the method to incorporate the `null` condition and added the reset logic


Screenshot 2023-01-17 at 4 58 49 PM

jonkeane:
> > But seeing this made me realize: it would be really nice if we could have drop down searches for low cardinality columns, and then free-form text search for all other columns. We can (and should!) do that as a follow on unless you think it's easier to tack on here.

> The existing Search bar is still functional and works well for free text searches. Probably we could utilize that?

IMO we should have both: the overall search bar searching is nice in cases where someone might not know which column an element is in. But it also would be fantastic to be able to let users search a specific column if they know that's what they're looking for. It would also allow this interface to degrade nicely in the case that a column that we think should be low cardinality but in reality end up being high cardinality. Additionally, I find it a bit bothersome visually that only some of the columns have an interface element (i.e. drop down or search bar) and others don't

conbench/templates/app.html
austin3dickey: This looks great, thanks!

austin3dickey (commented):
Overall, looks good to me! Maybe we should just remove the feature from the Hardware column.

sayantikabanik:
> > > But seeing this made me realize: it would be really nice if we could have drop down searches for low cardinality columns, and then free-form text search for all other columns. We can (and should!) do that as a follow on unless you think it's easier to tack on here.
> 
> > The existing Search bar is still functional and works well for free text searches. Probably we could utilize that?
> 
> IMO we should have both: the overall search bar searching is nice in cases where someone might not know which column an element is in. But it also would be fantastic to be able to let users search a specific column if they know that's what they're looking for. It would also allow this interface to degrade nicely in the case that a column that we think should be low cardinality but in reality end up being high cardinality. Additionally, I find it a bit bothersome visually that only some of the columns have an interface element (i.e. drop down or search bar) and others don't

Thanks for sharing, I could see the incremental value the additional feature could bring in. 
Shall create an issue to record this + a followup PR

jonkeane (approved):
I'm happy to merge these, and can't wait to see them in prod

austin3dickey (approved):
Thanks for this!

jonkeane:
Is there anything blocking us merging this?

conbench/templates/index.html
```diff
@@ -119,8 +119,12 @@ <h3>Language-independent Continuous Benchmarking (CB) Framework</h3>
           "searchPlaceholder": "commit hash works too",
       }
 });
-
 enable_search_query_string('#runs');
 
+/* specify the HTML table id (string) and the indexes of the columns for
+ which search functionality should be enabled in an array format
+*/
+specific_column_search('#runs', [6]);
```
austin3dickey, line 127: forgot to ask - can you add this to the Repository column too? It's theoretically possible to track a few repositories at once in one server.

sayantikabanik: Do we have couple of repos (from where we track), while I searched there was just one. 
Might be tad confusing to have a drop-down with only one option, which does not add any functionality

jonkeane: We've already talked about making each column searchable where the search that pops up is relevant to the contents (see below) of the column here and in #613 IMHO we should get that done and shipped in this process (and then we don't have to debate things like this!).

Relevant to the contents:
* < 10 values: a drop down
* < 2 values: nothing? with a small message saying "there's only one value"?
* > 10 values: free text search

That way each column has something to go there, and we can treat each column the same and not need to hand-manage "which column is a free text search? which column is a ...?")

sayantikabanik:
Merging this PR, thanks everyone for the feedback 🌟
Will be continuing on the finer details under #613

PR #585, opened 2023-01-10
Fix step change migration and branch limit

austin3dickey:
This PR fixes a few things:
- I accidentally committed [a migration](https://github.com/conbench/conbench/blob/c96741a742d253d1e6b858496f665fb81e48637a/migrations/versions/480dbbd48927_add_commit_branch_and_fork_point_sha.py#L20) that sets the character limit of `Commit.branch` to 100. It should be 510 as we discussed [here](https://github.com/conbench/conbench/pull/417#discussion_r1009621466).
- Alembic should pick up on column type changes like the above. I fixed its config so it does that.
- In the last PR (#574) the migration fails if there's any data in the database, because it's adding a non-nullable column. This PR deletes that migration (which never got deployed anywhere) and replaces it with the correct way to add a non-nullable column with a default value.


jonkeane (approved):
Thanks for the quick + clean fix as well as the improvement to catch this in the future.

jgehrcke (commented):
:muscle:

PR #587, opened 2023-01-10
Add environment variables for posting metadata

alistaire47:
Cleanup for passing metadata about a run to runners for posting via various environment variables. Specifically:

- Extends/replaces `BENCHMARKABLE_PR_NUMBER` with ~`BENCHMARKABLE_REPOSITORY` and `BENCHMARKABLE_COMMIT`~ `CONBENCH_PR_NUMBER`, `CONBENCH_REPOSITORY`, and `CONBENCH_COMMIT` so all github info can be passed via env vars. Since this makes everything automatable again, I changed the default for populating `github` in `benchadapt.BenchmarkResult` and `benchadapt.BenchmarkRun` to attempt to grab the metadata from the env vars instead of an empty dict (which will always fail if posted). If any are unspecified, it will raise a warning. All existing behavior (including pass `None` for detection) should continue as-is unless we're relying on that empty dict somewhere (which we shouldn't). I did not implement this for the old conbench version since nothing relying on it should need this new functionality (everything is already specifying this directly), though if we think I should I can.
- Adds `CONBENCH_HOST_NAME` as an env var to use as an alternative to detecting the machine host name in `machine_info()` (needed for cases where a consistent host name is needed where the actual one can vary, like on CI or cloud runners). Previously this was handled via the `.conbench` config; that will keep working for now, but this is intended to be the preferred path for this in the future. Implemented in conbench version as well because it's simple.
- Documents everything in docstrings and benchconnect, which are the two important places they'll be needed.
- Not here but related: I haven't introduced them externally anywhere (yet), but in an upcoming arrowbench PR I'll start using `CONBENCH_RUN_NAME`, `CONBENCH_RUN_ID`, and `CONBENCH_RUN_REASON` internally, and I'll likely reuse those again when I rework arrow-benchmarks-ci (again internally). I don't see a reason yet to codify them anywhere since usages are all internal, but consistency in implementations is nice, and anybody implementing benchmarks that will be sent to a Conbench server will likely need similar. Once things are settled, these should probably go in an implementation guidebook or CI template or however we decide to document implementation.

### DIFF
--- a/benchadapt/python/benchadapt/_machine_info.py
+++ b/benchadapt/python/benchadapt/_machine_info.py
@@ -96,7 +96,7 @@ def github_info():
             (
                 "Both CONBENCH_PROJECT_REPOSITORY and CONBENCH_PROJECT_COMMIT must be "
                 "set if `github` is not specified. CONBENCH_PROJECT_PR_NUMBER should be "
-                "null for builds on the default branch."
+                "not be set for builds on the default branch."
                 f"\nValues: `{gh_info}`"
             ),
         )
--- a/benchadapt/python/benchadapt/_machine_info.py
+++ b/benchadapt/python/benchadapt/_machine_info.py
@@ -91,11 +91,12 @@ def github_info():
         "pr_number": pr_number,
     }
 
-    if not (repo and pr_number and commit):
+    if not (repo and commit):
         warnings.warn(
             (
-                "All of CONBENCH_REPOSITORY, CONBENCH_PR_NUMBER, and "
-                "CONBENCH_COMMIT must be set if `github` is not specified. "
+                "Both CONBENCH_REPOSITORY and CONBENCH_COMMIT must be set if `github` "
+                "is not specified. CONBENCH_PR_NUMBER should be null for builds on the "
+                "default branch."
                 f"\nValues: `{gh_info}`"
             ),
         )
@@ -154,7 +155,7 @@ def detect_github_info():
     }
 
 
-def machine_info(host_name: Optional[str]):
+def machine_info(host_name: Optional[str] = None):
     os_name, os_version = platform.platform(terse=True).split("-", maxsplit=1)
 
     host_name = host_name or os.environ.get("CONBENCH_HOST_NAME")
--- a/benchadapt/python/benchadapt/_machine_info.py
+++ b/benchadapt/python/benchadapt/_machine_info.py
@@ -78,11 +78,11 @@ class GitParseWarning(RuntimeWarning):
 
 def github_info():
     """Get github metadata from environment variables"""
-    repo = os.environ.get("CONBENCH_REPOSITORY")
-    pr_number = os.environ.get("CONBENCH_PR_NUMBER") or os.environ.get(
+    repo = os.environ.get("CONBENCH_PROJECT_REPOSITORY")
+    pr_number = os.environ.get("CONBENCH_PROJECT_PR_NUMBER") or os.environ.get(
         "BENCHMARKABLE_PR_NUMBER"
     )
-    commit = os.environ.get("CONBENCH_COMMIT")
+    commit = os.environ.get("CONBENCH_PROJECT_COMMIT")
 
     gh_info = {
         "commit": commit,
@@ -94,9 +94,9 @@ def github_info():
     if not (repo and commit):
         warnings.warn(
             (
-                "Both CONBENCH_REPOSITORY and CONBENCH_COMMIT must be set if `github` "
-                "is not specified. CONBENCH_PR_NUMBER should be null for builds on the "
-                "default branch."
+                "Both CONBENCH_PROJECT_REPOSITORY and CONBENCH_PROJECT_COMMIT must be "
+                "set if `github` is not specified. CONBENCH_PROJECT_PR_NUMBER should be "
+                "null for builds on the default branch."
                 f"\nValues: `{gh_info}`"
             ),
         )
@@ -158,7 +158,7 @@ def detect_github_info():
 def machine_info(host_name: Optional[str] = None):
     os_name, os_version = platform.platform(terse=True).split("-", maxsplit=1)
 
-    host_name = host_name or os.environ.get("CONBENCH_HOST_NAME")
+    host_name = host_name or os.environ.get("CONBENCH_MACHINE_INFO_NAME")
     host_name = host_name or platform.node()
 
     info = {
--- a/benchadapt/python/benchadapt/machine_info.py
+++ b/benchadapt/python/benchadapt/machine_info.py
@@ -5,8 +5,6 @@ import subprocess
 import warnings
 from typing import Optional
 
-from benchclients.logging import fatal_and_log
-
 
 def _sysctl(stat):
     return ["sysctl", "-n", stat]
@@ -94,14 +92,14 @@ def github_info():
     if repo and pr_number and commit:
         return gh_info
     else:
-        fatal_and_log(
+        warnings.warn(
             (
                 "All of BENCHMARKABLE_REPOSITORY, BENCHMARKABLE_PR_NUMBER, and "
                 "BENCHMARKABLE_COMMIT must be set if `github` is not specified. "
-                f"Values: {gh_info}"
+                f"\nValues: `{gh_info}`"
             ),
-            etype=ValueError,
         )
+        return gh_info
 
 
 def detect_github_info():
--- a/benchadapt/python/benchadapt/machine_info.py
+++ b/benchadapt/python/benchadapt/machine_info.py
@@ -78,9 +78,11 @@ class GitParseWarning(RuntimeWarning):
 
 def github_info():
     """Get github metadata from environment variables"""
-    repo = os.environ.get("BENCHMARKABLE_REPOSITORY")
-    pr_number = os.environ.get("BENCHMARKABLE_PR_NUMBER")
-    commit = os.environ.get("BENCHMARKABLE_COMMIT")
+    repo = os.environ.get("CONBENCH_REPOSITORY")
+    pr_number = os.environ.get("CONBENCH_PR_NUMBER") or os.environ.get(
+        "BENCHMARKABLE_PR_NUMBER"
+    )
+    commit = os.environ.get("CONBENCH_COMMIT")
 
     gh_info = {
         "commit": commit,
@@ -92,8 +94,8 @@ def github_info():
     if not (repo and pr_number and commit):
         warnings.warn(
             (
-                "All of BENCHMARKABLE_REPOSITORY, BENCHMARKABLE_PR_NUMBER, and "
-                "BENCHMARKABLE_COMMIT must be set if `github` is not specified. "
+                "All of CONBENCH_REPOSITORY, CONBENCH_PR_NUMBER, and "
+                "CONBENCH_COMMIT must be set if `github` is not specified. "
                 f"\nValues: `{gh_info}`"
             ),
         )
--- a/benchadapt/python/benchadapt/machine_info.py
+++ b/benchadapt/python/benchadapt/machine_info.py
@@ -5,6 +5,8 @@ import subprocess
 import warnings
 from typing import Optional
 
+from benchclients.logging import fatal_and_log
+
 
 def _sysctl(stat):
     return ["sysctl", "-n", stat]
@@ -77,6 +79,32 @@ class GitParseWarning(RuntimeWarning):
 
 
 def github_info():
+    """Get github metadata from environment variables"""
+    repo = os.environ.get("BENCHMARKABLE_REPOSITORY")
+    pr_number = os.environ.get("BENCHMARKABLE_PR_NUMBER")
+    commit = os.environ.get("BENCHMARKABLE_COMMIT")
+
+    gh_info = {
+        "commit": commit,
+        "repository": repo,
+        # "branch": None,
+        "pr_number": pr_number,
+    }
+
+    if repo and pr_number and commit:
+        return gh_info
+    else:
+        fatal_and_log(
+            (
+                "All of BENCHMARKABLE_REPOSITORY, BENCHMARKABLE_PR_NUMBER, and "
+                "BENCHMARKABLE_COMMIT must be set if `github` is not specified. "
+                f"Values: {gh_info}"
+            ),
+            etype=ValueError,
+        )
+
+
+def detect_github_info():
     """Attempts to inspect a locally cloned repository for git information that can be
     posted to the "github" key when creating a run.
 
@@ -130,8 +158,8 @@ def github_info():
 def machine_info(host_name: Optional[str]):
     os_name, os_version = platform.platform(terse=True).split("-", maxsplit=1)
 
-    if not host_name:
-        host_name = platform.node()
+    host_name = host_name or os.environ.get("CONBENCH_HOST_NAME")
+    host_name = host_name or platform.node()
 
     info = {
         "name": host_name,
--- a/benchadapt/python/benchadapt/machine_info.py
+++ b/benchadapt/python/benchadapt/machine_info.py
@@ -89,9 +89,7 @@ def github_info():
         "pr_number": pr_number,
     }
 
-    if repo and pr_number and commit:
-        return gh_info
-    else:
+    if not (repo and pr_number and commit):
         warnings.warn(
             (
                 "All of BENCHMARKABLE_REPOSITORY, BENCHMARKABLE_PR_NUMBER, and "
@@ -99,7 +97,8 @@ def github_info():
                 f"\nValues: `{gh_info}`"
             ),
         )
-        return gh_info
+
+    return gh_info
 
 
 def detect_github_info():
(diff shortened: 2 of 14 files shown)


benchadapt/python/benchadapt/machine_info.py
```diff
@@ -77,6 +77,31 @@ class GitParseWarning(RuntimeWarning):
 
 
 def github_info():
+    """Get github metadata from environment variables"""
+    repo = os.environ.get("BENCHMARKABLE_REPOSITORY")
+    pr_number = os.environ.get("BENCHMARKABLE_PR_NUMBER")
+    commit = os.environ.get("BENCHMARKABLE_COMMIT")
```
jonkeane: Would it make sense to prefix these with `CONBENCH_` so that it's clear that they belong to the conbench (family)?

benchadapt/python/benchadapt/machine_info.py
```diff
@@ -77,6 +77,31 @@ class GitParseWarning(RuntimeWarning):
...
+    }
+
+    if not (repo and pr_number and commit):
+        warnings.warn(
+            (
+                "All of BENCHMARKABLE_REPOSITORY, BENCHMARKABLE_PR_NUMBER, and "
+                "BENCHMARKABLE_COMMIT must be set if `github` is not specified. "
+                f"\nValues: `{gh_info}`"
+            ),
+        )
```
jonkeane: Should this be an error or a warning? Or is it something we want as a warning for now while we transition and will become an error? 

The docs below seem to suggest it should be an error

```
        By default, metadata will be obtained from ``BENCHMARKABLE_REPOSITORY``,
        ``BENCHMARKABLE_COMMIT``, and ``BENCHMARKABLE_PR_NUMBER`` environment variables.
        If any are unset, an error will be raised.
```

jonkeane (commented):
This looks good. Only two comments (one is really more clarification, the other on everyone's favorite topic of naming)

benchadapt/python/benchadapt/machine_info.py
alistaire47: I thought (@austin3dickey ?) we were using `BENCHMARKABLE_PR_NUMBER` somewhere already ([here?](https://github.com/voltrondata-labs/arrow-benchmarks-ci/blob/e4de700016e0d473d82199733b8e4dbe89112b88/models/run.py#L115) but that's internal to arrow-benchmarks-ci, if I understand what's happening there correctly) so the `BENCHMARKABLE_*` prefix was for consistency there, but maybe it's fine if we just make it `CONBENCH_*` everywhere? That would be preferable.

jonkeane: Ah I didn't realize we had these already somewhere 😬. So maybe we accept both (for that one, temporarily to make our rollout easier? Or we could set both in the CI for a little bit too? either way is fine...)

But yeah, we probably should have done `CONBENCH_...` from the beginning (oops!). But IMO we aren't _too far_ down the envvarification that catching this now is a good time

benchadapt/python/benchadapt/machine_info.py
alistaire47: Oops, I forgot to update when I changed it from an error to a warning; fixed now. An error would be nice, but that causes workflow problems because the error will get raised before the field can get overridden in adapters, so I changed it to a warning.

benchadapt/python/benchadapt/machine_info.py
alistaire47: Yeah, I'm not sure exactly how for the usage of `BENCHMARKABLE_PR_NUMBER` spread, but yeah, I like the plan of accepting it as a backup if we need, but preferring `CONBENCH_*`. I'll adjust.

alistaire47: changed

jgehrcke: Confirming what I read here, in my own words:

If an environment variable enters the stack primarily to configure an aspect about Conbench, then it should be prefixed with `CONBENCH_`. Yes! Super important.

If we use env vars that don't do this so far: let's slowly change them. Config pipelines become a madness otherwise.

We might do some dirty cheating here and there where an environment variable that has central meaning elsewhere is _also_ read by Conbench in order to pragmatically react to it. In that case, the CONBENCH prefix is not applicable.

benchconnect/README.md
```diff
@@ -18,6 +18,28 @@ With [pipx](https://pypa.github.io/pipx/):
 pipx install benchconnect@git+https://github.com/conbench/conbench.git@main#subdirectory=benchconnect
 ```
 
+## Environment variables
+
+benchconnect relies on a number of environment variables to obtain various
+metadata:
+
+- `CONBENCH_URL`: The URL of the Conbench server. Required.
```
jgehrcke: Let's show an example, and let's maybe comment on trailing slash?

benchconnect/README.md
```diff
@@ -18,6 +18,28 @@ With [pipx](https://pypa.github.io/pipx/):
...
+- `CONBENCH_EMAIL`: The email to use for Conbench login. Only required if the
+server is private.
+- `CONBENCH_PASSWORD`: The password to use for Conbench login. Only required
+if the server is private.
+- `CONBENCH_REPOSITORY`: The URL of the repo being benchmarked. Defaults
+to `"https://github.com/apache/arrow"` if unset.
```
jgehrcke: Is this GitHub-specific, is this expected to start with `https://github.com`?

benchconnect/README.md
```diff
@@ -18,6 +18,28 @@ With [pipx](https://pypa.github.io/pipx/):
...
+- `CONBENCH_PR_NUMBER`: Integer of pull request number associated with
+the run being benchmarked.
+- `CONBENCH_COMMIT`: Hash of commit being benchmarked. If missing, will
+attempt to obtain it from `arrow::arrow_info()$build_info$git_id`, though
+this may not be populated depending on how Arrow was built.
+- `CONBENCH_HOST_NAME`: By default, the running machine host name will be
```
jgehrcke: Let's maybe comment on where/how this is after all used (e.g., "sent as value to the `machine-name` key when POSTing benchmark results" or so)?

benchconnect/README.md
alistaire47: Oops, these docs are slightly wrong; I copied them over from arrowbench, where defaults are arrow-specific. `CONBENCH_REPOSITORY` can be `org/repo` or the full URL; see https://github.com/conbench/conbench/blob/main/conbench/entities/run.py#L287-L293 I'll update this to use those descriptions instead so everything is consistent.

benchadapt/python/benchadapt/machine_info.py
austin3dickey: >I'm not sure exactly how for the usage of BENCHMARKABLE_PR_NUMBER spread

The word "benchmarkable" is primarily an `arrow-benchmarks-ci` term that means "the project being benchmarked." Notably, it doesn't exist for other projects like Velox. So I think you'll only find it in https://github.com/voltrondata-labs/arrow-benchmarks-ci and the projects it manages, which right now is just https://github.com/voltrondata-labs/benchmarks.

Agreed; I like using the `CONBENCH_` prefix for these. But I also think these could be a little misleading:
- CONBENCH_PR_NUMBER
- CONBENCH_REPOSITORY
- CONBENCH_COMMIT
- CONBENCH_HOST_NAME

since it feels weird to put an Arrow PR number in a variable called "conbench PR number", or the Arrow repository name in "conbench repository". I also might accidentally put something like https://conbench.ursa.dev/ in CONBENCH_HOST_NAME, confusing it with CONBENCH_URL.

So maybe these should be
- CONBENCH_PROJECT_PR_NUMBER
- CONBENCH_PROJECT_REPOSITORY
- CONBENCH_PROJECT_COMMIT
- CONBENCH_MACHINE_NAME
?

benchadapt/python/benchadapt/machine_info.py
```diff
@@ -77,6 +77,33 @@ class GitParseWarning(RuntimeWarning):
...
+    }
+
+    if not (repo and pr_number and commit):
+        warnings.warn(
+            (
+                "All of CONBENCH_REPOSITORY, CONBENCH_PR_NUMBER, and "
+                "CONBENCH_COMMIT must be set if `github` is not specified. "
+                f"\nValues: `{gh_info}`"
+            ),
+        )
```
austin3dickey, line 102: I also sorta prefer an error here before we post a messy `{"commit": None, "repository": None, "pr_number": None}` to Conbench. But I understand the "adapter override" behavior makes this hard. Can we error out somewhere else?

austin3dickey: Oh and another thing: `CONBENCH_PR_NUMBER` is guaranteed to be null on default-branch runs. So we should probably remove that from this check

austin3dickey (commented):
looking great!

benchadapt/python/benchadapt/machine_info.py
alistaire47: Yeah, I appreciate these names are a little ambiguous; they're currently based on a `CONBENCH_` prefix and the key of the element in the request JSON schema—mostly. `pr_number`, `repository`, and `commit` are all a part of `github`, but `CONBENCH_GITHUB_REPOSITORY` is certainly no clearer. `CONBENCH_PROJECT_GITHUB_REPOSITORY`? `CONBENCH_PROJECT_REPOSITORY`? Idk.

Also `CONBENCH_HOST_NAME` really ends up as `machine_info.name`, but I'm not sure if `CONBENCH_MACHINE_INFO_NAME` is clearer that this is the name of a computer. `CONBENCH_MACHINE_INFO_HOST_NAME` implies that it's in `machine_info.host_name` though, which it isn't.

tl;dr: naming is still hard 😭

benchadapt/python/benchadapt/machine_info.py
alistaire47: > I also sorta prefer an error here before we post a messy `{"commit": None, "repository": None, "pr_number": None}` to Conbench. But I understand the "adapter override" behavior makes this hard. Can we error out somewhere else?

Because of that we really can't on object creation. We do have [some similar warnings on the `.to_publishable_dict()` method](https://github.com/conbench/conbench/blob/main/benchadapt/python/benchadapt/result.py#L160-L168). We could move this warning there, though here is better because it will also get raised in other places `machine_info()` gets called. We _might_ be able to escalate the warning to an error there? That would cause augmentation in benchconnect to error out if they're not set, which is probably desirable. I'll try it and see what I break.

alistaire47: Ok, made `.to_publishable_dict()` error on run and result if commit and repo aren't specified and fixed what broke. I'm now wondering if the other warnings there could be errors too; I think I tried and then relaxed at some point, but I forget why

benchadapt/python/benchadapt/machine_info.py
alistaire47: Ok, changed to

- CONBENCH_PROJECT_PR_NUMBER
- CONBENCH_PROJECT_REPOSITORY
- CONBENCH_PROJECT_COMMIT
- CONBENCH_MACHINE_INFO_NAME

Don't love them, but they're a balance between specificity and concision.

benchconnect/README.md
```diff
@@ -18,6 +18,28 @@ With [pipx](https://pypa.github.io/pipx/):
...
+benchconnect relies on a number of environment variables to obtain various
+metadata:
+
+- `CONBENCH_URL`: Required. The URL of the Conbench server without a trailing
+slash, e.g. `https://conbench.example.com`
+- `CONBENCH_EMAIL`: The email to use for Conbench login. Only required if the
+server is private.
+- `CONBENCH_PASSWORD`: The password to use for Conbench login. Only required
+if the server is private.
+- `CONBENCH_PROJECT_REPOSITORY`: The repository name (in the format `org/repo`) or the
```
jgehrcke, line 32: I was asking before about GitHub-specificity because that's really a question that I think users might have here, and the answer to that question helps with naming.

Must this be a repository that exists in GitHub? 

What is the implication of setting `org/repo` that does not exist in GitHub?

If we want to keep these parameters GitHub-specific, then I believe it makes a lot of sense to pull `GITHUB` into the names. If this discussion is already too annoying and long by now feel free to ignore :).

austin3dickey: Right now, there is a baked-in assumption everywhere in this repo that any project tracked by Conbench exists in GitHub. If it doesn't exist in GitHub, pretty much any feature using commit history is turned off, which reduces the usefulness of Conbench. 

It would probably take a lot of work to adjust Conbench code to accommodate for repositories hosted elsewhere. However, if we did, we'd probably still need the concepts of `CONBENCH_PROJECT_REPOSITORY` etc, and it would be a breaking change to remove the `GITHUB_` string out of these env vars. So I'm okay with leaving that string out in the first place.

benchadapt/python/benchadapt/result.py
```diff
@@ -167,6 +172,16 @@ def to_publishable_dict(self):
                 "Result not publishable! `stats` and/or `error` must be be specified"
             )
 
+        if not (
+            res_dict["github"].get("repository") and res_dict["github"].get("commit")
+        ):
+            raise ValueError(
```
austin3dickey: This one on BenchmarkResult _might_ be a little too strict, for the workflow where someone first creates a Run manually and then wants to post results to it without worrying about the `github` object (since `github` is ignored on results if the `run_id` already exists). But I'm okay leaving this as-is and handling that issue later if it pops up.

benchconnect/README.md
```diff
@@ -18,6 +18,28 @@ With [pipx](https://pypa.github.io/pipx/):
...
+URL (in the format `https://github.com/org/repo`)
+- `CONBENCH_PROJECT_PR_NUMBER`: [recommended] The number of the GitHub pull request that
+is running this benchmark, or ``None`` if for a run on the default branch
```
austin3dickey: nit: maybe instead of None here (which could result in `CONBENCH_PROJECT_PR_NUMBER="None"`) say something like "do not supply this if it's for a run on the default branch"

austin3dickey (approved):
Looks great!

benchadapt/python/benchadapt/result.py
alistaire47: Yeah, that's true. We're already sort of moving in this direction—before we could post runs directly, all the run metadata had to be on the result. Now that we've normalized a bit, it would make sense to just take `run_id` in cases where the run is sent first, but we've still got a ton of legacy code that isn't opening runs first that still needs the duplicated fields. But as all that gradually gets migrated, we should at least think about deprecating `run_name`, `run_reason`, `github`, etc.

For now, I don't really mind this particular duplication, because if users are using env vars to populate this on the run, it will get picked up here without them doing anything, and benchconnect checks that they don't disagree (I believe missing is fine there). But we could reduce this one to a warning and keep the run one an error while we're here anyway.

alistaire47: changed

benchconnect/README.md
jonkeane: > It would probably take a lot of work to adjust Conbench code to accommodate for repositories hosted elsewhere.

Though we don't need to support this right now (so we should continue to defer it until it's actually necessary), I'm not sure this would be _a lot_ of work to adjust. Ideally the connections into github in the backend are isolated enough that it would be "just" writing new methods to connect to the gitlab or bitbucket or wherever APIs and plop the git relations into the database all the same, and then some `s/"github.com"/config.git_host_setting/g` for URLs. Of course that's not trivial, but it shouldn't be a lot of work (and if it is, that might be a sign that our factorization of our current github integration isn't ideal).

jonkeane: Not anything we need to change now (and certainly not in this PR), but something to think about as we continue to maintain this codebase

benchadapt/python/benchadapt/_machine_info.py
```diff
@@ -77,6 +77,34 @@ class GitParseWarning(RuntimeWarning):
...
+        # "branch": None,
+        "pr_number": pr_number,
+    }
+
+    if not (repo and commit):
+        warnings.warn(
+            (
+                "Both CONBENCH_PROJECT_REPOSITORY and CONBENCH_PROJECT_COMMIT must be "
+                "set if `github` is not specified. CONBENCH_PROJECT_PR_NUMBER should be "
+                "null for builds on the default branch."
```
jonkeane: Very minor, and very pedantic: when you say "CONBENCH_PROJECT_PR_NUMBER should be null" does that mean unset or `""`?

jonkeane (approved):
This is looking good! I'm excited to see this merged

benchadapt/python/benchadapt/_machine_info.py
alistaire47: Digging through, I'm fairly sure that because [this line](https://github.com/conbench/conbench/blob/0efd2b9132c73034572e13ea74527b24ecb12288/conbench/entities/commit.py#L499) and both `not None` and `not ""` will return `True`, it doesn't matter. But I'll change this to recommend not setting it so nobody tries quoted `"null"` or anything else silly

Issue #588, opened 2023-01-11
history endpoint: expose raw samples


jgehrcke:
Example entry:
```
conbench-app-1  |   {
conbench-app-1  |     "benchmark_id": "02fdb19e506e4de7b7fb152a89e6466f",
conbench-app-1  |     "case_id": "8069d1a31bd844269195ba6dd5304f09",
conbench-app-1  |     "context_id": "c6f559f8eeec4b3cad107d9a2b805d5a",
conbench-app-1  |     "distribution_mean": 19.987086157760995,
conbench-app-1  |     "distribution_stdev": 0.38970190762365126,
conbench-app-1  |     "hardware_hash": "machine1-0-16-16-64424509440",
conbench-app-1  |     "is_step_change": false,
conbench-app-1  |     "mean": 20.049251096779297,
conbench-app-1  |     "message": "ARROW-16728: [Python] Switch default and deprecate use_legacy_dataset=True in ParquetDataset (#14052)",
conbench-app-1  |     "repository": "https://github.com/apache/arrow",
conbench-app-1  |     "run_name": "commit: 5a9805807456fa1b50671afded557044ab6cc8e6",
conbench-app-1  |     "sha": "5a9805807456fa1b50671afded557044ab6cc8e6",
conbench-app-1  |     "timestamp": "2022-12-23T21:25:10",
conbench-app-1  |     "unit": "s"
conbench-app-1  |   },
```
That `mean` is probably derived directly from the raw samples in this run, but those are not exposed. Exposing them is valuable for doing custom analysis on data returned by the history endpoint.

I think exposing raw samples might trigger worries about size/performance topics, but I'd say let's not be held back by those and tackle challenges as they come.

jgehrcke:
I think we can consider this as 'done' after having merged https://github.com/conbench/conbench/pull/591.

Issue #571, opened 2023-01-05
history plot: show raw data points


jgehrcke:
One wants to see where the individual data points are along this 'history' line:

![image](https://user-images.githubusercontent.com/265630/210776250-d6db2822-1ab5-4c27-a198-8876399a37d8.png)

Line + scatter is usually more informative.

Other ideas, while I have them on my mind (independent tickets, ideally):
- add right padding so that the newest data point does not collide with right boundary of plot.
- rework legend to explain better how mean and sigma were built.
- add more date ticks

jonkeane:
Love this. I also think it might be nice to fill in the +/- 5 sd area to make it a little more obvious that that is a range we expect the other values to land (again, this would ideally be a separate ticket, but while we're collecting improvement ideas)

jgehrcke:
https://github.com/bokeh/bokeh/pull/2186

jgehrcke:
https://github.com/conbench/conbench/pull/591 implements some ideas. Specifically:

> Line + scatter is usually more informative.

done

> add right padding so that the newest data point does not collide with right boundary of plot.

done

> rework legend to explain better how mean and sigma were built.

done (important improvement, but still ambiguity left!)

> add more date ticks

done

> fill in the +/- 5 sd area to make it a little more obvious that that is a range we expect the other values to land

not yet done

jgehrcke:
This is done.

Issue #579, opened 2023-01-06
Macrobenchmarks versus microbenchmarks


jonkeane:
This has come up in a few places recently and I think having a dedicated space to talk about these issues might help. I'm not sure there's much of a to-do here (yet, possibly ever) but this is a good place to have a dsicussion.

Conbench is designed to work with both micro- and macro-benchmarks. Internally we treat this incredibly similarly (though our runners sometimes treat them quick differently)[^1]. 

So far our distinction is as follows (note, that many of these are in reference to the Arrow benchmarks, but are applicable to other implementations, but of course each setup is unique)

| Micro benchmarks | Macro benchmarks |
|---|---|
| Measures a small (ideally tiny!) piece of code |  Tends to be a larger chunk of code (or at least the code touches more layers of our stack) | 
| Uses relatively small inputs | Users relatively large inputs  | 
| Iterations are very quick (mili or nano seconds) | Iterations can be much longer (seconds, minutes) | 
| Typically measures that code a massive number of times | Typically measures that code a smaller number of times | 
| Generally reports aggregated data (though, note that we do not currently pull all the aggregate details into conbench yet!) | Reports each individual iteration | 
| Is very good for targeted developer attention (e.g. "this benchmark got 50% slower, it's touching this specific code, figure out what happened") | Is very good for top-line user-facing numbers "you can ingest this parquet file 2x faster on this common dataset than it was 2 years ago!" | 
| Can and should be very tuned to the code being executed | (sometimes) purposefully designed to be a generic external task that can be compared across implementations (TPC benchmarks, other industry standards) |
| Tends to be in lower-level languages (e.g. C++) | Tends to be in higher-level languages (e.g. Python, R) | 

But @jgehrcke has mentioned a few times (very rightfully! [one example](https://github.com/voltrondata-labs/benchmarks/issues/123#issuecomment-1372215463)) that there can be a tension in the macrobenchmarks between being helpful for driving development (targeted, not impacted by "uninteresting" things like disk speed, easily repeatable many times) and some of the properties of the macro benchmarks above (and most specifically, the external-validity that makes macro benchmarks good for user-facing questions of performance. 

[...]

Issue #530, opened 2022-12-12
plotting and analysis: derive per-sample uncertainty from N repetitions, show error bars / distr / box plot / violine plot


jgehrcke:
Taking into account per-sample uncertainty is in my opinion table stakes for visual and automatic performance regression detection. 

Each (duration) sample is subject to measurement uncertainty.

We should attempt (and find ways to require/encourage users) to estimate the per-sample uncertainty.

Given a per-sample uncertainty estimate we should then visualize it in corresponding plots (via error bars or violin plot, or comparable) and use it in corresponding analyses.

A plot showing samples without error bars can often be rather misleading -- a plot with meaningful error bars on the other hand allows for drawing stronger conclusions.

How to get there? Maybe for serious workflows it should not be allowed to submit runs with `iterations = 1`. Maybe Conbench should encourage its API users to not submit single duration values, but to always submit N duration values that come from repeating the _same thing_ N times (under *seemingly constant conditions*). A low-ish meaningful N would be 6 (as a rule of thumb, from six repetitions onwards a statistical analysis starts to become sound-ish).

#### Thoughts about about random vs. systematic errors

A systematic/methodic error can often not be properly represented until it's _known_. Once it's known, it can hopefully be fixed by changing measurement method.

However, random/non-deterministic measurement error can and should be quantified along the way for each sample. In practice, that means to repeat each experiment N times under *seemingly constant conditions*, to look at the distribution of outcomes, and to then build sample _plus_ error information for plotting and further analysis.

This can be looked at as
- doing per-sample quality assurance.
- measuring the _current_ volatility very local to the experiment

#### Thoughts about working with N > 1: min, max, mean?

In the context of computing it often makes sense to pick the 'best case' scenario instead of the mean. One could then do asymmetric error bars, basically visualizing spread between min and max.

#### Notes about current plotting:

![image](https://user-images.githubusercontent.com/265630/207088632-0c2c18d5-cfe4-4630-871b-a4029492004b.png)

That plot can be improved in terms of legend, and also in terms of showing raw samples plus uncertainty. Some thoughts:

- does not clearly reveal where individual samples are (only-line plots have this disadvantage)
- does not try to reveal per-sample uncertainty
- shows a mean value curve, but does not clarify from which values -- mean is constant over time, i.e. this is the mean over _all_ samples in history?
[...]

PR #591, opened 2023-01-11
history plot: add min-over-time, show raw mean&min, tweak size&spacing

jgehrcke:
I have done a 'quick' round of what I think are valuable changes to the history plot. Some ideas were expressed here: https://github.com/conbench/conbench/issues/571

I did this work using simulated benchmark data in a local deployment of Conbench.

Summary:
- Tweaks to dimensions and spacing. Add right padding so that the newest data point does not collide with right boundary of plot. Add left padding so that the legend has more space.
- Legend: renamed "History" to "individual benchmark (mean)". I needed to spend quite a bit of time in the code to actually learn what "History" was.
- Legend: added the 'rolling window' term which is still under-specified, but better than before IMO. (is this mean of means? how wide is the rolling window? at which part of the window is the window aggregate plotted?)
- Added scatter for mean-over-time, removed line.
- Added a min-over-time plot (related: https://github.com/conbench/conbench/issues/594).
- Adjusted style so that mean-over-time is a little less prominent than min-over-time.
- Made clear that the red dot for the current benchmark represented the mean. Added another circle for the min.
- As the plotting architecture uses the history API endpoint under the hood, I added raw samples to that. That would tick off https://github.com/conbench/conbench/issues/588 if we agree we want that.

**Before:**

Screenshot from 2023-01-11 15-19-12

**After:**
Screenshot from 2023-01-11 15-16-29


jgehrcke:
For the record, https://github.com/conbench/conbench/pull/592 shows how I chose to synthetically generate quasi-realistic benchmark data (which is not realistic yet at all, but it's a start).

conbench/entities/history.py
```diff
@@ -12,16 +12,36 @@
...
+        # `data` was representing durations then this vector was not duplicated
+        # as `times`.
+
         return {
             "benchmark_id": history.id,
             "case_id": history.case_id,
             "context_id": history.context_id,
             "mean": float(history.mean),
+            "data": [float(d) if d is not None else None for d in history.data],
+            "times": [float(t) if t is not None else None for t in history.times],
```
jgehrcke: @austin3dickey I'd love to have your careful analysis on these two lines: do we think that they can break in view of existing data? As you can see, I rely on values being either floaty or None.

jgehrcke:
Important note: this relates strongly to the discussion in https://github.com/conbench/conbench/issues/530. The way I have thought about this tweaked plot is by **assuming that min and mean are different**, because each underlying benchmark is based on N>1 samples. For N=1, mean and min become the same (and more importantly: they become weird names for the thing that they really are).

conbench/app/_plots.py
```diff
@@ -186,72 +198,168 @@ def _source(
...
-        }
-    ]
     with_dist = [h for h in history if h["distribution_mean"]]
     formatted, axis_unit = _should_format(history, unit)
 
-    source = _source(history, unit, formatted=formatted)
-    source_x = _source(current, unit, formatted=formatted)
+    # Note(JP): `history` is an ordered list of dicts, each dict has a `mean`
+    # key which is extracted here by default.
+    source_mean_over_time = _source(history, unit, formatted=formatted)
```
austin3dickey, line 227: This is pedantic but shouldn't this variable name be switched with the variable name `source_mean`? To me, `source_mean` would imply the mean of each benchmark (i.e.`distribution_mean=False`) and `source_mean_over_time` would imply the rolling mean (i.e. `distribution_mean=True`)

conbench/app/_plots.py
```diff
@@ -283,6 +391,8 @@ def time_series_plot(history, benchmark, run, height=250, width=1000):
         )
         spacer.outline_line_color = None
         spacers.append(spacer)
-        print(f"Ignore any MISSING_RENDERERS warning for id={spacer.id}")
+        # Note(JP): I cannot make sense of this warning. I think the printed
+        # text is not acted on by anyone. Outcommeting for now.
+        # print(f"Ignore any MISSING_RENDERERS warning for id={spacer.id}")
```
austin3dickey: I put this print line in originally because there's an expected bokeh MISSING_RENDERERS warning that's logged every time these empty spacer objects are rendered, and I didn't want it to be confused with any true warnings that might be thrown if a plot doesn't render in the webapp. The true way to fix it would be to either 1. figure out a better way to add a spacer, or 2) add a filter to the `bokeh` logger that filters out specific warning messages.

image

conbench/entities/history.py
austin3dickey: Yes, I believe the column itself is nullable, so you might need to handle the case where `history.data` is None instead of a list, too. Maybe just treat it as an empty list in that case.

Also, if you'd like you can use `from ..entities._entity import to_float`.

austin3dickey (commented):
This is looking really nice! I love the improvements.

austin3dickey:
>For N=1, mean and min become the same

This is of course true, and this is a common trait of macrobenchmarks (see #579). Even in this case, the legend might be a little misleading, but I think the plot will look nice. It will just plot the min and mean scatters on top of each other.

boshek:
❤️ - this is gorgeous. The margin tweaks alone are worth the change. A couple thoughts which are hopefully useful:

> For N=1, mean and min become the same

Would it be possible to detect N here and visualize appropriately? That is, if N=1 then don't show any summary statistics. If N>1 (or maybe 2) then present the plot as you have. 

- I wonder if you would consider replacing the term `individual benchmarks (*)` with something like `benchmark *`. AFAIK we don't use the term "individual" elsewhere so it introduces some new parlance. If someone has clicked this far IMO, they already know what "benchmark" refers to.  
- Following on from that, I'm not sure we need coloured dots for the "current" demarkation. My suggestion would either be a dot with a different shape _or_ maybe a vertical line mark the "current" date. Whatever the eventual solution, I think it would be worth replacing that red as a colour as I think that would be different for some colour blindness to distinguish
- `individual benchmark (mean)` and `rolling window mean +/- 5 σ` are the same colour. Is that to connect the fact that they are both means? It might be nice to distinguish those by colour. 
- i like the idea of emphasizing the min. Nice :)

conbench/app/_plots.py
jgehrcke: Ha : ) Yeah, I can see that.

I did not want to re-write everything here. I found `source_x` particularly un-expressive, and felt the strong inclination to change the variable name after having understood what it means. You see this inconsistency because I did not rename `source_mean` (but it also should be renamed for clarity, would maybe call it `source_rollingmean_over_time`).

In my mental naming model `source_mean_over_time` means  "this is the bokeh source object for a mean-over-time plot (where mean is on the high/y axis and time is on the right/x axis)".

I will leave this for now. Curious: when I say "let's plot mean over time", what would you draw? :)

Personally, I try to always say "rolling window ..." when there's a rolling window analysis involved, see for example the legends in the screenshot in https://github.com/jgehrcke/ci-analysis or in https://raw.githubusercontent.com/jgehrcke/covid-19-germany-gae/master/plots/daily-change-plot-latest.png.

conbench/app/_plots.py
jgehrcke: Thank you for clarifying!

jgehrcke:
Thank you @boshek!

> I wonder if you would consider replacing the term individual benchmarks (*) with something like benchmark *. AFAIK we don't use the term "individual" elsewhere so it introduces some new parlance. If someone has clicked this far IMO, they already know what "benchmark" refers to.

Yeah, that's fair!

> Following on from that, I'm not sure we need coloured dots for the "current" demarkation. [...]  think it would be worth replacing that red as a colour as I think that would be different for some colour blindness to distinguish

Love the feedback. The red color was there previously, but I also like those principles more where color is used very carefully and purposefully. Will update this.

> individual benchmark (mean) and rolling window mean +/- 5 σ are the same colour. Is that to connect the fact that they are both means? It might be nice to distinguish those by colour.

same color yes, but different plot styles (line vs. scatter). will think about this one :).

> i like the idea of emphasizing the min. Nice :)

:muscle:

conbench/entities/history.py
jgehrcke: done

conbench/app/_plots.py
jgehrcke: pushing a commit with more renames, consistent with the style I described above

conbench/app/_plots.py
jgehrcke: adding a commit that uses https://docs.bokeh.org/en/latest/docs/reference/layouts.html#spacer (removes original warning)

jgehrcke:
Pushed commits in response to great feedback. New screenshot:
image

jgehrcke:
Merging, let's discuss and iterate based on this. There are many many ideas about what can be tweaked about this presentation next. I am also super open to re-considering what was changed in this PR. Would love to see it in action in staging now.

conbench/app/_plots.py
austin3dickey: Nice, this Spacer object looks like a great solution.

alistaire47:
One thing to keep an eye on coming out of this PR is plot performance due to data size. AFAIK bokeh is a lot better than altair/vegalite (not sure if it uses webgl, which tends to help?), but it will still get slow at some point. We don't limit any history here, so for old benchmarks that run a lot of iterations on busy repos, we could potentially see some rendering speed issues. It may be worth a bit of experimentation to see how much data it takes to make things unpleasantly slow so we can consider whether we might want to truncate data at a certain size.

alistaire47:
Small thought: for rolling metrics, I find it helpful to see how long the window is, so I prefer a label like `7d rolling mean` to the more generic `rolling window mean`

jgehrcke:
> so I prefer a label like 7d rolling mean to the more generic rolling window mean

Yes, me too. However, in our case, we do not have a fixed time width. Instead, we have some other width that I still find hard to explain to users because I don't have the confidence yet to say I 100 % understand how current code behaves. I have also not yet seen a documentation that really explained to me concisely and precisely how the rolling window analysis is performed (maybe it exists! :)).

I added a code comment [here](https://github.com/conbench/conbench/blob/0c9adbc8bc19e9aede6d6ffe2f9492e671fbcd07/conbench/app/_plots.py#L299):
```
    # Note(JP). The `source_rolling_*` data is based on the "distribution"
    # analysis in conbench which I believe is a rolling window analysis where
    # the time width of the window is variable, as of a fixed commit-count
    # width.
```

A rolling window size of fixed observation count would make sense (e.g. containing N samples / samplegroups). Setting the commit count, however, is not easy for me to interpret. This indirection raises questions like "does each commit have at least one benchmark associated"?

jgehrcke:
> keep an eye on coming out of this PR is plot performance due to data size

Yeah, fair point! I want to say that this PR only changes a constant factor to the number of samples being shown, and that factor is 2 or less I think. It does not introduce a new quality/scaling characteristic, in a rough order of magnitude estimation this is a noop.

I assume we are orders of magnitude away from reaching such a point, but I have not tested anything, and we'll see. I have used so many different plotting solutions over the years and was typically surprised by how late these problems came. I typically wait for such a hard/soft wall to pop up, so that one can then do very targeted performance analysis and build a low effort high impact fix. With some good resampling.

Btw, I really like this kind of concern from a 'tidiness' point of view. Of course a great plot does not need to be built from a billion data points. That's rarely a great idea, indeed. Oh, and I don't even think that we need particularly interactive plots. Often, the ability to pan and zoom sets bad incentives, I find.

austin3dickey:
Re: the rolling window discussion, for what it's worth, I intend to clarify this when addressing #473 next. :) I think we'll end up using an "expanding window" instead of a true "rolling window" which is easier to explain anyway.

PR #592, opened 2023-01-11
populate_local_conbench: simulate benchmark samples via semi-realistic distribution

jgehrcke:
Here: sample from normal distribution with aritificial lower bound left to the mean.
Add two different mechanisms for 'rare' instabilities.
This is an extremely rude approximation of reality.
This is not a PR I want to land, but this is just to show how I generated the data used for the screenshots in https://github.com/conbench/conbench/pull/591

### DIFF
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -12,6 +12,7 @@ from sqlalchemy.orm import Query
 
 from conbench import util
 
+from ..config import Config
 from ..db import Session
 from ..entities._entity import (
     Base,
@@ -326,13 +327,12 @@ def backfill_default_branch_commits(repository: str, new_commit: Commit) -> None
 
     if last_tracked_commit:
         since = last_tracked_commit[0].timestamp
-    elif os.getenv("DB_HOST") == "localhost":
+    elif Config.TESTING:
         log.info(
-            "Your DB_HOST is localhost, so assuming you're running "
-            "conbench/tests/populate_local_conbench.py. Backfilling the DB only from "
-            "2022-01-01 in order to save time."
+            "Config.TESTING is true Backfilling the DB only from "
+            "2022-05-01 in order to save time."
         )
-        since = datetime(2022, 1, 1)
+        since = datetime(2022, 8, 1)
     else:
         since = datetime(1970, 1, 1)
 
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -2,7 +2,7 @@ import functools
 import json
 import logging
 import os
-from datetime import datetime
+from datetime import datetime, timedelta
 from typing import Dict, List, Optional
 
 import flask as f
@@ -330,9 +330,9 @@ def backfill_default_branch_commits(repository: str, new_commit: Commit) -> None
     elif Config.TESTING:
         log.info(
             "Config.TESTING is true Backfilling the DB only from "
-            "2022-05-01 in order to save time."
+            "the last 60 days in order to save time."
         )
-        since = datetime(2022, 8, 1)
+        since = datetime.today() - timedelta(days=60)
     else:
         since = datetime(1970, 1, 1)
 
--- a/conbench/tests/populate_local_conbench.py
+++ b/conbench/tests/populate_local_conbench.py
@@ -1,6 +1,8 @@
 import datetime
 import logging
 import os
+import random
+import statistics
 import time
 import uuid
 
@@ -25,10 +27,10 @@ session = requests.Session()
 def main():
     register()
     login()
+    log.info("start generate_synthetic_benchmark_history()")
+    generate_synthetic_benchmark_history()
     log.info("start create_benchmarks_data()")
     create_benchmarks_data()
-    log.info("start create_benchmarks_with_history()")
-    create_benchmarks_with_history()
 
 
 def generate_benchmarks_data(
@@ -254,6 +256,9 @@ def post_benchmarks(data):
         f"Received status code {res.status_code}. "
         f"Took {attempt} attempt(s). Last attempt took {time.monotonic() - t0 :.5f} s."
     )
+
+    if str(res.status_code).startswith("4"):
+        log.info("4xx response body: %s", res.text)
     return benchmark_id
 
 
@@ -339,7 +344,7 @@ def create_benchmarks_data():
     update_run_with_info(run_id, timestamp)
 
 
-def create_benchmarks_with_history():
+def create_be():
     # 7 commits in a row in apache/arrow, the commented one is missing
     commits = [
         "17d6fdc0e9c00534e4de7bfb193c33c86cab7e15",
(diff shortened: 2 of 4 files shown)


austin3dickey:
I'm not opposed to merging something like this. The better our dummy data in `populate_local_conbench.py`, the better we can make local changes. We should just make sure we're not losing anything by commenting out the other two `create_*()` lines.

conbench/tests/populate_local_conbench.py
```diff
@@ -254,6 +256,9 @@ def post_benchmarks(data):
         f"Received status code {res.status_code}. "
         f"Took {attempt} attempt(s). Last attempt took {time.monotonic() - t0 :.5f} s."
     )
+
+    if str(res.status_code).startswith("4"):
+        log.info("4xx response body: %s", res.text)
```
jonkeane, line 274: ❤️ Oooph I've wanted this before! Thank you

conbench/tests/populate_local_conbench.py
```diff
@@ -407,5 +412,143 @@ def create_benchmarks_with_history():
...
+    benchmark_name = "dummy-bench"
+
+    distr_mean = 20.0
+    # This is to simulate the real-world effect of there being a theoretical
+    # optimum: the fastest benchmark duration time with all sources of noise
+    # being silenced.
+    lower_bound = 17.5
+    slowdown_offset = distr_mean * 0.1
+    slowdown_lin = distr_mean * 0.1
+    distribution = statistics.NormalDist(mu=20.0, sigma=2)
```
jonkeane, line 441: I really love using simulated data. And I don't think we are _just yet_ using the shape of this data to test our techniques, so this comment is slightly less relevant just now but will be soon.

Could we make this generator function generate some data and save that as a fixture (we could also set a seed, I believe which would effectively be the same thing, we would get the same "random" numbers each time)? Even with pretty liberal z-scores, in principle there is a (very small!) chance that an observation is outside our band of variability when generated this way. Of course, we could re-try those tests and we would see that it's back to normal, but depending on what we are changing, the failure might start a wild goose chase! With a fixture (or seed), the tests outcomes themselves are deterministic, but we can re-shuffle the data if we want to or need to easily cause we have the generator helper

jgehrcke: Great point with respect to reproducibility in a test suite. Will certainly be of relevance. I agree that something like a regression detection method is best tested against specific data sets. Testing against certain distributions that become re-sampled upon every test run sounds pretty cool too, though (_if_ it leads to a stable outcome, it says a lot about the method in use).

jgehrcke:
I have briefly cleaned this up, and we can now land this to have better history plot screenshots created by CI! Also motivated by https://github.com/conbench/conbench/pull/611#discussion_r1082866456.

This increases the time it takes to do `make db-populate` because of the GitHub API interaction. On my machine:
```
$ time make db-populate
...
real	2m29.229s
```

On HEAD of main:
```
$ time make db-populate
...
real	0m59.869s
```

At first I wanted to address this, but it's good to keep that work independent. This will be measurable win from improving architecture around fetching data about commits.

jgehrcke:
I have reviewed the CI-generated history plot screenshot and it shows an error. This needs debugging and fixing before we can land this.

jgehrcke:
Yeah, this seems to result in many GitHub API requests, hitting the quota in CI:
```
conbench-app-1  | [230126-16:43:46.393] [9] [conbench.entities.commit] INFO: got unexpected HTTP response with code 403: {"message":"API rate limit exceeded for installation ID 19416693.","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
conbench-app-1  | [230126-16:43:46.393] [9] [conbench.entities.run] INFO: treat as unknown commit: error during get_github_commit(): 403 Client Error: Forbidden for url: https://api.github.com/repos/apache/arrow/commits/e26a88fd9fecca876073c5a18cf571c0daca6b3b
```

conbench/tests/populate_local_conbench.py
```diff
@@ -25,10 +27,12 @@
...
     register()
     login()
+    log.info("start generate_synthetic_benchmark_history()")
+    generate_synthetic_benchmark_history()
     log.info("start create_benchmarks_data()")
     create_benchmarks_data()
-    log.info("start create_benchmarks_with_history()")
-    create_benchmarks_with_history()
+    log.info("start create_benchmarks_data()")
+    create_benchmars_data_with_history()
```
austin3dickey: nit: typo in `benchmars`

conbench/tests/populate_local_conbench.py
```diff
@@ -398,14 +417,155 @@ def create_benchmarks_with_history():
...
+
+    commits = [c.strip() for c in reversed(ARROW_COMMIT_HASH_LINES_50.splitlines())]
+    benchmark_name = "dummy-bench"
+
+    distr_mean = 20.0
+    # This is to simulate the real-world effect of there being a theoretical
+    # optimum: the fastest benchmark duration time with all sources of noise
+    # being silenced.
+    lower_bound = 17.5
+    slowdown_offset = distr_mean * 0.1
```
austin3dickey, line 439: Another slowdown idea: Could we slow down by a lot more, for a short period of time? To test the z-score code I'm working on, I've been using this synthetic data but essentially doubling the results for indexes 30-40. That's given me a great distribution-change to work with.

conbench/tests/populate_local_conbench.py
```diff
@@ -398,14 +417,155 @@ def create_benchmarks_with_history():
...
 
-                benchmark_id = post_benchmarks(benchmark_data)
+                benchmark_id = post_benchmark_result(benchmark_data)
                 benchmark_ids.append(benchmark_id)
                 runs.append((run_id, timestamp))
 
+
+def generate_synthetic_benchmark_history():
+
+    commits = [c.strip() for c in reversed(ARROW_COMMIT_HASH_LINES_50.splitlines())]
```
austin3dickey: Can you remove the empty strings out of this list? I think a few are showing up, because of how the string is formatted

conbench/tests/populate_local_conbench.py
```diff
@@ -398,14 +417,155 @@ def create_benchmarks_with_history():
...
+    slowdown_offset = distr_mean * 0.1
+    slowdown_lin = distr_mean * 0.1
+    distribution = statistics.NormalDist(mu=20.0, sigma=2)
+
+    benchmark_ids = []
+
+    def sample_slowdown(s):
+        return s + slowdown_offset + slowdown_lin * random.random()
+
+    for idx, commit_hash in enumerate(commits, 1):
```
austin3dickey, line 453: TIL the second argument to `enumerate()`!
>returns a tuple containing a count (from `start` which defaults to 0)

conbench/tests/populate_local_conbench.py
```diff
@@ -398,14 +417,155 @@ def create_benchmarks_with_history():
...
+
+        benchmark_ids.append(post_benchmark_result(bdata))
+
     log.info("now, emit a benchmark ID on stdout")
     # That benchmark ID is consumed and used by CI.
     print(benchmark_ids[0])
 
 
+"""
+Every 20th commit in the apache/arrow repository, walking the commit tree
```
austin3dickey: Instead of every 20th commit, if you do consecutive commits it should hit the GitHub API a lot less often, because it won't need to fill in intermediate commits every time.

austin3dickey (approved):
Looking good!

conbench/tests/populate_local_conbench.py
jgehrcke: Thanks :)

But Mars is also cool! :D

conbench/tests/populate_local_conbench.py
jgehrcke: But I wanted to have a wider date range on the history plot! :-) 

For people reading along: when we depict an individual benchmark result on the history plot as of today then the position on the time axis is chosen by the _commit time_ of the commit corresponding to the benchmark run.

Related: https://github.com/conbench/conbench/issues/633

conbench/tests/populate_local_conbench.py
jgehrcke: will do!

conbench/tests/populate_local_conbench.py
jgehrcke: sure, that's a great starting point for simulating a distribution change (picking a certain time window that is transformed by a certain shift, instead of picking an individual sample)

conbench/tests/populate_local_conbench.py
austin3dickey: haha, that makes sense. Another option would be to choose a repo that has fewer commits over a longer time range. Maybe conbench itself?

jgehrcke:
> Instead of every 20th commit, if you do consecutive commits it should hit the GitHub API a lot less often, because it won't need to fill in intermediate commits every time.

I did consecutive commits now towards actually landing this, and iterating later. I did not investigate too deeply, but the duration is about the same and I think the usage of the GitHub HTTP API is still in the same ballpark. Individual POST requests are responded to rather slowly. I think that is because there is a constant minimum number of GitHub HTTP API requests done for every posted result.

As of
```
default_branch = github.get_default_branch(...)
```
and also
```
commits = github.get_commits_to_branch(...)
```

```
conbench-app-1  | [230130-13:23:03.672] [42] [conbench.entities.commit] INFO: Finding all commits to the master branch of apache/arrow between  2023-01-03T09:51:12Z and 2023-01-03T12:23:30Z
conbench-app-1  | [230130-13:23:03.873] [42] [conbench.entities.commit] INFO: Backfilling 0 commit(s)
```

As we know there is a lot that can be architecturally optimized here for faster responses and fewer failure paths. But that work should not happen in this PR here.

I have added a change so that if Config.TESTING we fetch commits for the last 60 days only. Reduced the duration of `make db-populate` for me locally to 1 minute 40 s. 

Will merge this when CI passes.

jgehrcke:
`make db-populate` was faster in CI, as of reduced latency to GitHub HTTP API:
```
Mon, 30 Jan 2023 13:34:25 GMT
230130-13:34:25.900 INFO: register via: http://127.0.0.1:32768/api/register/
Mon, 30 Jan 2023 13:34:26 GMT
230130-13:34:26.124 INFO: login via: http://127.0.0.1:32768/api/login/
Mon, 30 Jan 2023 13:34:26 GMT
230130-13:34:26.316 INFO: login succeeded

...
Mon, 30 Jan 2023 13:34:35 GMT
230130-13:34:35.454 INFO: Posted a benchmark with run_id 'history_machine12' and commit 88b42ef66fe664043c5ee5274b2982a3858b414e. Received status code 201. Took 1 attempt(s). Last attempt took 0.01044 s.
```

As of the few commits pushed in the past hour I did manage to consume CI's API quota again. Example from last CI run:
```
Mon, 30 Jan 2023 13:36:18 GMT
conbench-app-1  | [230130-13:34:35.405] [12] [conbench.entities.run] INFO: treat as unknown commit: error during get_github_commit(): 403 Client Error: Forbidden for url: https://api.github.com/repos/apache/arrow/commits/88b42ef66fe664043c5ee5274b2982a3858b414e
...
```

conbench/entities/commit.py
```diff
@@ -326,13 +327,12 @@ def backfill_default_branch_commits(repository: str, new_commit: Commit) -> None
...
+    elif Config.TESTING:
         log.info(
-            "Your DB_HOST is localhost, so assuming you're running "
-            "conbench/tests/populate_local_conbench.py. Backfilling the DB only from "
-            "2022-01-01 in order to save time."
+            "Config.TESTING is true Backfilling the DB only from "
+            "the last 60 days in order to save time."
         )
-        since = datetime(2022, 1, 1)
+        since = datetime.today() - timedelta(days=60)
```
austin3dickey, line 335: Nice, thank you!

PR #593, opened 2023-01-11
api/index.py: address warning: "alembic_version" does not ...

jgehrcke:
When working locally, this warning appears periodically in the containerized stack log output. Sometimes the log is central for being productive. And then this noise wastes time. 

Pragmatic solution here. A better solution with less divergence between testing and prod would be to have this table in local deployment as well as rightfully noted by @austin3dickey.

### DIFF
--- a/conbench/api/index.py
+++ b/conbench/api/index.py
@@ -85,20 +85,28 @@ class PingAPI(ApiEndpoint):
           - Ping
         """
         now = datetime.datetime.now(datetime.timezone.utc)
-        query = text("SELECT version_num FROM alembic_version")
-        try:
-            # In local deployment the PostgreSQL server logs: `ERROR:  relation
-            # "alembic_version" does not exist at character 25`. That is
-            # probably because not schema migration has yet been performed?
-            # Should we make it so that this table exists also for those
-            # dev/from-scratch deployments?
-            version = list(Session.execute(query))[0]["version_num"]
-        except:
-            version = "unknown"
+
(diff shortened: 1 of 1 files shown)


conbench/api/index.py
```diff
@@ -85,20 +85,28 @@ def get(self):
...
+
+        if Config.TESTING:
+            alembic_version = "not-set-in-testing"
+        else:
+
+            query = text("SELECT version_num FROM alembic_version")
+            try:
+                # In local deployment the PostgreSQL server logs: `ERROR:  relation
+                # "alembic_version" does not exist at character 25`. That is
+                # probably because not schema migration has yet been performed?
```
jonkeane, line 97: Typo?

```suggestion
                # probably because no schema migration has yet been performed?
```

jonkeane: Also I know this isn't your text — just moving, but it stood out anyawy!

jonkeane (approved):
I'm good with this + thankful for the detailed comments we have + adding to them. This _could_ be confusing at some point in the future, but I think this is a good balance between making ourselves more productive + documenting the slight side step we're taking in case it bumps into us in the future.

austin3dickey (approved):
Nice, I like the distinction between `not-set-in-testing` and `err`.

conbench/api/index.py
jgehrcke: Yes! Will change that in another branch.

Issue #589, opened 2023-01-11
history endpoint: specify result object schema?


jgehrcke:
I think the object properties of the individual objects returned by the history endpoint are not yet specified, at least not in the object schema in /api/docs as far as I have seen. Maybe we don't want to do this as of now because things are in flux. Just wanted to leave a reminder ticket.

jgehrcke:
In local testing I just got a case where this endpoint returned an object that had just three properties:
```
{
'mean': 19.585377722298556,
'message': 'GH-15205: [R] Fix a parquet-fixture finding in R tests (#15207)',
'timestamp': '2023-01-06T14:25:20'
}
```

I noticed this because I was expecting the `sha` key to be set here, too. It looks like when a commit message is set, a commit hash isn't necessarily set. That's rather wild.

I would expect either complete commit info to be set, or none -- not partial. Maybe that is something that can be enforced in the input validation side of things.

austin3dickey:
@jgehrcke Can you clarify which endpoint you used to get that object? Something like that should never come from `GET /api/history`; that should only return an object that looks like [this](https://github.com/conbench/conbench/blob/0c9adbc8bc19e9aede6d6ffe2f9492e671fbcd07/conbench/entities/history.py#L43-L61).

Your response looks suspiciously like one of these objects instead 🤔 
https://github.com/conbench/conbench/blob/0c9adbc8bc19e9aede6d6ffe2f9492e671fbcd07/conbench/app/_plots.py#L275-L279

jgehrcke:
> Your response looks suspiciously like one of these objects instead

Oh! Thank you so much for thinking this through where I (seemingly) didn't. I forgot for a moment how often we call `_source()` and that it's not always a list of these history objects that we inject.

jgehrcke:
Closing, because this is not relevant anymore.

austin3dickey:
Oh, I still think this would be a nice feature. We document many aspects of the API, but the history endpoint is still not documented.

jgehrcke:
Yes, we should keep working on that. I realize every now and then how this underspecification a lot of costs time when touching code.

jgehrcke:
Closing this as 'done'! This was am important step forward. We should make the next-obvious tweaks to the schema though, making use of new names.

Issue #477, opened 2022-11-30
Click on a point in the timeseries plot and travel to that BenchmarkResult's webpage


austin3dickey:
This will be a big UX improvement

jgehrcke:
Previously I looked at https://docs.bokeh.org/en/latest/docs/user_guide/interaction/callbacks.html#openurl. In a single scatter plot this can be used to open a URL in a new tab when clicking a glyph. This would work although rather non-obvious (not the greatest UX I think). I tried to implement this but didn't get this to work within ~10 minutes of trying, the shown code snippet uses a single subplot per figure and uses `p.select(type=TapTool)` where `p` is the figure object. We have more than one subplot in the figure (that's why this method doesn't work directly) and I could at least not quickly find a way around that (but I trust that there is a way!).

I also found https://stackoverflow.com/questions/68449634/can-i-add-an-a-tag-to-a-hovertool-tooltip-in-bokeh rather funny because "You can, technically, but it is completely useless. The tooltip follows next to the mouse at all times, so there is not way to click on a link inside the tooltip." is an important answer to "Can I add an `<a>` to a HoverTool tooltip in Bokeh".

Right now I think that it might be good idea to use a 'custom JS' callback to update a part of the DOM to show run details including a link to the corresponding results view.

jgehrcke:
I begin to understand some relevant bits and pieces from Bokeh that will help here.

I found that `bokeh.events.Tap` is triggered by each click in a plot, regardless of whether that's on a glyph or not.

I then found after quite a bit of forum reading that `TapTool` can be useful for pre-filtering clicks so that only those events bubble up that correspond to a click on a glyph (under-documented primary feature of TapTool).

So I wondered how one can hook arbitrary JavaScript code into events emitted by TapTool.

In [this](https://docs.bokeh.org/en/latest/docs/user_guide/interaction/callbacks.html#openurl) snippet I found that one can overwrite `taptool.callback`. I played with setting a `bokeh.models.CustomJS` object here and got things wired up so that indeed clicking a glyph led to custom JavaScript code execution. That was tricky because of a lack of docs/explanation. But the approach appeared robust and promising.

[...]

PR #597, opened 2023-01-13
history plot: show commit hash in addition to msg

jgehrcke:
Before: tooltip showed commit msg prefix:
Screenshot from 2023-01-13 14-56-24

After: tooltip shows short commit hash:
Screenshot from 2023-01-13 15-20-07

I think that's a tiny improvement. Of course, more importantly, we may want to link to that benchmark result (that's tracked in a different ticket which I just didn't find quickly -- update: https://github.com/conbench/conbench/issues/477).

Interestingly, I found that the internal state may not be consistent, probably as of a lack of input data validation: https://github.com/conbench/conbench/issues/589#issuecomment-1381918202

### DIFF
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -184,7 +184,8 @@ def _source(
     # Change this to use short commit hashes. The long commit message prefix
     # does not unambiguously specify the commit. Ideally link to the commit, in
     # the tooltip?
-    commits = [x["message"] for x in data]
+    commits = [d["message"] for d in data]
+
     dates = [dateutil.parser.isoparse(x["timestamp"]) for x in data]
 
     points, means = [], []
@@ -211,20 +212,12 @@ def _source(
         # the tooltip string labels?
         points = [x.split(" ")[0] for x in means]
 
-    commit_hashes_short = []
-    for d in data:
-        try:
-            commit_hashes_short.append("#" + d["sha"][:7])
-        except KeyError:
-            # https://github.com/conbench/conbench/issues/589
-            commit_hashes_short.append("not-set")
-
     return bokeh.models.ColumnDataSource(
         data={
             "x": dates,
             "y": points,
             "commits": commits,
-            "commit_hashes_short": commit_hashes_short,
+            "commit_hashes_short": ["#" + d["sha"][:7] for d in data],
             "means": means,
         }
     )
@@ -291,6 +284,7 @@ def time_series_plot(history, benchmark, run, height=380, width=1100):
                 "mean": benchmark["stats"]["mean"],
                 "message": run["commit"]["message"],
                 "timestamp": run["commit"]["timestamp"],
+                "sha": run["commit"]["sha"],
             }
         ],
         unit,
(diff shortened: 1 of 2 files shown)


conbench/app/_plots.py
```diff
@@ -211,8 +211,23 @@ def _source(
...
         points = [x.split(" ")[0] for x in means]
 
-    source_data = dict(x=dates, y=points, commits=commits, means=means)
-    return bokeh.models.ColumnDataSource(data=source_data)
+    commit_hashes_short = []
+    for d in data:
+        try:
+            commit_hashes_short.append("#" + d["sha"][:7])
+        except KeyError:
+            # https://github.com/conbench/conbench/issues/589
```
austin3dickey: As we talked about in this issue, you can probably adjust the code down below that sends a different list of dicts to the `data` input!

austin3dickey (commented):
Nice adjustment! I agree that #477 will make this experience much better. Feel free to do that one, or I'll get around to it eventually!

jonkeane:
I _love_ that this is cleaner and less all-consuming on the screen. But I wonder if we should do this at the same time as #477 Or maybe we have confidence that #477 will land soon enough that we can merge this now?

One of the biggest (and possibly only?) use cases for that github info in the plot is so that one can look at change points and see if the change looks relevant or not (if I'm looking at Python benchmarks, and the commit message is a docs change for go, I know that something weird is up, compared to something like `...[Python] refactor all the things` where I can be more confident the code change led to the big change. With #477 it's one click to get to that commit message (which isn't too bad), but without it there's a manual transposing of the short hash and looking up. So I think it would be good to minimize the time between these two — what do y'all think?

conbench/app/_plots.py
jgehrcke: done, commit pushed

jgehrcke:
> But I wonder if we should do this at the same time as https://github.com/conbench/conbench/issues/477 Or maybe we have confidence that https://github.com/conbench/conbench/issues/477 will land soon enough that we can merge this now? [...] if I'm looking at Python benchmarks, and the commit message is a docs change for go, I know that something weird is up, compared to something like ...[Python] refactor all the things where I can be more confident the code change led to the big change.

@jonkeane if I understand your feedback here correctly I think you are saying that for now you'd love to keep the commit msg (prefix) in the tooltip. Is that right? I assume so, and have pushed a corresponding commit. So that with this PR we show commit hash _and_ msg prefix.

With #477 we can then remove the msg again (it's kind of too much information for the tooltip).

jgehrcke:
Thanks for the feedback everyone!

jonkeane:
> @jonkeane if I understand your feedback here correctly I think you are saying that for now you'd love to keep the commit msg (prefix) in the tooltip. Is that right? I assume so, and have pushed a corresponding commit. So that with this PR we show commit hash and msg prefix.
> 
> With https://github.com/conbench/conbench/issues/477 we can then remove the msg again (it's kind of too much information for the tooltip).

Yes, exactly! Thanks for this

PR #601, opened 2023-01-17
Accept `"pr_number": ""`

alistaire47:
Accept `"pr_number": ""` (in addition to `"pr_number": null` and not populating it) when posting runs and results (`run.github.pr_number` and `benchmark.github.pr_number`), as this metadata is very likely to come from env vars and the intent here is unambiguous. I don't think passing `""` is desired behavior, but it's a worse reason for failing.

### DIFF
--- a/conbench/entities/run.py
+++ b/conbench/entities/run.py
@@ -59,7 +59,7 @@ class Run(Base, EntityMixin):
 
         if github_data := data.pop("github", None):
             repository = repository_to_url(github_data["repository"])
-            pr_number = github_data.get("pr_number")
+            pr_number = github_data.get("pr_number") or None
             branch = github_data.get("branch")
             sha = github_data["commit"]
 
--- a/conbench/entities/run.py
+++ b/conbench/entities/run.py
@@ -59,7 +59,7 @@ class Run(Base, EntityMixin):
 
         if github_data := data.pop("github", None):
             repository = repository_to_url(github_data["repository"])
-            pr_number = github_data.get("pr_number") or None
+            pr_number = github_data.get("pr_number")
             branch = github_data.get("branch")
             sha = github_data["commit"]
 
@@ -278,6 +278,13 @@ def commit_hardware_run_map():
 
 
 class GitHubCreate(marshmallow.Schema):
+    @marshmallow.pre_load
+    def change_pr_number_empty_string_to_none(self, data, **kwargs):
+        if "pr_number" in data:
+            data["pr_number"] = data["pr_number"] or None
+
+        return data
+
     commit = marshmallow.fields.String(
         required=True,
         metadata={
--- a/conbench/tests/api/test_benchmarks.py
+++ b/conbench/tests/api/test_benchmarks.py
@@ -780,6 +780,9 @@ class TestBenchmarkPost(_asserts.PostEnforcer):
         data = copy.deepcopy(self.valid_payload)
         data["run_id"] = _uuid()
         data["github"]["pr_number"] = pr_number
+        # viable `github` submissions without `pr_number`:
+        # - just `repository` and `commit` (assumed to be on default branch)
+        # - submit `branch` directly instead (mostly discouraged, but possible)
         if pr_number == "<absent>":
             data["github"].pop("pr_number")
 
--- a/conbench/tests/api/test_benchmarks.py
+++ b/conbench/tests/api/test_benchmarks.py
@@ -774,6 +774,21 @@ class TestBenchmarkPost(_asserts.PostEnforcer):
         location = "http://localhost/api/benchmarks/%s/" % new_id
         self.assert_201_created(response, _expected_entity(benchmark_result), location)
 
+    @pytest.mark.parametrize("pr_number", [12345678, "12345678", None, "", "<absent>"])
+    def test_create_allow_pr_number_variations(self, pr_number, client):
+        self.authenticate(client)
+        data = copy.deepcopy(self.valid_payload)
+        data["run_id"] = _uuid()
+        data["github"]["pr_number"] = pr_number
+        if pr_number == "<absent>":
+            data["github"].pop("pr_number")
+
+        response = client.post("/api/benchmarks/", json=data)
+        new_id = response.json["id"]
+        benchmark_result = BenchmarkResult.one(id=new_id)
+        location = "http://localhost/api/benchmarks/%s/" % new_id
+        self.assert_201_created(response, _expected_entity(benchmark_result), location)
+
     def test_create_benchmark_distribution(self, client):
         for payload in [self.valid_payload, self.valid_payload_for_cluster]:
             self.authenticate(client)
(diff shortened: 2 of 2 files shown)


jonkeane (changes requested):
Wow, that looks easy! Would you mind adding a test for this behavior too?

austin3dickey (commented):
Thank you! Agreed this is a better experience.

Could you add a unit test [around here](https://github.com/conbench/conbench/blob/main/conbench/tests/api/test_benchmarks.py#L776) by the other `github` unit tests? I don't think there are any tests around `pr_number` so it would be really nice to test `None`, `""`, not included, an integer, and maybe even an integer inside of a string.

I feel like some of those might still fail due to API schema validation but we'd want them to succeed, since they're unambiguous like you said.

alistaire47:
> Could you add a unit test [around here](https://github.com/conbench/conbench/blob/main/conbench/tests/api/test_benchmarks.py#L776) by the other `github` unit tests? I don't think there are any tests around `pr_number` so it would be really nice to test `None`, `""`, not included, an integer, and maybe even an integer inside of a string.

Ah thank you! I was looking for where these were but couldn't find them; the stuff in `tests/entities/test_run.py` is nice, but not this

alistaire47:
Ok, updated with a test, which turned up that marshmallow validation would error out before it got to the fix anyway, so I had to add some precleaning in the schema. Maybe there's a more elegant way to do this (open to suggestions), but this is what I could figure out.

conbench/tests/api/test_benchmarks.py
```diff
@@ -774,6 +774,21 @@ def test_create_allow_just_sha(self, client):
         location = "http://localhost/api/benchmarks/%s/" % new_id
         self.assert_201_created(response, _expected_entity(benchmark_result), location)
 
+    @pytest.mark.parametrize("pr_number", [12345678, "12345678", None, "", "<absent>"])
```
austin3dickey, line 777: Love it!!

austin3dickey (approved):
Thank you!

conbench/tests/api/test_benchmarks.py
```diff
@@ -774,6 +774,21 @@ def test_create_allow_just_sha(self, client):
...
+    def test_create_allow_pr_number_variations(self, pr_number, client):
+        self.authenticate(client)
+        data = copy.deepcopy(self.valid_payload)
+        data["run_id"] = _uuid()
+        data["github"]["pr_number"] = pr_number
+        if pr_number == "<absent>":
+            data["github"].pop("pr_number")
```
jonkeane, line 788: Would you mind adding a comment about what the behavior is here that we're expecting? It looks like in the `"<absent>"` case, we remove `"<absent>"` before we send it to the client? Or am I missing something? 

We would expect that that would error, right? Maybe we should make a separate test that tests and confirms that? Or we can leave it out if it's effectively the same thing as sending nothing?

austin3dickey: Since the beginning of `pr_number` I always intended that you could just send `{"commit": "abc", "repository": "org/repo"}` and it would assume you're running benchmarks on the main branch. If you think that's unclear we could change it, but this test is the expected behavior for me.

alistaire47: Yeah, I put that test here because not passing `pr_number` is documented as acceptable behavior. There's both the case Austin mentioned, plus the case where the user specifies `branch` directly (which is discouraged, but possible, and if for some reason someone had builds on a branch without a PR, the only way to submit it).

jonkeane: That behavior makes sense — the thing I was confused by is _how_ this code maps onto that (TBF, I didn't get to there reading the code it looked more like a mistake or testing some strange behavior where we expect a user to send `"<absent>"` and wanted to catch that). 

Something like what you say here in a comment (maybe along with "the `"<absent>"` sentinel here  is used so that we can test that case where we don't send pr_number at all" if that's accurate) would be fantastic to make that less confusing

alistaire47: Added a [somewhat verbose] comment to that effect

jonkeane: Awesome, that's great

jonkeane (approved):

conbench/entities/run.py
```diff
@@ -278,6 +278,13 @@ def commit_hardware_run_map():
 
 
 class GitHubCreate(marshmallow.Schema):
+    @marshmallow.pre_load
+    def change_pr_number_empty_string_to_none(self, data, **kwargs):
+        if "pr_number" in data:
+            data["pr_number"] = data["pr_number"] or None
```
jgehrcke, line 284: Quick note: that is not as precise as the method name suggests :)

I'd do:
```
if data.get("pr_number") == "":
    data["pr_number] = None
```
because `data["pr_number"]` may evaluate to False for more values than just `""`.

conbench/tests/api/test_benchmarks.py
jgehrcke: What is supposed to happen if a floaty value is provided? Rejected (error), silently discarded, or transparently converted to int?

jgehrcke:
Cool. Lot's of discussion for a simple change. Hope you let me join the party :)

Nits:
- a PR title prefix showing the area/component is always nice, something like `run POST handler: ...` or so. Here I actually care about the resulting commit message after squash-merge.
- in the PR description it's nice to briefly describe the actual issue that triggered the patch, or link elsewhere.

I want to leave some more thoughts here. Because this is about interfaces, and I care a lot about interfaces :). Great interfaces make great software, weak interfaces are always a bit of a nightmare for everyone involved.
 
I think some of the most important work when thinking through changes like this is to define correct/expected behavior in terms of specification. We have (at least) three places of code that have an opinion on this `pr_number` parameter:
- actual POST handler code
- unit tests
- documentation

Right now I feel like all these three pieces tell a little bit of a different story.

It's important to clarify which one of those is currently 'the most correct', and then to push for achieving _convergence_ between those three. I always favor documentation to be most correct, and like to compare code and tests against documentation (in particular when talking about APIs). That makes changes like this easiest to review, and also is after all best for users.

I feel like the current docs are wrong or incomplete, or somehow rather off:

```
             "description": "[recommended] The number of the GitHub pull request that "
            "is running this benchmark, or `null` if it's a run on the default branch"
```
This docstring raises questions.

Required or not? I think "recommended" is an unusual way to say "it's optional, and if you don't provide it you will miss out on something". What is the something? What if it's a run on the default branch and I don't provide `null`?

From a user's perspective I'd like to understand which feature I am getting when I provide this value (and which feature I am potentially missing out on when I don't provide this value, knowing that it seems to be entirely optional).

Once we understand the underlying feature, we can write docs like this:

"Setting a `pr_number` is optional. If you set it, then .... . If provided, a stringified integer is expected. All other value types are treated as if no value was provided."

PR #603, opened 2023-01-18
history plot: show run details when clicking a glyph

jgehrcke:
Motivated by #477 -- this is now ready to review.

A few notes:
- Do we have suggeststions for a better panel heading ("Details for selected benchmark (green data point in plot)")?
- Would appreciate: content feedback (words, sentences, details shown...).
- Happy also about color/placement feedback, will do visual changes in follow-up patches. For example, let's remove the zoom slider below the plot, then this detail panel will be closer to the plot (taken from Slack thread with @jonkeane). 
- This introduces a rather complex topic: Bokeh's custom JavaScript callbacks (they are cool, they are powerful, but they are also maintenance burden). This PR in particular introduces JavaScript code as literal Python strings. That code is not covered by linting, not even by any kind of CI. I still think it's for now good to take this approach (as of a lack in alternatives, and the approach is simple enough to still be manually testable and explainable). 

An example view:

image

### DIFF
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -193,45 +193,61 @@ def _source(
     # the `BenchmarkCreate` schema in the Conbench API). Are these tz-aware or
     # tz-naive but in UTC?
     datetimes = [dateutil.parser.isoparse(x["timestamp"]) for x in data]
-    date_strings = [d.strftime("%Y-%m-%d %H-%M %Z") for d in datetimes]
+    date_strings = [d.strftime("%Y-%m-%d %H:%M %Z") for d in datetimes]
 
-    points, means = [], []
+    points, values_with_unit = [], []
     if alert_min:
         for x in data:
             alert = 5 * float(x["distribution_stdev"])
             points.append(float(x["distribution_mean"]) - alert)
-            means.append(unit_fmt(points[-1], unit))
+            values_with_unit.append(unit_fmt(points[-1], unit))
     elif alert_max:
         for x in data:
             alert = 5 * float(x["distribution_stdev"])
             points.append(float(x["distribution_mean"]) + alert)
-            means.append(unit_fmt(points[-1], unit))
+            values_with_unit.append(unit_fmt(points[-1], unit))
     else:
         points = [x[key] for x in data]
         # Note(JP): If `means` is just string-formatted `points` then this is
         # kind of acknowledging that `points` is always a collection of mean
         # values. It seems that this is a string value field that is only
         # ever used for tooltip generation?
-        means = [unit_fmt(float(x[key]), unit) for x in data if x[key]]
+        values_with_unit = [unit_fmt(float(x[key]), unit) for x in data if x[key]]
 
     if formatted:
         # Note(JP): why would we want to calculate the raw numeric data from
         # the tooltip string labels?
-        points = [x.split(" ")[0] for x in means]
+        points = [x.split(" ")[0] for x in values_with_unit]
 
-    return bokeh.models.ColumnDataSource(
-        data={
-            "x": datetimes,
-            "y": points,
-            "date_strings": date_strings,
-            "commit_messages": commit_messages,
-            "commit_hashes_short": ["#" + d["sha"][:7] for d in data],
-            "relative_benchmark_urls": [
-                f'/benchmarks/{d["benchmark_id"]}' for d in data
-            ],
-            "means": means,
-        }
-    )
+    dsdict = {
+        "x": datetimes,
+        # Note(JP): maybe rename `points` into `y_values_for_plotting`
+        "y": points,
+        # List of human-readable date strings, corresponding to
+        # the `datetimes` objects for the time axis
+        "date_strings": date_strings,
+        # List of (truncated) commit messages.
+        "commit_messages": commit_messages,
+        # List of short commit hashes with hashtag prefix
+        "commit_hashes_short": ["#" + d["sha"][:8] for d in data],
+        "relative_benchmark_urls": [f'/benchmarks/{d["benchmark_id"]}' for d in data],
+        # Stringified values (truncated, with unit)
+        "values_with_unit": values_with_unit,
+    }
+
+    multisample, multisample_count = _inspect_for_multisample(data)
+    if multisample:
+        # Add a column to the ColumnarDataSource: stringified individual
+        # samples (with units).
+        strings = []
+        for d in data:
+            samples = d["data"]
+            log.info("samples: %s", samples)
+            strings.append(", ".join(unit_fmt(s, unit) for s in samples))
+
+        dsdict["multisample_strings_with_unit"] = strings
+
+    return bokeh.models.ColumnDataSource(data=dsdict)
 
 
 def _inspect_for_multisample(items) -> tuple[bool, Optional[int]]:
@@ -247,7 +263,14 @@ def _inspect_for_multisample(items) -> tuple[bool, Optional[int]]:
     """
 
     # Get number of samples for each benchmark.
-    samplecounts = [len(i["data"]) for i in items]
+    try:
+        samplecounts = [len(i["data"]) for i in items]
+    except KeyError:
+        # Likely, the provided `items` for testing here are not strictly of the
+        # required shape. It's a programming bug, but do not crash in this
+        # case. Return an answer: not multisample (at least not in the way as
+        # expected).
+        return False, None
 
     multisample = True
     multisample_count = None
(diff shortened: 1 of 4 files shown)


conbench/app/_plots.py
```diff
@@ -186,41 +187,67 @@ def _source(
...
+    # Note(JP): isoparse() returns a `datetime.datetime` object. And I think
+    # that the `timestamp` property corresponds to the utc-local invocation (or
+    # finish) time of the corresponding benchmark case run (the `timestamp`
+    # property on the `BenchmarkCreate` schema in the Conbench API). Are these
+    # tz-aware or tz-naive but in UTC?
+    datetimes = [dateutil.parser.isoparse(x["timestamp"]) for x in data]
+    date_strings = [d.strftime("%Y-%m-%d %H:%M %Z") for d in datetimes]
 
-    points, means = [], []
+    points, values_with_unit = [], []
```
jgehrcke, line 205: A variable name change attempt towards more clarity. Not perfection, but improvement :)
`points` also needs a better name, I think.

conbench/app/_plots.py
```diff
@@ -186,41 +187,67 @@ def _source(
...
+        "values_with_unit": values_with_unit,
+    }
+
+    multisample, multisample_count = _inspect_for_multisample(data)
+    if multisample:
+        # Add a column to the ColumnarDataSource: stringified individual
+        # samples (with units).
+        strings = []
+        for d in data:
+            samples = d["data"]
```
jgehrcke, line 251: There is an example for where type annotations will help so so so soooooooooo much. Without type annotations one has to keep a lot of mental overhead about what `data` and `data["data"]` even are. With type annotations and proper editor setup a mouse-over hover effect simply provides the answer.

https://github.com/conbench/conbench/issues/605

conbench/app/_plots.py
```diff
@@ -236,7 +263,15 @@ def _inspect_for_multisample(items) -> tuple[bool, Optional[int]]:
     """
 
     # Get number of samples for each benchmark.
-    samplecounts = [len(i["data"]) for i in items]
+    try:
+        samplecounts = [len(i["data"]) for i in items]
+    except KeyError:
```
jgehrcke, line 274: I did this 'guard' because I have no confidence if `items` always is what I think it is. Type annotations and type checking will help with that.

conbench/app/_plots.py
```diff
@@ -258,8 +293,102 @@ def _inspect_for_multisample(items) -> tuple[bool, Optional[int]]:
...
+        # `s1.selected.indices` seems to always be available no matter where
+        # one clicks, and in case no glyph was clicked the array is zero
+        # length.
+        args={"s1": source},
+        code="""
+        // console.log("cb_data:", cb_data);
+        // console.log("cb_obj:", cb_obj);
+        // console.log("s1.selected.indices: ", s1.selected.indices);
+
+        if (s1.selected.indices.length == 0){
```
jgehrcke: todo: add space

conbench/app/_plots.py
```diff
@@ -258,8 +293,102 @@ def _inspect_for_multisample(items) -> tuple[bool, Optional[int]]:
...
+            console.log("nothing selected, remove detail");
+            // Make the panel invisible.
+            const e = document.getElementsByClassName("conbench-histplot-run-details")[0];
+            e.style.display = 'none'
```
jgehrcke: todo: add semicolon

conbench/app/_plots.py
```diff
@@ -258,8 +293,102 @@ def _inspect_for_multisample(items) -> tuple[bool, Optional[int]]:
...
+def gen_js_callback_click_on_glyph_show_run_details(repo_string):
+    """
+    Dynamically manipulate Bootstrap panel conbench-histplot-run-details`.
+
+    Show run details if a corresponding glyph was clicked in the plot.
+
+    Warning: requires manual testing, not covered by tests/CI.
+    """
+    return bokeh.models.CustomJS(
+        code=f"""
```
jgehrcke, line 344: Watch out: this is an f-string that dynamically assembles JavaScript code :-(
It's going to be fine. Maybe. :)

conbench/app/_plots.py
```diff
@@ -258,8 +293,102 @@ def _inspect_for_multisample(items) -> tuple[bool, Optional[int]]:
...
+            run_samples_with_units = cb_data.source.data['multisample_strings_with_unit'][i];
+        }}
+
+        // JavaScript code generated in a Python f string -- templating hell, yeah! :)
+        // I don't know if repo string is always a URL, if it's always
+        // pointing to GitHub. But if it does, we can do some UX sugar.
+
+        var commit_repo_string =  run_commit_hash_short + ' in {repo_string}';
+
+        const repo = '{repo_string}';
```
jgehrcke, line 370: That's the only place where it's dynamic. Pragmatic solution, I think it's actually fine for now.

conbench/app/_plots.py
```diff
@@ -333,11 +467,27 @@ def time_series_plot(history, benchmark, run, height=380, width=1100):
...
     )
     p.toolbar.logo = None
 
+    # TapTool is not responding to each click event, but but only triggers when
+    # clicking a glyph:
+    # https://discourse.bokeh.org/t/how-to-trigger-callbacks-on-mouse-click-for-tap-tool/1630
+    taptool = p.select(type=bokeh.models.TapTool)
+    taptool.callback = gen_js_callback_click_on_glyph_show_run_details(
+        # Assume that the repository is constant across all data points in
+        # this plot. Is that a good assumption
```
jgehrcke: I'd love to have feedback on this! 

Do all samples in a history plot have the same repository specified in their contexts? Any guarantee on that? CC @austin3dickey

conbench/app/_plots.py
```diff
@@ -353,8 +503,15 @@ def time_series_plot(history, benchmark, run, height=380, width=1100):
...
-        size=4,
+        size=6,
         color="#ccc",
+        line_width=1,
+        selection_color="#76bf5a",  # like bootstrap panel dff0d8, but darker
+        selection_line_color="#5da540",  # same green, again darker
+        # Cannot change the size upon selection
+        # selection_size=10,
+        nonselection_fill_alpha=1.0,
+        nonselection_line_alpha=1.0,
```
jgehrcke, line 524: This is a pragmatic way to define different styles for "selected" vs "non-selected". Documented at https://docs.bokeh.org/en/3.0.3/docs/user_guide/styling/plots.html#selected-and-unselected-glyphs.

conbench/app/_plots.py
```diff
@@ -450,11 +607,7 @@ def time_series_plot(history, benchmark, run, height=380, width=1100):
         bokeh.models.HoverTool(
             tooltips=[
                 ("date", "$x{%F}"),
-                # Note(JP): this is where the `means` name becomes special,
-                # I think.
-                ("mean", "@means"),
-                ("commit", "@commit_hashes_short"),
-                ("commit msg", "@commit_messages"),
+                ("value", "@values_with_unit"),
```
jgehrcke, line 620: Now the word "value" appears in the tooltip where before it said "mean".

The `value_with_unit` list contains value-with-unit strings, one item is for example `"1.2 s"` So, this effectively says to show something like `value: 1.2 s` in the tooltip.

conbench/templates/benchmark-entity.html
```diff
@@ -30,6 +30,22 @@
 </div>
 <br>
 
+<div class="row">
+  <div class="col-md-12">
+    Click on a gray glyph in the plot to see corresponding benchmark run details.
```
boshek: My inclination is that "glyph" is too jargon-y. What if one directly linked it to legend for common language?

```suggestion
    Click on any grey point (benchmark mean) in the plot to see corresponding benchmark run details.
```

conbench/app/_plots.py
```diff
@@ -186,41 +187,67 @@ def _source(
     # the tooltip?
     commit_messages = [d["message"] for d in data]
 
-    dates = [dateutil.parser.isoparse(x["timestamp"]) for x in data]
+    # Note(JP): isoparse() returns a `datetime.datetime` object. And I think
+    # that the `timestamp` property corresponds to the utc-local invocation (or
+    # finish) time of the corresponding benchmark case run (the `timestamp`
```
austin3dickey: The timestamp property here comes from `/api/history`, which means it's actually the commit timestamp. That comes straight from GitHub and represents a UTC time. However, the `/api/history` endpoint just reports it as e.g. `"2022-03-03T14:16:51"` so this datetime will be tz-naive, assumed to be in UTC.

conbench/app/_plots.py
```diff
@@ -186,41 +187,67 @@ def _source(
...
+    }
+
+    multisample, multisample_count = _inspect_for_multisample(data)
+    if multisample:
+        # Add a column to the ColumnarDataSource: stringified individual
+        # samples (with units).
+        strings = []
+        for d in data:
+            samples = d["data"]
+            log.info("samples: %s", samples)
```
austin3dickey: Can you remove this before merging so we don't see a lot of noise in production logs?

conbench/app/_plots.py
austin3dickey: Can you create a followup issue or note to clean this up once we do #605? This is a bit of extra complexity for something we may not even be able to replicate in tests.

conbench/app/_plots.py
austin3dickey: 🥲

conbench/app/_plots.py
```diff
@@ -258,8 +293,102 @@ def _inspect_for_multisample(items) -> tuple[bool, Optional[int]]:
...
+        const run_result_value_with_unit = cb_data.source.data['values_with_unit'][i];
+
+        var run_samples_with_units = undefined;
+
+        if (cb_data.source.data['multisample_strings_with_unit'] !== undefined) {{
+            run_samples_with_units = cb_data.source.data['multisample_strings_with_unit'][i];
+        }}
+
+        // JavaScript code generated in a Python f string -- templating hell, yeah! :)
+        // I don't know if repo string is always a URL, if it's always
```
austin3dickey, line 363: Right now, repo string should either be None or a URL to the GitHub repo. I don't know if that will change in the future 🙂

conbench/app/_plots.py
austin3dickey: Yes! [Here's the guarantee](https://github.com/conbench/conbench/blob/main/conbench/entities/history.py#L105). 🙂

austin3dickey (approved):
This is looking great, thanks JP!

conbench/app/_plots.py
boshek: What is the thinking for this change? It _is_ the mean right? My inclination would be to actually expand it and match it to "benchmark mean" to align language with the legend.

conbench/templates/benchmark-entity.html
```diff
@@ -30,6 +30,22 @@
...
+    <div class="conbench-histplot-run-details" style="display: none;">
+      <br />
+      <div class="panel panel-success">
+        <div class="panel-heading">Details for selected benchmark (green data point in plot)</div>
```
boshek, line 39: I think this is good but I do think that we could standardize language - i.e. we use data point here and glyph elsewhere.

conbench/app/_plots.py
jgehrcke: TODO note added

conbench/app/_plots.py
jgehrcke: Thanks, especially for linking to it! :)

conbench/app/_plots.py
jgehrcke: Great question. These code sections are rather hard to see through, and I am slowly changing things in hopefully digestible chunks.

It's not always the mean. The same function as of today builds various Bokeh `ColumnDataSource`s, including for single sample benchmarks (n=1) and also for displaying min and max values for multi-sample benchmarks, and even for the 'alert min' and 'alert max' time series which seem to be the 5sigma band boundaries.

We should try to tidy this up step by step, not changing everything at once, but in controllable pieces.

Right now we still have code like

```
"mean": benchmark["stats"]["min"]
```
(wow)

and also
```
points.append(float(x["distribution_mean"]) + alert)
```

And in these cases we previously added the stringified-value-with-unit as `"mean", "@means"` to the `ColumnDataSource`. To make the code a tiny less misleading: that's why I changes this to the more general concept of `"value", "@values_with_unit"` here.

In other words, right now `_source()` is a generic function that is called for various y axis units and value meanings. 

The function is not too important anyway, we can move on from here and I am super OK with this being gone or looking completely different maybe already next week.

conbench/templates/benchmark-entity.html
jgehrcke: Thanks! I like that. And you also revealed your preference in the gray vs grey game :).

conbench/app/_plots.py
jgehrcke: Big thanks for this. I forgot to ping you on this specifically. Will update the comment and also the relevant pieces of text here.

> that comes straight from GitHub and represents a UTC time

Anecdote about this one (I think I have shared this before but :shrug:): the GitHub HTTP API shows tz-aware Zulu timestamps like `2011-04-14T16:00:49Z`. That is, those timestamps _unambiguously_ refer to UTC timezone. And then pygithub came in and in my opinion did a big mistake to transparently interpret these timestamps ans tz-naive ones. See `https://github.com/PyGithub/PyGithub/issues/512#issuecomment-1362654366.

When  we store and read commit timestamps -- let's make sure we treat the timezone information explicitly so that after all when data is displayed to users that there is no timezone ambiguity. I quickly created a reminder: https://github.com/conbench/conbench/issues/609.

conbench/app/_plots.py
jgehrcke: Absolutely, thanks for catching this. Debugging leftover I forgot to remove.

conbench/app/_plots.py
jgehrcke: Thanks!

jgehrcke: updated comment

conbench/app/_plots.py
jgehrcke: updated comment

conbench/templates/benchmark-entity.html
jgehrcke: thanks, updated to use "data point" instead of glyph

jgehrcke:
Thanks for the valuable feedback everyone. Will land this soon and then open a new PR for next chunk of small-ish changes.

Issue #599, opened 2023-01-17
Worker timeout when requesting compare/runs/<x>...<y>, 502 Bad Gateway seen by user agent


jgehrcke:
Seen by Antoine: 502 Bad Gateway response when GETting 
`conbench.ursa.dev/compare/runs/ef44eed05b61439ebfaf377a482d5f49...2d61d83ce5814bea96e47ebd04a625fe/`
`conbench.ursa.dev/compare/runs/a02a6dc9375543fea2f9a003f3bb341f...5c32ae93b4dd47ec854cb84ff040a770/`

Reproduced. Web application log around the time of request shows

```
Jan 17 15:39:29.254 i-0d9f4f0020ae49a35 conbench [2023-01-17 14:39:28 +0000] [389] [INFO] Booting worker with pid: 389
Jan 17 15:39:28.254 i-0d9f4f0020ae49a35 conbench [2023-01-17 14:39:27 +0000] [386] [INFO] Worker exiting (pid: 386)
Jan 17 15:39:28.254 i-0d9f4f0020ae49a35 conbench [2023-01-17 14:39:27 +0000] [1] [CRITICAL] WORKER TIMEOUT (pid:386)
```

We saw two symptoms, and can assess potential causes:

- The logs (datadog) show gunicorn worker timeouts around the time when requesting this compare view. That's probably because the worker process that is supposed to generate an HTTP response to the request doesn't do so within the deadline (30 seconds). In that case gunicorn seemingly does not generate an HTTP response at all, and terminates the worker process.

- The 502 Bad Gateway is then emitted from a reverse proxy (probably nginx, as part of k8s ingress controller) because the upstream (nginx term) probably terminates the TCP connection before it has sent a response.

I think there are two root issues:

- The HTTP response to those compare/runs/.... takes way too long to generate because of 'a lot of work to do'.
- There is no insight (via logs) about which part of that work takes very long.

We should explore why the response generation taskes so long (add logging, etc), and then envision architectural changes to make the compare/runs view 10x or 100x faster :).

jonkeane:
Separately from figuring out the root cause + ensuring that we rectify that. Is it possible to also add alerts (possibly in datadog?) when 502s?

austin3dickey:
I think it's likely that my history query PR added to the workload for requests like this. See discussion here: https://github.com/conbench/conbench/pull/507#issuecomment-1359892065

jgehrcke:
[...]

Issue #577, opened 2023-01-06
Move away from storing a rolling window distribution result in the database


jgehrcke:
Time series analysis / aggregation (such as rolling window analysis) for even many samples is extremely fast to calculate on the fly. 

Let's try to move to a world where there is no need to store these things in the database. 

Of course, if regression detection is based on such analysis then some results of such analysis may need to be stored in the database. But that's a different story.

jonkeane:
This will also come up when we start looking at multiple metrics for benchmarks. As it stands now storing one distribution isn't so bad (though I agree with JP that if we can make that faster by fetching the raw data and aggregating at runtime, we should do that!) but once we start needing to do that for the mean, median, max, and min of durations (and then also the min and max memory consumptions, the mean, median, max, min of through put ....) storing all of that starts to make less sense.

Issue #493, opened 2022-12-05
Introduce system metrics


jgehrcke:
Let's introduce the concept of system metrics to Conbench.

For better debuggability, for understanding the current system better, and for being able to health-monitor production instances.

Pick a library for providing a Prometheus metrics exporter endpoint. Start implementing basic counters so that we can measure HTTP request rate and error response rate for individual endpoints. Also start implementing basic timers so that we better understand expected timings around specific HTTP endpoints.

This relates to how we run the Flask app: currently we run N processes, each process processing at most one HTTP request at any given time. Should each process manage and export its own set of metrics? That would be quite a bit of overhead: for example in Kubernetes a single container would then expose N metrics endpoints that need to be scraped.

Maybe it's a good time to change the worker/deployment model where we use gunicorn's threaded workers. Per gunicorn instance we could run one process with N (actual) threads, each thread then processing at most one HTTP request at any given time.

austin3dickey:
I had the idle thought the other day that we could publish these kinds of metrics to a Conbench server itself. It would be a good source of training data for when we're testing alerting algorithms, etc. It could also help us ensure we don't introduce changes that drastically slow down the API.

jgehrcke:
> we could publish these kinds of metrics to a Conbench server itself

Interesting!

I have a very canonical stack in mind here that gets us to building useful dashboards right away. The typical flow is that the application provides a scrape endpoint using the Prometheus format (that need to be _pulled_ from, periodically). 

The data then needs to be written to a queryable timeseries database like Prometheus. Dashboards would then read from that data source. I trust that your idea has value, but my suspicion is that we should treat those ideas separately.

jgehrcke:
We also need to deploy a basic Prometheus to do the scraping, and set up a corresponding Grafana to explore/query and to build dashboards, alerts. Maybe deploy this into the k8s cluster that also runs Conbench?

austin3dickey:
I fleshed out the conbench-on-conbench idea in https://github.com/conbench/conbench/issues/606. I agree that this ticket is a completely different scope of work.

jgehrcke:
Path forward likely entails:
- Deploy Prometheus and Grafana on the conbench k8s via https://github.com/prometheus-operator/kube-prometheus (it's the standard, we used it exhaustively in our opstrace product).
[...]

PR #604, opened 2023-01-18
Speed up set_z_scores()

austin3dickey:
This PR speeds up `set_z_scores()` by using a better data structure. When there are many `benchmark_results`, each one was searching through the entire distribution query to find the 0 or 1 matches for its case/context, and that took a long time. We should have been using a hash lookup for this.

Helps with #599.

### DIFF
--- a/conbench/entities/distribution.py
+++ b/conbench/entities/distribution.py
@@ -207,31 +207,35 @@ def set_z_scores(benchmark_results: List["BenchmarkResult"]):
         if (baseline_commit, baseline_hardware) in cached_distributions:
             distributions = cached_distributions[(baseline_commit, baseline_hardware)]
         else:
-            distributions = (
-                Session.query(Distribution)
+            query_result = (
+                Session.query(
+                    Distribution.case_id,
+                    Distribution.context_id,
+                    Distribution.mean_mean,
+                    Distribution.mean_sd,
+                )
                 .filter_by(commit_id=baseline_commit, hardware_hash=baseline_hardware)
                 .all()
             )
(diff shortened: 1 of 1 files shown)


jonkeane:
We don't need to do this now (and it would be good to get this out quicker than later), but we _should_ make a benchmark for this (+ setup conbench on conbench to track it!) so that we don't accidentally revert this in the future. Would you mind making a separate issue (or issues) for that?

jgehrcke:
Cool! Really lovely analysis in https://github.com/conbench/conbench/issues/599#issuecomment-1387672510. Great work. I trust that this is a great algorithmic improvement (did not review the details). Let's land this.

So, we call `set_z_scores()` when GETting `compare/runs/<x>...<y>`, yes? It's a very important observation (architectural smell :-)). Relates to https://github.com/conbench/conbench/issues/577.

> We don't need to do this now (and it would be good to get this out quicker than later), but we should make a benchmark for this (+ setup conbench on conbench to track it!) so that we don't accidentally revert this in the future. 

Related to this: request duration measurements and alerts via observability for actual deployments: https://github.com/conbench/conbench/issues/493

jgehrcke (approved):
:rocket:

austin3dickey:
>We don't need to do this now (and it would be good to get this out quicker than later), but we should make a benchmark for this (+ setup conbench on conbench to track it!) so that we don't accidentally revert this in the future. Would you mind making a separate issue (or issues) for that?

Yes, good idea. I had mentioned this thought earlier in #493 but it belongs its own ticket. I just created #606.

PR #610, opened 2023-01-20
Bump to CPython 3.11, remove dateutil dep, introduce tznaive_iso8601_to_tzaware_dt()

jgehrcke:
This introduces a new function `tznaive_iso8601_to_tzaware_dt()` that I've hopefully specified precisely, and uses that across the code base. The underlying string parsing was changed to be stdlib-based, and an outcome is that we can remove the third-party dependency to python-dateutil.

There is also a change in behavior here that is important: the datetime objects returned by this new function are tz-aware, which is going to be helpful for code maintenance and for displaying correct values to users in the future.

This needs CPython 3.11 to make use of the valuable additions to `datetime.fromisoformat(s)` (e.g. to, not be tripped up by the 'Z'): https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat

### DIFF
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -5,13 +5,13 @@ import os
 from datetime import datetime
 from typing import List, Optional
 
-
 import flask as f
 import requests
 import sqlalchemy as s
 from sqlalchemy.orm import Query
 
 from conbench import util
+
 from ..db import Session
 from ..entities._entity import (
     Base,
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -5,12 +5,13 @@ import os
 from datetime import datetime
 from typing import List, Optional
 
-import dateutil.parser
+
 import flask as f
 import requests
 import sqlalchemy as s
 from sqlalchemy.orm import Query
 
+from conbench import util
 from ..db import Session
 from ..entities._entity import (
     Base,
@@ -546,7 +547,7 @@ class GitHub:
             "parent": commit["parents"][0]["sha"] if commit["parents"] else None,
             # Note(JP): this might need attention with respect to time zones.
             # Also see https://github.com/PyGithub/PyGithub/issues/512#issuecomment-1362654366
-            "date": dateutil.parser.isoparse(commit_author["date"]),
+            "date": util.tznaive_iso8601_to_tzaware_dt(commit_author["date"]),
             # Note(JP): don't we want to indicate if the msg was truncated,
             # with e.g. an ellipsis?
             "message": commit["commit"]["message"].split("\n")[0][:240],
--- a/conbench/util.py
+++ b/conbench/util.py
@@ -45,10 +45,17 @@ def tznaive_iso8601_to_tzaware_dt(
     """
 
     def _convert(s: str):
-        # Do some sanity-checking. If this happens, only emit the warning
-        # but then still hard-set UTC timezone.
-        if "Z" in s or "+" in s:
-            log.warning("expected tz-naive timestring, but saw: %s", s)
+        # Do some sanity-checking. If unexpected input is seen, only emit the
+        # warning but then still hard-set UTC timezone.
+        if "Z" in s:
+            # Input seems to be tz-aware but the timezone it specifies matches
+            # the one we want to set anyway.
+            log.warning("expected tz-naive timestring, but saw UTC: %s", s)
+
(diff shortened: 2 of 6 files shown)


conbench/util.py
```diff
@@ -19,6 +22,44 @@
...
+    knowledge).
+    """
+
+    def _convert(s: str):
+        # Do some sanity-checking. If this happens, only emit the warning
+        # but then still hard-set UTC timezone.
+        if "Z" in s or "+" in s:
+            log.warning("expected tz-naive timestring, but saw: %s", s)
+
+        return datetime.fromisoformat(s).replace(tzinfo=timezone.utc)
```
jgehrcke, line 60: Incredibly important: the `replace()` on tzinfo is non-magic, i.e. it does not try to convert from one timezone into another by tweaking the numerical values of the hour, etc.

Corresponding docs:

> If you merely want to attach a time zone object tz to a datetime dt without adjustment of date and time data, use dt.replace(tzinfo=tz)

from https://docs.python.org/3/library/datetime.html.

(a common mistake is to use `datetime.astimezone()` which tries to be smart, often too smart).

jgehrcke:
Super happy about feedback, but I am already going ahead with this to keep the chaos across my local branches small.

conbench/entities/commit.py
```diff
@@ -546,7 +547,7 @@ def _parse_commit(commit):
             "parent": commit["parents"][0]["sha"] if commit["parents"] else None,
             # Note(JP): this might need attention with respect to time zones.
             # Also see https://github.com/PyGithub/PyGithub/issues/512#issuecomment-1362654366
```
austin3dickey, line 549: I think you can remove this comment now 🙂

conbench/entities/commit.py
```diff
@@ -546,7 +547,7 @@ def _parse_commit(commit):
...
-            "date": dateutil.parser.isoparse(commit_author["date"]),
+            "date": util.tznaive_iso8601_to_tzaware_dt(commit_author["date"]),
```
austin3dickey, line 550: `commit_author["date"]` comes from the GitHub API and will always be tz-aware. I've always seen the form `"2011-04-14T16:00:49Z"` but the [API docs](https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#get-a-commit) have the example `"2007-10-29T02:42:39.000-07:00"` in the response schema, so that's either imprecise documentation or it could be in that form in some cases.

In any case, as the code stands we'll always be emitting a `log.warning()` when running this line. I think we should update the util function to expect that as an input without warning, or add a different util function that we could use here.

austin3dickey (commented):
Thanks!

conbench/entities/commit.py
jgehrcke: Oh, I also remember that at some point today I had the definite plan to remove that warning (because I also saw that sometimes we indeed hit this case, and it's actually fine), and then forgot. Thanks for catching this.

jgehrcke: Ah, more thinking:

> commit_author["date"] comes from the GitHub API and will always be tz-aware.

Yes, the GitHub HTTP API always sends tz-aware ISO 8601 strings. 

But pygithub _sometimes_ (?) turns these into tz-naive datetime objects. That's what https://github.com/PyGithub/PyGithub/issues/512 is about.

So I think you are saying here that `commit_author["date"]` is guaranteed to be the time string as returned by the GitHub HTTP API, yes?

I was kind of planning for the idea that this timestring underwent some lossy information flow with pygithub having been involved. If we know that this does not happen here that's great and that would be the main reason to remove or change my comment above.

austin3dickey: That's correct—we don't use `pygithub`, only `requests`, and receive a string in `commit_author["date"]` that's tz-aware straight from the HTTP API.

conbench/entities/commit.py
jgehrcke: Updating the comment in an upcoming PR

jgehrcke: done here: https://github.com/conbench/conbench/pull/612

conbench/entities/commit.py
jgehrcke: > That's correct—we don't use pygithub

Oh, that's somewhat funny. Thanks for clarifying, and I should have known/validated my assumption.

PR #611, opened 2023-01-20
history plot: remove slider, tweak spacing & ci: take screenshot of benchmark result view

jgehrcke:


jgehrcke:
This PR was a strong motivator to build a simple version of "CI, please, take a screenshot of the history plot view". I've done that here.

The artifact archive now contains a `benchmark-result-view.png` which indeed shows the result view and history plot. Super valuable.

.github/workflows/actions.yml
```diff
@@ -83,19 +83,21 @@ jobs:
...
-            make db-populate
-      - name: screenshot
-        run: |
+            export CONBENCH_BASE_URL=http://$(docker compose port app 5000)
+            export SOME_BENCHMARK_ID="$(make -s db-populate)"
+            echo "conbench base URL: $CONBENCH_BASE_URL"
+            echo "a valid benchmark ID: $SOME_BENCHMARK_ID"
             cd ci
             docker build . -t conbench-screenshot -f screenshot.Dockerfile
             export CONBENCH_BASE_URL=http://$(docker compose port app 5000)
```
austin3dickey: This is duplicated now, no?

jonkeane:
> The artifact archive now contains a benchmark-result-view.png which indeed shows the result view and history plot. Super valuable.

Love this! It's sooo much easier to confirm things like this now. Is it possible to _also_ include a selected-point screenshot? I know that might be more complicated than it sounds if it involves simulating mouse movements and activity, but if it's easy it might be nice!

ci/screenshot.py
```diff
@@ -41,6 +41,11 @@
 )
 
 
+# We bumped this to capture Bokeh plot renderings which sometimes take a little
+# while. It would however be better to dynamically respond to a render event.
+SLEEP_BEFORE_SCREENSHOT_SECONDS = 3.0
```
jonkeane: :100: We might even have that "it would be better" marked as a TODO

.github/workflows/actions.yml
```diff
@@ -83,19 +83,21 @@ jobs:
         uses: actions/checkout@v3
       - name: run-app-bg
         run: make run-app-bg
-      - name: populate DB
+      - name: populate DB & take screenshots
```
austin3dickey, line 86: Thanks for this change! To me, this suggests we should clean and merge https://github.com/conbench/conbench/pull/592 instead of letting it go. I like the fake data in that PR better than the existing fake data, especially for these history plots. :)

conbench/tests/populate_local_conbench.py
```diff
@@ -392,9 +398,13 @@ def create_benchmarks_with_history():
                         mean=mean,
                     )
 
-                post_benchmarks(benchmark_data)
+                benchmark_id = post_benchmarks(benchmark_data)
+                benchmark_ids.append(benchmark_id)
                 runs.append((run_id, timestamp))
 
+    log.info("now, emit a benchmark ID on stdout")
+    print(benchmark_ids[0])
```
jonkeane: Would you mind commenting on why we are sending the id to stdout instead of log here?

jonkeane (approved):
Thanks for this! A few (small) comments, but none are blocking

conbench/app/_plots.py
```diff
@@ -630,28 +630,4 @@ def time_series_plot(history, benchmark, run, height=380, width=1100):
     # https://github.com/bokeh/bokeh/pull/2186
     p.xaxis.ticker.desired_num_ticks = 9
 
-    range_tool = bokeh.models.RangeTool(x_range=p.x_range)
-    range_tool.overlay.fill_color = "gainsboro"
-    range_tool.overlay.fill_alpha = 0.2
-
-    select = bokeh.plotting.figure(
-        title="drag the middle and edges of the selection box to change the range above",
```
austin3dickey, line 638: We could do this in a separate PR if you wish, but would you mind enabling the feature of "click/drag on the plot to create a selection range to zoom in to"? I find myself constantly trying to do that on these pages and being disappointed when it pans instead

austin3dickey (approved):
Thanks!

.github/workflows/actions.yml
jgehrcke: Yeah absolutely. Had the very same thought, but want do this this in separate PRs, step by step. I look at the current populate_local_conbench.py as very temporary.

My plan is certainly to land https://github.com/conbench/conbench/pull/592, or a derivative thereof, asap.

.github/workflows/actions.yml
jgehrcke: Yes, thanks!

conbench/app/_plots.py
jgehrcke: Ack! Let's do this in a separate PR, yes. I also had this 'review/overhaul the Bokeh tools for the history plot' in my mind as wonderful topic for an individual, small PR.

conbench/tests/populate_local_conbench.py
jgehrcke: In this PR here it's used in CI via `export SOME_BENCHMARK_ID="$(make -s db-populate)"`. 

Will add a comment now, but I don't expect this particular code here to stay around for too long though -- https://github.com/conbench/conbench/pull/611/files#r1082918585

jgehrcke:
> Is it possible to also include a selected-point screenshot? I know that might be more complicated than it sounds if it involves simulating mouse movements and activity, but if it's easy it might be nice!

Yes! I have also done something like this extensively before with [playwright](https://playwright.dev/) and other tools. But this will indeed be bit of a bigger chunk of work and that should be done separately.

There will be the complication that 'clicking a glyph in a Bokeh plot' is a fancy special case! Normally you do CSS selectors and the likes to identify a particular element of the DOM tree (say, a div), and then you can send a click event exactly there. In case of the Bokeh plot we have a WebGL canvas that's a single entity from the DOM's point of view, and we might have to actually calculate virtual screen coordinates based on same fancy coordinate system transformation. Well :). I thought about adding this and found "too much work for now, must be done independently".

But yeah that kind of automated and precise UI testing is something we want to have at some point. It is a thing I have always supported or introduced myself because the amount of confidence it can yield.

ci/screenshot.py
jgehrcke: added

conbench/tests/populate_local_conbench.py
jgehrcke: added a code comment

.github/workflows/actions.yml
jgehrcke: removed

jgehrcke:
Thanks for the feedback! I have added another commit will land this now.

conbench/tests/populate_local_conbench.py
jonkeane: > In this PR here it's used in CI via export SOME_BENCHMARK_ID="$(make -s db-populate)".

AAAAH thanks for this. I grepped for the script name, but hadn't looked in the Makefile. That makes sense (and also very much explains the split behavior, now!)

jonkeane:
> Yes! I have also done something like this extensively before with [playwright](https://playwright.dev/) and other tools. But this will indeed be bit of a bigger chunk of work and that should be done separately.

Indeed, that's what I suspected. And strong agree it's out of scope for this if it's that complicated. I'm not sure if anyone even uses nightwatch anymore, but I have flashbacks to a testing system I worked with a bit years ago — magic when it works, very finicky when it doesn't.

PR #613, opened 2023-01-23
Search & UI: Unique data count based search type implementation for individual columns

### LINKED ISSUE #602: Free text search for columns
> > But seeing this made me realize: it would be really nice if we could have drop down searches for low cardinality columns, and then free-form text search for all other columns. We can (and should!) do that as a follow on unless you think it's easier to tack on here.

> The existing Search bar is still functional and works well for free text searches. Probably we could utilize that?

IMO we should have both: the overall search bar searching is nice in cases where someone might not know which column an element is in. But it also would be fantastic to be able to let users search a specific column if they know that's what they're looking for. It would also allow this interface to degrade nicely in the case that a column that we think should be low cardinality but in reality end up being high cardinality. Additionally, I find it a bit bothersome visually that only some of the columns have an interface element (i.e. drop down or search bar) and others don't

_Originally posted by @jonkeane in https://github.com/conbench/conbench/issues/584#issuecomment-1385768552_

### LINKED ISSUE #619: CSS updates for search box: conbench webapp
Currently #584 uses default search box styling

The updated CSS changes are present under #613.

### LINKED ISSUE #629: Decouple order & search feature: conbench UI
> I have disabled sorting/ordering for all the columns except Run, which helps improve the search behavior and disables the same on click action.

Instead of getting rid of the sorting feature (I'm not sure how widely it's used, but I know I use it _frequently_), is it possible to the search or picker option in a new row just below the header? That way the header can still be orderable, but we can also search or filter as well? This isn't the only way, but it would be mice to have both of these features at the same time (and I'm sure there's a way we can preserve them)

_Originally posted by @jonkeane in https://github.com/conbench/conbench/issues/613#issuecomment-1402746098_

### LINKED ISSUE #642: Conbench search: Author col with uncaptured `null`/`empty` instance
The `empty dropdown placeholder` should have been captured with the null condition under `dropdown_column_search`

Screenshot 2023-01-31 at 11 14 26 AM

sayantikabanik:
closes #602 
closes #629
closes #619 
closes #642 

#### Details & preview

- Adds additional capabilities for individual column search 

https://user-images.githubusercontent.com/17350312/214005656-f866bfe9-e030-453a-9bdf-d6ad431aa62d.mov

#### Search trigger options
1. Currently the free text is triggered on every keypress/input  
```js
on('keyup change clear', function() {
                that
                .search(this.value)
                .draw();
)};
```

2. trigger only when `enter` is pressed. Reduces the number of API pings + efficient. Little counter intuitive from the existing search option.
```js
.on('keydown', function(key_opt) {
   if (key_opt.keyCode == 13) { //search activated only on enter key-press
                that
                .search(this.value)
                .draw();
            }
)};
```

### DIFF
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -227,7 +227,7 @@ function column_search_implementation(data_table_id) {
                 free_text_column_search(data_table_id, index)
             }
             else if (item == 1) {
-                $(`<p class="message-header"> 👀 only one value found</p>`).appendTo(
+                $(`<p class="message-header"> 👀 only one/null value found</p>`).appendTo(
                     table.columns([index]).header())
             }
             else{
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -150,7 +150,7 @@ INPUTS:
  data_table_id: string
  column_indexes: array
  */
-function specific_column_search(data_table_id, column_indexes) {
+function dropdown_column_search(data_table_id, column_indexes) {
     $(document).ready(function() {
         var table = $(data_table_id).DataTable();
         table.columns(column_indexes).every(function() {
@@ -208,7 +208,12 @@ function free_text_column_search(data_table_id, column_indexes) {
 
 /*
 INFO:
-    Function to check unique value count and set the required search type
+    Function to check unique value count and set the required search type.
+        - Unique values <10 & >1 [specific_column_search]: drop-down based search is implemented
+        - Unique values >10 [free_text_column_search]: free text search is implemented
+        - If there is no data OR only single data point: a message is displayed ("only one/no value found")
+INPUT:
+    data_table_id: string
  */
 function column_search_implementation(data_table_id) {
     $(document).ready(function() {
@@ -221,7 +226,7 @@ function column_search_implementation(data_table_id) {
         });
         unique_count.forEach(function (item, index) {
             if (item < 10 && item > 1) {
-                specific_column_search(data_table_id, index)
+                dropdown_column_search(data_table_id, index)
             }
             else if (item > 10){
                 free_text_column_search(data_table_id, index)
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -230,6 +230,9 @@ function column_search_implementation(data_table_id) {
                 $(`<p class="message-header"> 👀 only one value found</p>`).appendTo(
                     table.columns([index]).header())
             }
+            else{
+                console.log("Unidentified value encountered")
+            }
         });
     })
 }
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -141,7 +141,8 @@ function enable_search_query_string(data_table) {
 /*
 INFO:
     Adds free text search ability for specific columns, based on selected column indexes.
-    Press enter after adding text to search box, which triggers column sorting/search operation
+    Based on the indexes of the columns, input boxes are created. The search is triggered on
+    every key press.
 INPUTS:
  data_table_id: string
  column_indexes: array
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -140,7 +140,8 @@ function enable_search_query_string(data_table) {
 
 /*
 INFO:
-    Adds free text search ability for specific columns in the DataTable
+    Adds free text search ability for specific columns, based on selected column indexes.
+    Press enter after adding text to search box, which triggers column sorting/search operation
 INPUTS:
  data_table_id: string
  column_indexes: array
@@ -149,18 +150,18 @@ function free_text_column_search(data_table_id, column_indexes) {
     $(document).ready(function() {
         var table = $(data_table_id).DataTable();
         table.columns(column_indexes).every(function() {
-            var title = $(this).text();
-            $(this).html('<input type="text" placeholder="Search ' + title + '" />');
+            var that = this;
+            var title = this.header().textContent;
+            $('<input type="text" placeholder="Search ' + title + '" />')
+                .appendTo(this.header())
+                .on('keydown', function(key_opt) {
+                    if (key_opt.keyCode == 13) { //only on enter keypress (code 13)
+                        that
+                            .search(this.value)
+                            .draw();
+                    }
+                });
         });
-
-        table.columns(column_indexes).every(function (){
-            var that= this;
-            $('input', this.headers()).on('keyup change', function (){
-                if (that.search()!== this.value){
-                    that.search(this.value).draw();
-                }
-            })
-        })
     })
 }
 
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -156,22 +156,22 @@ function specific_column_search(data_table_id, column_indexes) {
         table.columns(column_indexes).every(function() {
             var that = this;
             var reset = ""
-            var mySelectList = $("<select></select>")
+            var mySelectList = $("<select class='custom-search-box'></select>")
                 .appendTo(this.header())
                 .on("change", function() {
                     that.search($(this).val()).draw();
                 });
             mySelectList.append(
-                $(`<option class="custom-search-box" value="${reset}">${reset}</option>`)
+                $(`<option value="${reset}">${reset}</option>`)
             );
             this.cache("search").unique().sort().each(function(param) {
                 if (param !== "") {
                     mySelectList.append(
-                        $(`<option class="custom-search-box" value="${param}">${param}</option>`)
+                        $(`<option value="${param}">${param}</option>`)
                     );
                 } else {
                     mySelectList.append(
-                        $(`<option class="custom-search-box" value="${param}">null</option>`)
+                        $(`<option value="${param}">null</option>`)
                     );
                 }           
             });
@@ -227,7 +227,7 @@ function column_search_implementation(data_table_id) {
                 free_text_column_search(data_table_id, index)
             }
             else if (item == 1) {
-                $(`<p class="message-header"> 👀 only one/null value found</p>`).appendTo(
+                $(`<p class="message-header"> 👀 only one/no value found</p>`).appendTo(
                     table.columns([index]).header())
             }
             else{
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -138,6 +138,32 @@ function enable_search_query_string(data_table) {
     });
 }
 
+/*
+INFO:
+    Adds free text search ability for specific columns in the DataTable
+INPUTS:
+ data_table_id: string
+ column_indexes: array
+ */
+function free_text_column_search(data_table_id, column_indexes) {
+    $(document).ready(function() {
+        var table = $(data_table_id).DataTable();
+        table.columns(column_indexes).every(function() {
+            var title = $(this).text();
+            $(this).html('<input type="text" placeholder="Search ' + title + '" />');
+        });
+
+        table.columns(column_indexes).every(function (){
+            var that= this;
+            $('input', this.headers()).on('keyup change', function (){
+                if (that.search()!== this.value){
+                    that.search(this.value).draw();
+                }
+            })
+        })
+    })
+}
+
 $("#confirm-delete").on("show.bs.modal", function(e) {
     $(this).find("#confirm-delete-message").html($(e.relatedTarget).data("message"));
     $(this).find("#delete-form-id").attr("value", $(e.relatedTarget).data("form-id"));
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -142,7 +142,7 @@ function enable_search_query_string(data_table) {
 INFO:
     Adds free text search ability for specific columns, based on selected column indexes.
     Based on the indexes of the columns, input boxes are created. The search is triggered on
-    every key press.
+    key press.
 INPUTS:
  data_table_id: string
  column_indexes: array
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -153,9 +153,10 @@ function free_text_column_search(data_table_id, column_indexes) {
         table.columns(column_indexes).every(function() {
             var that = this;
             var title = this.header().textContent;
-            $('<input class="custom-input-text custom-search-box" type="text" placeholder="🔍 ' + title + '" />')
+            $(`<input class="custom-input-text custom-search-box" type="text" placeholder="🔍 ${title}"
+                onClick="if(event.stopPropagation!==undefined){event.preventDefault(); event.stopPropagation();}event.cancelBubble=true;"/>`)
                 .appendTo(this.header())
-                .on('keyup change clear', function() { //column search/sorting triggered on every input
+                .on('keyup change clear', function() { //column search/sorting triggered on keypress
                             that
                             .search(this.value)
                             .draw();
@@ -163,6 +164,7 @@ function free_text_column_search(data_table_id, column_indexes) {
         });
     })
 }
+
 
 $("#confirm-delete").on("show.bs.modal", function(e) {
     $(this).find("#confirm-delete-message").html($(e.relatedTarget).data("message"));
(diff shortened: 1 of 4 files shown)


jgehrcke:
Thank you for this and for preparing the video! Quick question: I saw that when you moved the cursor into the Commit search field the table contents changed (before you started typing). Is that expected?

austin3dickey:
>I saw that when you moved the cursor into the Commit search field the table contents changed (before you started typing).

Locally playing with this, I think this is because when you click into the text box, it sorts the column.

conbench/templates/app.html
```diff
@@ -138,6 +138,32 @@ <h4 class="modal-title"><span class="glyphicon glyphicon-trash submenu"></span>&
...
+ data_table_id: string
+ column_indexes: array
+ */
+function free_text_column_search(data_table_id, column_indexes) {
+    $(document).ready(function() {
+        var table = $(data_table_id).DataTable();
+        table.columns(column_indexes).every(function() {
+            var that = this;
+            var title = this.header().textContent;
+            $('<input type="text" placeholder="Search ' + title + '" />')
```
austin3dickey: (This is a nitpicky comment, feel free to ignore.) Is there a way to use a different CSS for the placeholder text? The big, bold letters distract me a little. I like the way the top-level search box is styled.

conbench/templates/index.html
```diff
@@ -121,6 +121,6 @@ <h3>Language-independent Continuous Benchmarking (CB) Framework</h3>
 });
 
 enable_search_query_string('#runs');
-
+free_text_column_search('#runs', [1,3,4]);
```
austin3dickey: I wouldn't mind being able to search the commit message too. I noticed that column is skinnier now, so maybe the search box would widen it a little too?

conbench/templates/app.html
```diff
@@ -138,6 +138,32 @@ <h4 class="modal-title"><span class="glyphicon glyphicon-trash submenu"></span>&
...
+                .appendTo(this.header())
+                .on('keyup change clear', function() { //column search/sorting triggered on every input
```
austin3dickey: I like this choice! I don't think it hits the API every keystroke; it searches through the 1000 results already in-memory.

austin3dickey (approved):
Looks great, thanks! Just some style comments, as usual 🙂

sayantikabanik:
> > I saw that when you moved the cursor into the Commit search field the table contents changed (before you started typing).
> 
> Locally playing with this, I think this is because when you click into the text box, it sorts the column.

Right, I will give a try to suppress this behavior

conbench/templates/app.html
sayantikabanik: great :)

conbench/templates/index.html
sayantikabanik: Do we have instances when we had to query the "commit message" col?
Trying to stretch my understanding here.

The width will remain same[even after addition of search]...
Since I am using the default search box setting from the Datatables. 

Screenshot 2023-01-24 at 10 55 49 AM

conbench/templates/app.html
sayantikabanik: ✅ totally, shall tweak the representation

sayantikabanik:
### TODO(s)

- [x] Suppress auto sorting of column on click 
- [x] Tweak the style of search boxes
- [x] More CSS to customize the search bar
- [x] fix commit col length 
- [x] add search for commit message

conbench/templates/index.html
sayantikabanik: CSS is updated, commit col maintains it's width.✅

sayantikabanik:
Ended up making couple of more changes, requesting a re-review.

I have disabled sorting/ordering for all the columns except `Run`, which helps improve the search behavior and disables the same on click action.

conbench/templates/index.html
```diff
@@ -113,14 +113,18 @@ <h3>Language-independent Continuous Benchmarking (CB) Framework</h3>
...
+  var table = $('#runs').DataTable({
       "order": [[0, 'desc']],
       "language": {
           "searchPlaceholder": "commit hash works too",
-      }
+      },
+      "columnDefs": [{
+          "targets": [1,2,3,4,5,6],
+          "orderable": false
+      }]
```
sayantikabanik: This patch of code disables the ordering of selected columns.

conbench/templates/app.html
```diff
@@ -138,6 +138,32 @@ <h4 class="modal-title"><span class="glyphicon glyphicon-trash submenu"></span>&
...
+ data_table_id: string
+ column_indexes: array
+ */
+function free_text_column_search(data_table_id, column_indexes) {
+    $(document).ready(function() {
+        var table = $(data_table_id).DataTable();
+        table.columns(column_indexes).every(function() {
+            var that = this;
+            var title = this.header().textContent;
+            $('<input class="custom-input-text custom-search-box" type="text" placeholder="🔍 ' + title + '" />')
```
sayantikabanik: The use of emoji, helped in capping the width of the column + visual treat for `searchability` feature

sayantikabanik:
### Updated preview

Screenshot 2023-01-24 at 2 52 19 PM

conbench/templates/app.html
austin3dickey: Love this!

austin3dickey (approved):
Looks fantastic!

jonkeane:
> I have disabled sorting/ordering for all the columns except Run, which helps improve the search behavior and disables the same on click action.

Instead of getting rid of the sorting feature (I'm not sure how widely it's used, but I know I use it _frequently_), is it possible to the search or picker option in a new row just below the header? That way the header can still be orderable, but we can also search or filter as well? This isn't the only way, but it would be mice to have both of these features at the same time (and I'm sure there's a way we can preserve them)

austin3dickey:
>'m not sure how widely it's used, but I know I use it frequently

Just curious, what columns do you use it for? After we merge this PR and #584, you'll still be able to sort the Run column, and either free-text search or dropdown filter all the other columns.

jonkeane:
> Just curious, what columns do you use it for? After we merge this PR and https://github.com/conbench/conbench/pull/584, you'll still be able to sort the Run column, and either free-text search or dropdown filter all the other columns.

Most of these on other views than the runlist one, but I want to make sure that we can roll this out to all of the other tables ASAP if not in this PR, so think doing this shortcut now is making it harder for us to ship this to the rest of the UI in the future. Off the top of my head: reason, benchmark name, language, mean, z-score, error,  batch. Some of those (like Error) would be just fine as a dropdown), but for others it can be super helpful to be able to order by them.

sayantikabanik:
I can try cloning the header to another row and see how it comes up. 
Not yet sure how quickly that cloud patch up. ~Probably both this PR and the drop-down search PR will require this update for design symmetry.~

sayantikabanik:
UPDATE: 

 https://github.com/conbench/conbench/issues/629#issuecomment-1406097388
I tested this in my local and worked real well + no major change in design

sayantikabanik:
Have updated the search type logic. The search type is based on the unique data count logic and dynamic in nature.

`observation:` for me the "commit message" column switched from free text to drop-down when I updated the data in my local. Might not be the same in prod.

conbench/templates/app.html
```diff
@@ -179,6 +179,69 @@ <h4 class="modal-title"><span class="glyphicon glyphicon-trash submenu"></span>&
     });
 }
 
+/*
+INFO:
+    Adds free text search ability for specific columns, based on selected column indexes.
+    Based on the indexes of the columns, input boxes are created. The search is triggered on
+    key press.
```
jgehrcke, line 188: Love those docstrings!

conbench/templates/app.html
```diff
@@ -179,6 +179,69 @@ <h4 class="modal-title"><span class="glyphicon glyphicon-trash submenu"></span>&
...
+        });
+    })
+}
+
+/*
+INFO:
+    Function to check unique value count and set the required search type.
+        - Unique values <10 & >1 [specific_column_search]: drop-down based search is implemented
+        - Unique values >10 [free_text_column_search]: free text search is implemented
+        - If there is no data OR only single data point: a message is displayed ("only one/no value found")
```
jgehrcke: :heart:

conbench/templates/app.html
```diff
@@ -179,6 +179,69 @@ <h4 class="modal-title"><span class="glyphicon glyphicon-trash submenu"></span>&
...
+        });
+        unique_count.forEach(function (item, index) {
+            if (item < 10 && item > 1) {
+                dropdown_column_search(data_table_id, index)
+            }
+            else if (item > 10){
+                free_text_column_search(data_table_id, index)
+            }
+            else if (item == 1) {
+                $(`<p class="message-header"> 👀 only one/no value found</p>`).appendTo(
```
austin3dickey: I don't think we need this placeholder; it would probably look cleaner to just do nothing. :)

image

conbench/templates/app.html
```diff
@@ -179,6 +179,69 @@ <h4 class="modal-title"><span class="glyphicon glyphicon-trash submenu"></span>&
...
+                    table.columns([index]).header())
+            }
+            else{
+                console.log("Unidentified value encountered")
```
austin3dickey, line 245: I think this branch is being taken for the Author column. Do you know why that is?

austin3dickey (commented):
Love these changes!

conbench/templates/app.html
jonkeane: Yeah, agreed. I'm fine with no placeholder if that's simplest (someone can sort and re-sort if they do believe that there should be more | should be a search bar). But if we do put one it would be nice if it was small enough that een on this column it didn't cause overflowing on lines (maybe a simple `-` or something?)

conbench/templates/run.html
```diff
@@ -189,7 +189,7 @@ <h6>Top Outliers</h6>
 } );
 
 enable_search_query_string('#benchmarks');
-specific_column_search('#benchmarks', [1]);
+dropdown_column_search('#benchmarks', [1]);
```
jonkeane: Should this be `column_search_implementation`? That would end in there being a dropdown if the number of elements is small(ish) right?

sayantikabanik: The functions `dropdown_column_search` & `free_text_column_search` works for now in standalone mode. Didn't make the change here yet (while testing)

Shall update :)

conbench/templates/app.html
sayantikabanik: Ah interesting, could you share the author col data loaded in your local. 
I am not able to replicate the above in my system

sayantikabanik:
TODO(s)
- [x] Remove the messaging for single data point search + related docstrings 
- [x] diagnose the author col search bug (data needed to investigate further)
      - The `whitespace` fix #642  might be the issue, but happy to dig further 
- [x] Update `run.html` with latest search changes

conbench/templates/app.html
sayantikabanik: Screenshot 2023-01-31 at 11 14 26 AM
I get a drop-down here

austin3dickey: Ah, if it's just my system let's not worry about it. If this issue persists in prod, we can look into it further.

austin3dickey (approved):

sayantikabanik:
The `lint` failure isn't related to this PR i believe: Would appreciate a heads-up before I merge this ✨
For any potential followup/fixes will create separate issues.

jgehrcke:
Cool :) The same thought as [before](https://github.com/conbench/conbench/pull/630#issuecomment-1406186211): the changes affect the UI's landig page, right? Can we add a corresponding prefix to the PR title before we merge? Something like `UI: landing page: ...`. That will result in a commit message that very quickly reveals the area in which the commit brings changes.

> The lint failure isn't related to this PR i believe: Would appreciate a heads-up before I merge this

Hm. `make lint` passed on `main` about 12 hours ago. Will double-check this and get back to this PR here. 

> `Merge branch 'main' into sb/free_text_search_datatable`

Personally I always do a rebase to catch up with main. `git rebase -i origin/main`, (de)select commits, resolve merge conflicts, then force-push.

jgehrcke:
>  Will double-check this and get back to this PR here.

The problem was https://github.com/conbench/conbench/issues/652. Once https://github.com/conbench/conbench/pull/653 lands our `main` branch is healthy again. You can then rebase on `main`, and this CI failure will be gone from this PR branch here, too.

jgehrcke:
@sayantikabanik I saw the `ci/ui-screenshots` failure here and investigated it. It's now tracked here: https://github.com/conbench/conbench/issues/654. It's unrelated to your PR, certainly.

jgehrcke:
In the current screenshot-created-by-CI I see we have a search box now for `Run creation time`:
image

Is that a deliberate goal of this patch? (was a little hard to extract from the above's discussion, my opinion is that we don't need a search box on this column)

I know I am a little late with this following kind of input, but it's important I think. I found that in the DataTables ecosystem this kind of functionality is called "search panes". From https://datatables.net/extensions/searchpanes/:

> SearchPanes adds panes to the DataTable with the capability to search the DataTable by selecting rows in the panes. [...] SearchPanes can search DataTables for multiple values that have been selected across multiple panes. They also provide the ability to define custom search functions which cannot be achieved through a simple searchBox.

There is a cool example here: https://datatables.net/extensions/searchpanes/examples/initialisation/viewTotal.html

The 'unique value count threshold' technique that we identified here over time is an established concept. From https://datatables.net/extensions/searchpanes/examples/initialisation/threshold.html:

> The ratio of uniqueness is the number of different values throughout the table divided by the total number of values, i.e. as the number approaches 1 the more unique the values are. ..... The default for [searchPanes.threshold](https://datatables.net/reference/option/searchPanes.threshold) is 0.6.

There are many nice "search pane" examples in their docs: https://datatables.net/extensions/searchpanes/examples

I don't have a strong recommendation or conclusion with respect to _this_ PR here! Let's keep this in the back of our minds for future development, and let's be aware of the pros and cons of using established building blocks vs. building our own :).

jonkeane:
> I know I am a little late with this following kind of input, but it's important I think. I found that in the DataTables ecosystem this kind of functionality is called "search panes". From https://datatables.net/extensions/searchpanes/

💯 suggestion here @jgehrcke when the first of these search PRs came along I thought the same thing and searched and searched for something like `searchpanes` but _could not find it_ at the time. I did find a bunch of suggestions that took this same approach as we did here. I totally agree we should use a standard building block if at all possible. I agree that we might not want to swap in `searchpanes` just now since this is pretty near completion  (though we should definitely create an issue to transition to that in the future), unless using `searchpanes` would actually end up being super quick and easy (in which case we can repurpose this PR and do that now)

> In the current screenshot-created-by-CI I see we have a search box now for Run creation time

So long as this matches on partial strings, I'm pro having this searchable. I've definitely wanted to find benchmarks from e.g. 2022-12-30 before and had to sort (well it's already reverse sorted...) and then page page page to get to what I was looking for.

jgehrcke:
> So long as this matches on partial strings, I'm pro having this searchable. I've definitely wanted to find benchmarks from e.g. 2022-12-30 before and had to sort (well it's already reverse sorted...) and then page page page to get to what I was looking for.

Oh, yeah makes perfect sense!

sayantikabanik:
> know I am a little late with this following kind of input, but it's important I think. I found that in the DataTables ecosystem this kind of functionality is called "search panes". From https://datatables.net/extensions/searchpanes/:

>I agree that we might not want to swap in searchpanes just now since this is pretty near completion (though we should definitely create an issue to transition to that in the future), unless using searchpanes would actually end up being super quick and easy (in which case we can repurpose this PR and do that now)

This is brilliant, I took a peek to evaluate if `searchpanes` could smoothly + quickly replace the current logic. We would require bit more customization over the defaults. Also there are some cool features that searchpanes brings in, might be interesting to implement (example: range filters in `z-score` value:[ similar example](https://datatables.net/extensions/searchpanes/examples/customFiltering/customOptionConditions.html)).

I will create a separate ticket to document the findings.

Issue #609, opened 2023-01-20
Commit/Run/BenchmarkResult/... timestamps: clarify timezone across codebase


jgehrcke:
Conbench should after all be useful for globally distributed teams that operate at the same time across the various timezones of the world. We have to address timezone confusion systematically.

Let's do one of or some of or all of:
- add code comments
- turn tz-naive in tz-aware datetime objects before storing in database

 Relates to https://github.com/conbench/conbench/issues/534.

jonkeane:
💯 on this.

an additional point (or maybe a slight addition to "turn tz-naive in tz-aware datetime objects before storing in database"). IMO we should heavily encourage folks to send timestamps that are TZ-aware, but if we _receive_ TZ-naive datetimes, we should turn those into UTC or assume UTC. That way we cover the most common cases (systems running CI somewhere where the TZ is typically UTC by default) and we're not rejecting data in cases where someone is sending something kind of off

jgehrcke:
> "turn tz-naive in tz-aware datetime objects before storing in database")

In https://github.com/conbench/conbench/pull/614 I decided against doing so because seemingly most (all?) of the DB columns we use for storing timestamps have been created so that they do _not_ track timezone information, example: https://github.com/conbench/conbench/blob/cc291e8566d77b527c27348fe7ba97ee8d44c2f5/conbench/entities/benchmark_result.py#L47

That's OK, too, if we introduce a strict convention that those always refer to UTC. I have started doing so by adding code comments etc.

jgehrcke:
> IMO we should heavily encourage folks to send timestamps that are TZ-aware, but if we receive TZ-naive datetimes, we should turn those into UTC or assume UTC.

Right! Did this for `finished_timestamp` for `RunCreate` in #614. The updated spec (feedback still welcome!) clarifies this behavior to users, and reads as

> A datetime string indicating the time at which the run finished. Expected to be in ISO 8601 notation. Timezone-aware notation recommended. Timezone-naive strings are interpreted in UTC. Fractions of seconds can be provided but are not returned by the API. Example value: 2022-11-25T22:02:42Z.

jgehrcke:
Next up I am doing #614 for `BenchmarkCreate`'s `timestamp` property.

jgehrcke:
More work for `Run.timestamp`: https://github.com/conbench/conbench/pull/647

jgehrcke:
TODO: find out if there are any more user-facing timestamp concepts left that need clarification.

PR #614, opened 2023-01-23
Run timestamp_finished: improved spec, consolidated timezone logic

jgehrcke:
This is motivated by https://github.com/conbench/conbench/issues/609.

I started with a rather harmless timestamp, the `timestamp_finished` property of a run.

I started by improving the specification.  My current proposal:

> A datetime string indicating the time at which the run finished. Expected to be in ISO 8601 notation. Timezone-aware notation recommended. Timezone-naive strings are interpreted in UTC. Fractions of seconds can be provided but are not returned by the API. Example value: `2022-11-25T22:02:42Z`"

To implement this kind of logic a convenient, productive, idiomatic place is to use marshmallow's `post_load()` hook, documented [here](https://marshmallow.readthedocs.io/en/stable/marshmallow.decorators.html#marshmallow.decorators.post_load). Initially this didn't work. The reason is that we used only `schema.validate()`, and not `schema.load()`. Docs [here](https://marshmallow.readthedocs.io/en/stable/quickstart.html#validation-without-deserialization).

I decided to do a bit of a risky surgery to replace `schema.validate()` with `schema.load()` in the code path that affects all mutating HTTP handlers. I think this was worth it, led to some good insights and cleanups here and there. Some tests needed to be adjusted slightly with respect to error messages -- I think the subtle changes in behavior are all OK.

I have added a test case with input/output pairs, maybe that's a good starting point for review.

There are more tiny changes in this PR that are hopefully explained in commit msgs, otherwise feel free to leave a comment/question!

Once we settle on this one here the plan is to treat other timestamp properties in the API similarly.

### DIFF
--- a/conbench/tests/api/test_runs.py
+++ b/conbench/tests/api/test_runs.py
@@ -308,6 +308,7 @@ class TestRunPut(_asserts.PutEnforcer):
 
     def _create_entity_to_update(self):
         _fixtures.benchmark_result(sha=_fixtures.PARENT)
+        # This writes to the database.
         benchmark_result = _fixtures.benchmark_result()
         return benchmark_result.run
 
--- a/conbench/tests/api/test_runs.py
+++ b/conbench/tests/api/test_runs.py
@@ -358,11 +358,9 @@ class TestRunPut(_asserts.PutEnforcer):
                 "finished_timestamp": timestring[0],
             },
         )
-        # print(resp.text)
         assert resp.status_code == 200
 
         resp = client.get(f"/api/runs/{before.id}/")
-        print(resp.json["finished_timestamp"])
         assert resp.json["finished_timestamp"] == timestring[1]
 
 
--- a/conbench/tests/api/test_runs.py
+++ b/conbench/tests/api/test_runs.py
@@ -358,10 +358,30 @@ class TestRunPut(_asserts.PutEnforcer):
                 "finished_timestamp": timestring[0],
             },
         )
-        assert resp.status_code == 200
+        assert resp.status_code == 200, resp.text
 
         resp = client.get(f"/api/runs/{before.id}/")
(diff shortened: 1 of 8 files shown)


conbench/util.py
```diff
@@ -25,6 +25,24 @@
...
+    the 'Z'.
+
+    Example output: 2022-11-25T16:02:00Z
+    """
+    if dt.tzinfo is not None:
+        # Programming error, but don't crash.
+        log.warning(
+            "tznaive_dt_to_aware_iso8601_for_api() got tz-aware datetime obj: %s", dt
+        )
+    return dt.isoformat(sep="T", timespec="seconds") + "Z"
```
jgehrcke, line 55: note to self:
- add unit tests
- do not do +Z if dt.tzinfo is not None

jgehrcke:
Landing this; feedback still highly highly appreciated. I understand that this might break something, but this would again be a forcing function towards CI-that-yields-more-confidence.

conbench/tests/api/test_runs.py
```diff
@@ -323,12 +328,61 @@ def test_update_allowed_fields(self, client):
...
+
         for key, value in self.valid_payload.items():
             if key == "finished_timestamp":
-                assert str(getattr(after, key)) == value
+                assert tznaive_dt_to_aware_iso8601_for_api(getattr(after, key)) == value
             else:
                 assert getattr(after, key) == value
 
+    @pytest.mark.parametrize(
+        "timestring",
```
austin3dickey, line 339: FYI: you can do something like `("input_timestring", "output_timestring")` as the first arg to `parametrize` and then include both of those in the function signature, and it will know how to unzip the pairs :)

jgehrcke: Thanks for reminding me! https://github.com/conbench/conbench/pull/626

PR #615, opened 2023-01-23
history plot: split plot lines before a distribution change

austin3dickey:
This PR breaks all the `p.line()`s on our timeseries plots before each distribution change. We don't want to visually connect, e.g., the min line from before a distribution change to after a distribution change.

Especially check out the black `min` lines on the following plots. The orange `rolling mean` and silver `alert boundary` lines will look better after I do #473.

Before:
image

After:
image

### DIFF
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -2,11 +2,12 @@ import collections
 import json
 import logging
 from datetime import timedelta
-from typing import Optional
+from typing import List, Optional
 
 import bokeh.events
 import bokeh.models
 import bokeh.plotting
+import numpy as np
 
 from conbench import util
 
@@ -168,6 +169,13 @@ def _should_format(data, unit):
     return should, axis_unit
 
 
+def _insert_nans(some_list: list, indexes: List[int]):
+    """Insert np.NaNs into a list before the given indexes."""
+    for ix in sorted(indexes, reverse=True):
+        some_list.insert(ix, np.NaN)
+    return some_list
+
+
 def _source(
     data,
     unit,
@@ -175,6 +183,7 @@ def _source(
     distribution_mean=False,
     alert_min=False,
     alert_max=False,
+    break_line_ixs: Optional[List[int]] = None,
 ):
     # Note(JP): important magic: if not specified otherwise, this extracts
     # the mean-over-time.
@@ -248,6 +257,13 @@ def _source(
             strings.append(", ".join(unit_fmt(s, unit) for s in samples))
 
         dsdict["multisample_strings_with_unit"] = strings
+
+    if break_line_ixs:
+        # This source will be used for a line, and we want to "break" the line between
+        # the given indexes and the immediately previous data points. Do this by
+        # inserting np.NaNs.
+        for key in dsdict:
+            dsdict[key] = _insert_nans(dsdict[key], break_line_ixs)
 
     return bokeh.models.ColumnDataSource(data=dsdict)
 
@@ -399,6 +415,11 @@ def time_series_plot(history, benchmark, run, height=380, width=1100):
     unit = history[0]["unit"]
     with_dist = [h for h in history if h["distribution_mean"]]
     formatted, axis_unit = _should_format(history, unit)
+    dist_change_ixs = [
+        ix
+        for ix, result in enumerate(history)
+        if result["change_annotations"].get("begins_distribution_change", False)
+    ]
 
     # Note(JP): `history` is an ordered list of dicts, each dict has a `mean`
     # key which is extracted here by default.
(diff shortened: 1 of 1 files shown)


conbench/app/_plots.py
```diff
@@ -168,13 +169,21 @@ def _should_format(data, unit):
...
+
+
 def _source(
     data,
     unit,
     formatted=False,
     distribution_mean=False,
     alert_min=False,
     alert_max=False,
+    break_line_ixs: Optional[List[int]] = None,
```
alistaire47, line 186: That I'm asking suggests we should have some docstrings, but what's `ixs`? "Indices"?

alistaire47 (approved):
Love the visual change! One minor question

conbench/app/_plots.py
```diff
@@ -168,13 +169,21 @@ def _should_format(data, unit):
     return should, axis_unit
 
 
+def _insert_nans(some_list: list, indexes: List[int]):
+    """Insert np.NaNs into a list before the given indexes."""
+    for ix in sorted(indexes, reverse=True):
+        some_list.insert(ix, np.NaN)
```
jgehrcke, line 175: Interesting! Maybe at some point these ideas might become relevant:
- https://wiki.python.org/moin/TimeComplexity
- https://pypi.org/project/blist/

conbench/app/_plots.py
```diff
@@ -249,6 +258,13 @@ def _source(
 
         dsdict["multisample_strings_with_unit"] = strings
 
+    if break_line_ixs:
+        # This source will be used for a line, and we want to "break" the line between
+        # the given indexes and the immediately previous data points. Do this by
+        # inserting np.NaNs.
```
jgehrcke, line 264: Did you find good docs explaining this behavior?

I at least found that it seems to be intentional :)
https://stackoverflow.com/a/67026207
https://github.com/PatrikHlobil/Pandas-Bokeh/issues/71

conbench/app/_plots.py
```diff
@@ -399,17 +415,26 @@ def time_series_plot(history, benchmark, run, height=380, width=1100):
     unit = history[0]["unit"]
     with_dist = [h for h in history if h["distribution_mean"]]
     formatted, axis_unit = _should_format(history, unit)
+    dist_change_ixs = [
+        ix
+        for ix, result in enumerate(history)
+        if result["change_annotations"].get("begins_distribution_change", False)
```
jgehrcke, line 421: is the `False` default important here? `d.get('unknown-key')` returns `None`.

conbench/app/_plots.py
```diff
@@ -399,17 +415,26 @@ def time_series_plot(history, benchmark, run, height=380, width=1100):
...
-            x=util.tznaive_iso8601_to_tzaware_dt([x["timestamp"] for x in history]),
+            # Insert NaNs to break the line at distribution changes
+            x=_insert_nans(
+                util.tznaive_iso8601_to_tzaware_dt([x["timestamp"] for x in history]),
+                dist_change_ixs,
+            ),
             # TODO: best-case is not always min, e.g. when data has a unit like
             # bandwidth.
-            y=[min(x["data"]) for x in history],
+            y=_insert_nans([min(x["data"]) for x in history], dist_change_ixs),
```
jgehrcke, line 437: I like the user-facing change that this PR brings. This implementation complexity of this module and also the readability might slow some people down in the future, but we can tweak this later.

jgehrcke:
Because it's fun: https://english.stackexchange.com/a/3126

> The plural of index is usually spelled indexes, but can also be spelled indices (as in the original Latin) in subjects like science and medicine.

> I was taught that books have indexes while maths has indices (similarly, people have appendixes while books have appendices).

jgehrcke (approved):
Lovely. Let's land this and iterate :rocket:

conbench/app/_plots.py
austin3dickey: Yes, good callout. I should have made this name a little more explicit

conbench/app/_plots.py
austin3dickey: Haha! I did think about performance here... but if we have enough data points in these lists that inserting a few values causes any noticeable performance degradation, I think rendering the plot itself is going to really blow up performance first 🙂

conbench/app/_plots.py
austin3dickey: Nope, I found it on StackOverflow as well, but I'll explore a little more!

conbench/app/_plots.py
austin3dickey: Nope, it's not important. None is falsey enough. 🙂 

I think this idiom was just copy/pasted from here, where it is important to be an actual boolean, because we're POSTing that value to the API.
https://github.com/conbench/conbench/blob/2b1653f5582bf397ffaa10278062949a89a2ff8b/conbench/app/benchmarks.py#L198-L200

conbench/app/_plots.py
austin3dickey: +100

conbench/app/_plots.py
austin3dickey: https://docs.bokeh.org/en/3.0.3/docs/user_guide/basic/lines.html#missing-points

PR #616, opened 2023-01-23
hotfix: python 3.8-compatible type hints

austin3dickey:
There are many infrastructures out there (such as [arrow-benchmarks-ci](https://github.com/voltrondata-labs/arrow-benchmarks-ci)) that download the `conbench` CLI from PyPI and use it to run benchmarks. Some of them are using python<3.9, so we can't use the new method of built-in object subscriptable type hinting like `list[str]`.

### DIFF
--- a/conbench/util.py
+++ b/conbench/util.py
@@ -4,7 +4,7 @@ import os
 import time
 import urllib.parse
 from datetime import datetime, timezone
-from typing import Union
+from typing import List, Union
 
 import click
 import requests
@@ -26,8 +26,8 @@ log = logging.getLogger()
 
 
 def tznaive_iso8601_to_tzaware_dt(
-    input: Union[str, list[str]]
-) -> Union[datetime, list[datetime]]:
+    input: Union[str, List[str]]
+) -> Union[datetime, List[datetime]]:
     """
     Convert time strings into datetime objects.
 
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ requirements_test = parse_requirements(path="requirements-test.txt")
 
 setuptools.setup(
     name="conbench",
-    version="1.62.0",
+    version="1.63.0",
     description="Continuous Benchmarking (CB) Framework",
     long_description=long_description,
     long_description_content_type="text/markdown",


austin3dickey:
We probably should consider running the test suite on 3.8. Actually, what we really should think about doing is splitting up the `conbench` CLI from the server code.

jonkeane:
> We probably should consider running the test suite on 3.8. Actually, what we really should think about doing is splitting up the conbench CLI from the server code.

Yeah, we absolutely should run python 3.8 on the CLI. Honestly, we also should test back as far as we can and support as many versions as can that aren't EOLed yet (for the CLI). I've had at least two conbench setups that had extra work involved because python 3.8 wasn't immediately available where we were trying to set it up. We worked around it, but it's a friction we shouldn't subject our users to if we can avoid it.

austin3dickey:
Love it. I think if we do https://github.com/conbench/conbench/issues/617 then it will be much easier to go way back for the CLI supported version of python .

conbench/util.py
```diff
@@ -26,8 +26,8 @@
 
 
 def tznaive_iso8601_to_tzaware_dt(
-    input: Union[str, list[str]]
-) -> Union[datetime, list[datetime]]:
+    input: Union[str, List[str]]
+) -> Union[datetime, List[datetime]]:
```
jgehrcke, line 30: Cool. And yeah, that was basically an oversight, and a consequence of me not yet having done tons of typing work in Python.

setup.py
```diff
@@ -22,7 +22,7 @@ def parse_requirements(path: str):
 
 setuptools.setup(
     name="conbench",
-    version="1.62.0",
+    version="1.63.0",
```
jgehrcke, line 25: It's a provocative question -- but what does this even mean? :)

jgehrcke (commented):
:rocket: thanks for the patch!

setup.py
austin3dickey: Not much at all. In this case, I wanted to ensure that Arrow benchmarks were using a version of the `conbench` CLI that included this change. We'll probably get more meaning out of versioning in the future once we're uploading tools to PyPI.

PR #621, opened 2023-01-24
replace print() with logging, or outcomment

jgehrcke:
I used `grep` to review a few more print() occurrences. Motivating ticket: https://github.com/conbench/conbench/issues/442


jgehrcke:
I assume this is good to go! :)

Issue #534, opened 2022-12-13
BenchmarkCreate schema: timestamp should be tz-aware or tz-agnostic


jgehrcke:
In a benchmark result I see for example
```
timestamp 2022-12-07 16:47
```
and I wonder what that means:
- is this submission time (as recorded by Conbench), or start time or end time (as recorded and freely set by the API user)?
- in which timezone is this to be interpreted? 

I went into docs and found that it's user-given, as defined by the `BenchmarkCreate` schema. It's currently documented with
```
string($date-time)
Timestamp the benchmark ran
```

No comment on time zone and no comment on start vs. end.

**Proposal A:** Maybe we should add documentation like "when the benchmark run was started" to guide users torwards consistent usage, and then also say what it means in the result page.

For distributed teams working on the same thing in the output there should be timezone information. But for that to work the input must have timezone information, too. This can be achieved by a breaking change (with a schema change, and stricter schema validation).

**Proposal B:** That `timestamp` property could be a unix timestamp or maybe an ISO 8601 variant which has tz information set (via. e.g. the `+00:00` method or via the `Z` Zulu method).

**Proposal C:** Conbench could take record of the submission time itself, and display it in the result page, too.

This relates to https://github.com/conbench/conbench/issues/484.

alistaire47:
For things run with the Conbench runner (and newer things, which mirror it in this regard), we're submitting UTC ISO 8601 with a tz offset: https://github.com/conbench/conbench/blob/main/conbench/runner.py#L36-L38 e.g. `"2022-11-23T16:28:08.359222+00:00"`. When this gets populated is a good and sometimes complicated question; we've got a lot of things running in a lot of ways, and while it should conceptually be start time, in practice it's probably frequently finish time when results get assembled, so it may be more "pretty close" than precisely correct. 

Not sure exactly how that gets parsed, stored, and presented, and if we're doing any frontend localization. I wouldn't be surprised if they're still UTC.

austin3dickey:
I've heard @jonkeane say in the past that we're leaving "start vs end" ambiguous on purpose, to allow for flexibility between different implementations. It's another one of those tradeoffs between easy implementation and data "meaningfulness". However, that doesn't mean we can't thoughtfully decide on a way to tempt users to congregate on one definition.

jgehrcke:
Thanks for the feedback above @alistaire47 and @austin3dickey!

This can almost get closed as 'done'. On the UI side of things however there is still a tiny bit of work to do.

[...]

PR #623, opened 2023-01-24
BenchmarkCreate timestamp: improved spec, explicit timezone conversion

jgehrcke:
Similar to https://github.com/conbench/conbench/pull/614.

This is motivated by https://github.com/conbench/conbench/issues/534 and https://github.com/conbench/conbench/issues/609.

This patch touches on the behavior of the `timestamp` property in the `BenchmarkCreate` schema.

Specifically, this patch

- changes the user-facing specification text
- adds more code comments around our DB timezone convention
- uses `marshmallow.fields.AwareDateTime` instead of `marshmallow.fields.DateTime`
- uses explicit timezone conversion code (to UTC, when users provide a non-UTC tz-aware time string)
- adds test cases for various special cases, including timezone conversion

I wanted to remove a bit of the ambiguity discussed in https://github.com/conbench/conbench/issues/534 about the meaning of this timestamp. New: "A datetime string indicating the time at which the benchmark _was started_". This takes some mental burden away from users and I think it makes sense in most of the cases. To advanced users it should be obvious that _any_ time can be communicated here. In this API spec I think it's also valuable to let users know where/how this timestamp is used (result sort oder); and so I added that.

### DIFF
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -419,9 +419,10 @@ class _BenchmarkFacadeSchemaCreate(marshmallow.Schema):
                 interpreted in UTC. Fractions of seconds can be provided but
                 are not returned by the API. Example value:
                 2022-11-25T22:02:42Z. This timestamp defines the default
-                sorting order when viewing runs via the UI or when enumerating
-                runs via the /api/runs/ HTTP endpoint.
-            """
+                sorting order when viewing a list of benchmarks via the UI or
+                when enumerating benchmarks via the /api/benchmarks/ HTTP
+                endpoint.
+                """
             )
         },
     )
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -418,7 +418,9 @@ class _BenchmarkFacadeSchemaCreate(marshmallow.Schema):
                 Timezone-aware notation recommended. Timezone-naive strings are
                 interpreted in UTC. Fractions of seconds can be provided but
                 are not returned by the API. Example value:
-                2022-11-25T22:02:42Z.
+                2022-11-25T22:02:42Z. This timestamp defines the default
+                sorting order when viewing runs via the UI or when enumerating
+                runs via the /api/runs/ HTTP endpoint.
             """
             )
         },
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -1,3 +1,5 @@
+from datetime import timezone
+
 import flask as f
 import marshmallow
 import numpy as np
@@ -5,6 +7,8 @@ import sqlalchemy as s
 from sqlalchemy import CheckConstraint as check
 from sqlalchemy.dialects import postgresql
 from sqlalchemy.orm import relationship
+
+import conbench.util
 
 from ..config import Config
 from ..entities._comparator import z_improvement, z_regression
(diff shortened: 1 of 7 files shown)


conbench/entities/benchmark_result.py
```diff
@@ -391,8 +403,27 @@ class _BenchmarkFacadeSchemaCreate(marshmallow.Schema):
...
+            "description": conbench.util.dedent_rejoin(
+                """
+                A datetime string indicating the time at which the benchmark
+                was started. Expected to be in ISO 8601 notation.
+                Timezone-aware notation recommended. Timezone-naive strings are
+                interpreted in UTC. Fractions of seconds can be provided but
+                are not returned by the API. Example value:
+                2022-11-25T22:02:42Z. This timestamp defines the default
+                sorting order when viewing runs via the UI or when enumerating
+                runs via the /api/runs/ HTTP endpoint.
```
jgehrcke, line 423: A note on this. For many years I have `Alt +Q` in my editor bound to doing `rewrap` (rewrap text so that it nicely fills up to N columns). The triple-quoted strings in Python allow for using that. But they contain literal newlines, and if indention is applied that there also is some leading whitespace. Those additional newlines and the indention can of course be removed with programmatic tools.

conbench/util.py
```diff
@@ -109,6 +110,34 @@ def _convert(s: str):
...
 
 
+def dedent_rejoin(s: str):
+    """
+    Remove common leading whitespace, replace newlines by spaces.
+
+    Useful for being able to write marshmallow property docstrings with
+    indented block paragraphs.
+    """
+    return " ".join(textwrap.dedent(s).strip().splitlines())
```
jgehrcke, line 120: A note on this. For many years I have `Alt +Q` in my editor bound to doing `rewrap` (rewrap text so that it nicely fills up to N columns). The triple-quoted strings in Python allow for using that. But they contain literal newlines, and if indention is applied that there also is some leading whitespace. Those additional newlines and the indention can of course be removed with programmatic tools.

austin3dickey: Luckily I think Swagger (like Sphinx) automatically collapses all these newlines and indentations before rendering docstrings into HTML. That's why I was able to get away with this triple-quoted string:
https://github.com/conbench/conbench/blob/e9fae1e603b28508cb79c2911e1c525fbf4ee23b/conbench/entities/benchmark_result.py#L363
without having to collapse things myself. It does insert the whitespace garbage into the OpenAPI json, but since that garbage is gone by the time it gets to Swagger I thought it wasn't worth cleaning up.

conbench/entities/benchmark_result.py
austin3dickey: I think `Run.timestamp` is what determines the sort order of the runs UI (aka the index) and `/api/runs`, not `BenchmarkResult.timestamp`. 

`BenchmarkResult.timestamp` only determines the sort order of `/api/benchmarks` with no argument, and the `/benchmarks` webpage (which I suspect not many people visit).

austin3dickey (approved):
Looks good! One small clarification

conbench/entities/benchmark_result.py
jonkeane: Yeah, I'll second what Austin said here: the sort order here would (likely) be on any list of benchmarks (like when displaying one run in the UI). Maybe we could say something along those lines "timestamps are used as the default order when listing items that contain them (benchmarks, runs, etc.) in the API and the UI" which helps explain without being too specific about which endpoints (so if we add some we don't need to come back and update this without it still being correct).

jgehrcke: > Run.timestamp is what determines the sort order of the runs UI (aka the index) and /api/runs, not BenchmarkResult.timestamp

Thanks! I confused those once again when writing the text.

conbench/util.py
jgehrcke: Oh, thanks for that feedback, I somehow missed that before. Personally, I like doing the explicit more-or-less tidy constructions of strings in Python, but yeah it's good to know that the Swagger toolchain is liberal with the input 'quality'.

conbench/entities/benchmark_result.py
jgehrcke: (added commits to fix that)

PR #624, opened 2023-01-24
history plot: clarify variables

austin3dickey:
This PR implements some feedback from #615.

- rename "ixs" to "indexes" for clarity
- add link to Bokeh docs for this behavior
- use `float("nan")` instead of `np.NaN` as recommended in the Bokeh docs
- remove redundant `.get()` default of False

### DIFF
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -7,7 +7,6 @@ from typing import List, Optional
 import bokeh.events
 import bokeh.models
 import bokeh.plotting
-import numpy as np
 
 from conbench import util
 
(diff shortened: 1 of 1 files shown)


conbench/app/_plots.py
```diff
@@ -7,7 +7,6 @@
 import bokeh.events
 import bokeh.models
 import bokeh.plotting
-import numpy as np
```
jgehrcke, line 10: I had the feeling we would not need numpy :)

jgehrcke (approved):
Nice!

PR #628, opened 2023-01-26
API docs: notes on implicit Run creation & metadata update, add a test

jgehrcke:
'Implicit run creation' was a bit of a mystery to me so far, and today's side quest was to understand what's actually happening (because current API documentation didn't really provide answers). While exploring the current behavior I documented findings on-the-fly in code and API docs, that's how this patch came to be.

### DIFF
--- a/conbench/api/benchmarks.py
+++ b/conbench/api/benchmarks.py
@@ -161,11 +161,13 @@ class BenchmarkListAPI(ApiEndpoint, BenchmarkValidationMixin):
         """
         ---
         description:
-            Submit a BenchmarkResult within a specific Run (as defined by its
-            Run ID). If the Run is not known yet in the database it gets
-            implicitly created, using details provided in this request. If the
-            Run is already known then Run-specific info in this request is
-            silently ignored.
+            Submit a BenchmarkResult within a specific Run.
+
+            If the Run (as defined by its Run ID) is not known yet in the
+            database it gets implicitly created, using details provided in this
+            request. If the Run ID matches an existing run, then the rest of
+            the fields describing the Run (such as name, hardware info, ...}
+            are silently ignored.
         responses:
             "201": "BenchmarkCreated"
             "400": "400"
--- a/conbench/api/benchmarks.py
+++ b/conbench/api/benchmarks.py
@@ -160,7 +160,12 @@ class BenchmarkListAPI(ApiEndpoint, BenchmarkValidationMixin):
     def post(self):
         """
         ---
-        description: Create a benchmark.
+        description:
+            Submit a BenchmarkResult within a specific Run (as defined by its
+            Run ID). If the Run is not known yet in the database it gets
+            implicitly created, using details provided in this request. If the
+            Run is already known then Run-specific info in this request is
+            silently ignored.
         responses:
             "201": "BenchmarkCreated"
             "400": "400"
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -152,7 +152,13 @@ class BenchmarkResult(Base, EntityMixin):
         if not info:
             info = Info.create({"tags": data["info"]})
 
-        # create if not exists
+        # Create a corresponding `run` entity in the database if it doesn't
+        # exist yet. Use the user-given `id` (string) as primary key. If the
+        # Run is already known in the database then only update the
+        # `has_errors` property, if necessary. All other run-specific
+        # properties provided as part of this BenchmarkCreate structure (like
+        # `machine_info` and `run_name`) get silently ignored. already known in
+        # the database.
         run = Run.first(id=data["run_id"])
         if run:
             if has_error:
@@ -388,18 +394,39 @@ or improvements, should we treat data from before this result as incomparable?
 class _BenchmarkFacadeSchemaCreate(marshmallow.Schema):
     run_id = marshmallow.fields.String(
         required=True,
-        metadata={"description": "Unique identifier for a run of benchmarks."},
+        metadata={
+            "description": conbench.util.dedent_rejoin(
+                """
+                Identifier for a Run (group of benchmarks. This can be the ID
+                of a known Run (as returned by /api/runs) or a new ID in which
+                case a new Run entity is created in the database.
+                """
+            )
+        },
     )
     run_name = marshmallow.fields.String(
         required=False,
         metadata={
-            "description": "Name for the run. When run in CI, this should be of the style '{run reason}: {commit sha}'."
+            "description": conbench.util.dedent_rejoin(
+                """
+                Name for the run. When run in CI, this should be of the style
+                '{run reason}: {commit sha}'. Ignored when run was previously
+                created.
+                """
+            )
         },
     )
     run_reason = marshmallow.fields.String(
         required=False,
         metadata={
-            "description": "Reason for run (commit, pull request, manual, etc). This should be low cardinality. 'commit' is a special run_reason for commits on the default branch which are used for history"
+            "description": conbench.util.dedent_rejoin(
+                """
+                Reason for run (commit, pull request, manual, etc). This should
+                be low cardinality. 'commit' is a special run_reason for
+                commits on the default branch which are used for history.
+                Ignored when run was previously created.
+                """
+            )
         },
     )
     batch_id = marshmallow.fields.String(required=True)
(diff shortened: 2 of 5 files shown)


conbench/api/benchmarks.py
```diff
@@ -160,7 +160,12 @@ def get(self):
     def post(self):
         """
         ---
-        description: Create a benchmark.
+        description:
+            Submit a BenchmarkResult within a specific Run (as defined by its
+            Run ID). If the Run is not known yet in the database it gets
+            implicitly created, using details provided in this request. If the
+            Run is already known then Run-specific info in this request is
+            silently ignored.
```
jonkeane: We might make this a bit more specific and say "If the Run ID {or whatever the key fields are that match} matches a existing run, than the rest of the fields describing a run {fields listed?} are silently ignored"

conbench/entities/benchmark_result.py
```diff
@@ -152,7 +152,13 @@ def create(data):
...
             info = Info.create({"tags": data["info"]})
 
-        # create if not exists
+        # Create a corresponding `run` entity in the database if it doesn't
+        # exist yet. Use the user-given `id` (string) as primary key. If the
+        # Run is already known in the database then only update the
+        # `has_errors` property, if necessary. All other run-specific
+        # properties provided as part of this BenchmarkCreate structure (like
+        # `machine_info` and `run_name`) get silently ignored. already known in
+        # the database.
```
jonkeane: Was this intended to be:

```suggestion
        # `machine_info` and `run_name`) get silently ignored. 
```

Or maybe the "already known in the database" can be expanded into a sentence? Or used as the "title" of this comment chunk?

And *thank you* for these detailed comments — they are fantastic

conbench/entities/benchmark_result.py
```diff
@@ -388,18 +394,39 @@ class BenchmarkResultSerializer:
...
     )
     run_name = marshmallow.fields.String(
         required=False,
         metadata={
-            "description": "Name for the run. When run in CI, this should be of the style '{run reason}: {commit sha}'."
+            "description": conbench.util.dedent_rejoin(
+                """
+                Name for the run. When run in CI, this should be of the style
+                '{run reason}: {commit sha}'. Ignored when run was previously
+                created.
```
jonkeane: We can probably weaken this point "When run in CI, this should be of the style '{run reason}: {commit sha}'." It used to be required, but it no longer is (for the most part). But it _is_ a nice easy way to navigate. Maybe "This is frequently of the style ... but can be anything that is helpful to the consumers of this data"? We don't have to do this here, of course

conbench/entities/hardware.py
```diff
@@ -202,11 +202,14 @@ class ClusterCreate(marshmallow.Schema):
             "benchmark runs performed on different sets of hardware"
         },
     )
+    # Note(JP): `optional` in the name conflicts with `required=True`.
+    # This is confusing for API users.
```
jonkeane, line 206: Feel free to add a TODO or make a ticket to do a whole deprecate + replace setting for this. In the UI it's called "additional info" and it probably always should have been something like that. Last I touched this I wanted to do that, but we needed to be more tactical than the whole cycle at the time!

conbench/entities/benchmark_result.py
```diff
@@ -426,8 +453,32 @@ class _BenchmarkFacadeSchemaCreate(marshmallow.Schema):
...
+            )
+        },
+    )
+    cluster_info = marshmallow.fields.Nested(
+        ClusterSchema().create,
+        required=False,
+        metadata={
+            "description": conbench.util.dedent_rejoin(
+                """
+                Precisely one of `machine_info` and `cluster_info` must be
```
jonkeane, line 473: I'm not sure if this is me only, but I find the "Precisely one of ... must be provided" a unnatural sounding (it's precisely accurate, but easy to skim over that and it sounds confusing to me). At the same time, I haven't been able to come up with anything better that's not massively verbose. Happy to leave it as is, but curious if others feel the same (or have a better rephrase than I do!)

conbench/tests/api/test_benchmarks.py
```diff
@@ -422,6 +422,92 @@ def test_create_benchmark_after_run_was_created(self, client):
...
+        behavior it's important that we think about specification first
+        (desired behavior), and then judge about the implementation.
+
+        I think this is an artifact of us allowing for potentially many
+        concurrent executors to submit results within a specific run without
+        requiring special coordination between those executors. That is, it is
+        expected that the submitted run metadata (like name and machine info)
+        are equivalent across all BenchmarkResult entities, and the fastest
+        submitter wins the price of DB insertion, while the other racers get
+        their info silently dropped.
```
jonkeane, line 442: This is certainly one kind of behavior this might supply. Honestly, the history here as far as I remember was basically: make it as easy as possible to post a benchmarkResult (e.g. make the run, hardware, etc. entities for you, don't make you make those manually) but still be able to lump benchmarkResults into runs and not have a proliferation of hardwares (hardware types do the same thing as runs, where if you post details of them that are keys, it re-uses the entry in the database — but if it's different you'll create a new database), etc.

Then the question basically is: is that ease of posting worth the confusion | complication in the API?

jonkeane (approved):
Thanks for this! ❤️ more, better docs on this stuff + more tests

conbench/entities/benchmark_result.py
```diff
@@ -388,18 +394,39 @@ class BenchmarkResultSerializer:
...
     )
     run_reason = marshmallow.fields.String(
         required=False,
         metadata={
-            "description": "Reason for run (commit, pull request, manual, etc). This should be low cardinality. 'commit' is a special run_reason for commits on the default branch which are used for history"
+            "description": conbench.util.dedent_rejoin(
+                """
+                Reason for run (commit, pull request, manual, etc). This should
+                be low cardinality. 'commit' is a special run_reason for
+                commits on the default branch which are used for history.
```
austin3dickey: oh shoot can you remove
>'commit' is a special run_reason for commits on the default branch which are used for history

That's no longer true 🙂

conbench/tests/api/test_benchmarks.py
austin3dickey: >is that ease of posting worth the confusion?

I tend to think so. For one thing, if we were to require creating a Run or Hardware before posting any results, it would be a breaking change that we'd have to fix in all CI systems. I think most CI systems are using this behavior precisely because it is easier.

austin3dickey (approved):
Nice!

conbench/entities/benchmark_result.py
jgehrcke: Sentence shuffling oversight, thanks!! :)

conbench/entities/benchmark_result.py
jgehrcke: lovely feedback, thanks

conbench/entities/hardware.py
jgehrcke: :+1:  https://github.com/conbench/conbench/issues/631

conbench/entities/benchmark_result.py
jgehrcke: I feel the same actually, and have always felt the same when I was in the position to describe that "exactly one of [] is required" constraint. Haha, this is a recurring problem all my programming life. I leave this as-is for now, but am also very open to more elegant, user-friendly wording.

conbench/tests/api/test_benchmarks.py
jgehrcke: Love the discussion. I hoped to trigger this kind of thought exchange by making that current behavior a little more explicit. 

I have not picked a side yet. 

There certainly would be a better middle ground API-wise if we had a well-specified `RunCreate` schema that would be part of the `BenchmarkCreate` schema.

>  it would be a breaking change that we'd have to fix in all CI systems.

Happy to share my opinion about this again and again :) Given the current state of the project I want to encourage us to not be afraid of making brutal changes. It's fair to say that we are in a prototyping phase.

jgehrcke:
Thanks for the great feedback @jonkeane @austin3dickey! Pushed another commit, will land this once CI goes green. Let's iterate from here.

conbench/entities/benchmark_result.py
austin3dickey: I'm a fan of "Either A or B", which in my English-speaking experience implies "A xor B", but I realize that it may not imply that for other people.

conbench/tests/api/test_benchmarks.py
austin3dickey: >I want to encourage us to not be afraid of making brutal changes.

I completely agree! Having done a few of these breaking changes in the past year, though, I've learned the hard way that they take more thought and work than they may seem on the surface, and so when we prioritize features by value-for-effort we should always be conscious of that.

PR #630, opened 2023-01-27
index.html template: full commit SHA search fix

### LINKED ISSUE #625: Search by full git commit SHA no longer working?
I am nearly positive we used to be able to search for runs in the main conbench list UI by the full Git commit SHA as well as the short one. But that doesn't seem to be the case anymore.

https://conbench.ursa.dev/?search=f3637aca7750f43f7e1766e4495254f98adbe40b should return results, but says there are none however https://conbench.ursa.dev/?search=f3637ac _does_ return results. 

This might have changed when we updated the styling on the commit column. We should figure out how to make it so that full SHAs are useable in search but maintain the nice-looking short SHA we have now.

sayantikabanik:
resolves #625

### DIFF
--- a/conbench/templates/index.html
+++ b/conbench/templates/index.html
@@ -71,9 +71,13 @@
                     <td>
                          <div class="table-entry">{{ run.hardware.name }}</div>
                     </td>
-
                     <td>
-                         <div class="table-entry"><a href="{{ run.commit.url }}">{{ run.commit.sha[:7] }}</a></div>
+                         <div class="table-entry">
+                             <!--NOTE: enables full commit SHA search while maintaining partial SHA display for UI -->
+                             <a href="{{ run.commit.url }}"> {{ run.commit.sha[:7] }}
+                             <p class="hide-text">{{ run.commit.sha }}</p>
+                         </a>
+                         </div>
                     </td>
 
                     <td>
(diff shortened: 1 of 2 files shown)


conbench/templates/index.html
```diff
@@ -71,9 +71,13 @@ <h3>Language-independent Continuous Benchmarking (CB) Framework</h3>
                     <td>
                          <div class="table-entry">{{ run.hardware.name }}</div>
                     </td>
-
                     <td>
-                         <div class="table-entry"><a href="{{ run.commit.url }}">{{ run.commit.sha[:7] }}</a></div>
+                         <div class="table-entry">
+                             <!--NOTE: enables full commit SHA search while maintaining partial SHA display for UI -->
```
jgehrcke, line 76: Thanks for the code comment, super valuable for future work in this.

jgehrcke (approved):
Cool! I trust that you manually confirmed that this works.

Automated testing for things like that at some point will be an exciting topic.

jgehrcke:
(I updated the title of this PR after merging, wanted to do that before squash-merging. I always like to have some kind of component/area prefix in commit messages / PR titles so that one can quickly understand in which area the change takes effect)

jonkeane:
> Automated testing for things like that at some point will be an exciting topic.

Agreed and glad we merged this now even if we don't yet have an automated test. This isn't _quite_ automated, but it would be relatively easy to add a auto screenshot to our screenshot CI that exercises this search since it's encoded in the URL. Then at least we can confirm if it's returning 0 or more than 0 results visually.

Issue #608, opened 2023-01-19
Alert on drifting means


alistaire47:
>              > If the mean is drifting, something is probably wrong with the benchmark or setup.
>
> 💯 this is something we actually _should_ automatically be monitoring and then alert separately that something is up. This is a classic death-by-a-thousand-slow-downs trap where no individual change makes things perceptibly slower, but over time you've introduced enough overhead with each change that you're now 2x or 3x slower. Conbench is in some ways _uniquely positioned_ to catch these. But that's a project for after we do this + have more facility to calculating things like the slope of the mean over time. Would you mind making an issue for this @alistaire47 so we can brainstorm there | figure out where to slot that in to new project(s)?
>
> _Originally posted by @jonkeane in https://github.com/conbench/conbench/issues/473#issuecomment-1397681592_
            
All our current alerting is predicated on the idea that one commit causes a significant regression or progression, but it is possible that benchmarks could drift (for code or setup reasons), and we should perhaps alert on this.

Issue #594, opened 2023-01-11
data interpretation: start emphasizing the meaning of best-case outcome (minimum duration)


jgehrcke:
In https://github.com/conbench/conbench/issues/535 we started discussing the relevance of min vs. mean in performance analysis. There is a plethora of literature out there that I think elaborates rather well on the fundamental ideas, such as

- https://math.mit.edu/~edelman/publications/robust_benchmarking.pdf
- https://blog.kevmod.com/2016/06/10/benchmarking-minimum-vs-average/

In plotting, I think we should move away from showing only mean values, but show both: min and mean (neutrally expose both pieces of information).

In addition, I think we can safely emphasize min over mean (but that might be controversial). Most importantly, there is no one-size-fits-all solution, and we should do a great job explaining related concepts.

jgehrcke:
Made good progress towards enabling that systematically via the introduction of the 'single value summary' concept in #1172  and #1183.

jgehrcke:
https://conbench.ursa.dev/compare/benchmark-results/06511cc82d757d868000ce4783696dc4...06511ed95f977e968000accebf5ff277/ good example where a 'best of N repetitions' would yield a more useful signal than the mean value. 


![image](https://github.com/conbench/conbench/assets/265630/f0c98ba8-c75a-4b9c-bb33-d7d25413ccde)

Given 

```
[5.366551, 5.2435, 5.283533]
```
vs
```
[5.278719, 6.832488, 5.97085]
```

Given that very simplistic view I think it's pretty obvious that [this alert for a potential performance regression](https://github.com/apache/arrow/pull/37826#issuecomment-1737570838) is a false positive. 

One could potentially emit a 'instability warning' instead because it seems like that the volatility extent is new.

Issue #535, opened 2022-12-13
detect multimodal distribution in sample submission, then do not use the `mean`


jgehrcke:
Two interesting cases:

![image](https://user-images.githubusercontent.com/265630/207337025-52022ef3-a37d-46ec-9acf-8b1e3b023035.png)

----------

![image](https://user-images.githubusercontent.com/265630/207337186-ae4dacd7-c0ac-463d-b71e-56c4a0adb20c.png)

---------

In both cases we got samples from three iterations.

In both cases it appears as if the mean value is either not meaningful or even misleading, or both. It appears to be largely affected by the magnitude of a warm-up effect. That is, when one plots the mean value of these scenarios over time, it seems likely that one actually mainly analyzes the severity of the warm-up effect (and not so much the 'good samples').

I think that it's obvious that given these samples the benchmark method should be investigated for methodological sanity by a human.

We could automatically detect this situation. And then do something useful. Warn users. Maybe do a fall-back by picking `min` values. Maybe emit something in the lines of a 'sample distribution alert'. 

I really like that we have examples like this in some of our environments, it shows the potential of attempting to perform per-sample distribution analysis from N iterations: https://github.com/conbench/conbench/issues/530

jonkeane:
I would be really curious what benchmark you found this in (I don't think I see it above — but maybe I'm missing it!).

One thing we've never totally taken seriously is the warm-up effect + how to discount it (or not, depending on what you care about!). This was one of the biggest reasons [we have the disk cache clearing](https://github.com/conbench/conbench/blob/508267f8159f0b932f81b942eb5fe0e5dccf4869/conbench/runner.py#L306-L323) between each iteration on our macro benchmarks was as a way to basically remove the warmed up benefit. It was expeditious (+ has seemingly served us well), but I'm not totally sure it's the right long term solution.

As we're working on history and comparison updates — I do think we should be able to (soon) start looking at comparing differences not just in means, but also in mins like you described. I could also imagine comparing the mean of iterations 2:N, the max, and the result of just iteration 1. Each of these would have slightly different interpretations of course ("what does this look like after warm up", "what does this look like in a more real-world scenario where I run one thing once and not repeatedly", "what's the worst-case scenario" etc.). I could even imagine having toggles on the various plots + comparison pages that let you change those comparisons.

jgehrcke:
Thanks for getting back on this!

> I would be really curious what benchmark you found this in (I don't think I see it above — but maybe I'm missing it!).

[...]

PR #636, opened 2023-01-27
Updated z-score logic and calculate them on read

### LINKED ISSUE #473: Change z-score calculations to incorporate `begins_distribution_change`

### LINKED ISSUE #567: Cache less aggressively in `set_z_scores()`
Currently in `set_z_scores()`, upon finding a baseline Run for a contender BenchmarkResult, we use that same baseline Run for each subsequent BenchmarkResult sharing the same contender Run. This might be inaccurate if, say, the baseline Run has fewer languages than the contender Run. (This isn't a huge deal because the problematic BenchmarkResults will just have a z-score of 0, but it is unintended behavior.)

See if there's a way to handle that situation (i.e. find the true baseline run of every given BenchmarkResult) without slowing the function down too much (i.e. don't call `benchmark_result.run.get_baseline_run()` on every single iteration of the loop).

One solution might be to figure out all the case/context pairs of the cached baseline runs, and only call `benchmark_result.run.get_baseline_run()` when the contender BenchmarkResult doesn't match one of those pairs.

_Originally posted by @jonkeane in https://github.com/conbench/conbench/pull/507#discussion_r1059462511_

### LINKED ISSUE #577: Move away from storing a rolling window distribution result in the database

austin3dickey:
Resolves #473. See that issue for greater detail, including the in-depth rationale for the new calculation.
Also resolves #577 and resolves #567, since they were implementation details of the old method. 

This PR:
- changes the logic of z-score calculation so that
  - the distribution mean doesn't use data from before a distribution change
  - the distribution standard deviation doesn't blow up after a distribution change
- calculates z-scores on-read for the `/history`, `/benchmarks`, and `/compare` APIs, instead of on-write
- removes `Distribution` from the database

Here's a nice plot of the new ±5 sigma lines!
image


austin3dickey:
TODO:
- add a db migration to delete the Distribution table
- ensure this change doesn't slow down the Arrow webapp or affected API endpoints too much

conbench/entities/history.py
```diff
@@ -85,31 +89,170 @@ def get_history(case_id, context_id, hardware_hash, repo) -> List[tuple]:
...
+    history = Session.query(
+        history,
+        func.coalesce(
+            func.avg(history.c.mean).over(
+                partition_by=history.c.segment_id,
+                order_by=s.extract("epoch", history.c.timestamp),
+                range_=(None, -1),
+            ),
+            history.c.mean,
+        ).label("rolling_mean_excluding_this_commit"),
```
jgehrcke, line 396: great naming!

conbench/entities/history.py
```diff
@@ -85,31 +89,170 @@ def get_history(case_id, context_id, hardware_hash, repo) -> List[tuple]:
...
+
+    # Add same column, excluding the current commit
+    history = Session.query(
+        history,
+        func.stddev(history.c.residual)
+        .over(
+            order_by=s.extract("epoch", history.c.timestamp),
+            range_=(None, -1),
+        )
+        .label("rolling_sd_excluding_this_commit"),
```
jgehrcke: I like `stddev` more than `sd`, but :shrug: :)

conbench/entities/history.py
```diff
@@ -85,31 +89,170 @@ def get_history(case_id, context_id, hardware_hash, repo) -> List[tuple]:
...
+        z_score = None
+
+    if z_score and _less_is_better(unit):
+        z_score = z_score * -1
+
+    return z_score
+
+
+def append_z_scores(benchmark_results: List[BenchmarkResult]):
+    """Populate the "z_score" attribute of each given BenchmarkResult, comparing it to
```
jgehrcke: I did not read the code yet and would hope to understand a little better what this is doing based on the docstring :).

"append" in the function name vs "populate": can we use the same verb, whichever is more appropriate? Or maybe just 'set'? Is this really just setting an attribute on a Python class instance with the assignment operator? :D

It would be nice if you could add a sentence about overwrite behavior? In the lines of error-on-conflict, silent  but innocent overwrite...

Could we add to this docstring

- explicitly that this tries _not to_ write to the database?  via sqlalchemy magic often times we have implicit write-to-db in the code base (where it isn't maybe immediately obvious), so if this should deliberately not write to the DB, it might be worth pointing that out.
- a recommendation or precise expectation for when this should be called / how often, and/or the disadvantages of not calling this?

conbench/entities/history.py
```diff
@@ -85,31 +89,170 @@ def get_history(case_id, context_id, hardware_hash, repo) -> List[tuple]:
...
+        history,
+        func.sum(history.c.begins_distribution_change.cast(s.Integer))
+        .over(order_by=history.c.timestamp)
+        .label("segment_id"),
+    ).subquery()
+
+    # Add column with rolling mean of the means (only inside of the segment)
+    history = Session.query(
+        history,
+        func.avg(history.c.mean)
```
jgehrcke: This `history.c.mean` here is the multisample mean which _might_ be for a single iteration (`n=1`), right?

jgehrcke (changes requested):
Great work. Super excited to see this. Thanks for the screenshot!

It's engineering pleasure when one invests work to _reduce_ complexity. 

I'd be happy to merge this as-is, trusting your judgement and also testing completely.

But because this touches on a few difficult-to-understand core aspects of what we do with Conbench I'd love to pay extra attention to code readability and documentation of methodology.

To that end, I'd love for us do clarify the exact nature of rolling windows used (and 'why' if possible). When I see 'rolling window' then I have three important questions that I'd love to see answered via text/code comment, instead of having to "decode" this from code :)

- is it a sliding rolling window (I guess so) or [tumbling](https://stackoverflow.com/questions/12602368/sliding-vs-tumbling-windows)?
- what's the width? also: fixed number of observations, or in time?
- is the result of the window aggregate used at the left, center, or right? (when doing fixed-time-width rolling window then it's important to draw the result at the center to not retain with symmetry of the input)

Another point that I'd love to get your quick opinion on: library-driven time series analysis (via. e.g. pandas) vs. doing this with SQL on-the-fly in the database. My feeling is that given the order of magnitude of time series length one can get the individual samples-over-time from the db (without the rolling window aggregates), and then build small time series objects in code/mem and do the rolling window analysis with pandas on the fly, w/o noticeable performance change over the current rolling-window-in-db method that you have chosen here. To be quite honest, when I talked about #577 I always had the pandas approach in mind. I didn't think it through too much though. I really like that you chose a different method, and would love to read your thoughts about that decision.

A few more specific questions.

One might think that i) "the distribution mean doesn't use data from before a distribution change" specifically implies ii) "the distribution standard deviation doesn't blow up after a distribution change". If the second point cannot be fully explained by the first point, I'd love to see one sentence about the methodological sugar that explains the second point.

> removes Distribution from the database and DISTRIBUTION_COMMITS from the Config

I believe that this configuration variable in some way determined the window width of the previous rolling window analysis. We did not do a great job explaining (in written form, for users) how this can be used and how important that is. I love that we remove this now. But I'd love to read more about _why_ we do not need to expose the window width as a configuration parameter anymore.

> add a db migration to delete the Distribution table

Yeah! Recently I had a thought (it's orthogonal to the migration you mention, just quickly dumping that): over time there may accumulate many orphaned objects in a Conbench database, right? Commit objects that are not needed anymore, etc.

> ensure this change doesn't slow down the Arrow webapp or affected API endpoints too much

I don't think we need to overdo it here. I think it's enough to have brief manual confirmation that this is in the same ballpark as before -- for my taste that would be enough to merge, for sure!

For proper observation, again https://github.com/conbench/conbench/issues/493 would deliver that 'for free'.

Again, super nice work! And thanks for going through my points and questions in advance!

jgehrcke:
The CI failure in `db-migrations` as of `ImportError: cannot import name 'distribution' from 'conbench.entities' (/app/conbench/entities/__init__.py)` is probably expected to you as of now, right?

conbench/entities/history.py
austin3dickey: >append vs populate vs set

I can't find the GitHub PR comment, but I remember there was some confusion in the past when it was called `set_z_scores()`, because that name sounded like z-scores were being permanently stored in the database. That's not the case in this function; the attribute is just being set on the Python object, but it's not being saved into the DB. I'll definitely clarify that in the docstring. I think the verb `append` might be a little clearer in that regard than `set`.

conbench/entities/history.py
austin3dickey: Yes, that's correct. This is how we've always done this, but I'm open to switching it up.

conbench/entities/history.py
jonkeane: > explicitly that this tries not to write to the database? via sqlalchemy magic often times we have implicit write-to-db in the code base (where it isn't maybe immediately obvious), so if this should deliberately not write to the DB, it might be worth pointing that out.

💯 on this point. sqlalchemy is fantastic in the way it just takes care of many of these things, but coming back and trying to maintain something + remember if it's using sqlalchemy's magic (I guess that would be just alchemy!) or not can be really hard | frustrating

austin3dickey:
Thanks for your thorough review!! Here are some thoughts:

>The CI failure is probably expected to you as of now, right?

Yes, I need to fix up the migrations still. :)

>is it a sliding rolling window or tumbling? what's the width?

Sure, I'll comment here (and in the code)! As implemented right now, the rolling windows are actually _expanding windows_. They are not fixed-width, because they use as much data from the past as possible. I justified this by considering that (1) for our known use cases, the number of data points per window is usually less than 100 and will probably never go above 10000 before something changes, (2) there's no reason to not use "old" data if we alert on drifting means (#608) and manage changes in distributions via annotations, and (3) setting a window size means we have to continue carrying a Config parameter around to store that window size, vs the alternative of getting rid of it in this PR.

>is the result of the window aggregate used at the left, center, or right?

It's used at the right of the window. I disagree that it's important to draw at the center of the window for time-series analysis, because that would be using future data. If we draw at the right, then we retain the property of knowing at a glance what the z-score would have been at that time, without having access to any data that might have come in afterwards. (Maybe I misunderstood what you're saying here.)

>calculation in pandas vs the database

I thought about this for a while, and still could be swayed to the other side. I agree that performance isn't a huge factor in this decision based on the order of magnitude of the data. What swayed me is that we often need to request the z-score for just one BenchmarkResult (e.g. when hitting `/api/benchmarks/X`). That calculation needs access to _all_ the historic data for that case/context/hardware/repo. So when writing the data query for that calculation, we'd still need to bring all that same data into a pandas dataframe, and we are already mostly done. I thought adding a few columns to the query in SQL would be easier than adding the pandas dependency. (though: technically we already have a pandas dependency due to `bokeh` 🙃)

>complexity

I had a thought over the weekend that my current query has too many columns. Whenever we call `get_history()`, we either want the rolling mean and standard deviation including the current commit (for `append_z_scores()`) OR excluding the current commit (for plots). So I think I'll add an input parameter that returns the calculation whichever way you choose. Code to come.

conbench/entities/history.py
```diff
@@ -85,31 +89,170 @@ def get_history(case_id, context_id, hardware_hash, repo) -> List[tuple]:
...
+    the PR.
+
+    Should not raise anything. If we can't find a z-score for some reason, the z_score
+    attribute will be None on that BenchmarkResult.
+    """
+    # {(case, context, hardware, repo, commit): (mean, sd)}
+    cached_stats: Dict[
+        Tuple[str, str, str, str, str],
+        Tuple[Optional[float], Optional[float]],
+    ] = {}
```
jonkeane: I don't think I saw comments about this above, I'm curious if you could share a bit of how frequently this ends up getting hit | used? My first reaction is that it might be filling in a cache that gets used exactly once for each entry. But I might be missing something!

conbench/entities/history.py
```diff
@@ -85,31 +89,170 @@ def get_history(case_id, context_id, hardware_hash, repo) -> List[tuple]:
...
+        repo = benchmark_result.run.commit.repository
+        baseline_sha = baseline_run.commit.sha
+
+        if (case, context, hardware, repo, baseline_sha) not in cached_stats:
+            # Fill in the cache with all the stats for this history
+            history = get_history(case, context, hardware, repo)
+            for res in history:
+                cached_stats[(case, context, hardware, repo, res.sha)] = (
+                    res.rolling_mean_including_this_commit,
+                    res.rolling_sd_including_this_commit,
```
jonkeane: Does `rolling_sd_including_this_commit` include the current contender commit value in this calculation? I think yes, but want to double check.

conbench/entities/history.py
```diff
@@ -85,31 +89,170 @@ def get_history(case_id, context_id, hardware_hash, repo) -> List[tuple]:
...
+        .over(
+            partition_by=history.c.segment_id,
+            order_by=history.c.timestamp,
+        )
+        .label("rolling_mean_including_this_commit"),
+    ).subquery()
+
+    # Add same column, excluding the current commit
+    history = Session.query(
+        history,
```
alistaire47: Nit: Instead of redefining `history` repeatedly, it might be nice to name the subqueries more specifically like CTEs (though I have no idea if they're translated that way). 

Though the comment implies this is mutation (in the dplyr sense), I don't think it quite is in the SQL—this table really is getting redefined. Although maybe that's how mutation gets translated sometimes too? Idk; as I said, a nit

conbench/entities/history.py
austin3dickey: Oh yeah good callout - for most use cases (get one BenchmarkResult, get all BenchmarkResults with one `batch_id`, get all BenchmarkResults with one `run_id`) the cache will only be used once per BenchmarkResult. Not super useful there. But there are a few use cases (get all BenchmarkResults sharing a case name, get the last 1000 BenchmarkResults) where multiple BenchmarkResults could share the same case/context/hardware/repo, and we'd already have the history for the subsequent ones in the cache.

conbench/entities/history.py
```diff
@@ -85,31 +89,170 @@ def get_history(case_id, context_id, hardware_hash, repo) -> List[tuple]:
...
+        func.coalesce(
+            func.avg(history.c.mean).over(
```
alistaire47: In an ideal state I'd rather do this on the raw iteration data instead of the mean (or weight by the number of iterations). But since the number of iterations should almost always be constant I'm grudgingly alright with this, with the caveat that there are some minor corner cases.

conbench/entities/history.py
austin3dickey: Actually, your comment gave me an idea: we could further take advantage of this cache in the `/api/compare` endpoint, by appending all z-scores at once instead of calling `append_z_scores()` on all the baseline BenchmarkResults and then all the contender BenchmarkResults. I'm sure many comparisons share a case/context/hardware/repo. That's sorta the point!

conbench/entities/history.py
```diff
@@ -85,31 +89,170 @@ def get_history(case_id, context_id, hardware_hash, repo) -> List[tuple]:
...
+        ).label("rolling_mean_excluding_this_commit"),
+    ).subquery()
+
+    # Add column with the residuals from the exclusive rolling mean
+    history = Session.query(
+        history,
+        (history.c.mean - history.c.rolling_mean_excluding_this_commit).label(
+            "residual"
+        ),
+    ).subquery()
```
alistaire47: When we were talking about this, I was thinking of using a single mean for the entire segment instead of the rolling mean. The advantage of rolling is it won't change if recalculated later; the disadvantage is at the beginning of a segment it will be crap even if calculated later. But stability and reproducibility are arguably more important than that, so maybe this is better.

conbench/entities/history.py
austin3dickey: Yeah, it does. I think the way it's written now is confusing. I'm gonna rewrite it a little.

conbench/entities/history.py
```diff
@@ -85,31 +89,170 @@ def get_history(case_id, context_id, hardware_hash, repo) -> List[tuple]:
...
+
+    # Add column with the rolling standard deviation of the residuals
+    # (these can go outside the segment since we assume they don't change much)
+    history = Session.query(
+        history,
+        func.stddev(history.c.residual)
+        .over(order_by=history.c.timestamp)
+        .label("rolling_sd_including_this_commit"),
+    ).subquery()
```
alistaire47: I do worry without any recency weighting or windowing bounds will essentially become fixed in long histories even if they shouldn't. That can be a later problem, but I don't want us to lose sight of it.

conbench/entities/history.py
```diff
@@ -85,31 +89,170 @@ def get_history(case_id, context_id, hardware_hash, repo) -> List[tuple]:
...
+                range_=(None, -1),
+            ),
+            history.c.mean,
+        ).label("rolling_mean_excluding_this_commit"),
+    ).subquery()
+
+    # Add column with the residuals from the exclusive rolling mean
+    history = Session.query(
+        history,
+        (history.c.mean - history.c.rolling_mean_excluding_this_commit).label(
```
alistaire47: So in this case, using the mean instead of the raw data will have implications for the stats for z-score thresholds—5 sigma is not 5 sigma if the number of iterations is sometimes 1 and sometimes 100. Not sure if this is an in-practice problem, but this is worth looking into.

conbench/entities/history.py
```diff
@@ -85,31 +89,170 @@ def get_history(case_id, context_id, hardware_hash, repo) -> List[tuple]:
...
+
+        case = benchmark_result.case_id
+        context = benchmark_result.context_id
+        hardware = benchmark_result.run.hardware.hash
+        repo = benchmark_result.run.commit.repository
+        baseline_sha = baseline_run.commit.sha
+
+        if (case, context, hardware, repo, baseline_sha) not in cached_stats:
+            # Fill in the cache with all the stats for this history
+            history = get_history(case, context, hardware, repo)
```
alistaire47: Checking my understanding: By this comment, I'm assuming that this in practice runs once and then uses cached results, even though this gets called for every `BenchmarkResult`? Otherwise I was worried about this function repeatedly doing the same work, but if there's caching it should be fast.

alistaire47 (approved):
A few thoughts on what's here and what's in the future, but overall this looks pretty clean and tight!

conbench/entities/history.py
austin3dickey: Haha, yes I think this all gets translated into CTEs. So it's like
```sql
select *, func2(y) as new_col2 from (
  select *, func(x) as new_col from (
    ...
  )
)
```

I could rename these python variables to different things if you think that's cleaner. I originally had it like that, but then thought it would be easier to reorder them or, like, delete a line, if they were all named the same.

conbench/entities/history.py
austin3dickey: Yeah that's fair. Actually we should really be using something smarter than the mean. We've discussed this in https://github.com/conbench/conbench/issues/594, https://github.com/conbench/conbench/issues/535, etc... I think `min` could be a good candidate here (except it should be `max` for benchmarks where `not _less_is_better(unit)`). I'll make an issue to come back to this.

austin3dickey: https://github.com/conbench/conbench/issues/640

conbench/entities/history.py
austin3dickey: Yeah I thought stability and reproducibility were more important. Since we're making the assumption that the mean is always pretty stable in during one segment (and if it's not we need to break up the segment) I think it shouldn't get too crappy at the beginnings.

conbench/entities/history.py
jonkeane: Ok, I don't think that we want that to be how we calculate it, we probably want to use `rolling_sd_excluding_this_commit` since we don't want the contender to influence the distribution. With enough observations in history the impact here should be small, but we don't want something that is a big outlier to make it less likely we would detect that because we're including it in the distribution we're testing against.

conbench/entities/history.py
austin3dickey: https://github.com/conbench/conbench/issues/641

conbench/entities/history.py
austin3dickey: Oh I see what you were asking. Yeah sorry, this is what's confusing about how I wrote this code. We aren't including the contender commit in the distribution when calculating z-scores. We ARE including the baseline commit though, and this code is accessing the distribution AT the baseline commit, so we use `_including_this_commit` not `_excluding_this_commit`.

I'll rewrite this to make it more clear. :)

conbench/entities/history.py
jonkeane: > Actually we should really be using something smarter than the mean. We've discussed this in https://github.com/conbench/conbench/issues/594, https://github.com/conbench/conbench/issues/535, etc... I think min could be a good candidate here (except it should be max for benchmarks where not _less_is_better(unit)). I'll make an issue to come back to this.

I'll comment on the issue there as well. Which measure to use isn't totally clear and is somewhat up to interpretation (and add more details about all the possible ones there too). For now, keeping this is probably the most pragmatic thing to do — and might be good enough for a while still, but as we (re)visit (multiple) metrics I suspect we will bump into this and end up needing to be more flexible and allow multiple measures here

conbench/entities/history.py
austin3dickey: Changed it for you :)

conbench/entities/history.py
austin3dickey: I think actually I'll keep it this way. It's been a pain to rearrange the query when there are multiple variable names.

conbench/entities/history.py
austin3dickey: Luckily this could be nice and parametrizable!

conbench/entities/history.py
austin3dickey: Yeah at this point "z-score" is a very loose term. :) I don't think there's anything "normal" about a lot of benchmarking data

conbench/entities/history.py
austin3dickey: Yes, see this discussion: https://github.com/conbench/conbench/pull/636#discussion_r1090929658

conbench/entities/history.py
```diff
@@ -68,15 +72,42 @@ class HistorySerializer:
...
+    repo: str,
+    include_current_commit_in_rolling_stats: bool = True,
+) -> list:
     """Given a case/context/hardware/repo, return all non-errored BenchmarkResults
     (past, present, and future) on the default branch that match those criteria, along
     with information about the stats of the distribution as of each BenchmarkResult.
-    Order is not guaranteed.
+    Order is not guaranteed. Used to power the timeseries plots and z-scores on
+    BenchmarkResults.
+
```
austin3dickey: @jgehrcke Hopefully the new docstring below clarifies many of your questions!

austin3dickey:
The only CI failure is #637.

austin3dickey:
I connected to the Arrow database from my local laptop with these changes, and ran the code corresponding to different API calls on this branch. Most endpoints perform about the same. The only thing is I see a big slowdown in the `/api/benchmarks?run_id=ABC` pattern, which is a pretty well-used endpoint. Most of the time is spent in `get_history()` in every iteration, because each result has a different case/context.

Like we discussed above, we don't really need ALL the history if we just want 1 run (1 commit)'s worth. I think I could add a different function that re-uses a lot of the same code but is optimized for this query pattern.

Ordinarily I'd say let's just ship this and speed up that endpoint with a followup patch. But if we deploy it now, I think the Arrow conbench webapp will start to give 500s again. So I'd better fix this before we merge.

conbench/entities/history.py
jgehrcke: > I think the verb append might be a little clearer in that regard than set.

"Append" is a rather specific verb in programming which implies that a value is appended to some kind of ordered container object. That is, doing "append" multiple times usually adds elements, retaining the previous ones. When I saw "append" it told me that when I call `append_z_scores(input)` twice on the same `input` that the result is different the second time than it was the first time.

IMO when one says "set the `z_score` attribute on the Python object" then that's a precise, explicit description of the behavior using established terminology. But even then it makes sense to add a statement about the intended behavior when calling this more than once for the same input. That is why I asked about overwrite behavior, and how you would specify it. Should this skip (to save time) when it sees an object that has this attribute already set? Should it error out to help reveal programming errors? Should it  re-calculate and replace the previous value?

It's not too important to further change this. 

I just wanted to clarify where I am coming from when I _read_ code and try to understand the interface of this function, the intended behavior.

conbench/entities/history.py
jgehrcke: great exchange, thanks

conbench/entities/history.py
jgehrcke: > For now, keeping this is probably the most pragmatic thing to do

:muscle:

conbench/entities/history.py
```diff
@@ -68,15 +72,42 @@ class HistorySerializer:
...
+        - ``rolling_mean`` - the rolling mean of the current segment (calculated using
+          the mean of each BenchmarkResult). This window is an "expanding window." That
+          is, the window will use unlimited past data (as long as it doesn't go outside
+          the segment).
+        - ``residual`` - the difference of the current BenchmarkResult's mean from the
+          rolling mean.
+        - ``rolling_stddev`` - the rolling standard deviation of the residuals. This is
+          also an expanding window, and it always goes back to the beginning of history,
+          even beyond the start of the segment.
```
jgehrcke: > the rolling standard deviation of the residuals. This is also an expanding window, and it always goes back to the beginning of history, even beyond the start of the segment

Hm :). Skeptical about this methodological choice (calculating mean and stddev from different sample collections). It's probably fine given the goals (which are -- currently at least -- maybe mainly around visualization). It will be easier to discuss details like this with  https://github.com/conbench/conbench/issues/644 in mind.

conbench/entities/history.py
jgehrcke: Thank you, this is really a great improvement. Yes!

jgehrcke:
> The only CI failure is https://github.com/conbench/conbench/issues/637.

That should be gone when you rebase on current HEAD of main.

jonkeane:
Following up on a few 1:1 convos we've had: we probably cannot simply remove the lookback window just yet. If we included all of history now, we would need to go back and label all of the discontinuities that we've observed in our long running conbenches (e.g. the arrow conbench) or else the standard deviation calculation would be the entire distribution from all time (and thus be super wide for any benchmark that has ever had variation).

If we keep the 100 or 1000 commit limit for now and encourage folks to use the manual regression button, we will have at worst the same performance we have today (if no one touches the button) and possibly much improved (if folks do use the button). Once we have recency scaling or auto labelling we (ideally both) we should be able to open this up without as much of a worry.

conbench/entities/history.py
austin3dickey: Yeah, this is a sugary technique to allow us to start returning z-score information immediately after a marked distribution change. Basically, we're baking in the assumption that while the mean might change after a distribution change, the spread does not change, at least by the magnitude that would make a difference in analysis. This unlocks the ability to provide somewhat-meaningful alerts immediately after recognized regressions or improvements.

With #641 the impact of this baked-in assumption will lessen. It probably will never be perfectly rigorous. But I think the tradeoff is worth it.

austin3dickey: I added a little to the docstring to explain this choice.

austin3dickey:
>We probably cannot simply remove the lookback window just yet.

Okay, I added a configuration parameter back in (and updated the docs). Now instead of `DISTRIBUTION_COMMITS`, the parameter is `ROLLING_WINDOW_SIZE_SEC`. It actually represents a time interval instead of number of commits. The default is 60 days, which for Arrow and Velox is probably around 350 commits or so at their current speeds. We can always change it in their respective servers. I just didn't want too small a default.

The reason for the change is the way the query works; we'd like to return the distribution mean/stddev excluding all BenchmarkResults on the current commit. If we rolled with a window of e.g. (-1000, -1) _rows_, this could accidentally include other BenchmarkResults on the current commit. So we have to roll by commit timestamp, not rows. 

(This problem would exist using either SQL or pandas. The reason it didn't exist before is we were calculating stats for 1 BenchmarkResult at a time, so we could filter to the last 100 commits before doing aggregate calculations, instead of computing a true rolling window function.)

If we still really want `DISTRIBUTION_COMMITS` instead, I could get it to work, but it would take a little more time. (I'd probably make a monotonically increasing commit_number column or something and roll by that.)

Thoughts @jonkeane @alistaire47 @jgehrcke ?

jonkeane:
> Okay, I added a configuration parameter back in (and updated the docs). Now instead of DISTRIBUTION_COMMITS, the parameter is ROLLING_WINDOW_SIZE_SEC. It actually represents a time interval instead of number of commits. The default is 60 days, which for Arrow and Velox is probably around 350 commits or so at their current speeds. We can always change it in their respective servers. I just didn't want too small a default.

I think we really want to do this by number of commits and not date. A few reasons: 

1. With a date range like this there are varying window sizes. For large and quickly moving repositories, this is "probably fine" but if a repo were stable and didn't see frequent changes, these windows could be 1-2 commits large. But even just describing this in the documentation would need a decent chunk of words to describe (which to me is a smell that somethings wrong here).
2. Similar to the above, but there are a bunch of new edgecases we would need to deal with here: what if there haven't been _any_ commits for 60 days?
3. When we look into our next step of weighting distribution values by recency, we _really_ don't want those to be weighted be second or days since so we'll have to back this change out then. If this were the only reason, we might consider doing it anyway, but given the others here I don't think we want to do this.

I haven't yet looked at the code (but I thought ^^^ were important to get out before actually reviewing), but I wonder if that it's seemingly hard to compile the list of commits and then filter out this commit, leaving the rest of the list before, this is showing a larger issue with the way these queries are structured?

jonkeane:
Would it help to decouple the rolling mean in the visualization(s) from what's necessary for the z-score calculation? 

Ideally these use the same concepts | calculations | code, but for the z-score calculations, the means and SDs aren't really rolling, it's that observations we are considering (n commits back), and then the means are by group, the SDs on the residuals from those means. 

The visualizations will need to calculate those things commit by commit, but maybe we can do that in the visualization code instead? IMO if we have to roll back some of the visualization to land the z-score improvements that's an ok trade off. Or maybe we store the values we need in the visualizations for now? Ideally we won't need to do that, but the history project + z-score improvements are blocking improved alerting, fewer false positives, etc.

conbench/entities/history.py
```diff
@@ -85,31 +119,188 @@ def get_history(case_id, context_id, hardware_hash, repo) -> List[tuple]:
...
+            order_by=s.extract("epoch", history.c.timestamp),
+            range_=(
+                -Config.ROLLING_WINDOW_SIZE_SEC,
+                0 if include_current_commit_in_rolling_stats else -1,
+            ),
+        )
+        .label("rolling_stddev"),
     )
+
+    return history.all()
```
jonkeane: It might be helpful to render the SQL for these queries here so we can think about ways to adapt this to having a window of fixed commit sizes. Is that something you could do and post here?

Or maybe it would be helpful for the two of us to pair on this tomorrow and see if we can get this calculation going

austin3dickey:
>I think we really want to do this by number of commits and not date

Yes, the reasons you gave are quite convincing. After my restructuring for performance, I don't think it will be too hard to switch back to number of commits.

austin3dickey:
Okay, I pushed a "final" version of this PR. It includes the `DISTRIBUTION_COMMITS` configuration again. API "get" performance is much more reasonable. There's still a slight slowdown (expected since we're doing more work on read) but it's not a huge order of magnitude.

I also noticed a pretty big warm-up effect when executing the same queries repeatedly, suggesting to me that we could better tweak the database to expect these kinds of queries. But I think we can put that work off until we start more discretely measuring performance in #493.

austin3dickey:
Build is red because we exhausted the API rate limit again.

austin3dickey:
I'm going to go ahead and merge this to monitor the staging deployment before the weekend. Thanks for everyone's thoughtful feedback!

jgehrcke:
:rocket: glad that we landed this, congrats! Lovely work, great discussion.

PR #620, opened 2023-01-24
util: consolidate tznaive_dt_to_aware_iso8601_for_api, add tests

jgehrcke:

PR #647, opened 2023-01-31
Run.timestamp: identify and document current behavior with comments and tests

jgehrcke:
This patch is for https://github.com/conbench/conbench/issues/609, analogue to https://github.com/conbench/conbench/pull/620 and https://github.com/conbench/conbench/pull/623.

What was not clear to me, but is now:

- `Run.timestamp` is set upon DB insertion
- Cannot be set/defined by API clients.

Added a corresponding test to lock this in.

TODO: we should document the Run objects returned by the API (as part of API spec) and then document `timestamp` to correspond to the Run DB insertion/creation time.

This PR makes it so that the API emits `Run.timestamp` now via a tz-aware ISO 8601 string, indicating UTC timezone (before, it was ambiguous). This is achieved by using the new `conbench.util.tznaive_dt_to_aware_iso8601_for_api()` instead of just `naivedatetimeobject.isoformat()`.

This is magic, and brittle:
```python
s.DateTime(timezone=False), server_default=s.sql.func.now()
```

What really happens when `s.sql.func.now()` has the notion of a non-UTC timezone and the result is then stored in a Column defined by `DateTime(timezone=False)`? Is the timezone information brutally discarded, or properly used in a timezone transformation calculation? I have added a test that I think confirms that the conversion is done properly.

### DIFF
--- a/conbench/entities/run.py
+++ b/conbench/entities/run.py
@@ -220,8 +220,9 @@ class _Serializer(EntitySerializer):
             "id": run.id,
             "name": run.name,
             "reason": run.reason,
-            # TODO: also use tznaive_dt_to_aware_iso8601_for_api
-            "timestamp": run.timestamp.isoformat(),
+            "timestamp": conbench.util.tznaive_dt_to_aware_iso8601_for_api(
+                run.timestamp
+            ),
             "finished_timestamp": conbench.util.tznaive_dt_to_aware_iso8601_for_api(
                 run.finished_timestamp
             )
--- a/conbench/entities/run.py
+++ b/conbench/entities/run.py
@@ -37,7 +37,16 @@ class Run(Base, EntityMixin):
     id = NotNull(s.String(50), primary_key=True)
     name = Nullable(s.String(250))
     reason = Nullable(s.String(250))
-    # tz-naive timestamp expected to refer to UTC time.
+    # `timestamp`  is never set by API clients, i.e. the
+    # `server_default=s.sql.func.now()` is always taking effect. That also
+    # means that this property reflects the point in time of DB insertion (that
+    # should be documented in the API schema for Run objects). A more explicit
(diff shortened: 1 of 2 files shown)


conbench/entities/run.py
```diff
@@ -211,8 +220,9 @@ def _dump(self, run):
             "id": run.id,
             "name": run.name,
             "reason": run.reason,
-            # TODO: also use tznaive_dt_to_aware_iso8601_for_api
-            "timestamp": run.timestamp.isoformat(),
+            "timestamp": conbench.util.tznaive_dt_to_aware_iso8601_for_api(
+                run.timestamp
+            ),
```
jgehrcke, line 225: It feels good when TODO items disappear :).

conbench/tests/api/test_runs.py
```diff
@@ -21,7 +23,7 @@ def _expected_entity(run, baseline_id=None, include_baseline=True):
         run.hardware_id,
         run.hardware.name,
         run.hardware.type,
-        run.timestamp.isoformat(),
+        tznaive_dt_to_aware_iso8601_for_api(run.timestamp),
```
jgehrcke, line 26: Just FYI. Without this change, the corresponding test failure would be something like
```
E         -  'timestamp': '2023-01-31T18:10:31.686148',
E         ?                                   ^^^^^^^
E         +  'timestamp': '2023-01-31T18:10:31Z',
```

The first line is the return value if `run.timestamp.isoformat()` (showing fractions of seconds and _no_ timezone information).

The second line is the return value of the new `tznaive_dt_to_aware_iso8601_for_api()`.

jgehrcke:
I assume that this is uncontroversial after https://github.com/conbench/conbench/pull/620 and https://github.com/conbench/conbench/pull/623. Would love to get post-merge feedback on this! Will land this now, and then open another PR with corresponding UI changes.

Issue #650, opened 2023-01-31
ci instability: screenshot job: `NoSuchElementException` may happen


jgehrcke:
This seems to be rare, but now that we wait explicitly for certain DOM elements to pop up (within a timeout period), this does not always seem to happen:
```
230131-18:51:10.027 root INFO: Found div#plot-history-0. Wait for Bokeh-generated div.bk-Column
230131-18:51:20.066 root ERROR: NoSuchElementException during _wait(): Message: no such element: Unable to locate element: {"method":"css selector","selector":"div.bk-Column"}
```
The log message timestamps show that 10 seconds passed waiting for this element. My intuition is that this element would not have popped up later. Something else went wrong here. Hm. It's probably important to take a screenshot _especially_ in this case, for better debuggability :).

Seen in https://github.com/conbench/conbench/actions/runs/4057323192/jobs/6982929333. Let's see how frequent this is.

jgehrcke:
> It's probably important to take a screenshot especially in this case, for better debuggability :).

doing that work in #651.

jgehrcke:
In https://github.com/conbench/conbench/pull/651 I found that the same symptom was as of a legit bug.

I retried the failed job on main on the same commit, and the result was positive (https://github.com/conbench/conbench/actions/runs/4057323192/jobs/6984413048).

jgehrcke:
This seems to have significantly improved. Cool.

alistaire47:
Happened here: https://github.com/conbench/conbench/actions/runs/4812113775/jobs/8566941131#step:4:1656

Log excerpt:

```log
230426-18:36:51.714 WDM INFO: ====== WebDriver manager ======
230426-18:36:51.755 WDM INFO: Driver [/root/.wdm/drivers/chromedriver/linux64/112.0.5615.49/chromedriver] found in cache
230426-18:36:52.421 root INFO: webdriver is set up
230426-18:36:52.525 root INFO: driver.get(): http://127.0.0.1:32770/benchmarks/064496ddaf7076aa8000c7f263013b60/
230426-18:37:03.103 root ERROR: NoSuchElementException during _wait(): Message: no such element: Unable to locate element: {"method":"css selector","selector":"div#plot-history-0"}
  (Session info: headless chrome=112.0.5615.165)
```

jgehrcke:
[...]

PR #651, opened 2023-01-31
UI: clarify Run creation time(zone), and Benchmark (start) time(zone)

jgehrcke:
This is for https://github.com/conbench/conbench/issues/534 and https://github.com/conbench/conbench/issues/609.

Landing page: this changes the column header from "Run" to "Run creation time", and starts showing the timezone in the individual table rows.

Benchmark result view: this now shows the Benchmark (start) time timezone.

Will add screenshots once they have been created by CI.

### DIFF
--- a/conbench/app/_util.py
+++ b/conbench/app/_util.py
@@ -20,8 +20,16 @@ def dataset_name(name):
     return name.replace("_", " ")
 
 
-def display_time(t):
-    return t.split(".")[0].replace("T", "  ").rsplit(":", 1)[0] if t else ""
+def display_time(t: str):
+    """
+    Expect `t` to be an ISO 8601 compliant string
+    - that encodes the UTC timezone with a Z suffix
+    - that does not contain fractions of seconds
+
+    Input example:  "2023-01-31Z05:36:45Z"
+    Output example: "2023-01-31 05:36:45 UTC"
+    """
+    return t.replace("T", " ").replace("Z", " UTC")
 
 
 def set_display_language(benchmark, contexts):
--- a/conbench/app/benchmarks.py
+++ b/conbench/app/benchmarks.py
@@ -236,6 +236,9 @@ class Benchmark(AppEndpoint, BenchmarkMixin, RunMixin, TimeSeriesPlotMixin):
 
 class BenchmarkList(AppEndpoint, ContextMixin):
     def page(self, benchmarks):
+        # Note(JP): What type is benchmarks? As of `response =
+        # self.api_get("api.benchmarks")` down below it's seemingly the result
+        # of JSON-decoding, i.e. not the DB model type.
         contexts = self.get_contexts(benchmarks)
         for benchmark in benchmarks:
             augment(benchmark, contexts)
--- a/conbench/app/benchmarks.py
+++ b/conbench/app/benchmarks.py
@@ -109,6 +109,9 @@ class RunMixin:
 
     def _augment(self, run):
         self._display_time(run, "timestamp")
+
+        # Note(JP): `run["commit"]["timestamp"]` can be `None`, see
+        # https://github.com/conbench/conbench/pull/651
         self._display_time(run["commit"], "timestamp")
         repository = run["commit"]["repository"]
         repository_name = repository
@@ -122,7 +125,14 @@ class RunMixin:
         run["commit"]["display_message"] = commit_message
 
     def _display_time(self, obj, field):
-        obj[f"display_{field}"] = display_time(obj[field])
+        timestring = obj[field]
+
+        # Seemingly this can be `None`.
+        if isinstance(timestring, str):
+            obj[f"display_{field}"] = display_time(timestring)
+
+        else:
+            obj[f"display_{field}"] = ""
 
     def _get_run(self, run_id):
         response = self.api_get("api.run", run_id=run_id)
--- a/conbench/templates/benchmark-entity.html
+++ b/conbench/templates/benchmark-entity.html
@@ -104,7 +104,7 @@
       {% if benchmark.stats %}
       <li class="list-group-item active">Result</li>
       <li class="list-group-item" style="overflow-y: auto;">
-        <b>timestamp</b>
+        <b>timestamp (start time)</b>
         <div align="right" style="display:inline-block; float: right;">
           {{ benchmark.display_timestamp }}
         </div>
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -99,12 +99,6 @@ jobs:
                 python screenshot.py "${CONBENCH_BASE_URL}/benchmarks/${SOME_BENCHMARK_ID}/" \
                   /artifacts benchmark-result-view --wait-for-canvas
             /bin/ls -ahltr ci-artifacts/
-            # If the browser-initiated GET requests led to e.g an Internal Server Error
-            # then we see that in the screenshot(s). For debugging that, however,
-            # it's crucial to get web application logs. Note that this shows
-            # container logs interleaved, but prefixed so that it's after all
-            # unambiguous which line came from which container.
-            docker compose logs --since 30m
       - name: upload-artifacts
         uses: actions/upload-artifact@v3
         # Upload screenshots _especially_ when result was unexpected.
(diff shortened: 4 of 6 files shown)


jgehrcke:
This PR is right now also affected by https://github.com/conbench/conbench/issues/650.

Piggybacking an improvement for this on this PR here. The GHA `if: always()` technique is described here: https://stackoverflow.com/a/58859404

Update: confirmed that the screenshots show "Internal Server Error". Nice. Hope this reproduces in the next run which should reveal the container logs (which should show the specific exception occurring).

jgehrcke:
After enhancing debuggability I saw that `ci/ui-screenshots` here failed with a genuine bug introduced by this PR:
```
conbench-app-1  |   File "/app/conbench/app/benchmarks.py", line 175, in get
conbench-app-1  |     benchmark, run = self._get_benchmark_and_run(benchmark_id)
conbench-app-1  |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
conbench-app-1  |   File "/app/conbench/app/benchmarks.py", line 233, in _get_benchmark_and_run
conbench-app-1  |     run = self.get_display_run(benchmark["run_id"])
conbench-app-1  |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
conbench-app-1  |   File "/app/conbench/app/benchmarks.py", line 88, in get_display_run
conbench-app-1  |     self._augment(run)
conbench-app-1  |   File "/app/conbench/app/benchmarks.py", line 112, in _augment
conbench-app-1  |     self._display_time(run["commit"], "timestamp")
conbench-app-1  |   File "/app/conbench/app/benchmarks.py", line 125, in _display_time
conbench-app-1  |     obj[f"display_{field}"] = display_time(obj[field])
conbench-app-1  |                               ^^^^^^^^^^^^^^^^^^^^^^^^
conbench-app-1  |   File "/app/conbench/app/_util.py", line 32, in display_time
conbench-app-1  |     return t.replace("T", " ").replace("Z", " UTC")
conbench-app-1  |            ^^^^^^^^^
conbench-app-1  | AttributeError: 'NoneType' object has no attribute 'replace'
conbench-app-1  | 172.18.0.1 - - [31/Jan/2023:19:33:17 +0000] "GET /benchmarks/45b8120bbc744ed4b277821b7bcbcb11/ HTTP/1.1" 500 0 "-" "-"
```

It's interesting that something as simple as HTTP `GET /benchmarks/45b8120bbc744ed4b277821b7bcbcb11/` is seemingly doing important functional test coverage not provided by the core test suite.

jgehrcke:
OK. Looks good. Screenshots from CI:

- Benchmark result view: https://user-images.githubusercontent.com/265630/215873391-62e466df-e867-4981-ab7f-349c322d42a5.png
- landing page: https://user-images.githubusercontent.com/265630/215873492-e0025eed-9ef6-47c4-b16b-82278896b03d.png

conbench/app/_util.py
```diff
@@ -20,8 +20,16 @@ def dataset_name(name):
...
+def display_time(t: str):
+    """
+    Expect `t` to be an ISO 8601 compliant string
+    - that encodes the UTC timezone with a Z suffix
+    - that does not contain fractions of seconds
+
+    Input example:  "2023-01-31Z05:36:45Z"
+    Output example: "2023-01-31 05:36:45 UTC"
+    """
+    return t.replace("T", " ").replace("Z", " UTC")
```
jonkeane, line 32: 😍  Love this

conbench/templates/benchmark-entity.html
```diff
@@ -104,7 +104,7 @@
       {% if benchmark.stats %}
       <li class="list-group-item active">Result</li>
       <li class="list-group-item" style="overflow-y: auto;">
-        <b>timestamp</b>
+        <b>timestamp (start time)</b>
```
jonkeane, line 107: I'm slightly hesitant to add this kind of strictness of "these times must be when they start" here since not all types of benchmark runs have that kind of data on a per-benchmark basis (e.g. our microbenchmarks will all have the same timestamp for either when the macro runner like `archery` is kicked off, or when the results are being adapted and uploaded, see https://conbench.ursa.dev/runs/4c9036a68cab402c94f8942634b59fca/ for example, each language has effectively one timestamp per). 

We could encourage folks to adapt these more accurately (though some of that might be an uphill battle to send PRs to google benchmark, folly, etc. in the fullness of time we should do that, but I don't think that's a today issue). But maybe leaving this ambiguous (or up to implementation specification) gives us space to be better about the precision when we can.  

Or maybe we merge the change in this PR for now and deal with any possible fallout if it happens (I think it's unlikely that anyone is relying on those columns being precise right now).

jonkeane (approved):

conbench/templates/benchmark-entity.html
jgehrcke: Love the feedback, it was rather expected.

Let me summarize what we talked about so far.

In https://github.com/conbench/conbench/issues/534 I described that as a fresh Conbench user I was confused by the lack of documentation. I asked:

> is this submission time (as recorded by Conbench), or start time or end time (as recorded and freely set by the API user)?

@austin3dickey said

> I've heard @jonkeane say in the past that we're leaving "start vs end" ambiguous on purpose, to allow for flexibility between different implementations. It's another one of those tradeoffs between easy implementation and data "meaningfulness". However, that doesn't mean we can't thoughtfully decide on a way to tempt users to congregate on one definition.

As an API consumer (I have programmed with many APIs and read API documentation of various qualities) I felt that this is too much ambiguity. I thought that if we don't want to specify anything about the meaning of this property then we have to at least document _that_, but I didn't feel good about adding that to the spec.

I really liked Austin's

> thoughtfully decide on a way to tempt users to congregate on one definition.

because when I build code to submit data and when I _can_ set this property freely, then if the documentation tells me how to set it then this is a no-brainer to implement. In contrast, if the documentation tells me "you are free to introduce any convention here" then I get uncertain and overthink this, and try to acquire more information so that I can make a good choice.

In https://github.com/conbench/conbench/pull/623 I added this statement to the _specification_ of the `BenchmarkCreate.timestamp` property:

> A datetime string indicating the time at which the benchmark was started. 

I commented on that PR:

> This takes some mental burden away from users and I think it makes sense in most of the cases. **To advanced users it should be obvious that any time can be communicated here**. In this API spec I think it's also valuable to let users know where/how this timestamp is used (result sort oder); and so I added that.

(emphasis mine, here)

Now it's easier to reply.

> add this kind of strictness of "these times must be when they start" here since not all types of benchmark runs have that kind of data on a per-benchmark basis

Makes total sense, this general concern. But I think it's fine because we don't _actually_ introduce strictness here, but we merely describe what this property _usually_ means, removing mental burden. We did this on the specification side of things via #623 for the input, and now we're doing the corresponding part on the output. "(usually the start time)" would have maybe been more appropriate than "(start time)", but it gets a little long for little additional value.


> when the results are being adapted and uploaded, see https://conbench.ursa.dev/runs/4c9036a68cab402c94f8942634b59fca/ for example, each language has effectively one timestamp per).

Thanks for showing an example. It's good to know when and _why_ the `timestamp` property shows something vastly different from the benchmark start time, and we can see if we can or even want to do something about that in the future.

If this ever gets annoying then our discussion here is the right way towards a _better_ solution, better and _more_ specification. Maybe the answer is to not ask for a _required_ timestamp property, but to automatically keep track of the time when this was inserted in the DB (and maybe have an _optional_ `benchmark_start_time` property that has precise specification). This here is a I think an example for a metadata concept that has no well-defined use case yet.

> maybe leaving this ambiguous (or up to implementation specification) gives us space to be better about the precision when we can

In practice with people coming and going it's just hard to be precise w/o strictness in specification. Deployment-specific conventions sound a bit hard to maintain :). I want to say again that what I saw about what we have so far was confusing.

>  (I think it's unlikely that anyone is relying on those columns being precise right now)

That. I think that reliance is unlikely to happen as long as there is no rather clear expectation (via documentation, specification, and such).

Super long reply. Sorry, I know this wasn't necessary. 

I think the summary is that this corresponds to #623, and that I do think this adds a bit of hand-holding that will lead to more happy API consumers and no actual reduction in flexibility, and so I would now follow your last thought here:

> Or maybe we merge the change in this PR for now and deal with any possible fallout if it happens

Thanks for this thoughtful exchange @jonkeane :heart:

conbench/app/benchmarks.py
```diff
@@ -236,6 +246,9 @@ def _get_benchmark_and_run(self, benchmark_id):
 
 class BenchmarkList(AppEndpoint, ContextMixin):
     def page(self, benchmarks):
+        # Note(JP): What type is benchmarks? As of `response =
+        # self.api_get("api.benchmarks")` down below it's seemingly the result
+        # of JSON-decoding, i.e. not the DB model type.
```
austin3dickey, line 251: yeah from what I've seen, most python modules in `conbench/app/` only work with dicts (the responses from the API), not the entities

Issue #656, opened 2023-02-01
GitHub HTTP API 403 errors: more logging for the reason, attempt to retry


jgehrcke:
This is about

```
2023-02-01T09:31:20.5177944Z conbench-app-1  | [230201-09:29:13.405] [10] [conbench.entities.commit] INFO: got unexpected HTTP response with code 403: {"message":"API rate limit exceeded for installation ID 19416693.","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
2023-02-01T09:31:20.5179249Z conbench-app-1  | [230201-09:29:13.406] [10] [conbench.entities.run] INFO: treat as unknown commit: error during get_github_commit(): 403 Client Error: Forbidden for url: https://api.github.com/repos/apache/arrow/commits/1e8ca94fc3682eb97bcf243545dcb282c1aaa0b4
```
as it affects CI and local dev every now and then. See https://github.com/conbench/conbench/issues/654.

We have ideas. From https://github.com/conbench/conbench/issues/655#issue-1565953026:

> There is more intermediate work we can do. More retrying, more backing off, better logging, maybe adding support for more than one API token so that auto-rotation can happen upon quota exhaustion.

It's important to distinguish quota exhaustion or _actual_ rate limiting error. The latter is retryable (at the cost of driving http request processing duration up). This can be detected by inspecting the response header, it's expected to contain something in the lines of `x-ratelimit-remaining: 4987`, referring to the hourly quota.

jgehrcke:
Interesting: https://docs.github.com/en/developers/apps/building-github-apps/rate-limits-for-github-apps#user-to-server-rate-limits-for-github-enterprise-cloud

jgehrcke:
This has been taken care of via https://github.com/conbench/conbench/pull/657.

PR #657, opened 2023-02-01
github http api: better retrying, logging

jgehrcke:
This is for https://github.com/conbench/conbench/issues/656.

Example log messages that this creates:

```
 [230201-12:24:57.872] [98] [conbench.entities.commit] INFO: x-ratelimit headers: X-RateLimit-Limit: 60, X-RateLimit-Remaining: 0, X-RateLimit-Reset: 1675257726, X-RateLimit-Resource: core, X-RateLimit-Used: 60
```

```
[230201-12:39:23.514] [8] [conbench.entities.commit] INFO: attempt 5 failed, wait for 5.500 s
[230201-12:39:29.014] [8] [conbench.entities.run] INFO: treat as unknown commit: error during get_github_commit(): _get_response(): deadline exceeded, giving up after 16.061 s
```

With this we will know if we run into quota issues or actual rate limiting issues. If it's the latter then the additional retrying behavior might auto-heal the problem.

Note that locally I have tried to hit every code path of this. While this adds quite a few lines of code and maybe some new complexity, I am pretty confident that this does not introduce new bugs. linting/mypy feedback also helped. :pray: / hope :D

### DIFF
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -587,6 +587,8 @@ class GitHub:
         Gunicorn has a worker timeout behavior (as of the time of writing: 30
         seconds) and the retrying method below must come to a conclusion before
         that.
+
+        Return deserialized JSON-structure or raise an exception.
         """
         timeout_seconds = 20
 
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -2,6 +2,7 @@ import functools
 import json
 import logging
 import os
+import time
 from datetime import datetime, timedelta
 from typing import Dict, List, Optional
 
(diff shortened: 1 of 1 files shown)


conbench/entities/commit.py
```diff
@@ -578,26 +579,111 @@ def _parse_commit(commit):
...
-        # An HTTP response was received.
-        if response.status_code != 200:
-            # This is here to log details an make debugging easier.
+        timeout_seconds = 20
+
+        t0 = time.monotonic()
+        deadline = t0 + timeout_seconds
+        attempt: int = 0
+
+        while time.monotonic() < deadline:
```
jgehrcke, line 599: This is btw a standard recipe I have used many times over the years. In _this_ part of the code I rather have all the flexibility and also readability (better than using an involved retrying library IMO).

The only "success" way to leave this function is via `return result`. In all other cases we at some point break from this loop and raise the "deadline exceeded" exception.

jgehrcke:
I would love to get this into CI for the next PR. Post-merge review very welcome!

Issue #645, opened 2023-01-31
history plot: review Bokeh plot tools


jgehrcke:
> Personally I think it would still be handy to have access to a "box zoom" (https://docs.bokeh.org/en/2.4.0/docs/user_guide/tools.html#boxzoomtool) to more easily zoom into exactly the area you want.

(https://github.com/conbench/conbench/issues/335#issuecomment-1118659131)

> We could do this in a separate PR if you wish, but would you mind enabling the feature of "click/drag on the plot to create a selection range to zoom in to"? I find myself constantly trying to do that on these pages and being disappointed when it pans instead

(https://github.com/conbench/conbench/pull/611#discussion_r1082873420)

jgehrcke:
Closing this as 'done'.

PR #659, opened 2023-02-01
history plot: rework plot tools (add wheel zoom and box zoom, and more)

jgehrcke:
Motivated by https://github.com/conbench/conbench/issues/645

- Add `wheel_zoom` and remove `zoom_in` and `zoom_out`. I think all users can do `wheel_zoom`, including those that use a touch pad. The current wheel_zoom configuration affects x and y axis at the same time.
- Add `box_zoom` (I agree with @jorisvandenbossche and @austin3dickey that this is a useful tool).
- Explicitly define the active tool for each type of user input.
  - Enable `wheel_zoom` by default (upon scroll).
  - Keep `pan` enabled by default upon drag, require active user choice (button click) before `box_zoom` is enabled.

A quick video showing the current state:

https://user-images.githubusercontent.com/265630/216070836-ec06179c-9f52-46a0-92d7-3257ab761219.mov

### DIFF
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -513,7 +513,29 @@ def time_series_plot(history, benchmark, run, height=380, width=1100):
         x_axis_type="datetime",
         height=height,
         width=width,
-        tools=["pan", "zoom_in", "zoom_out", "reset", "tap"],
+        tools=[
+            # Allows for dragging the field of view, default drag action.
+            "pan",
+            # Allow for resetting the plot to original view.
+            "reset",
+            # This allows for benchmark-details-on-datapoint-click
+            "tap",
+            # Zoom in and out with mouse wheel, default wheel action
+            "wheel_zoom",
+            # Allow for drawing a box for zooming, not enabled by default,
+            # because the "pan" tool is enabled by default. Can be activated by
+            # clicking the icon on the toolbar (one can toggle manually between
+            # pan and box-zoom).
+            "box_zoom",
+        ],
+        # Bokeh toolbars can have at most one active tool from each kind of
+        # gesture (drag, scroll, tap).
+        # https://docs.bokeh.org/en/2.4.0/docs/user_guide/tools.html#setting-the-active-tools
+        active_drag="pan",
+        active_scroll="wheel_zoom",
+        active_tap="tap",
+        active_inspect=None,
+        toolbar_location="right",
         x_range=(t_start, t_end),
     )
     p.toolbar.logo = None
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -534,7 +534,7 @@ def time_series_plot(history, benchmark, run, height=380, width=1100):
         active_drag="pan",
         active_scroll="wheel_zoom",
         active_tap="tap",
-        active_inspect=None,
+        active_inspect="auto",  # this enables hover by default (tool added below)
         toolbar_location="right",
         x_range=(t_start, t_end),
     )
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -522,16 +522,17 @@ def time_series_plot(history, benchmark, run, height=380, width=1100):
             "tap",
             # Zoom in and out with mouse wheel, default wheel action
             "wheel_zoom",
-            # Allow for drawing a box for zooming, not enabled by default,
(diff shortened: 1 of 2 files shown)


jonkeane (approved):
I love the zoom box especially! I'm happy to merge this as is.

We might consider waiting on the scroll zoom — though I'm fine merging and seeing how folks feel about it. I sometimes get frustrated when scrolling and the page doesn't go cause I'm in some other area — but that might be just me

jgehrcke:
> I sometimes get frustrated when scrolling and the page doesn't go cause I'm in some other area — but that might be just me

Yeah, there's a trade-off here, I agree (we all like this full-width embedded map that starts to zoom when scrolling over it, right? :D :fire: ). In this particular case I think it's more of value than of annoyance. I think one rather easily can move the mouse pointer to a regime that scrolls the page (the plot is not huge). Happy to change this again of course if this doesn't pass the test of time/usage.

conbench/app/_plots.py
```diff
@@ -513,7 +513,29 @@ def time_series_plot(history, benchmark, run, height=380, width=1100):
...
+            # Allow for drawing a box for zooming, not enabled by default,
+            # because the "pan" tool is enabled by default. Can be activated by
+            # clicking the icon on the toolbar (one can toggle manually between
+            # pan and box-zoom).
+            "box_zoom",
+        ],
+        # Bokeh toolbars can have at most one active tool from each kind of
+        # gesture (drag, scroll, tap).
+        # https://docs.bokeh.org/en/2.4.0/docs/user_guide/tools.html#setting-the-active-tools
+        active_drag="pan",
```
austin3dickey: Since the default view allows us to see all data on the plot, I think the default drag should be box_zoom. There's not really a point to panning left or up when you can see everything already.

austin3dickey (approved):
looks good - just one nit

sayantikabanik:
More on the lines of interactivity and experience, I played around a bit with the options.

Initial I landed on the plot and while `tap` was enabled and the nice summary got displayed. 
Was hoping to `hover` and get just the overview of the points (which required clicking on the hover option). Was wondering if `tap` + `hover` could be enabled by default. 

It's was my experience and the way I interact with the plots, maybe different for other users :)

jgehrcke:
Thanks for the feedback @sayantikabanik!

> Was wondering if tap + hover could be enabled by default.

They are! :) (were before, and still are) I agree these are valuable.

sayantikabanik:
> They are! :) (were before, and still are) I agree these are valuable.

Ah, did I miss something here? when I land on this plot `hover` wasn't enabled
Screenshot 2023-02-02 at 2 51 30 PM

conbench/app/_plots.py
jgehrcke: This is a great point. Made me think and compare a little more.

I think after all it's the combination of two tools that creates the navigation comfort zone or not. I found wheel_zoom+pan rather powerful, gives a clear and intuitive method to navigate the plot very quickly along all dimensions. Right after having zoomed in it's good to be able to pan to correct/adjust the field of view slightly or by a lot.

I have just tried wheel_zoom+box_zoom locally as default combo and I found that there's a double click that after box_zoom let's you quickly zoom back out again, so you can re-adjust the next box zoom. That's OK, too.

For now, my personal favorite is  wheel_zoom+pan. But I choose to not care too much yet and have pushed a commit to make this wheel_zoom+box_zoom :) That's fine! Let's see how we use/like/find that in practice.

jgehrcke:
> Ah, did I miss something here? when I land on this plot hover wasn't enabled

It's enabled by default for current head of main (this is a cropped version of the last CI-generated screenshot of the history plot):
image

It has been enabled by default for a long while. Maybe what you're seeing is a bug/special case.  How did you get to that view that you screenshotted?

sayantikabanik:
I click a random run and landed there.
Though thanks for looking into this. Will create an issue (with more details) if I happen to stumble on this again.

jgehrcke:
> I click a random run and landed there.

Well, I was mainly interested in which environment. If local checkout: latest HEAD of main?

sayantikabanik:
Not sure if this helps
Previously I checkout out to this PR(s) branch last commit SHA was `72a487d`
(Now) Over main I could find the `hover` option enabled.

Issue #617, opened 2023-01-23
legacy CLI / webapp entanglement: separate legacy CLI into isolated package


austin3dickey:
We have systems like [arrow-benchmarks-ci](https://github.com/voltrondata-labs/arrow-benchmarks-ci) still using the `conbench` CLI and python package to define and run benchmarks. That package is legacy (to be replaced in the future with better tools like `benchrun` and `benchadapt`) but it should be stable, and especially not sensitive to changes in webapp code, for example.

If we split the server code into its own "package" (in quotation marks because it wouldn't ever need to be built/distributed) then we could feel a little freer to do certain things, like use python 3.11 features in the server without having to worry about 3.8 compatibility, or maintain different sets of dependencies without having to worry about incompatibility or build times.

jgehrcke:
Code organization is an important topic and I agree that CLI and web application code should be conceptually separated. But in view of the breakage, I think those four points I noted here are more relevant towards achieving confidence: https://github.com/conbench/conbench/issues/618#issuecomment-1401747409.

> That package is legacy (to be replaced in the future with better tools like benchrun and benchadapt) but it should be stable

Should really be version-pinned on consumer side. Changing the version/build requires an explicit patch with passing CI.

This here is also strongly relating to the relevance of unit tests for achieving actual confidence: if I want to make sure an HTTP API works for users, I want to test over actual HTTP. If I want to make sure that a CLI works for users, I want to test the actual CLI build.

jonkeane:
> Should really be version-pinned on consumer side. Changing the version/build requires an explicit patch with passing CI.

I'm fine to do this pinning now, but IMO we should be pragmatic about this pining if at all possible. As we iterate more quickly, it can be helpful to have arrow benchmarks ci "just" pull the latest means one less PR to implement changes. Of course that comes with more breakage. Now might be the right time to pin these things — but that will mean more PRs needed to get things out in prod. Thoughts about that weighing right now?

austin3dickey:
Right now, I think pinning to a commit or PyPI version is fine. I'm planning to give some love to PyPI and versioning after my current story. If we can be semi-rigorous about following semantic versioning, I've found that can be a really effective strategy to be able to get patches out without breaking users or causing them to constantly update pins.

jgehrcke:
We should prioritize this! I think of this like "isolate legacy CLI, tame importlib/version hell, and tame dependency management for legacy CLI" by now.

[...]

PR #662, opened 2023-02-02
Conbench on minikube with kube-prometheus (milestones 1, 2, 3)

### LINKED ISSUE #470: test suite: sometimes hangs in database create_all()
There is a transient problem when invoking the test suite -- I have seen that a handful of times in the last few days, out of probably hundreds of test invocations. Starts to bug me, so maybe this needs to be debugged at some point. Unclear if that is specific to my local environment or not. Notably, not seen in CI yet.

The symptom:

```
± docker-compose down && docker-compose build app && docker-compose run -e CONBENCH_LOG_LEVEL_STDERR=DEBUG app pytest -s -vv -k test_oidc_flow_against_dex conbench/tests
Stopping conbench_db_1  ... done
...
uccessfully tagged conbench_app:latest
Creating network "conbench_default" with the default driver
Creating conbench_dex_1 ... done
Creating conbench_db_1  ... done
Creating conbench_app_run ... done
[221130-13:55:22.256] [1] [conbench] DEBUG: Config.CREATE_ALL_TABLES appears to be set, call create_all()
[221130-13:55:22.256] [1] [conbench.db] DEBUG: Starting call to 'conbench.db.create_all', this is the 1st time calling it.
[221130-13:57:33.640] [1] [conbench.db] INFO: result after attempt 1 for <function create_all at 0x7f6c4dc685e0>: (psycopg2.OperationalError) connection to server at "db" (192.168.224.3), port 5432 failed: Connection timed out
	Is the server running on that host and accepting TCP/IP connections?

(Background on this error at: https://sqlalche.me/e/14/e3q8)
[221130-13:57:34.641] [1] [conbench.db] DEBUG: Starting call to 'conbench.db.create_all', this is the 2nd time calling it.
```

After all, the app is never able to connect to the containerized database, as of:

> connection to server at "db" (192.168.224.3), port 5432 failed: Connection timed out

This might have to do with the virtual network that is used under the hood, managed by docker-compose (seen many of these kinds of docker network instabilities in the past).

In that state the SIGINT is not quickly propagated, i.e. one cannot Ctrl+C out of this -- I do `docker ps`, then `docker kill ...` to kill the app container, and then I restart the docker-compose command.

jgehrcke:
This is supposed to implement the milestones 1, 2, 3 laid out roughly in https://github.com/conbench/conbench/issues/493#issuecomment-1413770175.

I'd like to write in more detail about what this patch is doing because code and code comments do not easily reveal the bigger picture.

This **instruments our web application with Prometheus metrics**. These metrics are exposed at the `/metrics` HTTP path of the gunicorn HTTP server that also serves the regular Flask app.

That metrics HTTP endpoint needs to be periodically pulled / "scraped" by Prometheus (in the Prometheus ecosystem, metrics are usually communicated in a pull paradigm, not a push paradigm), which then writes into its own time series database. Further below I describe how I chose to do that scraping.

I chose to use [prometheus_flask_exporter](https://github.com/rycus86/prometheus_flask_exporter) for now to get going quickly. It automatically records/measures traditional metrics around HTTP request handlers, providing insight into the health of a web application. For example: the duration required for building HTTP responses (histogram metric), the count of HTTP responses emitted with specific status codes (there's much more, this is just to give an impression).

This PR **introduces [minikube](https://minikube.sigs.k8s.io/docs/) to CI as a foundation to test k8s-specific parts of Conbench**. This allows for testing Conbench-related k8s YAML manifests in pull request CI (valuable because we embrace k8s as a/the recommended way to run Conbench, and we run it on k8s in prod ourselves). This PR introduces a CI check that after all runs Conbench on minikube and confirms that it works. 

That's really an implementation detail, but for running PostgreSQL on minikube next to Conbench I chose to use [postgres-operator](https://postgres-operator.readthedocs.io), to not spend time on configuring the stateful nature of running a DB on k8s. Great docs, this was a real time saver.

In production, on our k8s clusters running Conbench, I'd like for us to run [prometheus-operator/kube-prometheus](https://github.com/prometheus-operator/kube-prometheus) asap. Being able to roll this out with ~confidence soon is the main motivator for introducing minikube here in CI. prometheus-operator/kube-prometheus "just works", but it needs to be configured properly. The configuration surface is rather big. It's a mature and established project (boring tech, by now). It runs [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator), Grafana, Prometheus itself, [Alertmanager](https://prometheus.io/docs/alerting/latest/alertmanager/), and other proven components. All wired up for **generic k8s cluster system monitoring as well as for k8s application monitoring**. This provides a lot of value for free. I do not try to list all the value and use here. Two highlights though:
* Pre-built Grafana dashboards for per-node and per-container resource monitoring (CPU utilization etc). So far this has been a bit of a blind spot for our Conbench deployments.
* This automatically adds useful k8s metadata labels to metrics, i.e. one can for example precisely filter for the state of _one of many_ pods in a ReplicaSet. Example: `rate(flask_http_request_total{status="500", pod='conbench-deployment-79cc79bfbf-nzt54'}[2m])` shows the rate of HTTP responses with status code 500 emitted by _one specific pod_.

How does prometheus-operator/kube-prometheus know how and where to periodically scrape the /metrics HTTP endpoint served by Conbench containers? Prometheus Operator comes with custom resource definitions, one of which is the [ServiceMonitior](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#monitoring.coreos.com/v1.ServiceMonitor). This tells Prometheus Operator to use one of its managed Prometheus instances to start scraping a custom metrics endpoint.

An exaggerated perspective is that after introducing `prometheus_flask_exporter` and  `prometheus-operator/kube-prometheus` the only code required to connect the pieces together is this:

```yaml
# This is a CRD defined by the prometheus-operator/kube-prometheus project
kind: ServiceMonitor
metadata:
  name: conbench-service-monitor
spec:
  selector:
    matchLabels:
      app: conbench
  endpoints:
    # It's important that ServiceMonitor i) refers to a Service port  and ii)
    # refers to it via its name.
    - port: conbench-service-port
      path: /metrics
      scheme: http
      # This defines the time resolution.
      interval: 30s
```

The new CI check introduced in this PR deploys all mentioned components on minikube and then confirms that
- Conbench comes up health in the k8s cluster (can interact with DB, etc)
- Conbench's `/metrics` endpoint indeed gets scraped.

Areas of friction / limitations:

* This new CI check adds quite a bit of complexity and therefore new failure paths. There is a lot of "fetch from remote" and "wait until ready", and quite a bit of coordination happening. Plenty of room for new instabilities to pop up. We will iron those out over time.
* The `/metrics` endpoint is now served by the same gunicorn HTTP server that also serves the regular web application logic. That's nice and simple initially, but we might want to change that later.
* I have for now embraced the fact that we run gunicorn with process workers -- resulting in the "multiprocess challenge" where Prometheus client metrics need to be coordinated between processes -- that coordination happens through the file system. You see `PROMETHEUS_MULTIPROC_DIR` pop up here and there in the code, that's related to this topic. In the future this complexity can disappear I hope. It's totally fine for now.
* The new CI check took ~5-7 minutes, and is therefore for now the dominating check in terms of duration. I am pretty sure it can be sped up by doing smarter coordination / concurrent deployment. The GitHub Action that I picked for running minikube itself consumes between 1 and 2 minutes. That was a bit surprising.

More work is required before we can take advantage of all this in prod. The rough plan forward from here is laid out in https://github.com/conbench/conbench/issues/493#issuecomment-1413770175, milestones 4 and onwards.

-----

Quick architecture summary in terms of information flow. This is really basic stuff, but when I learned about the Prometheus ecosystem I wish someone had explained the simple things:

### Creating metrics, writing metrics to time series DB (the write path)

- Web app code continuously updates its internal metrics state (increases counters, etc). It's always doing that.
- A scraping Prometheus instances periodically scrapes /metrics to get a _current snapshot_ (current state of all counters, gauges, etc). It knows when it does that. It persists that current state (timestamp, and all those metrics) into its internal time series database. The time resolution is only ever as big as the scrape interval allows. The scrape interval is typically 30 s or 60 s. That is, one can typically say "the mean incoming HTTP request rate in this 30 s time window was ...". That's usually sufficient!

### Reading metrics for dashboarding, alerting etc (the read path)

- Grafana has the concept of "data sources". Prometheus can be configured as a data source for Grafana. The Grafana UI can be used to explore the contents of a Prometheus time series database. It can be used to construct queries, and it can be used to build dashboards. And of course it can be used to display dashboards that were built before.

Happy to also give a tiny presentation to go through the architecture.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,3 +101,40 @@ test-run-app-dev:
 	docker compose -f docker-compose.yml -f docker-compose.dev.yml \
 		up --build --wait --detach || (docker compose logs --since 30m; exit 1)
 	docker compose down
+
+# The version string representing the current checkout / working directory.
+# This for example defines the Docker image tags. The default `dev` suffix
+# represents a local dev environment. Override CHECKOUT_VERSION_STRING with a
+# different suffix (e.g. `ci`) in the CI environment so that the version string
+# attached to build artifacts reveals the environment that the build artifact
+# was created in.
+export CHECKOUT_VERSION_STRING ?= $(shell git rev-parse --short=9 HEAD)-dev
+# Set a different repo organization for pushing images to
+DOCKER_REPO ?= conbench
+
+CONTAINER_IMAGE_SPEC=$(DOCKER_REPO)/conbench:$(CHECKOUT_VERSION_STRING)
+
+$(info --------------------------------------------------------------)
+$(info CONTAINER_IMAGE_SPEC is $(CONTAINER_IMAGE_SPEC))
+$(info --------------------------------------------------------------)
+
+
+.PHONY: build-conbench-container-image
+build-conbench-container-image:
+	docker build . -f Dockerfile -t ${CONTAINER_IMAGE_SPEC}
+	echo "Size of docker image:"
+	docker images --format "{{.Size}}" ${CONTAINER_IMAGE_SPEC}
+	# docker push ${CONTAINER_IMAGE_SPEC}
+
+
+.PHONY: deploy-on-minikube
+deploy-on-minikube: build-conbench-container-image
+	minikube addons enable ingress
+	minikube status
+	mkdir -p _build
+	cat ci/minikube/deploy-conbench.yml.template > _build/deploy-conbench.yml
+	sed -i.bak "s|<CONBENCH_CONTAINER_IMAGE_SPEC>|${CONTAINER_IMAGE_SPEC}|g" _build/deploy-conbench.yml
+	rm _build/deploy-conbench.yml.bak
+	minikube image load ${CONTAINER_IMAGE_SPEC}
+	minikube kubectl -- apply -f  _build/deploy-conbench.yml
+
--- a/Makefile
+++ b/Makefile
@@ -105,12 +105,11 @@ test-run-app-dev:
 	docker compose down
 
 
-# The version string representing the current checkout / working directory.
-# This for example defines the Docker image tags. The default `dev` suffix
-# represents a local dev environment. Override CHECKOUT_VERSION_STRING with a
-# different suffix (e.g. `ci`) in the CI environment so that the version string
-# attached to build artifacts reveals the environment that the build artifact
-# was created in.
+# The version string representing the current checkout / working directory. The
+# default `dev` suffix represents a local dev environment. Override
+# CHECKOUT_VERSION_STRING with a different suffix (e.g. `ci`) in the CI
+# environment so that the version string attached to build artifacts reveals
+# the environment that the build artifact was created in.
 export CHECKOUT_VERSION_STRING ?= $(shell git rev-parse --short=9 HEAD)-dev
 # Set a different repo organization for pushing images to
 DOCKER_REPO_ORG ?= conbench
@@ -120,6 +119,7 @@ CONTAINER_IMAGE_SPEC=$(DOCKER_REPO_ORG)/conbench:$(CHECKOUT_VERSION_STRING)
 .PHONY: print-container-image-spec
 print-container-image-spec:
 	@echo $(CONTAINER_IMAGE_SPEC)
+
 
 .PHONY: build-conbench-container-image
 build-conbench-container-image:
@@ -145,22 +145,20 @@ start-minikube:
 		--extra-config=controller-manager.bind-address=0.0.0.0
 
 
-# This target is used by ci/minikube/test-conbench-on-mk.sh.
-# The `minikube image load` technique is rather new and allows for using local
-# Docker images in k8s deployments (as long as they specify `imagePullPolicy:
-# Never`). That command however takes a while for bigger images (about 1 min
-# per GB, on my machine).
+# This target is used by ci/minikube/test-conbench-on-mk.sh. The `minikube
+# image load` technique allows for using local Docker images in k8s deployments
+# (as long as they specify `imagePullPolicy: Never`). That command however
+# takes a while for bigger images (about 1 min per GB, on my machine).
 # https://minikube.sigs.k8s.io/docs/handbook/pushing/
 # https://stackoverflow.com/a/62303945
 .PHONY: deploy-on-minikube
 deploy-on-minikube:
 	minikube status
 	mkdir -p _build
-	cat ci/minikube/deploy-conbench.template.yml > _build/deploy-conbench.yml
+	cp ci/minikube/deploy-conbench.template.yml _build/deploy-conbench.yml
 	sed -i.bak "s|<CONBENCH_CONTAINER_IMAGE_SPEC>|${CONTAINER_IMAGE_SPEC}|g" _build/deploy-conbench.yml
-	rm _build/deploy-conbench.yml.bak
 	time minikube image load ${CONTAINER_IMAGE_SPEC}
-	minikube kubectl -- apply -f  _build/deploy-conbench.yml
+	minikube kubectl -- apply -f _build/deploy-conbench.yml
 
 .PHONY: minikube-bigflow
 minikube-bigflow: build-conbench-container-image start-minikube
--- a/Makefile
+++ b/Makefile
@@ -127,6 +127,21 @@ build-conbench-container-image:
 	docker images --format "{{.Size}}" ${CONTAINER_IMAGE_SPEC}
 	# docker push ${CONTAINER_IMAGE_SPEC}
 
+# Some of the cmdline flags are taken from
+# https://github.com/prometheus-operator/kube-prometheus#minikube
+# https://minikube.sigs.k8s.io/docs/commands/start/
+.PHONY: start-minikube
+start-minikube:
+	minikube delete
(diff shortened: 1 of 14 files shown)


docker-compose.dev.yml
```diff
@@ -7,14 +7,17 @@ services:
...
     command: [
       "gunicorn",
+      "-c", "conbench/gunicorn-conf.py",
       "-b", "0.0.0.0:5000",
       "-w", "2",
       "conbench:application",
       "--access-logfile=-",
       "--error-logfile=-",
-      "--reload"
+      "--preload"
```
jgehrcke: This is... interesting. Need to come back to that.

jgehrcke:
Huge progress locally. `kubectl logs deployment/conbench-deployment --all-containers` showed `"GET /metrics HTTP/1.1" 200 569 "-" "Prometheus/2.41.0"` in the gunicorn access log.

That means that some of the most important wires are set up 'correctly' now, or at least in a somewhat working way that can be iterated upon. As always when having so many moving pieces this is a tiny breakthrough moment.

What's somewhat working now:
- minikube itself, postgres itself on minikube, conbench on minikube
- prometheus-flask-exporter with a special setup to retain coordination between multiple gunicorn worker processes
- A `ServiceMonitor` object (this is a custom resource, as defined by the Prometheus operator ecosystem) properly configured against the `Service` (native k8s object) that represents the Conbench k8s deployment.

Makefile
```diff
@@ -97,7 +97,42 @@ run-app-bg:
 # fails then emit container logs before fast-failing the makefile target.
 .PHONY: test-run-app-dev
 test-run-app-dev:
-	docker compose down
+	mkdir -p /tmp/_conbench-promcl-coord-dir
+	docker compose down --remove-orphans
```
jgehrcke, line 110: I sneaked in an additional `--remove-orphans` here which fixes a rather annoying issue -- see #470.

Makefile
```diff
@@ -97,7 +97,42 @@ run-app-bg:
...
+
+
+# The version string representing the current checkout / working directory. The
+# default `dev` suffix represents a local dev environment. Override
+# CHECKOUT_VERSION_STRING with a different suffix (e.g. `ci`) in the CI
+# environment so that the version string attached to build artifacts reveals
+# the environment that the build artifact was created in.
+export CHECKOUT_VERSION_STRING ?= $(shell git rev-parse --short=9 HEAD)-dev
+DOCKER_REPO_ORG ?= conbench
+CONTAINER_IMAGE_SPEC=$(DOCKER_REPO_ORG)/conbench:$(CHECKOUT_VERSION_STRING)
```
jgehrcke, line 123: This is an attempt towards standardizing on globally applicable conbench web application container image naming. I don't think we upload to the Docker container hub (or something comparable) yet, but maybe we will. As long as we do not upload container images for the public to consume we can of course iterate on this naming convention.

The two make targets below use `CONTAINER_IMAGE_SPEC`.

ci/minikube/test-conbench-on-mk.sh
```diff
@@ -0,0 +1,181 @@
+#!/usr/bin/env bash
+set -o errexit
```
austin3dickey, line 2: wow, verbose names, fancy!

conbench/__init__.py
```diff
@@ -3,6 +3,8 @@
 import logging
 import os
 
+from prometheus_flask_exporter.multiprocess import GunicornInternalPrometheusMetrics
```
austin3dickey: As this stands, I think you'll need to add `prometheus_flask_exporter` to `requirements-cli.txt` as well, because this is the CLI's `__init__.py` too. Then we'll need to fix the [conda feedstock](https://github.com/conda-forge/conbench-feedstock). 

Or maybe you can trycatch this and do nothing if `prometheus_flask_exporter` isn't installed?

austin3dickey (approved):
This looks great! Obviously a lot of time put into it, so thank you. Just one note about this potentially breaking the conbench CLI, as usual. (#617)

conbench/__init__.py
jgehrcke: Ha! Thanks for catching that! Will think about this tomorrow. 

Looking forward to having more separation of concerns.

ci/minikube/test-conbench-on-mk.sh
jgehrcke: My brain refuses to learn which of those one-letter variants really means what (I think I got -e by now, but ...). I have always liked the long versions more :).

conbench/__init__.py
jgehrcke: I added a commit to deal with this, and also left a note in `conbench/__init__.py` to warn about the fact that this is used in two rather different contexts.

jgehrcke:
Thanks for the feedback. I am going to land this. I know it's a lot of new complexity. I take responsibility for all the challenges that this introduces and commit to making things smooth over time.

PR #663, opened 2023-02-02
Add an auto assigner when a PR is opened

jonkeane:
Though we should generally be better about assigning ourselves issues as we take them on, this cleans that up so that issues are automatically assigned when a PR is opened to the PR author (so long as the PR links to the issue, of course).

### DIFF
--- a/.github/workflows/auto-assign-issues-prs.yml
+++ b/.github/workflows/auto-assign-issues-prs.yml
@@ -1,0 +1,43 @@
+name: Auto assign pull requests and issues
+on:
+  pull_request:
+    types: [opened, reopened, edited]
+
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  assign-issue-and-pr-author:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: toshimaru/auto-author-assign@v1.6.2
+      - uses: octokit/graphql-action@v2.x
+        id: get_linked_issue
+        with:
+          query: |
+            query($repo: String!, $owner: String!, $pr_number: Int!) {
+              repository(owner: $owner, name: $repo) {
+                pullRequest(number: $pr_number) {
+                  id
+                  closingIssuesReferences (first: 50) {
+                    edges {
+                      node {
+                        id
+                        body
+                        number
+                        title
+                      }
+                    }
+                  }
+                }
+              }
+            }
+          owner: ${{ github.repository_owner }}
+          repo: ${{ github.event.repository.name }}
+          pr_number: ${{ github.event.pull_request.number }}
+      - run: "echo 'Issue output: ${{ steps.get_linked_issue.outputs.data }}'"
+      - run: |
+          echo '${{ steps.get_linked_issue.outputs.data }}' | \
+          jq -c '.. | objects | select(has("edges")) | .edges | .[].node.number' | \
+          xargs -I{} gh api repos/voltrondata/perfengtools/issues/{}/assignees -f assignees="${{ github.event.pull_request.user.login }}"
+


jgehrcke:
> so that issues are automatically assigned when a PR is opened to the PR author (so long as the PR links to the issue, of course).

I went into the README of this GitHub Action to understand the logic by which it tries to identify the issue(s) to mutate, based on the PR state.

I found

> In most cases, pull request author should be assigned an assignee of the pull request.

and this largely confused me -- what is this action really trying to do?

I went into [the code](https://github.com/toshimaru/auto-author-assign/blob/0fd2847dd52b3cb103d743049dcfb7f17d39c51f/src/index.js#L24) and it seems like this only ever mutates the pull request it acts on. That is, it adds an assignee to the pull request once that gets opened.

So, this is what this action is doing: the user that opens the pull request gets automatically set as assignee on that very pull request :detective: 

This action does not seem to try to inspect the pull request contents (description, commits, title) in order to identify related github issues. Maybe I missed that hidden magic :).

jonkeane:
> So, this is what this action is doing: the user that opens the pull request gets automatically set as assignee on that very pull request 🕵️

Yes it does do this. This is actually a bit tangential to the issue thing, but found it a bit weird that PR assignees aren't automatic it seems https://github.com/conbench/conbench/pulls only has some PRs assigned to people, for example – I haven't dug into if this is a configuration issue with either conbench or our user settings, cause I agree with the sentiment that PR authors should be the same as assignees generally, but was surprised that wasn't _already_ automatic in Gitubh. This behavior (assigning PR assignees) is technically not super important for automations around issues and PRs that I'm trying to expand (since PRs also have author fields, that can be used). But given that authors should generally be assignees like you mentioned, I figured add this for now so we follow that automatically.  

> This action does not seem to try to inspect the pull request contents (description, commits, title) in order to identify related github issues. Maybe I missed that hidden magic :).

It actually does inspect the pullrequest for issue linkage(s), but it'll be easier for me to show where in the code (which I'll do in just a sec)

.github/workflows/auto-assign-issues-prs.yml
```diff
@@ -0,0 +1,43 @@
...
+                        id
+                        body
+                        number
+                        title
+                      }
+                    }
+                  }
+                }
+              }
+            }
```
jonkeane, line 34: This query pulls from the PR that has triggered this event some details of the `closingIssuesReferences` https://docs.github.com/en/graphql/reference/objects#pullrequest (though you'll need to scroll down a little bit in the field reference

This is then passed (through `steps.get_linked_issue.outputs.data`) to the next step

.github/workflows/auto-assign-issues-prs.yml
```diff
@@ -0,0 +1,43 @@
...
+          owner: ${{ github.repository_owner }}
+          repo: ${{ github.event.repository.name }}
+          pr_number: ${{ github.event.pull_request.number }}
+      - run: "echo 'Issue output: ${{ steps.get_linked_issue.outputs.data }}'"
+      - run: |
+          echo '${{ steps.get_linked_issue.outputs.data }}' | \
+          jq -c '.. | objects | select(has("edges")) | .edges | .[].node.number' | \
+          xargs -I{} gh api repos/voltrondata/perfengtools/issues/{}/assignees -f assignees="${{ github.event.pull_request.user.login }}"
```
jonkeane, line 42: You might be wondering why I used GraphQL to retrieve the closing issue reference, but the standard API here.

As far as I could tell `closingIssueReferences` are _only_ available in GraphQL. I could have made a GraphQL mutate call to update the assignees, but that ends up being a lot more code and IMO iterating over the (possible) multiple issues was much less elegant (which is saying something cause this bash is _not particularly elegant_)

jgehrcke:
> It actually does inspect the pullrequest for issue linkage(s), but it'll be easier for me to show where in the code (which I'll do in just a sec)

Ah! That's in the code that you added on top of that GitHub Action (I only looked in their repo.... :-)). Thanks for clarifying.

Btw, I once went very deeep into the differences between their GraphQL API and the various HTTP API versions they have. It's a fascinating, but also irritating landscape.

jgehrcke (approved):
I'd say let's go for this and learn how it does in practice!

PR #667, opened 2023-02-03
Dockerfile: work on image size and cacheability

jgehrcke:
Breaking this out of https://github.com/conbench/conbench/pull/662.

This reduces the container image size:

before:
```
docker images --format "{{.Size}}" conbench/conbench:5a05aff71-dev 
1.49GB
```

after:
```
docker images --format "{{.Size}}" conbench/conbench:5a05aff71-dev
594MB
```

And, maybe even more importantly, it does not try to add the repo root directory content entirely into the image. This resulted in cache invalidation during local dev in many cases when it wasn't needed.

Why is this valuable for #662? There we use `minikube image load` to make a Docker image available in the k8s cluster. This command is slow, and the time it takes is linearly related to image size. The container image size reduction in this PR wins ~1 min!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,8 @@
 FROM python:3.11-slim
 
 # Curl is needed for docker-compose health checks.
-RUN apk add --update curl &&  rm -rf /var/cache/apk/*
+RUN apt-get update && apt-get install -y -q --no-install-recommends \
+    curl && apt-get clean && rm -rf /var/lib/apt/lists/*
 
 COPY requirements-webapp.txt /tmp/
 COPY requirements-dev.txt /tmp/
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,6 +25,9 @@ WORKDIR /app
 # to make most use of Docker container image layer caching.
 COPY conbench /app/conbench
 COPY migrations /app/migrations
+# TODO: make it so that .git is not needed
+# see https://github.com/conbench/conbench/pull/667
+COPY .git /app/.git
 COPY setup.py README.md requirements-cli.txt requirements-webapp.txt requirements-dev.txt alembic.ini /app/
 
 # Inspect contents of /app
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,8 @@
 FROM python:3.11-slim
 
+# Curl is needed for docker-compose health checks.
+RUN apk add --update curl &&  rm -rf /var/cache/apk/*
+
 COPY requirements-webapp.txt /tmp/
 COPY requirements-dev.txt /tmp/
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
 FROM python:3.11-slim
 
 # Curl is needed for docker-compose health checks.
+# and `git` is seemingly needed by some unit tests
 RUN apt-get update && apt-get install -y -q --no-install-recommends \
-    curl && apt-get clean && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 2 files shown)


jgehrcke:
Woof. In `test_runner.py` this PR sees validation errors like
```
{'github': {'commit': ['Missing data for required field.'], 'repository': ['Missing data for required field.']}}
```

https://github.com/conbench/conbench/blob/b524d43468bc4cfce04c7aba59bc2392a595355f/conbench/tests/benchmark/test_runner.py#L107


That is, the following code has some optional dependency:
```
    benchmark = CasesBenchmark()
    case = ("2", "10")
    [(result, output)] = benchmark.run(case=case, iterations=10)
```
It does not error out. But the `result` does not have `github` populated, and that's why validation starts to fail. But why is `github` not populated? I'd appreciate a clear error message or log message.

jgehrcke:
That was quite a bit of a time sink, but it's at least understood now what failed: https://github.com/conbench/conbench/issues/668.

We implicitly copied the `.git` directory into the container image. `machine_info.py` interacted with that, and found what it needed to find.

My patch here makes it so that `.git` is _not_ in the container image anymore. 

That's how those tests started failing.

I think as a quick workaround I think  will copy `.git` into the container image, together with a TODO note.

We should strive for having two container images anyway:
- one for deploying the Conbench webapp in prod. This should be slim, and not contain tests, lint dependencies, etc.
- one for CI with testing dependencies (test runner).

jgehrcke:
Current CI failure:
```
INFO     conbench.tests.entities.test_commit:test_commit.py:31 exc during backfill_default_branch_commits(): Hourly GitHub HTTP API quota exhausted
```
tracked here: https://github.com/conbench/conbench/issues/895#issuecomment-1473693995

Dockerfile
```diff
@@ -1,15 +1,44 @@
...
+RUN pwd && /bin/ls -1 .
+
+# Installing this package as of now is necessary not for installing
+# dependencies, but for preventing:
+# Answer: otherwise this happens:
+# importlib.metadata.PackageNotFoundError: No package metadata was found for conbench
 RUN pip install .
+
+#RUN echo "biggest dirs"
+#RUN cd / && du -ha . | sort -r -h | head -n 50 || true
```
austin3dickey, line 42: nit: remove comments unless you think they're useful to keep

conbench/machine_info.py
```diff
@@ -80,10 +80,6 @@ def r_info():
     }
 
 
-class GitParseWarning(RuntimeWarning):
-    pass
```
austin3dickey: I just [commented](https://github.com/conbench/conbench/issues/668#issuecomment-1416367880) on the issue you created about this. I'd rather not rip this out in this PR, because it creates some noise in the pytest summary. Maybe you can put it back and we can address your concerns in a followup PR?

austin3dickey (changes requested):
Looks good! I can't imagine the number of minutes I've spent adding `&& apt-get clean && rm -rf /var/lib/apt/lists/*` to Dockerfiles. Probably less than the number of minutes I've spent downloading big layers. :)

conbench/machine_info.py
jgehrcke: Thanks! I added my thoughts to that PR. You are right, it's better to not touch the warnings themselves here in this PR. I have just added a pragmatic check for the local git dir instead. As described in #668 I think that's the responsible thing to do, to potentially save others time.

Dockerfile
jgehrcke: I think this is useful. I have cleaned up comments in the Dockerfile a bit, and also added one here.

PR #677, opened 2023-02-06
gha: do not publish coverage report for now

### LINKED ISSUE #439: Coveralls is not performing optimally

jgehrcke:
See issue #439.

We can re-introduce when this signal is more interesting to us.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -33,21 +33,8 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v3
-      - name: Run tests, generate coverage data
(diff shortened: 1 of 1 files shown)


jonkeane (approved):
Thanks for this!

Issue #461, opened 2022-11-25
API endpoint showing build commit / version information, show commit in UI


jgehrcke:
Sometimes it's incredibly useful to quickly find out via HTTP API **which commit specifically a deployment is based on** (potentially requiring authentication). Producing an endpoint like this usually requires enhancing the complexity of the build chain a tiny bit.

jgehrcke:
This becomes more and more important with various slightly different versions running across deployments.

I looked into prior art and project-specific history on this topic. Found these discussions:
- https://github.com/conbench/conbench/issues/290
- https://github.com/conbench/conbench/issues/199
- https://github.com/conbench/conbench/pull/201


> Yup, revert back to version=1.33.0 in setup.py, and remove the setuptools_scm dependency. We can consider using setuptools_scm later, but for now I don't think we need it.

(https://github.com/conbench/conbench/pull/201#discussion_r734752095)

> I already implemented the setuptools_scm in setup.py. This has the added benefit that if you're working off of a dirty version -- it'll show the hash of the latest commit. 

(https://github.com/conbench/conbench/issues/199#issuecomment-949283046)

This was IMO absolutely the right intuition, but it was not followed because of inflating PyPI CLI releases with the web application versioning:

> but when I tested it I got the following which I don't think will work for PyPI

Related: https://github.com/conbench/conbench/issues/617

jgehrcke:
When the build artifact is a container image then the build/version information can and _should_ be baked into the image. 

Traditional tooling like setuptools_scm needed to adjust to this and the result is slightly ugly recipes like https://github.com/pypa/setuptools_scm/#usage-from-docker

I like to keep it simple and have CI create structured buildinfo in, say, a JSON file. That file can be baked into the container image. The web application can look it up at a path (by convention). If it finds it, it can display the version information in the UI, and it can expose it via API.

> (potentially requiring authentication)

Btw security concerns: yes, in established software there is the concern that exposing the precise version information implies telling attackers which attack vectors they can use. We have to choose not to care about this (yet!), otherwise we slow ourselves down. Security is never the result of analysis paralysis.

jgehrcke:
We show this in the UI now; and also (a little hidden) in the Ping API, example:
```
$ curl -s https://conbench.ursa.dev/api/ping/ | jq
{
  "alembic_version": "fe3b1041ed15",
  "commit": "8554cefdfbb6dcef4ca899a27f09c8496ebd4729",
  "conbench_version": "2023.2.8",
[...]

PR #678, opened 2023-02-06
UI: show short commit hash in footer (persist buildinfo into webapp container image)

jgehrcke:
This is for https://github.com/conbench/conbench/issues/461.

### DIFF
--- a/conbench/buildinfo.py
+++ b/conbench/buildinfo.py
@@ -1,11 +1,9 @@
-import logging
 import json
-
+import logging
+from dataclasses import dataclass
 from typing import Optional
 
-from dataclasses import dataclass
 import dacite
-
 
 log = logging.getLogger(__name__)
 
--- a/conbench/buildinfo.py
+++ b/conbench/buildinfo.py
@@ -1,0 +1,39 @@
+import logging
+import json
+
+from typing import Optional
+
+from dataclasses import dataclass
+import dacite
+
+
+log = logging.getLogger(__name__)
+
+
+@dataclass
+class Buildinfo:
+    # Full-length git commit hash.
+    commit: str
+    # Name of the branch carrying the commit.
+    branch_name: str
+    # Example value: 2023-02-06 12:01:01+00:00
+    build_time_rfc3339: str
+    build_hostname: str
+    # Canonical
+    version_string: str
+
+
+BUILD_INFO: Optional[Buildinfo] = None
+try:
+    # Try to discover additional build information at well-known path. For now,
+    # do not crash the application upon: file not found, file having unexpected
+    # contents, etc.
+    with open("/buildinfo.json", "rb") as fh:
+        data = json.loads(fh.read().decode("utf-8"))
+        log.info("decoded build info JSON: %s", data)
+
+    # `dacite.from_dict()` validates against dataclass types, rendering the
+    # type info on the dataclass reliable.
+    BUILD_INFO = dacite.from_dict(data_class=Buildinfo, data=data)
+except Exception as exc:
+    log.info("graceful degradation, could not read/parse buildinfo: %s", exc)
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -131,10 +131,15 @@ jobs:
         uses: actions/checkout@v3
       - name: Start PostgreSQL DB
         run: docker compose run --detach db
-      # Run `alembic upgrade head` via docker compose so that it is within
-      # the network that the DB is also in, reachable via DNS name `db`
+      # Run `alembic upgrade head` via docker compose. This runs in the web
+      # application container image. Running this via docker compose makes it
+      # so that it is within the virtual network that the DB is also in,
+      # reachable via DNS name `db`. Set CREATE_ALL_TABLES=false to defuse some
+      # webapp bootstrap code (not needed here, because these Python modules
+      # run in the context of alembic, not in the context of the actual web
+      # application bootstrapping itself).
       - name: Test database migrations
-        run: docker compose run -e CREATE_ALL_TABLES app alembic upgrade head
+        run: make set-build-info && docker compose run -e CREATE_ALL_TABLES app alembic upgrade head
         env:
           CREATE_ALL_TABLES: false
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,8 @@
+# conbench-specific things
+.buildinfo.json
+_build
+_new_api_docs_json
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,11 @@ RUN pwd && /bin/ls -1 .
 # importlib.metadata.PackageNotFoundError: No package metadata was found for conbench
 RUN pip install .
 
+# Bake build information into container image. This invalidates the layer
+# cache for every commit and therefore it is important to do this as late as
+# possible in this Dockerfile.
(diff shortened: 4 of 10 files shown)


jgehrcke:
`ci/dev-cmds` failed with an instability which I am now tracking here: https://github.com/conbench/conbench/issues/680. I have triggered a re-run.

conbench/buildinfo.py
```diff
@@ -0,0 +1,37 @@
...
+class Buildinfo:
+    # Full-length git commit hash.
+    commit: str
+    # Name of the branch carrying the commit.
+    branch_name: str
+    # Example value: 2023-02-06 12:01:01+00:00
+    build_time_rfc3339: str
+    build_hostname: str
+    # Canonical
+    version_string: str
```
jgehrcke, line 21: I think it's useful to share my thinking here.

My goal was: quickly define a type so that
- an IDE can show what a `Buildinfo` object looks like
- `mypy` can use it for type checking.
- a JSON structure read from a file can nicely be validated at runtime against that type

About validation at runtime: pydantic, dacite, marshmallow? In this particular case here I think either pydantic or dacite. Which one? https://github.com/konradhalas/dacite/issues/42 :). dacite is small and robust. Maybe we want to use pydantic later / soon.

Oh and I briefly wondered: TypedDict or dataclass? I found https://dev.to/meeshkan/typeddict-vs-dataclasses-in-python-epic-typing-battle-onb useful.

jgehrcke:
Screenshot of a bottom-right fragment of the CI-generated screenshot, showing the commit hash prefix next to the version number:
image

conbench/buildinfo.py
Fatal1ty: If you let me make a suggestion, I'd recommend that you take a look at [mashumaro](https://github.com/Fatal1ty/mashumaro) as a robust and efficient alternative. It doesn't have pydantic-style custom validators and constrains out of the box yet but dacite, which you're considering an alternative to pydantic, doesn't have them either.

jgehrcke: @Fatal1ty -- thanks! Can I ask: how did you get here? : ) Get an event notification for all GitHub comments containing certain keywords? :P Smart move, in any case. Love mashumaro's README. Will give this serious consideration, just maybe not here in this PR anymore.

conbench/buildinfo.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+BUILD_INFO: Optional[Buildinfo] = None
+try:
+    # Try to discover additional build information at well-known path. For now,
+    # do not crash the application upon: file not found, file having unexpected
+    # contents, etc.
+    with open("/buildinfo.json", "rb") as fh:
+        data = json.loads(fh.read().decode("utf-8"))
```
austin3dickey, line 30: Any reason not to use `load`?
```suggestion
    with open("/buildinfo.json", "r") as fh:
        data = json.load(f)
```

austin3dickey (approved):

conbench/buildinfo.py
Fatal1ty: > @Fatal1ty -- thanks! Can I ask: how did you get here? : ) Get an event notification for all GitHub comments containing certain keywords? :P

Haha, you're right on the money :D Actually I do have my own notification bot that searches through GitHub for certain keywords but in this particular case it didn't go well :) But despite this, I still regularly open GitHub in search of something new related to open source projects that I inspired by just to be kept in the loop. This is how I got here :) Anyway, thanks for taking your time and have a great day!

conbench/buildinfo.py
jgehrcke: Oh. Thanks for asking! I could choose to not care, we could do that here, and it's unlikely that we would ever note the difference. That is true.

But this topic is important to me, and you asked for a reason why I wrote the code the way I wrote it. There is a pretty strong reason.

This is a principle I apply. "my version" is less complex and gives a useful error messages when it fails.

Opening files in "text mode" in Python uses an unbelievable amount of magic under the hood. This magic makes use of operating system features/environment specifics (locale). Magic that's great for the unaware developers that quickly want to get something working (I mean that!) -- but for thorough engineering and smooth maintenance it gets in the way, because it increases complexity, and wastes time when things go pear-shaped.

The following changelog entry shows that this topic is far from trivial and undergoes evolution -- `json.load(f)` can (since CPython 3.6) _also_ receive file objects that were opened with 'rb', and then the json module attempts to perform codec guessing:

> Changed in version 3.6: fp can now be a [binary file](https://docs.python.org/3/glossary.html#term-binary-file). The input encoding should be UTF-8, UTF-16 or UTF-32.

(When you do `open(path, 'r') as f: json.load(f)` you get a different codec guessing magic applied than as if you were doing `open(path, 'rb') as f: json.load(f)`.)

It is almost always better to be explicit about assumptions.  As a software engineer I can take burden from my future self and also fellow devs when I leave an explicit note about the character encoding that my code expects. That is why I typically do `rb` (give me the bytes!) and then specify the codec by name (describing the exact table by which I want bytes transformed into code points).

What's the magic that I don't want to have? From the Python docs, first a general statement (emphasis mine):

> Normally, files are opened in text mode, that means, you read and write strings from and to the file, which are encoded in a specific encoding. If encoding is not specified, **the default is platform dependent** (see [open()](https://docs.python.org/3/library/functions.html#open)). Because UTF-8 is the modern de-facto standard, **encoding="utf-8" is recommended unless you know that you need to use a different encoding**.

(https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files)

The docs for open() say:

> encoding is the name of the encoding used to decode or encode the file. This should only be used in text mode. **The default encoding is platform dependent** (whatever [locale.getencoding()](https://docs.python.org/3/library/locale.html#locale.getencoding) returns), but any [text encoding](https://docs.python.org/3/glossary.html#term-text-encoding) supported by Python can be used.

(https://docs.python.org/3/library/functions.html#open)

And now what even I find impressive now after having worked with CPython rather intensely for ~15 years, locale.getencoding() is _new_, it's documented with:

> Get the current [locale encoding](https://docs.python.org/3/glossary.html#term-locale-encoding): On Android and VxWorks, return "utf-8". On Unix, return the encoding of the current [LC_CTYPE] https://docs.python.org/3/library/locale.html#locale.LC_CTYPE) locale. Return "utf-8" if nl_langinfo(CODESET) returns an empty string: for example, if the current LC_CTYPE locale is not supported. On Windows, return the ANSI code page. The [Python preinitialization](https://docs.python.org/3/c-api/init_config.html#c-preinit) configures the LC_CTYPE locale. See also the [filesystem encoding and error handler](https://docs.python.org/3/glossary.html#term-filesystem-encoding-and-error-handler). This function is similar to [getpreferredencoding(False)](https://docs.python.org/3/library/locale.html#locale.getpreferredencoding) except this function ignores the [Python UTF-8 Mode](https://docs.python.org/3/library/os.html#utf8-mode). **New in version 3.11.**

(https://docs.python.org/3/library/locale.html#locale.getencoding)

"text mode" files in Python also have tremendous complexity with respect to newline-representing byte sequences.

Another wall of text :scroll:  It's just that I have spent a lot of time on the topic of character encoding in software before, in particular in Python:
- https://stackoverflow.com/a/17733233
- https://gehrcke.de/2015/12/how-to-raise-unicodedecodeerror-in-python-3/
- https://gehrcke.de/2014/07/mojibake-beatports-id3-text-encoding-is-broken/
- https://gehrcke.de/2014/02/command-line-argument-binary-data/

austin3dickey: Wow, thanks for the in-depth explanation! TIL a safer Python idiom.

Issue #563, opened 2022-12-22
commit author prominence: rethink?


jgehrcke:
Faces are nice! But maybe the author of a commit does not always have tooo much to do with a benchmark run. Maybe this is a bit of a philosophical question but so far I think the benchmark run itself might be a (slightly) more important entity than a commit, for Conbench UX considerations.

Looking at for example:
![image](https://user-images.githubusercontent.com/265630/209174069-56741511-5aa1-4b53-9f8c-34931d4ef0d8.png)

If we think about users, service accounts, identities in this context: I think it might be more interesting to show in this view _which identity submitted the benchmark results_ than who is the author of the commit.

jgehrcke:
In the meantime we should rename the column title "Author" to "Commit author".

A column titled "Author" next to a column titled "Run" implies that this is the Run author.

jonkeane:
I'm 💯 behind renaming to be more accurate with "Commit author" 

One positive side of having the commit authors on this page is I believe (though we could | should measure this in the fullness of time) that it increases developer engagement + makes it easier for a developer to find which benchmarks correspond to their patches. 

I wonder if there's something we could do with the information hierarchy that would make all of the Commit details more cohesive (and make it clear that all of those are information about the same thing | from the same source) that would make it clearer that one way we think about runs is by the data we get from git.

jonkeane:
> If we think about users, service accounts, identities in this context: I think it might be more interesting to show in this view which identity submitted the benchmark results than who is the author of the commit.

Agreed that having this information in a more standardized way (especially when it varies) would be helpful (but IMO  would be good in addition to the commit details including (commit) author). With our current setups, these are almost always sent to Conbench by a single robot user per conbench instance, so a publisher column there would be a lot of the same. This will become much more important (and in fact critical!) if we have setups that have more arbitrary submission of results.

jgehrcke:
> I'm 100 behind renaming to be more accurate with "Commit author"

Thanks for the feedback!

> makes it easier for a developer to find which benchmarks correspond to their patches.

Right! 

> I wonder if there's something we could do with the information hierarchy that would make all of the Commit details more cohesive (and make it clear that all of those are information about the same thing | from the same source) 

Love this. Thought the same. Commit hash, author, commit msg all are "commit information", could and probably should be shown so that they are visually nicely grouped. For example in the ~same table field, potentially revealing more detail when toggling a 'detail view'.

jgehrcke:
Column title has been changed in #697.

[...]

PR #679, opened 2023-02-06
ui: front page: cosmetic improvements for column heading, width, search box content, orderability

jgehrcke:
Originally I only wanted to change "Author" into "Commit author" (see https://github.com/conbench/conbench/issues/563), but then I added a few more tiny-ish changes, see comment and screenshots below.

### DIFF
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -196,7 +196,7 @@ function free_text_column_search(data_table_id, column_indexes) {
         table.columns(column_indexes).every(function() {
             var that = this;
             var title = this.header().textContent;
-            $(`<input class="custom-input-text custom-search-box" type="text" placeholder="🔍 ${title}"
+            $(`<br /><input class="custom-input-text custom-search-box" type="text" placeholder="🔍 filter"
                 onClick="if(event.stopPropagation!==undefined){event.preventDefault(); event.stopPropagation();}event.cancelBubble=true;"/>`)
                 .appendTo(this.header())
                 .on('keyup change clear', function() { //column search/sorting triggered on keypress
--- a/conbench/templates/index.html
+++ b/conbench/templates/index.html
@@ -30,7 +30,7 @@
             <thead>
                 <tr>
                     <th scope="col">Run creation time</th>
-                    <th scope="col">Author</th>
+                    <th scope="col">Commit author</th>
                     <th scope="col">Repository</th>
                     <th scope="col">Hardware</th>
                     <th scope="col">Commit</th>
--- a/conbench/templates/index.html
+++ b/conbench/templates/index.html
@@ -30,12 +30,12 @@
             <thead>
                 <tr>
                     <th scope="col">Run creation time</th>
-                    <th scope="col">Commit author</th>
+                    <th scope="col" class="author-col" style="width: 195px">Commit author</th>
                     <th scope="col">Repository</th>
                     <th scope="col">Hardware</th>
-                    <th scope="col">Commit</th>
-                    <th scope="col">Commit message</th>
-                    <th scope="col">Reason</th>
+                    <th scope="col" class="commit-col" style="width: 65px">Commit</th>
+                    <th scope="col" style="width: 150px">Commit message</th>
+                    <th scope="col" class="reason-col" style="width: 80px">Reason</th>
                 </tr>
             </thead>
             <tbody>
@@ -129,6 +129,8 @@
           "searchPlaceholder": "commit hash works too",
       },
 });
+
+
 enable_search_query_string('#runs');
 
 /* specify the HTML table id (string) for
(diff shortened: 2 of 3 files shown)


jgehrcke:
Not a fan of how this currently looks:
image

I am looking into pragmatically changing this on this PR. The least that needs to be done: make it so that the heading does not line-wrap

But of course that screenshot triggers many ideas.

Update: I have added some commits to this branch to make the columns on the landing page appear a little more tidy. Screenshot created on my machine, directly comparable to the screenshot above:

image

conbench/templates/index.html
```diff
@@ -30,12 +30,12 @@ <h3>Language-independent Continuous Benchmarking (CB) Framework</h3>
             <thead>
                 <tr>
                     <th scope="col">Run creation time</th>
-                    <th scope="col">Author</th>
+                    <th scope="col" class="author-col" style="width: 195px">Commit author</th>
```
sayantikabanik: `author-col` is missing under `app.css` ?

conbench/templates/index.html
```diff
@@ -30,12 +30,12 @@ <h3>Language-independent Continuous Benchmarking (CB) Framework</h3>
...
                     <th scope="col">Repository</th>
                     <th scope="col">Hardware</th>
-                    <th scope="col">Commit</th>
-                    <th scope="col">Commit message</th>
-                    <th scope="col">Reason</th>
+                    <th scope="col" class="commit-col" style="width: 65px">Commit</th>
+                    <th scope="col" style="width: 150px">Commit message</th>
```
sayantikabanik, line 37: Thinking loud here, as commit message length is variable. Maybe having it fixed to the max content-length might help avoid any overflows.

conbench/templates/index.html
jgehrcke: Thanks for careful review! I have removed `class="author-col"` here again, I think I wanted to do that before and forgot.

conbench/templates/index.html
jgehrcke: > the max content-length

Interesting thought! We know that this is 2xx characters (can look this up in code -- what we show here is typically a commit message prefix; truncated at a max character count length).

In terms of geometrical width, the max width is somewhat unpredictable. Even if it was predictable, it would be "too wide" to default to this. That is, for the longer messages in particular we _want_ the text to automatically wrap.

I thought:
- this column has the somewhat least predictable content width, which is why giving this column a fixed width keeps the overall layout somewhat predictable. 
- it's likely for commit messages to contain multiple short-ish words separated by whitespace. Conservative text wrapping automatically happens, and prevents overflowing in sane cases. 

For more exotic cases we can do ` overflow-wrap: break-word;` which brutally cuts even long words into pieces: https://css-tricks.com/almanac/properties/o/overflow-wrap/

sayantikabanik: > For more exotic cases we can do ` overflow-wrap: break-word;` which brutally cuts even long words into pieces: https://css-tricks.com/almanac/properties/o/overflow-wrap/

Love this, we can definitely revisit(need basis) and iterate :)

sayantikabanik (approved):

jonkeane:
Thank you for cleaning this up! This is super minor, but would it be possible to have the column heads all top aligned as well? I know bottom-aligned or centered is traditional, but given that some have sorting icons and some have search | toggle boxes, top alignment should make all the labels be in line with each other for quicker scanning?

jgehrcke:
> Thank you for cleaning this up! This is super minor, but would it be possible to have the column heads all top aligned as well?

Thanks for asking! (really). I had the inclination, but didn't touch this because it felt like over-loading this PR with too many unrelated things. But yeah that's totally one of those visual/design offenders now! :eye: 

I have quickly cone some cosmetic cleanup here and pushed another commit. Screenshot from local state:
image

conbench/templates/app.html
```diff
@@ -196,7 +196,7 @@ <h4 class="modal-title"><span class="glyphicon glyphicon-trash submenu"></span>&
         table.columns(column_indexes).every(function() {
             var that = this;
             var title = this.header().textContent;
-            $(`<input class="custom-input-text custom-search-box" type="text" placeholder="🔍 ${title}"
+            $(`<br /><input class="custom-input-text custom-search-box" type="text" placeholder="🔍 filter"
```
jgehrcke, line 199: Note: the additional break here is so that the input field is guaranteed to go below the column heading.

jonkeane: 😂 This feels like one of those "Design is my passion ❤️ CSS" memes right here. 

In seriousness: this is totally fine and I like that the sorting is next to the inputboxes when they are there

jgehrcke:
Thanks for the quick feedback here @jonkeane and @sayantikabanik this was a pleasure. I am merging this now, and we can and will iterate on this more and more I am sure.

PR #687, opened 2023-02-06
[benchadapt] GoogleBenchmarkAdapter: Put gbench_context in optional_benchmark_info

### LINKED ISSUE #686: [benchadapt][bug]: In `GoogleBenchmarkAdapter`, `gbench_context` should not be in `tags`
In `benchadapt.adapters.GoogleBenchmarkAdapter`, `gbench_context` is appended to `tags`, but it's not constant (it contains at least a `load_avg` array that will vary), so it will always break history. We should put it somewhere else instead, probably `optional_benchmark_info`.

alistaire47:
Closes #686 by moving `gbench_context` in `benchadapt.adapter.GoogleBenchmarkAdapter` results from `tags` (where it was breaking history) to `optional_benchmark_info`, so it's still there if you want it, but not breaking anything.

### DIFF
--- a/benchadapt/python/benchadapt/adapters/gbench.py
+++ b/benchadapt/python/benchadapt/adapters/gbench.py
@@ -181,7 +181,17 @@ class GoogleBenchmarkAdapter(BenchmarkAdapter):
 
     @staticmethod
     def _parse_gbench_json(raw_json: dict) -> Tuple[dict, list]:
-        """Parse gbench result json into a context dict and a list of grouped benchmarks"""
+        """
+        Parse gbench result json into a context dict and a list of grouped benchmarks
+
+        See https://github.com/google/benchmark/blob/main/docs/user_guide.md#output-formats
+        for a (very minimal!) example of gbench output json. This method splits out
+        the `context` attribute, which "contains information about the run in general,
+        including information about the CPU and the date" from the `benchmarks` one, which
+        contains a dict for all benchmarks in the run.
+
+        Aggregate benchmarks are excluded, as they are duplicative of the raw benchmarks.
+        """
         gbench_context = raw_json.get("context")
 
         # Follow archery approach in ignoring aggregate results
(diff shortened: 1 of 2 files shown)


benchadapt/python/benchadapt/adapters/gbench.py
```diff
@@ -175,6 +174,7 @@ def _parse_results(self, results: dict, extra_tags: dict) -> List[BenchmarkResul
                 batch_id=batch_id,
                 extra_tags=extra_tags,
             )
+            result_parsed.optional_benchmark_info = {"gbench_context": gbench_context}
```
jgehrcke, line 177: Now or later: I think it makes sense to add a code comment here explaining what `gbench_context` is, or where one can read about it.

jgehrcke (approved):
Cool. I suppose we are aligned on the idea that the red CI check `Auto assign pull requests ....` can for now stay red, and we try to get it greener and greener over time :).

I have left a remark about a code code comment.

Curious: how do we roll out changes like this to users? Do they check out some `HEAD` of some `main`, or do we need to make a release, or ...? :)

benchadapt/python/benchadapt/adapters/gbench.py
alistaire47: Good idea; added it to the `_parse_gbench_json()` docstring, as that's where things are really happening.

alistaire47:
> Curious: how do we roll out changes like this to users? Do they check out some `HEAD` of some `main`, or do we need to make a release, or ...? :)

Right now they're all installing from GitHub with pip ([labs/benchmarks example](https://github.com/voltrondata-labs/benchmarks/blob/main/requirements.txt#L4)), so they'll get updates on their next run after changes are merged to main. After Austin's work on getting everything (including benchadapt and its dep benchclients) released to PyPI better is done, installing from there will definitely be the recommended approach.

jgehrcke:
> Right now they're all installing from GitHub with pip ([labs/benchmarks example](https://github.com/voltrondata-labs/benchmarks/blob/main/requirements.txt#L4)), so they'll get updates on their next run after changes are merged to main.

Thanks! I like that.

Just for the record, we will also have intermediate options (between installing from HEAD of main and referring to a full/proper PyPI release), such as
- Refer to a release branch, e.g. `benchadapt@git+https://github.com/conbench/conbench.git@cli-release` -- which we could move forward more consciously than `main`.
- Refer to a specific commit on `main` (pinning).

austin3dickey (approved):
Thanks!

jonkeane:
Sorry for the noise there — wanted to trigger the auto assigner to confirm it's working (and I should stop cause clearly it took too long for those typos to be correct!)

PR #689, opened 2023-02-06
[docs][benchadapt] Overhaul README

alistaire47:
A big overhaul of the README for benchadapt laying out the core components of the package, usage, and context, following up on user questions. A couple small tweaks to docstrings to clarify.

Feedback very welcome; this is a starting point. If there are things that are still unclear, that you have questions about, or you think deserve some attention here, let's iterate.

### DIFF
--- a/benchadapt/python/README.md
+++ b/benchadapt/python/README.md
@@ -20,10 +20,10 @@ like populating `machine_info` with metadata on the current machine. If you are 
 on a cluster instead, you will need to populate `cluster_info` yourself, and
 `machine_info` will remain empty.
 
-There is light validation, but [for now] the server is the ultimate validator; it is
+There is light validation, but [for now] the API is the ultimate validator; it is
 possible to make payloads that will be rejected.
 
-If you need to interact directly with a Conbench server's API instead of letting adapters
+If you need to interact directly with a Conbench webapp's API instead of letting adapters
 (see below) or another tool manage sending results for you, you can use
 [benchclients.ConbenchClient](https://github.com/conbench/conbench/blob/main/benchclients/python/benchclients/conbench.py)
 to make requests. As benchclients is a dependency of benchadapt, you should not need to
@@ -33,7 +33,7 @@ install anything new, and it is nicely set up to handle auth and such for you.
 
 The concept of Conbench adapters is for when you already have a benchmarking method that
 produces a pile of results (say in JSON files, though anything works), and you need to
-transform them into an appropriate form that can be sent to a Conbench server.
+transform them into an appropriate form that can be posted to a Conbench API.
 
 The `benchadapt.adapters.BenchmarkAdapter` abstract class defines a basic workflow:
 
@@ -42,7 +42,7 @@ are already guaranteed to exist, this can be set to do nothing.
 2. Transform results produced by the benchmarks into a list of `BenchmarkResult` instances.
 3. Postprocess results to ensure a consistent `run_id` and override any metadata fields
 not already set correctly.
-4. Post each result to a Conbench server.
+4. Post each result to a Conbench API.
 
 Classes that inherit from the abstract class need to define
 
@@ -66,7 +66,7 @@ Adapters have separate `.run()` and `.publish_results()` methods; the former run
 benchmarks, transforms the results, and stores them in a `.results` attribute of the
 instance. It does not post them, so is useful for looking at results interactively before
 sending them. `.publish_results()` takes the results from the `.results` attribute and
-posts them to a Conbench server.
+posts them to a Conbench API.
 
 The whole instance also has a `__call__()` method defined so it can be called like a
 function that both runs and publishes, so a somewhat minimal script for running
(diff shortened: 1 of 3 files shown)


alistaire47:
@joosthooz It won't let me add you as a reviewer, but if anything here is unclear or you see opportunities to improve, please jump in!

benchadapt/python/README.md
```diff
@@ -1,3 +1,129 @@
...
+```
+
+Of note:
+
+- `result_fields_override` will replace the whole attribute with a new value. This works
+with all types (strings, dicts, etc.), so here `run_reason` will be set for all results.
+- `result_fields_append` will append the new values to dicts which may already have data.
+Here, `build_version` will be appended to the `info` dict. (In this case it is empty
+anyway, so this is equivalent to
+`result_fields_override={"info": {"build_version": os.getenv("MY_BUILD_VERSION")}})`.)
```
jonkeane: It might be nice to have (another?) example where `result_fields_append` does actually append

benchadapt/python/README.md
```diff
@@ -1,3 +1,129 @@
...
+`machine_info` will remain empty.
+
+There is light validation, but [for now] the server is the ultimate validator; it is
+possible to make payloads that will be rejected.
+
+If you are managing everything yourself, to actually send the results to a Conbench
+server, use
+[benchclients.ConbenchClient](https://github.com/conbench/conbench/blob/main/benchclients/python/benchclients/conbench.py).
+As benchclients is a dependency of benchadapt, you should not need to install anything
+new, and it is nicely set up to handle auth and such for you.
```
jonkeane: Could you explain more here about what "managing everything yourself" means in this case?

benchadapt/python/README.md
alistaire47: Yeah, I was trying to figure out a good example here. `tags` is populated, but for the most part people shouldn't be messing with that at this level. `context` might be a better example; `"benchmark_language"` will already be set, but we could append `"compiler_flags"` or something

benchadapt/python/README.md
alistaire47: Yeah, this is inexact; I'll reword. Really I meant any situation in which you're managing your own results directly instead of using adapters or benchrun or the conbench runner or whatever; that could just be running experiments locally or something.

benchadapt/python/README.md
alistaire47: Left this one (I think seeing how they connect is helpful), but added a new `context` one to make a use-case more obvious

benchadapt/python/README.md
alistaire47: Reworded. It feels a little weird putting low-level API details before the higher-level adapters API, but in practice adapters make way more sense with some context around what `BenchmarkResult` looks like even though it's part of the lower-level API. 

Open to ideas if there's a better way to structure this; I don't want most users to use `ConbenchClient`, but nor do I want someone to go roll their own if they do want to interact with the API manually, so mentioning it seems important

benchadapt/python/README.md
jonkeane: Awesome, yeah that sounds good

benchadapt/python/README.md
```diff
@@ -1,3 +1,132 @@
 # Python {benchadapt}
 
-A small python package with utilities for running Conbench benchmarks.
+A small python package with utilities for getting benchmark results into a Conbench server.
```
jgehrcke, line 3: I wanted to mention this before; many times. I think the term "server" is not ideal

I can see that this is a bit of a cultural thing in computing terminology, but "server" here sounds a bit naive to me.

Conbench is a web application in the most traditional sense, and one component thereof is the WSGI server (gunicorn). There also is an HTTP server (technically a reverse proxy) in front of the Conbench web application. 

We can say things like "into Conbench". Or to "a Conbench instance". Or to "a Conbench deployment". Or "to the Conbench API".

benchadapt/python/README.md
```diff
@@ -1,3 +1,132 @@
...
+
+## Useful components in this package
```
jgehrcke, line 5: Makes me curious: is there a "not so useful components ..." section below? :D

benchadapt/python/README.md
```diff
@@ -1,3 +1,132 @@
...
+
+### `BenchmarkResult` and `BenchmarkRun` dataclasses
```
jgehrcke, line 7: I love that, and I was hoping we have something like that!

benchadapt/python/README.md
jgehrcke: > small python package with utilities

Can you add here whether this is a CLI? Or a set of CLIs? Or a library and CLI?

benchadapt/python/README.md
```diff
@@ -1,3 +1,132 @@
...
+But the `context` dict will already contain a `"benchmark_language"` key; this will be
+retained, and `compiler_flags` will be appended.
+- For this to work, a lot of environment variables have to be set! This includes ones
+with information about the Conbench server and the current git metadata. See the
+"Environment Variables" section below for a full list.
+
+
+## Environment variables
+
+benchadapt relies on a number of environment variables. The Conbench server ones
```
jgehrcke: > benchadapt relies on a number of environment variables

Can we say "the `benchadapt` CLI" or "the `benchadapt` library" here, depending on what's correct?

jgehrcke: >  The Conbench server ones

I think this is confusing. Does a consumer of these docs need to know that the Conbench webapp can be configured with environment variables that have the same names?

benchadapt/python/README.md
```diff
@@ -1,3 +1,132 @@
...
+(`CONBENCH_*`) are used by `benchclients.ConbenchClient`; the git ones
+(`CONBENCH_PROJECT_*`) are used to populate run and result metadata if not specified
```
jgehrcke, line 113: > the git ones (`CONBENCH_PROJECT_*`)

Hm. For future work on this. If these are always related to git then should we call those `CONBENCH_GIT_*`? Or maybe these are not "the git ones"?

jgehrcke (approved):
Thanks for putting :heart: into docs! This is an improvement, no matter what. I just left a few comments, but we can merge this anytime and iterate further later on.

benchadapt/python/README.md
```diff
@@ -1,3 +1,132 @@
...
+directly; and `CONBENCH_MACHINE_INFO_NAME` is for overriding the machine name in
+automatically gathered machine info when necessary:
+
+- `CONBENCH_URL`: Required. The URL of the Conbench server without a trailing
+slash, e.g. `https://conbench.example.com`
+- `CONBENCH_EMAIL`: The email to use for Conbench login. Only required if the
+server is private.
```
austin3dickey: For the purposes of this library I think `CONBENCH_EMAIL` and `CONBENCH_PASSWORD` are required. They need to be set to post things.

austin3dickey (approved):
Looking great, thanks!

benchadapt/python/README.md
alistaire47: I've been trying to avoid "into Conbench" because we have a lot of things named Conbench (the runner CLI, the runner, various classes, the webapp, etc), and I'm always referring to a particular one. "Conbench instance" sounds like a class instance to me (of which there are a handful named `Conbench`). "Conbench API" is probably most apt here.

benchadapt/python/README.md
alistaire47: There is no benchadapt CLI; for a certain use-case, that's benchconnect. benchadapt is just a library.

benchadapt/python/README.md
alistaire47: There was previous conversation on the PR where these were introduced; "PROJECT" was added to make it clear these are for the repo being benchmarked, not Conbench's repo itself.

PR #690, opened 2023-02-06
webapp: increase worker timeout, fix KeyError

austin3dickey:
After deploying the last 34 commits to our production environments, I noticed a couple of problems.

1. #636 introduced an expected performance slowdown. This causes 502s on pages like https://conbench.ursa.dev/compare/runs/a02a6dc9375543fea2f9a003f3bb341f...5c32ae93b4dd47ec854cb84ff040a770/ because of the 30-second default gunicorn worker timeout. I propose lifting that timeout to 120 seconds. Based on some local testing, I think this will solve the issue for now while we work on performance of those pages.
2. After the deployment (could be the same PR) we started receiving a 500 on this page: https://velox-conbench.voltrondata.run/compare/runs/GHA-4071695296-1...GHA-4106991798-1/
The logs show a KeyError when we're constructing the URLs to link from the outlier plots. I tried it on a commit before that PR, and the plots simply don't show up. So I'm not sure this code ever worked properly. As a band-aid, I set the default `outlier_url` to "", which just refreshes the page if you click on the plot title. At least we won't be returning a 500 anymore.

### DIFF
--- a/conbench/app/compare.py
+++ b/conbench/app/compare.py
@@ -72,7 +72,8 @@ class Compare(AppEndpoint, BenchmarkMixin, RunMixin, TimeSeriesPlotMixin):
                 augment(benchmark)
             outliers, outlier_ids, outlier_names = self.get_outliers(benchmarks)
             outlier_urls = [
-                comparisons_by_id[x]["compare_benchmarks_url"] for x in outlier_ids
+                comparisons_by_id.get(x, {}).get("compare_benchmarks_url", "")
+                for x in outlier_ids
             ]
             plot_history = [
                 self.get_history_plot(b, contender_run, i)
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -584,7 +584,7 @@ class GitHub:
 
         Do not try for too long because there is an HTTP client waiting for
         _us_ to generate an HTTP response in a more or less timely fashion.
-        Gunicorn has a worker timeout behavior (as of the time of writing: 30
+        Gunicorn has a worker timeout behavior (as of the time of writing: 120
         seconds) and the retrying method below must come to a conclusion before
         that.
 
--- a/deploy.yml
+++ b/deploy.yml
@@ -20,7 +20,7 @@ spec:
       containers:
       - name: conbench
         image: "{{DOCKER_REGISTRY}}/{{FLASK_APP}}:{{BUILDKITE_COMMIT}}"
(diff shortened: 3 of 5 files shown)


jgehrcke:
:heart: Thanks for rolling things out and for watching out for problems! 

> I propose lifting that timeout to 120 seconds. ... , I think this will solve the issue for now while we work on performance of those pages.

:+1: I am OK with that for now. 30 s and 120 s are both painful, so why not pick the latter :).

I trust that we all agree that users typically do not wait 2 minutes for a loading tab to complete loading. They hit refresh or close the browser tab before that. For ~all (converging to 100 %) browser-oriented views we need to get HTTP response generation times to below N seconds (converging to 1 second or less). We will get there, step by step, and https://github.com/conbench/conbench/issues/493 is the foundation for that.

It's important to distinguish robots from humans here. Robots seem to meditate more often and are sometimes OK with with waiting for a minute until they get a response. :lotus_position: 

> The logs show a KeyError when we're constructing the URLs to link from the outlier plots.

I would really really love to see such a log excerpt here in this issue! Can you please add that?

Update: saw this on Slack, copying it here:
image


I try to not get stressed out about this and I know we will figure this out over time -- but the code base that we deal with here is full of such ambiguities and uncertainties, and it has become hard to do _quick_ code changes with confidence. Need more testing, more invariants (more typing, etc).

> 502s on pages like

Just want to connect the dots: https://github.com/conbench/conbench/issues/599

> The 502 Bad Gateway is then emitted from a reverse proxy (probably nginx, as part of k8s ingress controller) because the upstream (nginx term) probably terminates the TCP connection before it has sent a response.

conbench/app/compare.py
```diff
@@ -72,7 +72,8 @@ def page(self, comparisons, regressions, improvements, baseline_id, contender_id
                 augment(benchmark)
             outliers, outlier_ids, outlier_names = self.get_outliers(benchmarks)
             outlier_urls = [
-                comparisons_by_id[x]["compare_benchmarks_url"] for x in outlier_ids
+                comparisons_by_id.get(x, {}).get("compare_benchmarks_url", "")
```
jgehrcke, line 75: Yeah... :D those `KeyError`s are obliterated now :fire:. Thanks.

More strictly defined objects will really help make this more sane. 

Just for the record (I believe we all agree): code like this will not be maintainable, and is a rather significant burden for new developers entering a code base. We need to launch an initiative to make it easier again to understand the state space by just reading code. I am writing this now because I hit a situation like this almost daily, and am frequently slowed down spending time on trying to imagine what an object _could be_ during runtime.

PR #696, opened 2023-02-07
benchadapt: GoogleBenchmarkAdapter: Set batch_id at benchmark level

### LINKED ISSUE #695: benchadapt: GoogleBenchmarkAdapter: Set batch_id at benchmark level
Right now in benchadapt's `GoogleBenchmarkAdapter`, everything is being dumped into one batch, instead of one `batch_id` per benchmark name (`result["tags"]["name"}`) like everything else. We just need to move the `batch_id` generation into the loop right below here: https://github.com/conbench/conbench/blob/main/benchadapt/python/benchadapt/adapters/gbench.py#L167

alistaire47:
Closes #695 by moving `batch_id` into the `for` loop so each benchmark group gets its own new `batch_id`. Adds tests to ensure the number of unique values for `result.tags.name` and `result.batch_id` correspond.


jonkeane (approved):
Thanks for catching this!

PR #698, opened 2023-02-07
deployment: fix deploy script

### LINKED ISSUE #697: deploy failure: buildinfo.json: file does not exist
As of #678, the `conbench-staging-deploy` Buildkite pipeline is failing when it gets to this line:
https://github.com/conbench/conbench/blob/22f00b779b878009099a0cc01e571087d12b2db8/.buildkite/utils.sh#L5
with the error message
```
Step 13/13 : COPY ./buildinfo.json /buildinfo.json
COPY failed: file not found in build context or excluded by .dockerignore: stat buildinfo.json: file does not exist
```

austin3dickey:
After #678, we always need to `make set-build-info` before we `docker build`. In PR CI, we always do that through the Makefile. Since the Buildkite deployment scripts do not use the Makefile, we need to explicitly call that then. Fixes #697

### DIFF
--- a/.buildkite/utils.sh
+++ b/.buildkite/utils.sh
@@ -2,6 +2,7 @@
 
 build_and_push() {
   set -x
+  make set-build-info
   docker build -t ${FLASK_APP} .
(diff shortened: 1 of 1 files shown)


austin3dickey:
This PR fixed this issue!

PR #703, opened 2023-02-08
db.py: lift libpq statement timeout from 30 to 60 s

jgehrcke:
Follow-up from a Slack discussion. Our users won't find this particularly beautiful, and maybe they won't even notice. But this is a very pragmatic attempt to make a bad situation a tiny bit better, in the spirit of what we did before in https://github.com/conbench/conbench/pull/690.


austin3dickey (approved):
Thank you!!

PR #704, opened 2023-02-08
Upload all libraries to PyPI on merge-to-main

### LINKED ISSUE #684: monorepo: upload all packages to PyPI with better deployment strategy
Currently we are only uploading the `conbench` package to PyPI, and we do that on Buildkite whenever we do a deployment to a server. I propose the following deployment strategy:

Upon each merge to `main`, after the rest of CI succeeds, GitHub Actions runs a `deploy-packages` step. For each package (today this includes `conbench`, `benchadapt`, `benchclients`, `benchconnect`, and `benchrun`, and soon may include `benchalerts` and `datalogistik`), this step:
- checks to see if the package version has already been deployed to PyPI
- if not,
  - builds the wheel
  - deploys it to PyPI
  - tags the commit with the version tag, formatted `<pkg>-<version>` (e.g. `conbench-1.2.3`)

This means that in order to release a new version of a package, we must remember to bump its version in a PR. In the past I've worked on projects that use [GitHub Releases](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository) to enable developers to release packages from the GitHub website (with CI handling the version bumps) but I think that's overkill for this repository right now.

austin3dickey:
Resolves #684.

That issue became a little controversial 😉 , but here's my proposal in code.

This PR enables CI to upload built wheels of each applicable library to PyPI upon a merge-to-main build, if and only if a developer changed the library's version number in the merge commit (and the rest of the build is green). The proposal is to start with CalVer.


austin3dickey:
TODO: update main README, update the various `requirements.txt` files that pin dependencies to the main branch of the github repo

jgehrcke:
> if and only if a developer changed the library's version number in the merge commit (and the rest of the build is green)

I see! So, in my mental model: this is "lightweight releases", to get a release machinery rolling. 

When would we typically want to make a release?

For consumers: if they want the latest dev version (HEAD of main) "now" -- how would they do that? Would we _need to_ make a 'quick release' before this is possible, or are there / will there be tested docs for them to be able to use the latest dev version? (the important part in this question is "they" vs "we")

> That issue became a little controversial wink 

Thanks for the controversy! :) I think in the process we have learned more about the other person's point of view, and we can be proud of that.

I do like that we do not assign too much meaning in version numbers for now.

.buildkite/utils.sh
```diff
@@ -79,32 +79,6 @@ s/{{BUILDKITE_COMMIT}}/${BUILDKITE_COMMIT}/g;\
...
-  conda create -y -n conbench --force -c conda-forge python=3.8
-  conda activate conbench
-  pip install -r requirements-dev.txt -r requirements-webapp.txt -r requirements-cli.txt
-  pip install -e .
-  python -m pip install --upgrade build
-  python -m pip install --upgrade twine
-  rm -rf dist
-  python -m build
-  python -m twine upload dist/*
-}
```
jgehrcke, line 106: :muscle:

.github/workflows/actions.yml
```diff
@@ -178,3 +178,86 @@ jobs:
           memory: max
       - name: ci/minikube/test-conbench-on-mk.sh
         run: export CONBENCH_REPO_ROOT_DIR=$(pwd) && bash ci/minikube/test-conbench-on-mk.sh
+
+  deploy-packages:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
```
jgehrcke, line 184: I always forget so many things about GitHub Actions. Does this here take effect only for merges to main? If yes -- how does the 'branch filtering' work?

.github/workflows/actions.yml
```diff
@@ -178,3 +178,86 @@ jobs:
...
+
+          cd ${{ matrix.pkg_root }}/${{ matrix.pkg_name }}
+          version=$(
+            python -c 'from _version import __version__; print(__version__)'
+          )
+
+          echo "Last ${{ matrix.pkg_name }} version released to PyPI: $last_released_version"
+          echo "Current ${{ matrix.pkg_name }} version in repo: $version"
+
+          if [[ $last_released_version < $version ]]; then
```
jgehrcke: Does this comparison work? If yes -- why? :D 

The messy topic of comparing version numbers. Has haunted so many software projects.

Others have built so much code to compare version numbers.
https://stackoverflow.com/a/11887885
https://packaging.pypa.io/en/latest/version.html#packaging.version.parse

.github/workflows/actions.yml
austin3dickey: Yes, and that's because of this configuration:
https://github.com/conbench/conbench/blob/411c8132c6a9936e45473602f35c41a528172110/.github/workflows/actions.yml#L4-L6
I'll add a code comment!

benchadapt/python/benchadapt/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.0.1"
+__version__ = "23.2.8"
```
jonkeane: If we're going to use calversioning (which I'm totally fine with) could we please use `YYYY.MM.DD` (or `YYYY.M0.D0` if we really hate the month and date 0 padding) so that it's clear(er) that this isn't a semvar?

jonkeane: _technically_ it's possible we're on the 2023rd major release, of course, but I think seeing `2023.` will trigger year more than `23.` in someone's eyes

.github/workflows/actions.yml
austin3dickey: Ah, I copied this from the former `.buildkite/utils.sh`. That was when they were "sort of" using semver for the legacy Conbench CLI. But the bash `<` just sorts alphabetically. That isn't necessarily what `pip` might do, as you point out.

I think the more intuitive behavior here would just be "check if the version exists on PyPI, and if not, upload it."

jonkeane (commented):
I'm fine with merging this approach. 

If we're using calendar versions, this whole thing could be flipped around pretty easily, no? Instead of inspecting the source code for a change in the version number, we setup a workflow that cuts a release (and bumps that version number, commits that back to the repo).

Pros:
* the version sniffing code can 🚮 
* the magic of the version sniffing is avoided
* anyone can cut a release 🎉 empowerment (with appropriate permissions in the repo, of course)
* we don't need a PR or branch _just_ to release

Cons:
* we don't have a PR or branch to _discuss_ the release (we could enforce a process like having an issue before we press the button, of course)
* anyone can cut a release 😬 someone might accidentally do this without realizing the consequences
* we need version bumping code (though I suspect this is both less code and has fewer (external) dependencies than the version sniffing code!)

benchadapt/python/benchadapt/_version.py
austin3dickey: We definitely can use the full year. I was a little hesitant about 0-padding because of https://peps.python.org/pep-0440/#integer-normalization . In some cases, the leading 0 will be stripped, which could be confusing. I also slightly think `2023.2.8` looks more like a date than `2023.20.80` but I will definitely not die on that hill!

jonkeane: This might have been a typo, but in case not: the padding would go on the left, so `2023.02.08`, but yeah those zeros are definitely less important than `YYYY.`

austin3dickey:
>we setup a workflow that cuts a release

This is exactly what I've done in previous projects. It's really nice! Also: once we start hosting documentation, the "doc regeneration workflow" would pair well with this workflow.

austin3dickey:
>For consumers: if they want the latest dev version (HEAD of main) "now" -- how would they do that? Would we need to make a 'quick release' before this is possible, or are there / will there be tested docs for them to be able to use the latest dev version? (the important part in this question is "they" vs "we")

I think recommending `pip install benchadapt@git+https://github.com/conbench/conbench.git@main#subdirectory=benchadapt/python` or from a locally-cloned repo is still fine in this case.

.github/workflows/actions.yml
alistaire47: Doesn't alphabetical sorting imply we need zero-padding then so `2023.10.10` > `2023.02.02`?

.github/workflows/actions.yml
```diff
@@ -178,3 +178,86 @@ jobs:
...
+            echo "Releasing new ${{ matrix.pkg_name }} version $version"
+            echo 'changed=true' >> $GITHUB_OUTPUT
+            echo "version=$version" >> $GITHUB_OUTPUT
```
alistaire47: Oh I quite like this use of env vars to separate steps

benchadapt/python/benchadapt/_version.py
jgehrcke: I like the full year. For the record, the section https://calver.org/#other-notable-projects is related. Inspiration. After all, all these variants (zero-padded or not) exists out there.

jgehrcke:
> I think recommending pip install benchadapt@git+https://github.com/conbench/conbench.git@main#subdirectory=benchadapt/python or from a locally-cloned repo is still fine in this case.

:+1: -- thanks for the reply, @austin3dickey!

.github/workflows/actions.yml
jgehrcke: > I think the more intuitive behavior here would just be "check if the version exists on PyPI, and if not, upload it."

:+1:

benchadapt/python/benchadapt/_version.py
austin3dickey: Yep, I looked at that site, and that's what convinced me that 0-padding is scary! Because it claims that `youtube_dl` uses 0M.0D (ex https://github.com/ytdl-org/youtube-dl/blob/2021.06.06/youtube_dl/version.py) but PyPI remembers those version numbers without 0-padding (ex. https://pypi.org/project/youtube_dl/#history)

austin3dickey:
I addressed all feedback except the idea to make this a UI-triggerable workflow instead. I still like that idea so I'm going to work on it now and leave this PR open in case there's more feedback.

austin3dickey:
Actually, I'll merge this now so it's not blocking anything. I'll timebox my work on the other idea to tomorrow morning.

jgehrcke:
I'd still love to know your current opinion about

> When would we typically want to make a release?

This is just to see how this evolves!

Issue #702, opened 2023-02-08
benchadapt: GoogleBenchmarkAdapter not separating name and params for grouping


alistaire47:
`benchadapt.adapters.GoogleBenchmarkAdapter` got changed yesterday to fix `batch_id` grouping, but it is using the gbench name, which includes both what becomes `tags["name"]` and `tags["params"]` instead of just the former like it should be, leading each case to be grouped separately instead of all cases for a benchmark name being grouped together.

The fix here is to split the name before grouping, while making sure params still end up in tags correctly.

alistaire47:
#705 attempted to fix this, but grouping is still wrong; as of writing, it's lumping across cases as if they were iterations. I'm going to make a new PR that adds another layer of grouping, so there's one at the `tags["params"]` level (to collapse iterations) and one at the `tags["name"]` level (to collapse cases into a batch).

PR #705, opened 2023-02-08
benchadapt: gbench grouping and run_name default

### LINKED ISSUE #701: benchadapt: BenchmarkResult.run_name not defaulted if github set after instantiation
In `benchadapt.BenchmarkResult`, if `github` is set after instantiation (e.g. via `result_fields_override` in an adapter), `run_name` will not get generated like it would if `github` were provided on instantiation, because the setting code is in `__post_init__()` because it depends on both `run_reason` and `github`.

A solution here would be to also run the defaulting code in the `@_github_property.setter`. 

There are some oddities here around `run_reason`—the default for `run_name` will get whatever it is when the defaulting code is run, even if it is set/changed afterwards—but I don't think this is worth fixing, as what `run_name` is mostly doesn't matter. The `github` situation _is_ worth fixing, though, because the API still requires `run_name` to be set, so this will prevent results from being rejected, even if the `run_name` itself is sometimes `"None: <hash>"`.

alistaire47:
Fixing a couple little benchadapt bugs; closes #701 and #702. Apologies for not splitting this in two, I initially thought they were more related than they turned out to be.

### DIFF
--- a/benchadapt/python/benchadapt/result.py
+++ b/benchadapt/python/benchadapt/result.py
@@ -139,6 +139,12 @@ class BenchmarkResult:
         self._maybe_set_run_name()
 
     def _maybe_set_run_name(self) -> None:
+        """
+        Set a default value for `run_name` if not populated and `github["commit"]` is.
+        Uses `run_reason`, but does not check if it's set, so may produce
+        `None: <commit hash>`. Since all three are required by the API, this should in
+        most situations produce a reasonably useful `run_name`.
+        """
         if not self.run_name and self.github.get("commit"):
             self.run_name = f"{self.run_reason}: {self.github['commit']}"
 
--- a/benchadapt/python/benchadapt/result.py
+++ b/benchadapt/python/benchadapt/result.py
@@ -136,6 +136,9 @@ class BenchmarkResult:
     github: Dict[str, Any] = field(default_factory=_machine_info.github_info)
 
     def __post_init__(self) -> None:
+        self._maybe_set_run_name()
+
+    def _maybe_set_run_name(self) -> None:
         if not self.run_name and self.github.get("commit"):
             self.run_name = f"{self.run_reason}: {self.github['commit']}"
 
@@ -148,6 +151,7 @@ class BenchmarkResult:
         if value is None:
             value = _machine_info.detect_github_info()
         self._github_cache = value
+        self._maybe_set_run_name()
 
     @property
     def _cluster_info_property(self) -> Dict[str, Any]:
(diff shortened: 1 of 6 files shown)


benchadapt/python/benchadapt/result.py
```diff
@@ -148,6 +151,7 @@ def _github_property(self, value: Optional[dict]):
         if value is None:
             value = _machine_info.detect_github_info()
         self._github_cache = value
+        self._maybe_set_run_name()
```
jgehrcke, line 160: A docstring on `_maybe_set_run_name()` might help, buuut I wonder: why does this not try to set a run name when `self.github.get("commit")` evaluates to `False`?

alistaire47: > A docstring on `_maybe_set_run_name()` might help, 

👍 

> buuut I wonder: why does this not try to set a run name when `self.github.get("commit")` evaluates to `False`?

The value would be fairly useless; probably `"commit: None"` or `"None: None"` in most situations. A result is not publishable without a commit hash anyway, so this change effectively enforces both will be set by the time anything gets posted. 

There's nothing special about the default format anymore aside from the fact it needs to be _something_ since `run_name` is still required by the API. Using a UUID instead of this is possible, but since `run_name` needs to match 1:1 with `run_id`, which almost always gets set after instantiation, we'd need to turn `run_id` into a property and either copy `run_id` or hash the hash. And from a user utility perspective it's not really better.

If we want to change the API so it's no longer required, I'd be very happy to just let it be `None` and rip out all this mess.

jgehrcke: Lovely. Thanks for the thoughtful reply. I am now slightly unconfused. :)

austin3dickey: To bring our various conversations together: technically, `run_name` is not required by the API (you can omit it from the payload) but right now the API doesn't like `"run_name": None`. I consider that a bug in the API.

austin3dickey (approved):
Looks good to merge as-is. If you want to pop off `run_name` from the payload instead, go for it. Else we could wait until we fix the bug in the API to clean this up a little.

alistaire47:
Realized I need to fix `BenchmarkRun.name` in the same way, so added that. Also changed both to pop off [run_]name if it's unpopulated, since that's how benchadapt is treating other non-required fields already, even if we want to change the API too.

jgehrcke:
```
        for attr in [
            "name",
            "machine_info",
            "cluster_info",
            "finished_timestamp",
            "error_type",
            "error_info",
        ]:
            if not res_dict[attr]:
                res_dict.pop(attr)
```

For the future: the `if not res_dict[attr]` should have a code comment or have a more precise criterion. So many things evaluate to `False` an Python. I look at this code snippet and think "this might contain bugs, or not -- I cannot really know without going much deeper into the code and tests". That's where a code comment or clearer code can help.

E.g., if we want to remove _really_ all of `0, "", None,` etc -- then this deserves a code comment. If we _know_ that there is type safety and this really only ever catches `None`, then it's better to say `if res_dict[attr] is None:`, and so on.

PR #707, opened 2023-02-08
releases: add a dispatchable workflow

austin3dickey:
This was inspired by the comment at https://github.com/conbench/conbench/pull/704#pullrequestreview-1289777437.

This PR removes the "deploy-packages" step from the normal CI build and adds a dispatchable GHA workflow. We would kick this off whenever we need to release a new version of a package.

It would generate a new version number (like 2023.2.9), overwrite the version in the repo by creating/merging a PR (that we could discuss on if necessary), and build/upload the wheel.


austin3dickey:
I haven't tested this fully but I'd love to get your thoughts.

.github/workflows/release.yml
```diff
@@ -0,0 +1,126 @@
...
+              head: '${{ steps.variables.outputs.branch_name }}',
+              base: 'main',
+              body: 'Beep, boop. I am a bot, requested to open this PR in preparation to release to PyPI.'
+            })
+
+            await github.rest.pulls.merge({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: pull.number
+            })
```
jonkeane: I'm fine to do this in a pull request like this, but we could also just push to main (that's effectively what's going on here since it opens and closes without anything in between), no?

austin3dickey: Yes we could! I thought the PR might give a little more observability, and create a place if we want discussion. But if it's too much noise we can just push to main.

.github/workflows/release.yml
```diff
@@ -0,0 +1,126 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v3
+      - name: Define some variables to use later
+        id: variables
+        run: |
+          # Generate the new version number (YYYY.m.d.H.M, from the date in UTC)
+          # We don't want zero-padding. Use space-padding, then delete the spaces.
+          version=$(date -u +'%Y.%_m.%_d.%_H.%_M' | sed 's/ //g')
```
jgehrcke: Oh! I don't want to slow you down, but am inclined to add thoughts again here for the record.

This version number evolution across PRs is telling us something. We can call this 'weird', ack? :) I think this is not surprising. When translating thoughts into code the universal truth becomes more apparent. This is telling us that we didn't really align deeply yet on what we want to achieve and what exactly we need why. We knew that before, and we decided that we're OK with that.

It relates to my high-level question 

> When would we typically want to make a release?

https://github.com/conbench/conbench/pull/704#issuecomment-1423777215

And that of course needs an answer to the question: "what do we expect from a release?"

The code here suggests that you maybe wanted to add the ability to make more than one release per day?

If we want fast pace (get changes out quickly to consumers), then I don't think we want to do that with releases in our current project phase. But instead with 

> For consumers: if they want the latest dev version (HEAD of main) "now" -- how would they do that?

which we discussed in 
https://github.com/conbench/conbench/pull/704#issuecomment-1423086472.

Maybe for dev/testing you wanted to have the ability to run this more than once per day w/o running into a PyPI conflict. In that case:  add a suffix for the special case, not affecting the common case? maybe use `MICRO`(/`PATCH`) instead of `d`? Or add the commit hash prefix? 

(discussing version numbers is inflicting overhead, like this :mag: )

austin3dickey: Oops, I forgot to answer your question in GitHub, but had put my thoughts here:
https://github.com/conbench/conbench/blob/9199cc042f4f8bc415df80f349d90510f2c353fe/README.md?plain=1#L76-L77
My idea is that we almost never will have more than one release per day. At our current rate of development, probably a few per month.

However, I anticipate the possibility that we'd need to push out a hotfix on code that was released earlier that day. For example, #696 was merged on the same UTC day as #687. If Edward had released `benchadapt` after #687 to provide value to a user, he'd have to wait to release the second bugfix until the next day.

But I agree this is a rare case! I like your suffix idea.

austin3dickey: Okay, I added a workflow input to allow for a suffix, but defaulted back to `YYYY.m.d`

jonkeane: I really _do not_ want to beat a dead horse here (in the spirit of disagree and commit, we've merged the releasing, so let's go with it). 

> However, I anticipate the possibility that we'd need to push out a hotfix on code that was released earlier that day. For example, https://github.com/conbench/conbench/pull/696 was merged on the same UTC day as https://github.com/conbench/conbench/pull/687. If Edward had released benchadapt after https://github.com/conbench/conbench/pull/687 to provide value to a user, he'd have to wait to release the second bugfix until the next day. 

If this happens frequently, it's a sign that something is wrong with our process. It could be a number of things, but ones that come to mind (and I want to be super clear I'm not trying to point fingers or blame anyone or any thing for the specific instance brought up up above, these are general and not criticisms of that past work!): 

* we're lacking test coverage for the features or we're shipping not-quite-fully baked features
* we need a more rigorous process for checking our integration points on releases
* we're iterating and prototyping with this code and moving quickly is of importance 

And many of those points are very similar what @jgehrcke (though please feel free to disagree with me on my characterization here, I am not trying to speak for you — but do want to make sure I acknowledge what I saw you say in the prior thread!) and I have been saying when we voiced our hesitancy to set up this process. If the third point is the concern, having a process here holds us back and slows us down. If the second point is the concern, we should have more comprehensive checks in our release process to ensure that we don't break things like that. Which is what led us to pushing for a more lightweight "let's push artifacts somewhere, and not label things a release since that comes with guarantees that we might need to scramble to fix — which happen in part because we haven't _really_ set up a true release process with extra checks that we don't do that".

austin3dickey: Yep, I completely agree with you both. I don't expect anything beyond `YYYY.m.d` to happen more than very, very rarely. If it does, we should reevaluate this process.

I still think there's value in the suffix solution I just committed. But if you disagree, we can take it out.

jonkeane: I'm fine with that format + needing to specify it manually. I couldn't quite tell just from reading the code, what would happen if there's a release 2023.01.01 and then on the same day I try and release without adding the manual suffix? Surely it would error somewhere, but where?

jgehrcke: > it would error somewhere, but where

Proper PyPI (there also is testing PyPI) does not allow for overwriting releases. There simply is no way. So that would be where it fails if nothing else fails before. Symptom probably is some 409 HTTP conflict error.

austin3dickey: Oh good catch! It would probably either error during the git or PR steps (since there would be no change to the version file) or all the way in the upload step (since PyPI would reject the same version). I'll add a check that the version file was actually changed.

alistaire47: To be clear, I've got no desire to do more than one release a day; for new things that are iterating quickly, installing from the tip of main or another branch is preferable. But being able to release when we're confident we've got a stable version that can then easily be pulled by users and build systems is still useful. I don't want to try too hard to define what our confidence threshold is, but I suspect it's somewhere around when we're saying "cool this works, let's put it in x production system"

austin3dickey: Alright, I added the check if the version file would be changed. I'm going to test this workflow on a fork.

austin3dickey:
Okay @jgehrcke @jonkeane @alistaire47 this is ready to ship if you'd like a final review!

alistaire47 (approved):
I'm good with this. A couple thoughts: 

1. I'm on the fence about whether it's good to make a new PR for each deployment and then merge it right away (will this break if we start requiring approvals?) since any conversation on the PR would be after-the-fact anyway, but it's fine.
2. I like this better than deploying based on version bumps, but does this mean if we intend to release, in a prior PR we should bump the version to what it will be so the tagged commit will have the right version in the code?

austin3dickey:
>I'm on the fence about whether it's good to make a new PR for each deployment and then merge it right away (will this break if we start requiring approvals?) since any conversation on the PR would be after-the-fact anyway, but it's fine.

Yeah, I figure let's just see how this goes to start with. I'd like to get the emails, at least. We can always change it later.

>does this mean if we intend to release, in a prior PR we should bump the version to what it will be so the tagged commit will have the right version in the code?

Nope! This workflow tags the commit *after* the version bump:
https://github.com/conbench/conbench/blob/42fa2824ed1733d554cd045cfc32fd1e67ef001a/.github/workflows/release.yml#L130-L137
https://github.com/conbench/conbench/blob/42fa2824ed1733d554cd045cfc32fd1e67ef001a/.github/workflows/release.yml#L155-L164

Issue #709, opened 2023-02-09
`make run-app-dev`: psutil fails building on macOS (gcc not found)


sayantikabanik:
#### Error message

```shell
failed to solve: executor failed running [/bin/sh -c pip install -r /tmp/requirements-webapp.txt &&     pip install -r /tmp/requirements-dev.txt &&     rm -rf ./root/.cache]: exit code: 1
make: *** [run-app-dev] Error 17
```

jgehrcke:
Hey. `failed to solve: executor failed running` is interesting, but I think above we're not seeing everything that we need to see to learn more about the problem.

Can you run
```
$ cd /tmp && git clone https://github.com/conbench/conbench && cd conbench && make run-app-dev
```
and put the complete output into a gist? Thanks!

sayantikabanik:
<details>
<summary> Complete log </summary>

```shell
(conbench) sayantikabanik@Sayantikas-MacBook-Pro conbench % cd /tmp && git clone https://github.com/conbench/conbench && cd conbench && make run-app-dev 
Cloning into 'conbench'...
remote: Enumerating objects: 4787, done.
remote: Counting objects: 100% (459/459), done.
remote: Compressing objects: 100% (225/225), done.
remote: Total 4787 (delta 289), reused 318 (delta 216), pack-reused 4328
Receiving objects: 100% (4787/4787), 2.37 MiB | 3.62 MiB/s, done.
Resolving deltas: 100% (3507/3507), done.
# Write a file buildinfo.json to the root of the repository. This file is
# later meant to be added to build artifacts / container images. For
# discovery during runtime (via convention on the path in the container
# file system). During non-containerized local dev, either do not fail when
# this file does not exist or use path convention on host file system or
# introduce an environment variable (sth like CONBENCH_BUILDINFO_PATH).
#
# Use `sed` for the replacements below. Exit code is 0 for both cases:
# replacement made, or replacement not made. Expect non-zero exit code only
# for e.g. file not existing.
#
# Notes:
# - `buildinfo.json` is not tracked in the repository.
# - `git branch --show-current` requires at least git 2.22.
# - use GITHUB_REF_NAME in GHA, documented with "The short ref name of the
#   branch or tag that triggered the workflow run. This value matches the
#   branch or tag name shown on GitHub"
# - use in-place editing with `sed` to make this portable across Linux and
#   macOS: https://stackoverflow.com/a/16746032/145400
cat ci/buildinfo.json.template > buildinfo.json
[...]

PR #710, opened 2023-02-09
ui: `onclick` propagation event trigger condition updated and search fixes

### LINKED ISSUE #671: ui: unique value filter per column: drop-down contains empty field, and also null
I think the following screenshots show unintended behavior in various ways. Just leaving this here to hopefully save some time.

Screenshot from 2023-02-04 14-24-38
Screenshot from 2023-02-04 14-27-28

Found in current staging environment, `runs/96d6f941f20e456a9d60b68699121084/`

### LINKED ISSUE #708: ui: Drop-down filter columns still change the sort when used
This is a follow on to #613 when I click on a drop down to filter, the ordering of the column with the dropdown is changed. I know we fixed this for text-fields, but I'm seeing it for dropdowns too. 

For example, the reason column on the main page: https://conbench.ursa.dev/ (the screen shot doesn't show much other than the changed state, but in case it's helpful!)

image

We should fix this up along with any of the other column types. If #661 is going to come soon, I'm fine just doing that which should _also_ solve this problem. But if we're not going to work on that soon let's fix this bug soon.

### LINKED ISSUE #734: Search: free text search not enabled for `commit message col` staging env
Happen to notice search wasn't enabled due to a missed condition

sayantikabanik:
closes #708 
  - Event trigger handler added for `click` action 
  
closes #734 
  - condition updated (>=10)

closes #671 
  - error col text added -> adding it drop-down
  - minor UI fix : setting static width
  - search logic updated for dropdown

~I have to test the condition after the web-app execution comes though.
Related issue #709~

### DIFF
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -157,11 +157,13 @@ function dropdown_column_search(data_table_id, column_indexes) {
             var that = this;
             var reset = ""
             var mySelectList = $("<select class='custom-search-box'></select>")
+                //stops ordering event when clicked on a menu selection
+                .on("click", function (e){
+                    e.stopImmediatePropagation()
+                })
                 .appendTo(this.header())
-                .on("click", function(e) {
+                .on("change", function() {
                     that.search($(this).val()).draw();
-                    //stops ordering event when clicked on a menu selection
-                    e.stopPropagation();
                 });
             mySelectList.append(
                 $(`<option value="${reset}">${reset}</option>`)
@@ -178,7 +180,6 @@ function dropdown_column_search(data_table_id, column_indexes) {
                     );
                 }
             });
-            // console.log(mySelectList)
         });
     });
 }
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -160,6 +160,7 @@ function dropdown_column_search(data_table_id, column_indexes) {
                 .appendTo(this.header())
                 .on("click", function(e) {
                     that.search($(this).val()).draw();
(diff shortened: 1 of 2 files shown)


sayantikabanik:
testing a few edge-cases

More on various event propagation handlers -https://www.carlrippon.com/stoppropagation-v-stopimmediatepropagation/

sayantikabanik:
I tried testing most combinations , but definitely  not all of them. 
The `null` selection still feels buggy, apart from this rest should work as expected. 

UPDATE:

The bug for null value selection is hopefully now resolved.

conbench/templates/app.html
```diff
@@ -157,26 +157,46 @@ <h4 class="modal-title"><span class="glyphicon glyphicon-trash submenu"></span>&
...
                 });
             mySelectList.append(
-                $(`<option value="${reset}">${reset}</option>`)
+                $(`<option value="All">${reset}</option>`)
             );
             this.cache("search").unique().sort().each(function(param) {
-                // check for multiple whitespaces and empty string
-                if (/^\s*$/.test(param) == false) {
+                //trim the whitespaces (setting to empty string) before appending to drop-down list
+                param = param.trim()
```
jgehrcke, line 186: this trims leading and trailing whitespace, yes?

jgehrcke (approved):
:rocket: thank you for the detailed PR description. I think we should land this and then iterate further as we see problems/challenges arise!

conbench/templates/app.html
sayantikabanik: Yup :)

PR #715, opened 2023-02-09
Makefile: fix run-app-dev (tmpdir), do not run db populate on minikube for now

jgehrcke:
This is to address https://github.com/conbench/conbench/issues/709#issuecomment-1424029036 observed by @sayantikabanik.


jgehrcke:
@sayantikabanik confirmed that this helped.

Issue #712, opened 2023-02-09
buildkite deploy: branch name empty in buildinfo.json


jgehrcke:
```
++ make set-build-info
...
cat buildinfo.json
{
    "branch_name": "",
...
```
seen in e.g. `conbench-staging-deploy/builds/231#018635b3-e5cd-4702-a97b-fc86425ebdc2/105-107`.

jgehrcke:
https://github.com/conbench/conbench/pull/718 helped!

From the build log of `conbench-staging-deploy/builds/235`:
```
cat buildinfo.json
{
    "branch_name": "main",
    "version_string": "37d7dea2b-dev",
    "commit": "37d7dea2bfeff2e0493cedb0ab804d9d0ea64ba5",
    "build_time_rfc3339": "2023-02-09 17:38:07Z",
    "build_hostname": "ip-10-1-5-94.us-east-2.compute.internal"
}
```

PR #718, opened 2023-02-09
`make set-built-info`: identify branch name with new logic

jgehrcke:
This is to address https://github.com/conbench/conbench/issues/712.

Sources, and their priority:

- `BUILDKITE_BRANCH` env var, if set
- `GITHUB_REF_NAME` env var, if set
- local `git branch --show-current`

Code looks ugly. I think I tested all combinations locally.


austin3dickey (approved):
Thanks!

Issue #713, opened 2023-02-09
make conbench-on-minikube: illegal `date` option, not enough CPUs, psutil build fails


jgehrcke:
@amoeba tried running `make conbench-on-minikube`. Helpful build log. I am going through parts here step by step.

`make conbench-on-minikube` triggers `make set-build-info` through dependencies, and that failed:
```
...
date: illegal option -- -
usage: date [-jnRu] [-r seconds|file] [-v[+|-]val[ymwdHMS]]
            [-I[date | hours | minutes | seconds]]
            [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]
```

Cause: `date --rfc-3339=seconds --utc` requires GNU date 5.9 upwards. TODO: make this more compatible.

Another error during starting minikube:

```
Exiting due to RSRC_INSUFFICIENT_CORES: Requested cpu count 4 is greater than the available cpus of 2
```
TODO: do not assume 4 CPUs. :)

And then the build log also contains
```
Building wheel for psutil (pyproject.toml): finished with status 'error'
```
during `make build-conbench-container-image`. This is the same problem as tracked in https://github.com/conbench/conbench/issues/709, solved in `main`.

austin3dickey:
I'm still getting `date: illegal option -- -` on macos. I think the only way to specify UTC is with `-u`.

jgehrcke:
That should be resolved now. Thanks for testing.

PR #721, opened 2023-02-09
Make targets: change date and sed args for Darwin compatibility

jgehrcke:
See https://stackoverflow.com/a/16746032/145400. Thank you @austin3dickey also for https://github.com/conbench/conbench/issues/713#issuecomment-1424580171.


jgehrcke:
OK, this helped. Addressing other problems in other PRs. I really do like how CI enables us to go quickly right now.

PR #723, opened 2023-02-09
benchadapt: fix GoogleBenchmarkAdapter grouping

### LINKED ISSUE #702: benchadapt: GoogleBenchmarkAdapter not separating name and params for grouping

alistaire47:
Closes #702 by adding a level of grouping, so grouping at the case level collapses iterations (repetitions in microbenchmark-speak), and grouping at the name level collapses cases into a batch.

Adds some new test data that exercises all the levels of grouping better; left the old because it has some other mess like aggregates that we still want to test. Duplicated and tweaked the test instead of parameterizing everything because I find it easier to debug this way when something breaks.

### DIFF
--- a/benchadapt/python/benchadapt/adapters/gbench.py
+++ b/benchadapt/python/benchadapt/adapters/gbench.py
@@ -164,20 +164,21 @@ class GoogleBenchmarkAdapter(BenchmarkAdapter):
         gbench_context, benchmark_groups = self._parse_gbench_json(results)
 
         parsed_results = []
-        for benchmark in benchmark_groups:
-            # all results for a benchmark name share a batch id
-            batch_id = uuid.uuid4().hex
-            result_parsed = self._parse_benchmark(
-                result=benchmark,
-                batch_id=batch_id,
-                extra_tags=extra_tags,
-            )
-            result_parsed.optional_benchmark_info = {"gbench_context": gbench_context}
-            parsed_results.append(result_parsed)
+        for name, benchmark in benchmark_groups.items():
+            for case in benchmark["cases"]:
+                result_parsed = self._parse_benchmark(
+                    result=case,
+                    batch_id=benchmark["batch_id"],
+                    extra_tags=extra_tags,
+                )
+                result_parsed.optional_benchmark_info = {
+                    "gbench_context": gbench_context
+                }
+                parsed_results.append(result_parsed)
 
         return parsed_results
 
-    def _parse_gbench_json(self, raw_json: dict) -> Tuple[dict, list]:
+    def _parse_gbench_json(self, raw_json: dict) -> Tuple[dict, dict]:
         """
         Parse gbench result json into a context dict and a list of grouped benchmarks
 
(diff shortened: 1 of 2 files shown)


austin3dickey (approved):
Thanks for the tests!

benchadapt/python/benchadapt/adapters/gbench.py
```diff
@@ -164,20 +164,21 @@ def _parse_results(self, results: dict, extra_tags: dict) -> List[BenchmarkResul
...
-            # all results for a benchmark name share a batch id
-            batch_id = uuid.uuid4().hex
-            result_parsed = self._parse_benchmark(
-                result=benchmark,
-                batch_id=batch_id,
-                extra_tags=extra_tags,
-            )
-            result_parsed.optional_benchmark_info = {"gbench_context": gbench_context}
-            parsed_results.append(result_parsed)
+        for name, benchmark in benchmark_groups.items():
```
jonkeane, line 167: I'm not sure if this is better or worse, but `name` here goes unused, right? Would it be preferable to do `for _, benchmark in ...`? to show that?

benchadapt/python/tests/adapters/test_gbench.py
```diff
@@ -232,3 +321,46 @@ def test_transform_results(self, gbench_adapter) -> None:
...
+            assert result.tags["name"].startswith("DoThingAndThenUndoIt")
+            assert result.context == {"benchmark_language": "C++"}
+            assert "params" in result.tags
+            assert result.machine_info is not None
+            assert result.stats["iterations"] == 3
+            assert "gbench_context" in result.optional_benchmark_info
+
+    def test_run(self, gbench_adapter) -> None:
+        results = gbench_adapter.run()
+        assert len(results) == 4
```
jonkeane, line 366: ❤️

jonkeane (approved):

benchadapt/python/benchadapt/adapters/gbench.py
alistaire47: `name` does get used as a key on L219 (though really that could be a list instead of a dict; the keys are just so it's easier to see what each element is), but `name_params` on L223 never gets used. I left it so it's obvious what the grouping level here is (something like `MyBenchmark/param1=1/param2=2`), though by convention it should be `_`. I mildly prefer writing it out so it's easier to see why it's structured like this, but it's not a strong preference if someone disagrees.

jonkeane: _nods_ yeah, I saw that it was used elsewhere, but was (mildly) confused about this specific loop here. Just to be extra clear: the loop down below that you mention (~L219) is totally separate from this one, yeah?

> I left it so it's obvious what the grouping level here is (something like MyBenchmark/param1=1/param2=2), though by convention it should be _. I mildly prefer writing it out so it's easier to see why it's structured like this, but it's not a strong preference if someone disagrees.

Maybe let's put a comment that says that about this loop? Just so that anyone coming back here doesn't think "hey this `name` in this loop definition is totally superfluous, that's dumb!"

alistaire47: Renamed a bit and rearranged the comments to try to make it clearer what both the outer and inner grouping and loops are doing

PR #732, opened 2023-02-10
benchclients: Improve README and docstrings

### LINKED ISSUE #731: [benchclients] [docs]: Improve benchclients README
[The benchclients README](https://github.com/conbench/conbench/blob/main/benchclients/python/README.md) is sad, and could use some love. More context around what methods are on clients would help, as looking at the source for `ConbenchClient` hides that it has a pile of useful methods from its parent class.

alistaire47:
Closes #731 by filling in benchclients' README and tweaking its docstrings a tad. Not trying to make it perfect, just get enough there that users looking at it can read docs to get the bigger picture quickly.

### DIFF
--- a/benchclients/python/README.md
+++ b/benchclients/python/README.md
@@ -1,3 +1,104 @@
 # benchclients
 
 A lightweight Python package with clients for working with APIs in the Conbench ecosystem
+
+## Installation
+
+The benchclients package can be installed [from PyPI](https://pypi.org/project/benchclients/)
+with
+
+``` sh
+pip install benchclients
+```
+
+or direct from GitHub with
+
+``` sh
+pip install benchclients@git+https://github.com/conbench/conbench.git@main#subdirectory=benchclients/python
+```
+
+or with your preferred environment management system. Package code should rarely change,
+and the number of dependencies should remain very minimal.
+
+## Components
+
+The package is structured such that all users should need is client classes that it
+exposes. As of writing, there is only one such useful class: `benchclients.ConbenchClient`.
+However, clients for other APIs can easily be created by inheriting from the
+`benchclients.BaseAdapter` class, which defines core methods for making HTTP requests
+(e.g. `.get()`, `.post()`, etc.) set up to handle retries and logging and so on.
+
+### `ConbenchClient`
+
+#### Environment variables
+
+Configuration and credentials for `ConbenchClient` are handled by three environment
+variables set before class instantiation:
+
+- `CONBENCH_URL`: The URL of the Conbench server. Required.
+- `CONBENCH_EMAIL`: The email to use for Conbench login. Only required for GETting if the
+server is private.
+- `CONBENCH_PASSWORD`: The password to use for Conbench login. Only required for GETting
+if the server is private.
+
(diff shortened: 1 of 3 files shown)


benchclients/python/README.md
```diff
@@ -1,3 +1,104 @@
 # benchclients
 
 A lightweight Python package with clients for working with APIs in the Conbench ecosystem
```
jgehrcke: Within the first few lines I'd love to understand whether this is a library or a CLI or both. I do not care so much about the "lightweight". 

>  for working with APIs in the Conbench ecosystem

This sounds like we have various APIs to work with. I think there is just one, but maybe I don't know something here. Let's add precision :-).

If we can narrow down "working with the API" a tiny little further, I'd also love to see that. 

Examples: is this supposed to support the _complete_ Conbench API surface? Or does it have the goal to help with submitting benchmark data?

Right now I still have the luxury to ask these questions w/o actually knowing the answers, but I'd love to find them here :).

alistaire47: > Within the first few lines I'd love to understand whether this is a library or a CLI or both. 

My mental model is that "package" implies "library" unless CLI is mentioned, but yeah, we can switch to "library" if it's clearer.

> I do not care so much about the "lightweight".

Some users do; part of why this package exists is because the legacy runner required installing all the webapp deps, which not everybody wants to install in their CI.

> > for working with APIs in the Conbench ecosystem
> 
> This sounds like we have various APIs to work with. I think there is just one, but maybe I don't know something here. Let's add precision :-).

Yeah, there's a little ambiguity here because primarily this library addresses the one right now, but is designed to work with more if/when they exist; `benchalerts` extends the base client here to make a `GitHubClient`. If that becomes a thing others need, it should live here. So there's a difference between "what is currently here" and "what belongs here". I don't want to make this too verbose, but I'll see if I can clarify.

> If we can narrow down "working with the API" a tiny little further, I'd also love to see that.
> 
> Examples: is this supposed to support the _complete_ Conbench API surface? Or does it have the goal to help with submitting benchmark data?

It's very general and doesn't get specific to what endpoints exist, it just handles auth and retries and logging and such. So it's a slightly higher level than `requests` (though not much), but definitely isn't a mapping of every endpoint. Users will still probably need to reference the Swagger or another docs source to actually make useful requests with the client.

jgehrcke: >  we can switch to "library" if it's clearer.

Yes, please! 

> part of why this package exists is because the legacy runner required installing all the webapp deps, which not everybody wants to install in their CI.

That's a great point to know to understand the history of this library. 

But for new people coming to this ecosystem they don't know the legacy runner and "lightweight" means something different depending on the person who reads it.

> addresses the one right now, but is designed to work with more if/when they exist;

It's this level of ambiguity which makes it hard for new users and developers to understand what we build, and what the individual components do. One can describe the state of today with precision, without imposing road blocks for the future :). Precisely by embracing what you write:

>  a difference between "what is currently here" and "what belongs here". I don't want to make this too verbose, but I'll see if I can clarify.

Yes yes please. This is really valuable.

Issue #693, opened 2023-02-07
ci instability: cb-on-minikube may fail during postgres-operator deployment


jgehrcke:
Seen here: https://github.com/conbench/conbench/pull/692#issuecomment-1420616849
```
Wait for port forwarding to take effect
The command curl --location --silent --output /dev/null http://127.0.0.1:8080/clusters timed out
==== FAILURE: OPERATOR DID NOT START OR PORT FORWARDING DID NOT WORK
If you observe issues with minikube VM not starting/not proceeding, consider deleting the .minikube dir and/or rebooting before re-running the script
Error: Process completed with exit code 1.
```

Went away upon re-run, proving that this is a non-deterministic effect.

Maybe trying longer than 1 minute is a solution: https://github.com/zalando/postgres-operator/blob/933b45aba0d64243368e1ed8619ce1cbfb3e7ac6/run_operator_locally.sh#L34

Maybe _not_ swallowing stderr is of help: https://github.com/zalando/postgres-operator/blob/933b45aba0d64243368e1ed8619ce1cbfb3e7ac6/run_operator_locally.sh#L170

jgehrcke:
@austin3dickey had this twice yesterday. Right?
Depending on the platform this is seemingly way more likely to hit in. Needs tackling.

jgehrcke:
Seen by @alistaire47 here: https://github.com/conbench/conbench/actions/runs/4138141592/jobs/7154181896 but that was before adding the latest improvements on this front.

@austin3dickey I'd really love to see the current state on your machine. Thanks!

jgehrcke:
Just observed this again locally:
```
Wait for port forwarding to take effect
Wait for port forwarding to take effect
Wait for port forwarding to take effect
Wait for port forwarding to take effect
Wait for port forwarding to take effect
Wait for port forwarding to take effect
The command curl --location --silent --output /dev/null http://127.0.0.1:8080/clusters timed out
==== FAILURE: OPERATOR DID NOT START OR PORT FORWARDING DID NOT WORK
```

Means that https://github.com/conbench/conbench/pull/725 did not help, which was one of two expected outcomes. We've learned that this needs further debugging.

It's entirely possible that the port-forward was never set up properly before going into the curl loop. We wouldn't see a corresponding error as I've pointed out in the initial comment above:

> Maybe not swallowing stderr is of help

jgehrcke:
[...]

PR #736, opened 2023-02-13
ci: use patched postgres-operator for robustness

jgehrcke:
This attempts to address issue #693. It's likely that this heals the unknown problem, but if it doesn't then as of better debug output this will shed light on the cause of the instability.

Also adds to unrelated Makefile changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -187,6 +187,7 @@ minikube-conbench-url:
 start-minikube:
 	minikube delete --profile mk-conbench
 	minikube start --profile mk-conbench --cpus=2 --memory=4g \
+		--driver=docker \
 		--disable-metrics \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -187,6 +187,7 @@ minikube-conbench-url:
 start-minikube:
 	minikube delete --profile mk-conbench
 	minikube start --profile mk-conbench --cpus=2 --memory=4g \
+		--driver=docker \
```
jgehrcke, line 190: On my machine minikube suddenly started prompting me for a superuser password while picking the specific 'driver' it wanted to use. I got this prompt non-deterministically. I have the hope that being explicit about the driver helps.

PR #738, opened 2023-02-13
kube-prometheus: prepare config to remote_write to external system

jgehrcke:
This is a patch towards https://github.com/conbench/conbench/issues/737, relevant research/discussion and also references can be found in there. 

Aspects that matter:

* The goal is to get the `remoteWrite` key into the specification for the Prometheus k8s object, as defined here: https://github.com/prometheus-operator/prometheus-operator/blob/c237d26b62ee5e29087e01f173e94886ada5b2ec/Documentation/api.md#monitoring.coreos.com/v1.RemoteWriteSpec. One can do that before JSONNET compilation or thereafter. The choice between both approaches is not so easy to make as of today. I have picked _before_ for now.
* The invalid remote_write URL and credentials are as intended (indicating that by default this the pusher proms do _not_ write to an external system).

### DIFF
--- a/k8s/kube-prometheus/conbench-flavor.jsonnet
+++ b/k8s/kube-prometheus/conbench-flavor.jsonnet
@@ -9,7 +9,7 @@ local kp =
   // Uncomment the following imports to enable its patches
   // (import 'kube-prometheus/addons/anti-affinity.libsonnet') +
   // (import 'kube-prometheus/addons/managed-cluster.libsonnet') +
-  // (import 'kube-prometheus/addons/node-ports.libsonnet') +
+  (import 'kube-prometheus/addons/node-ports.libsonnet') +
   // (import 'kube-prometheus/addons/static-etcd.libsonnet') +
   // (import 'kube-prometheus/addons/custom-metrics.libsonnet') +
   // (import 'kube-prometheus/addons/external-metrics.libsonnet') +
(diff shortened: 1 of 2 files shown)


k8s/kube-prometheus/conbench-flavor.jsonnet
```diff
@@ -28,6 +28,25 @@ local kp =
...
+              },
+              password: {
+                name: 'kubepromsecret',
+                key: 'password',
+              },
+            },
+          }],
+        },
+      },
+    },
```
jgehrcke, line 49: For the record; CI confirms validity of this as part of the ci/minikube check on multiple levels:
- we perform JSONNET compilation (output: k8s manifest files). JSONNET compilation fails for unexpected structure/keys, etc.
- the k8s manifest files are submittes to k8s, and test-conbench-on-mk.sh confirms that after all, Prometheus deploys correctly (it would not if its config was invalid).

k8s/kube-prometheus/conbench-flavor.jsonnet
```diff
@@ -9,7 +9,7 @@ local kp =
   // Uncomment the following imports to enable its patches
   // (import 'kube-prometheus/addons/anti-affinity.libsonnet') +
   // (import 'kube-prometheus/addons/managed-cluster.libsonnet') +
-  // (import 'kube-prometheus/addons/node-ports.libsonnet') +
+  (import 'kube-prometheus/addons/node-ports.libsonnet') +
```
jgehrcke, line 12: I am not sure if this is needed.

jgehrcke:
OK, this is a working checkpoint that I'd love to land to base further work on.

PR #739, opened 2023-02-14
test-conbench-on-mk: pin run_operator_locally.sh

jgehrcke:
This makes explicit use of the current state of https://github.com/zalando/postgres-operator/pull/2218. To address https://github.com/conbench/conbench/issues/693.


jgehrcke:
Landing this now; would love to see you trying `make conbench-on-minikube` with this again @austin3dickey.

PR #741, opened 2023-02-14
conbench webapp grafana dashboard: dynamic datasource

jgehrcke:
This adds a dashboard-local variable to the dashboard which is consumed in each panel. This makes it so that the same dashboard JSON model works as-is in different Grafana instances (which have different data sources attached), either by doing precisely nothing, or by just having to pick a data source from a drop down.

The following screenshot shows that in action:
image

On the right: this is the Grafana UI as served by grafana-on-minikube on my `http://localhost:3000/`. It shows the dashboard as defined by `k8s/kube-prometheus/conbench-grafana-dashboard.json` in this repo as of today.

On the left: this is Grafana UI as served by my personal Grafana Cloud tenant `https://jgehrckevdgrafana.grafana.net/`. It shows the same dashboard after me having only copy/pasted the JSON into a field. 

A notable difference between both screens around the top-left corner.
- Left it says `datasource: grafanacloud...-prom` (which is the _receiving end_ of where my local minikube cluster pushes metrics). 
- Right it says `datasource: prometheus` (which is the minikube-local Prometheus managed by kube-prometheus, actually running in a k8s StatefulSet).

In the background behind the two browser windows is a photo of the moon :full_moon: :woman_shrugging:

### DIFF
--- a/k8s/kube-prometheus/conbench-flavor.jsonnet
+++ b/k8s/kube-prometheus/conbench-flavor.jsonnet
@@ -63,10 +63,14 @@ local kp =
             // Reference docs for the mechanism used here:
             // https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config
             // https://github.com/prometheus-operator/prometheus-operator/blob/c237d26b62ee5e29087e01f173e94886ada5b2ec/Documentation/api.md#relabelconfig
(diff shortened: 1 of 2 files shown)


austin3dickey (approved):
very nice!

Issue #742, opened 2023-02-14
makefile: minikube-conbench-url during conbench-on-minikube fails on darwin


austin3dickey:
Discovered this small bug while running `make conbench-on-minikube` on Mac M1. It only seems to affect printing out the last helpful statement, not necessarily the behavior of setting everything up.

At the end, I see this:
```
Grafana UI port-forward:
     run: kubectl --namespace monitoring port-forward svc/grafana 3000
     then open the Grafana UI at: http://localhost:3000
     log in with admin/admin

Conbench UI port-forward:
     run: kubectl port-forward svc/conbench-service 8000:conbench-service-port
     then open the Conbench UI at: http://localhost:8000
❗  Because you are using a Docker driver on darwin, the terminal needs to be open to run it.
```

Based on the emoji, I'm assuming that this command is failing for some reason.
https://github.com/conbench/conbench/blob/c7528b43024e124710f4b7c150e248c1b749fb6b/Makefile#L172

I tried the subshell command by itself in a separate iTerm2 window, and got this, which hung:
```
> minikube --profile mk-conbench service conbench-service --url
http://127.0.0.1:50893
❗  Because you are using a Docker driver on darwin, the terminal needs to be open to run it.
```

jgehrcke:
Some background here:
https://github.com/kubernetes/minikube/issues/13788#issuecomment-1068640594
https://github.com/kubernetes/minikube/issues/14039

On the Linux setup the 'magic' is set up permanently; and the command in question here can immediately return to only display the URL.

On MacOS the required magic is not set up permanently, and the command in question here adds magic (and therefore becomes long-running).

> I tried the subshell command by itself in a separate iTerm2 window, and got this, which hung:

Thanks for trying.

In that case there is little additional value over running `kubectl port-forward svc/conbench-service 8000:conbench-service-port`.

jgehrcke:
This is resolved.

PR #753, opened 2023-02-16
Makefile: do not run `minikube service` on darwin

jgehrcke:
For issue #742.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -167,14 +167,24 @@ conbench-on-minikube: build-conbench-container-image start-minikube
 	@make -s minikube-conbench-url
 
 
+# The `minikube ...service conbench-service --url` command returns immediately
+# on Linux, but is a long-running process on Darwin.
+# See https://github.com/conbench/conbench/issues/742.
+# Only run it on Linux.
 .PHONY: minikube-conbench-url
 minikube-conbench-url:
-	@CONBENCH_BASE_URL=$$(minikube --profile mk-conbench service conbench-service --url) && \
+	@if [[ "$$OSTYPE" == "darwin"* ]]; then \
+		echo "Darwin detected. Not running automatically, because it's long-running: " && \
+		echo "  minikube --profile mk-conbench service conbench-service --url" && \
+		echo ""; \
+	else \
+		CONBENCH_BASE_URL=$$(minikube --profile mk-conbench service conbench-service --url); \
 		echo "" && \
 		echo "Depending on what you'd like to do next:" && \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -167,14 +167,24 @@ conbench-on-minikube: build-conbench-container-image start-minikube
...
+# The `minikube ...service conbench-service --url` command returns immediately
+# on Linux, but is a long-running process on Darwin.
+# See https://github.com/conbench/conbench/issues/742.
+# Only run it on Linux.
 .PHONY: minikube-conbench-url
 minikube-conbench-url:
-	@CONBENCH_BASE_URL=$$(minikube --profile mk-conbench service conbench-service --url) && \
+	@if [[ "$$OSTYPE" == "darwin"* ]]; then \
+		echo "Darwin detected. Not running automatically, because it's long-running: " && \
+		echo "  minikube --profile mk-conbench service conbench-service --url" && \
```
austin3dickey: Would this URL be necessary to use `make db-populate`? If so, could you add small instructions here for using the output of this command (like you do in the linux branch)?

austin3dickey (approved):
Hooray, now `make conbench-on-minikube` exits 0 for me!

Makefile
jgehrcke: I think it's OK to merge without this (it's additional value that we can add later). I agree that it's maybe nice. But I can't do that anymore today. Can you maybe do so, on this branch? :)

austin3dickey: Sure! I'll merge it afterwards.

jgehrcke: > Would this URL be necessary to use make db-populate?

Ah, and to your question: not necessary. In the sense that `minikube ..` is not the only way to construct this URL. Maybe you can get a bit of a feel which of both methods you like more, on your system:

- `kubectl port-forward ...` (as a long-running process)
- `minikube ... service ...` (as a long-running process)

On mac, both are long-running processes, and I don't know which of both I would like more. I think the `minikube ... service ...` might make it slightly easier to assign a port dynamically. Which could be a pro.

On linux, the latter is not long-running, which is why it adds value compared to  `kubectl port-forward ...`, which is why I started documenting this method in the first place.

jgehrcke: > Sure! I'll merge it afterwards.

Thanks!!

austin3dickey: I ended up recommending the `kubectl port-forward ...` way for Mac because it *doesn't* dynamically assign the port, and therefore we don't have to get the stdout of a long-running process. If a Mac user runs into problems here we can always take another look.

jgehrcke: good choice!

PR #761, opened 2023-02-20
Add benchmarking development process doc

jonkeane:
Adds a document describing how we use benchmarks in the development process.

I also had to add to the rst structure to get this on the main page (I created a new section at the top since this is the most important part of the docs we should be concentrating on right now). Part of me wonders if we should move that other TOC elsewhere or lower, but have left it be for now.

While trying to iterate on this I ran into the following (semi-stream of consciousness, but to show some of the bumps), we should probably add to our developer docs what the right way to do this + iterate is.

1. attempting to render the docs: docs for how to do this?
2. `make` from the Makefile, I guess? 
3. An error about sphinx not found: Oh I need to install sphinx I guess? `pip install '.[dev]'`
4. oh, there are lots of `make` options, `make html`, I guess? 
5. a bunch of errors cause I don't have benchadapt, adapters, benchalerts, etc. installed -> this is probably fine
6. not totally sure that the image will be copied over when rendered (it didn't seem to be the first time I tried it, but maybe it is automatically?)

PR #763, opened 2023-02-21
docs: add developer target for building docs

### LINKED ISSUE #762: docs: add developer makefile target for building docs
The command for building HTML docs is simple but non-trivial. We should add a Makefile target for doing so, such that developers can iterate locally and confirm that the docs are linted.

austin3dickey:
To address @jonkeane 's thoughts in https://github.com/conbench/conbench/pull/761#issue-1592417932, this PR adds a Makefile target and instructions for iterating locally with docs. Closes #762 .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,13 @@ rebuild-expected-api-docs: run-app-bg
 	rm _new_api_docs.json
 	git diff ./conbench/tests/api/_expected_docs.py
 
+
+# Build HTML docs, exiting non-zero if there are warnings. Those can be very bad.
+.PHONY: build-docs
+build-docs:
+	$(MAKE) -C docs html SPHINXOPTS='-W --keep-going'
+	echo 'To see the generated docs, open docs/_build/html/index.html'
+
 # The targets above this comment have been in use by humans. Change
 # conservatively.
 
--- a/Makefile
+++ b/Makefile
@@ -76,10 +76,10 @@ rebuild-expected-api-docs: run-app-bg
 	git diff ./conbench/tests/api/_expected_docs.py
 
 
-# Build HTML docs, exiting non-zero if there are warnings. Those can be very bad.
+# Build HTML docs locally
 .PHONY: build-docs
 build-docs:
-	$(MAKE) -C docs html SPHINXOPTS='-W --keep-going'
+	$(MAKE) -C docs html
 	echo 'To see the generated docs, open docs/_build/html/index.html'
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -75,6 +75,13 @@ rebuild-expected-api-docs: run-app-bg
 	rm _new_api_docs.json
 	git diff ./conbench/tests/api/_expected_docs.py
 
+
+# Build HTML docs, exiting non-zero if there are warnings. Those can be very bad.
```
jonkeane: While I agree that erroring on warnings in CI builds | deploys is good, not sure we want that when iterating locally. Could we make that an option?

README.md
```diff
@@ -115,6 +115,18 @@ Deploys the Conbench API server to a local minikube-powered Kubernetes cluster.
...
+* `make docs-build`: Builds HTML docs locally so you may check that they render
+  correctly with no linting problems. You may install dependencies with
+  ```
+  pip install \
+      -r requirements-dev.txt \
+      -e ./benchclients/python \
+      -e ./benchalerts \
+      -e ./benchadapt/python \
+      -e ./benchconnect \
+      -e ./benchrun/python \
```
jonkeane: We can keep this here, but we should probably also mention that installing _all the_ dependencies (eg benchalerts, etc.) isn't totally necessary if one isn't touching that. 

If I were contributing to the philosophy + general docs having to have functional installs of all of those is a pretty big hurdle to clear that isn't in principle important for us (right now at least).

Makefile
austin3dickey: Done: removed it from the Makefile command since one can just look at the warnings locally

README.md
austin3dickey: Yeah great point. Clarified!

README.md
```diff
@@ -115,6 +115,15 @@ Deploys the Conbench API server to a local minikube-powered Kubernetes cluster.
...
 
+* `make docs-build`: Builds HTML docs locally so you may check that they render
+  correctly with no linting problems. Dependencies can be installed with
+  `pip install -r requirements-dev.txt`. Also, if you're working on the docstrings of
+  any of this repo's python packages, ensure the package is installed locally before
+  using this command.
+
+  In CI, we use `make build-docs SPHINXOPTS='-W --keep-going'` to fail the build if
+  there are Sphinx warnings. When using this command locally, you can just do
+  `make build-docs`, but keep an eye on the warnings.
```
jgehrcke, line 126: :heart:

PR #765, opened 2023-02-21
benchalerts: fix bug in regression_details

### LINKED ISSUE #764: benchalerts: bug when rendering markdown details
If a commit has multiple runs, and some runs don't have a baseline run, `benchalerts` handles those runs separately when formatting a markdown summary.

But a bug exists in the implementation of that intended behavior:
https://github.com/conbench/conbench/blob/8f2523afb28687c8d2d29457937de8b1b65fe434/benchalerts/benchalerts/parse_conbench.py#L180-L185
If the run without a baseline happened to be the first run in the list, this last line fails with `TypeError: 'NoneType' object is not subscriptable`. If we get to that point in the code, we know that at least one comparison has `compare_details` that are populated. We should find that comparison instead of assuming it's the first one.

austin3dickey:
Closes #764 . See issue for details.

### DIFF
--- a/benchalerts/benchalerts/parse_conbench.py
+++ b/benchalerts/benchalerts/parse_conbench.py
@@ -17,6 +17,7 @@ from dataclasses import dataclass
 from typing import List
 
 from .clients import CheckStatus
+from .log import fatal_and_log
 from .talk_to_conbench import RunComparison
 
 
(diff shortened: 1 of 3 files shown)


benchalerts/benchalerts/parse_conbench.py
```diff
@@ -182,7 +182,13 @@ def regression_details(comparisons: List[RunComparison]) -> str:
     if not any(comparison.baseline_info for comparison in comparisons):
         return None
 
-    z_score_threshold = comparisons[0].compare_results[0]["threshold_z"]
+    # Find the z-score threshold that was used so we can display it here
+    z_score_threshold = None
+    for comparison in comparisons:
+        if comparison.compare_results:
+            z_score_threshold = comparison.compare_results[0]["z_score_threshold"]
```
jonkeane: These should all be the same, yeah? would something like this (+ possibly a check that the set is a single element) make sense + be easier to debug?

```
z_score_threshold  = set([x.compare_results[0]["threshold_z"] for x in comparisons if x is not None])
```

austin3dickey: Yes, they should be the same. This should be guaranteed, but there's not one nice code line I can point to to show that guarantee, so I'll do your `set()` suggestion so that it's obvious!

austin3dickey:
Going to merge to unblock future work.

Issue #745, opened 2023-02-14
conbench.ursa.dev > /compare/runs/... is too slow when comparing many with many results, often returns 502/504


ElenaHenderson:
Steps to reproduce this issue:
1. Go to https://conbench.ursa.dev/compare/runs/17345921fa2e410e83b12f1bfba89c87...ccc22b5568cd4ff189a158c908947801/

We should look at performance of db queries on this page and figure out what can be done.

I would have looked at this performance issue myself but I am swamped for at least 1-2 months.

These are the tools I love for investigating Postgres performance issues and they served me well for many years now in case anybody wants to take a stub:
https://github.com/ankane/pghero
https://tatiyants.com/pev/#/plans/new

We can also change gunicorn request timeout as a short term solution for this.

jgehrcke:
Did you get a 504 emitted by nginx or an 502?

I typically get 502 Bad Gateway on this endpoint:
![image](https://user-images.githubusercontent.com/265630/218966264-ec8a80aa-4fae-4ca7-a10e-61d2c6345b1b.png)


But the symptom varies over time. With the URL you provided I right now got a 500 response with this response body:
```
{"code": 500, "name": "Internal Server Error"}
```

![image](https://user-images.githubusercontent.com/265630/218966101-2541f2d3-4433-4356-8114-46e45073c4a1.png)

austin3dickey:
With https://github.com/conbench/conbench/pull/766 merged and deployed, this page loads more often for me. However, it still could be refactored.

austin3dickey:
I looked into this a little bit over the last hour or two. Most of the time spent when loading this page is (as I expected) in `set_z_scores()`, waiting for this query to complete:
https://github.com/conbench/conbench/blob/275617881da25ebe744d7b0502290ed44f7c3a21/conbench/entities/history.py#L271
I used EXPLAIN ANALYZE and your cool visualization tool to generate this:
<img width="1453" alt="image" src="https://user-images.githubusercontent.com/16600275/221015130-0cb767a6-9192-46d6-adce-2b872a0dc42c.png">

[...]

Issue #744, opened 2023-02-14
conbench.ursa.dev > /runs/run-id/ page returns intermittent 504


ElenaHenderson:
Steps to reproduce this issue:
1. Go to https://conbench.ursa.dev/runs/17345921fa2e410e83b12f1bfba89c87/
This reproduces on and off.

We should look at performance of db queries on this page and figure out what can be done.

jgehrcke:
reverse proxy gateway timeout hit in before gunicorn worker timeout:
```
$ time curl -v https://conbench.ursa.dev/runs/17345921fa2e410e83b12f1bfba89c87/
...
<html>
<head><title>504 Gateway Time-out</title></head>
<body>
<center><h1>504 Gateway Time-out</h1></center>
</body>
</html>
* Connection #0 to host conbench.ursa.dev left intact
real	1m0.599s
```

This is pretty precisely the nginx upstream request timeout (nginx not waiting for the webapp anymore).

austin3dickey:
With #766 merged and deployed, this page loads more often for me. However, it still could be refactored.

jgehrcke:
I visited the landing page of https://conbench.ursa.dev/ and tried to open the run overviews for the last few runs in new tabs. Most requests errored out with a 504 Gateway Timeout:
![image](https://user-images.githubusercontent.com/265630/221827298-36249340-4fa5-48fb-be89-f8453a4d3744.png)

PR #766, opened 2023-02-21
history: convert history stats calculations to pandas

austin3dickey:
May address #744 and #745.

This PR replaces the SQL query logic for calculating history stats (rolling mean, rolling standard deviation, etc) with pandas logic. The SQL query was incredibly complex and often timing out. Since we're operating on a relatively small number of rows, it's much faster to grab the raw data and do these calculations in pandas. I think this will reduce the number of timeouts on Conbench instances with large data.

In previous PRs, we consolidated all this logic into `conbench.entities.history` so I'm happy that we only have to change it in one place!

### DIFF
--- a/conbench/entities/history.py
+++ b/conbench/entities/history.py
@@ -376,7 +376,9 @@ def _add_rolling_stats_columns_to_df(
     """
     # pandas likes the data to be sorted
     df.sort_values(
-        ["case_id", "context_id", "hash", "repository", "timestamp"], inplace=True
+        ["case_id", "context_id", "hash", "repository", "timestamp"],
+        inplace=True,
+        ignore_index=True,
     )
 
     # Clean up begins_distribution_change so it's a non-null boolean column
@@ -386,16 +388,21 @@ def _add_rolling_stats_columns_to_df(
     ]
 
     # Add column with cumulative sum of distribution changes, to identify the segment
+    print("segment_id")
     df["segment_id"] = (
         df.groupby(["case_id", "context_id", "hash", "repository"])
         .rolling(
-            _CommitIndexer(window_size=len(df) + 1), on="timestamp", closed="right"
+            _CommitIndexer(window_size=len(df) + 1),
+            on="timestamp",
+            closed="right",
+            min_periods=1,
         )["begins_distribution_change"]
         .sum()
         .values
     )
 
     # Add column with rolling mean of the means (only inside of the segment)
+    print("rolling_mean_excluding_this_commit")
     df["rolling_mean_excluding_this_commit"] = (
         df.groupby(["case_id", "context_id", "hash", "repository", "segment_id"])
         .rolling(
@@ -403,6 +410,7 @@ def _add_rolling_stats_columns_to_df(
             on="timestamp",
             # Exclude the current commit first...
             closed="left",
+            min_periods=1,
         )["mean"]
         .mean()
         .values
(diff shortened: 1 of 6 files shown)


conbench/entities/history.py
```diff
@@ -342,112 +343,89 @@ def _add_rolling_stats_columns_to_history_query(
...
+        # HACK: pandas needs a datetime-like column if you want windows with
+        # potentially-different numbers of data points.
+        # TODO: use a subclass from pd.api.indexers.BaseIndexer instead
+        .astype("datetime64[s]")
+    )
+
+    # Add column with cumulative sum of distribution changes, to identify the segment
+    df["segment_id"] = (
+        df.groupby(["case_id", "context_id", "hash", "repository"])
+        .rolling(window=f"{Config.DISTRIBUTION_COMMITS}s", on="commit_rank")[
```
alistaire47: Shouldn't you just use the int as the window so it goes by observation instead of relying on a sequential 1s index since it's not time?

(Though to be honest, this exact parameter is one of the parts of the pandas API I've complained about the most; it's not just a union-typed param, it actually returns different classes with different methods depending on the type.)

austin3dickey: Yeah this API is really bad. The problem (as before) is we might have multiple observations per commit. So let's say `DISTRIBUTION_COMMITS=2`. I want something like this:
```
 ix | timestamp  | commit_rank | mean | rolled_ixs 
----+------------+-------------+------+------------
 0  | 2020-01-01 | 1           | 10   | -          
 1  | 2020-01-02 | 2           | 11   | 0          
 2  | 2020-01-02 | 2           | 12   | 0          
 3  | 2020-01-05 | 3           | 13   | 0,1,2      
 4  | 2020-01-07 | 4           | 14   | 1,2,3      
 5  | 2020-01-07 | 4           | 15   | 1,2,3      
```
Looking at the last row, the indices we want to use for `rolling_mean_excluding_this_commit` are 1,2,3. That's because we're on the Jan 7 commit, and we want to use data from the previous 2 commits (Jan 5 and Jan 2). That corresponds to indxes 1,2,3.

I was able to do this in SQL with a window function over `commit_rank` with RANGE BETWEEN 2 PRECEDING AND 1 PRECEDING. 

In pandas it's tricky. I can't do `window=2` because that assumes it's a constant 2 rows per window; i.e. the `rolled_ixs` for the last row would be 3,4. So I tried this hack. I'm not even sure it's working as intended; with this hack, the last two rows are returning different `rolled_ixs`.

austin3dickey: This is solved in the latest commit using [custom window rolling](https://pandas.pydata.org/docs/user_guide/window.html#custom-window-rolling).

requirements-webapp.txt
```diff
@@ -16,6 +16,7 @@ marshmallow
 numpy
 oauthlib
 openapi-spec-validator
+pandas
```
austin3dickey, line 19: This shouldn't affect our conda-feedstock

austin3dickey:
This is ready for review! I am still analyzing the performance speedup locally, but initial results look to be more than a 10x speedup.

austin3dickey:
Okay yes, running the webapp locally with a connection to our Arrow database shows promising results. Page loads always complete within 30 seconds. I will merge this now and deploy to Arrow first. Feedback is still welcome!

jgehrcke:
Quick feedback: I am super pleased to see that you tried this out! Didn't look at the code here yet.

I always love to connect the dots in a ticketing system for posterity; this patch relates to:
https://github.com/conbench/conbench/pull/636#issuecomment-1408918954
https://github.com/conbench/conbench/pull/636#pullrequestreview-1275014089
https://github.com/conbench/conbench/issues/577

PR #767, opened 2023-02-22
hotfix: bug in nullable column

austin3dickey:
This line was previously sometimes throwing `AttributeError: 'NoneType' object has no attribute 'get'`.

I'm not going to add a test in this PR, but we should eventually test for the case that `change_annotations` is null rather than `{}`.

### DIFF
--- a/conbench/entities/history.py
+++ b/conbench/entities/history.py
@@ -372,7 +372,7 @@ def _add_rolling_stats_columns_to_df(
 
     # Clean up begins_distribution_change so it's a non-null boolean column
     df["begins_distribution_change"] = [
-        bool(x.get("begins_distribution_change", False))
+        bool(x.get("begins_distribution_change", False)) if x else False
         for x in df["change_annotations"]
     ]
 


jgehrcke:
Would love to have a ticket about the symptom (probably a 500 response while GETting a specific endpoint).

austin3dickey:
Oops, yes I should have made a ticket. More background on this PR: after deploying a number of commits including https://github.com/conbench/conbench/pull/766 to https://conbench.ursa.dev/ , I checked the logs of that server. There were a number of tracebacks caused by the exception in this PR's description.

When I explored the webapp myself, everything was running fine. But as usual, lots of SEO bots were crawling our server and trying to hit pages that need to get older BenchmarkResults than the pages I was exploring. All benchmark results since https://github.com/conbench/conbench/pull/582 have `change_annotations={}` instead of `change_annotations=None`, so I never noticed a problem in the pages I explored. The poor webscrapers were the only ones seeing the 5XXs.

Issue #720, opened 2023-02-09
Add integration tests for libraries


austin3dickey:
This is a placeholder issue for what may be many issues in the future.

Because this is a monorepo and we build all our libraries during CI, we can, for example, test that `benchadapt` correctly publishes results to a running Conbench server during CI. This gives us better confidence than the unit tests in that library.

More to come.

jgehrcke:
Yes! Most tests should use the client tooling we build for users. The Flask tests with the fake HTTP client use a testing approach that I have never liked. When I built API servers before I did even 'unit tests' with an _actual_ HTTP client and client tooling, ensuring conceptual isolation between _test runner_ and _tested system_. That isolation makes sure that tests focus on the right thing. We have many tests that use a combination of fake HTTP client and _direct_ database interaction -- hard to see through.

austin3dickey:
As we discussed in recent PRs/issues, `benchalerts` has many unit tests that mock Conbench API responses, and "integration" (more like end-to-end) tests that hit live production servers. Both of these methods will have issues in CI if we change Conbench API responses. Implementing this issue correctly will be much safer and less brittle.

austin3dickey:
This is becoming so important with rapid changes to the Conbench API endpoints that `benchalerts` uses, and also the changes to `benchalerts`'s usage itself. I need to set this up for `benchalerts` soon.

PR #771, opened 2023-02-23
benchalerts: implement composable pipelines

### LINKED ISSUE #770: benchalerts: allow for composable pipelines
Every project implementing `benchalerts` may want different input and output channels for their alerts. As an example, one project may want to use the `z_regression` field to send alerts to Slack, while another may want to use the `error` field to create a GitHub Check and make a comment on a pull request. The combinatorial explosion would be too many to maintain as separate functions in [workflows.py](https://github.com/conbench/conbench/blob/275617881da25ebe744d7b0502290ed44f7c3a21/benchalerts/benchalerts/workflows.py).

I propose a relatively simple (and [familiar](https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html)) framework: a user composes alerting pipelines out of common steps. Each step can be a Python class that defines a unit of behavior, with optional configuration upon instantiation. Then an `AlertPipeline` instance would wrap a list of steps, and know how to run them in order, handling any errors that may arise.

IMO, completing this issue would mean implementing whatever is necessary to replace the guts of the current `workflows.py`, and not more. We can consider new features later (like [filtering steps](https://github.com/conbench/conbench/issues/475)? [how to branch the pipeline](https://github.com/conbench/conbench/issues/756)? [more integrations](https://github.com/conbench/conbench/issues/769)?). 

I think completing this issue sooner will make it much easier to onboard future (picky) projects and add new features like PR comments next.

austin3dickey:
Fixes #770. See that issue for further details. Basically, this PR enables `benchalerts` workflows to be a lot more composable, by using pipelines of interchangeable steps.

Sorry this PR got so large. It is almost entirely just moving code around. The net number of lines of code is just +51. 😄 

I started this PR by designing my ideal interface. That's what exists in the new README. Then I moved code around to match that interface. Here's a summary of what I did:

- Deleted `workflows.py`
- Added `alert_pipline.py`. This is pretty much the only new code.
- Moved the dataclasses that were in `talk_to_conbench.py` to `conbench_dataclasses.py`
- Moved the rest of the code in `talk_to_conbench.py` to a new `GetConbenchZComparisonStep`
- Moved the formatting helpers that were in `parse_conbench.py` to `message_formatting.py`
- Moved the rest of the code in `parse_conbench.py` to a new set of steps, `GitHubCheckStep` and `GitHubStatusStep`
- Moved tests around to match


austin3dickey:
This is ready for review.

benchalerts/benchalerts/pipeline_steps/conbench.py
```diff
@@ -0,0 +1,106 @@
...
+        The (positive) z-score threshold to send to the Conbench compare endpoint.
+        Benchmarks with a z-score more extreme than this threshold will be marked as
+        regressions or improvements in the result. Default is to use whatever Conbench
+        uses for default (at the time of writing, this is 5).
+    conbench_client
+        A ConbenchClient instance. If not given, one will be created using the standard
+        environment variables.
+    step_name
+        The name for this step. If not given, will default to this class's name.
+
```
austin3dickey, line 33: I forgot to move over the environment variable docs. Will do in the next commit.

benchalerts/benchalerts/alert_pipeline.py
```diff
@@ -0,0 +1,95 @@
...
+        try:
+            for step in self.steps:
+                step_outputs[step.step_name] = step.run_step(
+                    previous_outputs=step_outputs
+                )
+
+        except Exception as exc:
+            for error_handler in self.error_handlers:
+                error_handler.handle_error(exc=exc, traceback=format_exc())
+            raise
```
jgehrcke, line 95: Did you explore exception chaining? Would that be of relevance/value here?

benchalerts/benchalerts/conbench_dataclasses.py
```diff
@@ -0,0 +1,254 @@
...
+    """Track and organize specific info about a comparison between a contender run and
+    its baseline run.
+
+    Parameters
+    ----------
+    contender_info
+        The dict returned from Conbench when hitting /runs/{contender_run_id}. Contains
+        info about the run's ID, commit, errors, links, etc.
+    baseline_info
+        The dict returned from Conbench when hitting /runs/{baseline_run_id}, if a
```
jgehrcke, line 34: Once again strong motivator for introducing proper types for these objects, so that in my IDE can make use of rich type information.

benchalerts/tests/unit_tests/test_pipeline_steps/test_github.py
```diff
@@ -0,0 +1,116 @@
...
+        exc = e
+        traceback = format_exc()
+
+    handler = GitHubStatusErrorHandler(
+        commit_sha="abc",
+        github_client=GitHubRepoClient(repo="some/repo", adapter=MockAdapter()),
+        build_url="https://google.com",
+    )
+    handler.handle_error(exc=exc, traceback=traceback)
+    assert True
```
jgehrcke: `assert True` appears to be a noop :)

benchalerts/tests/unit_tests/test_pipeline_steps/test_github.py
```diff
@@ -0,0 +1,116 @@
...
+@pytest.mark.parametrize("github_auth", ["pat", "app"], indirect=True)
+def test_GitHubStatusErrorHandler(github_auth: str):
+    try:
+        1 / 0
+    except Exception as e:
+        exc = e
+        traceback = format_exc()
+
+    handler = GitHubStatusErrorHandler(
+        commit_sha="abc",
```
jgehrcke: talked about terminology before -- would this be an opportunity to do less of 'sha' and more of 'hash'?

benchalerts/tests/unit_tests/test_pipeline_steps/test_github.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+    handler.handle_error(exc=exc, traceback=traceback)
+    check_posted_markdown(caplog, [("summary_builderror", "details_builderror")])
+
+
+@pytest.mark.parametrize("github_auth", ["pat", "app"], indirect=True)
+def test_GitHubStatusErrorHandler(github_auth: str):
+    try:
+        1 / 0
+    except Exception as e:
```
jgehrcke, line 106: :) but please give me infinityyyyy! :infinity:

benchalerts/benchalerts/alert_pipeline.py
austin3dickey: Since this code will be run during the CI build, I thought it might be nice to raise the same exception (without chaining) that would have been raised if there weren't any error handling. This behavior is mostly meant as an additional layer of notification beyond the failed CI build—we might want to post a message to a specific slack channel or update the GitHub Check or something before we exit.

Chaining might be valuable if we wanted to handle any errors in the handler code itself. But I think that might be too complicated for now. Do you see any value in chaining that I'm missing?

benchalerts/benchalerts/conbench_dataclasses.py
austin3dickey: +1! That's one reason these dataclasses are helpful, so we can have typed, useful attributes on top of the raw responses.

benchalerts/tests/unit_tests/test_pipeline_steps/test_github.py
austin3dickey: haha, true. I thought it would be nice to have an explicit "assert we got to this point in the code." But that is pretty redundant. I wonder if this pattern is stuck in my head from writing R tests a long time ago.

benchalerts/benchalerts/alert_pipeline.py
jgehrcke: > Do you see any value in chaining that I'm missing?

No, but I also took the really easy way out in the sense that wanted to see what your thoughts are before thinking too much myself :-). Completely happy with that answer for now!! Thanks.

benchalerts/tests/unit_tests/test_pipeline_steps/test_github.py
austin3dickey: Yeah, that's pretty interesting. I think this PR is touching every use of `sha` in this package, so we could just go ahead and change it in this PR. I don't think it would be too confusing for conbench+benchalerts users; it's obvious what `commit_hash` means even if Conbench uses a different term.

benchalerts/tests/unit_tests/test_pipeline_steps/test_github.py
austin3dickey: This is my favorite way to raise an error in Python. Even better than `assert False`! `ZeroDivisionError` is such a specific exception class and I love it

benchalerts/tests/unit_tests/test_pipeline_steps/test_github.py
austin3dickey: oh, here's one potential clue for why `sha` was chosen in the past: the [GitHub REST API](https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#get-a-commit) also refers to the commit's ID/hash as "sha" in the response body and in query parameters. Interesting!

I changed everything in `benchalerts` to refer to `hash` because IMO it's still unambiguous.

benchalerts/benchalerts/alert_pipeline.py
```diff
@@ -0,0 +1,97 @@
...
+
+        Parameters
+        ----------
+        previous_outputs
+            A dict of previous steps' outputs in the pipeline, keyed by step name.
+
+        Returns
+        -------
+        Any
+            Anything (sent to subsequent steps via the ``previous_outputs`` dict).
```
alistaire47, line 35: So, um, you've written a DAG structure here (though execution will still be linear). Like this DAG from [the Wikipedia article](https://en.wikipedia.org/wiki/Directed_acyclic_graph)

wiki dag

can be written as

```python
class A(AlertPipelineStep):
    def run_step(self, previous_outputs: Dict[str, Any]) -> Any:
        return "a"

class B(AlertPipelineStep):
    def run_step(self, previous_outputs: Dict[str, Any]) -> Any:
        assert previous_outputs.get("A")
        return "b"

class C(AlertPipelineStep):
    def run_step(self, previous_outputs: Dict[str, Any]) -> Any:
        assert previous_outputs.get("A")
        return "c"

class D(AlertPipelineStep):
    def run_step(self, previous_outputs: Dict[str, Any]) -> Any:
        assert previous_outputs.get("A")
        assert previous_outputs.get("B")
        assert previous_outputs.get("C")
        return "d"

class E(AlertPipelineStep):
    def run_step(self, previous_outputs: Dict[str, Any]) -> Any:
        assert previous_outputs.get("A")
        assert previous_outputs.get("C")
        assert previous_outputs.get("D")
        return "e"

dag = AlertPipeline(steps=[A(), B(), C(), D(), E()])
```

A true pipeline would only pass the results of the previous step, so you need some contract of what inputs and outputs are (like dplyr's contract that inputs and outputs will always be dataframes). There's no obvious candidate for that datastructure contract here, so this is probably more sensible, but let's maybe try not to write complicated DAGs with it even if we can. And maybe find a way to make it obvious which steps depend on which.

benchalerts/benchalerts/conbench_dataclasses.py
```diff
@@ -0,0 +1,254 @@
...
+    """Track and organize specific info about one BenchmarkResult."""
+
+    name: str
+    link: str
+    has_error: bool
+    has_z_regression: bool
+    run_id: str
+    run_reason: str
+    run_time: str
+    run_link: str
```
alistaire47, line 20: We should integration test this; I suspect we're going to mess with that structure at some point (maybe soon). Or make it a habit of popping off elements from the dict that don't correspond to attributes when instantiating from a dict so additive changes don't break anything.

alistaire47: ...also from this name I'd assume this is `result.info`. Maybe there's something more descriptive than `info`?

benchalerts/benchalerts/alert_pipeline.py
austin3dickey: Yes, you're right that this can represent DAGs. This was a change from an earlier idea I had for this PR. Just to clarify, are you suggesting we shouldn't use the term "pipeline" to describe this concept?

I agree people should be careful about complexity, but if they'd like to notify a GitHub PR and a slack channel (with different logic for both) we might need to handle that sort of complexity. But I totally hear your feedback that someone should be able to glance at an instance of one of these and know exactly what's going on.

benchalerts/benchalerts/conbench_dataclasses.py
austin3dickey: Yes! #720 

>Maybe there's something more descriptive than info?

Good callout that `info` is a field of benchmark results already. Honestly I think that's the one that needs to be clarified, not this one 😉 Since this is an internal-only data structure I'm not too worried.

benchalerts/benchalerts/alert_pipeline.py
alistaire47: I don't care tremendously what we call it, but I do think it's important with this structure that it's very obvious what dependencies on previous steps are.

Issue #786, opened 2023-03-01
api: GET /api/benchmarks/ -- improve specification, make it responsive (less DoSsy)


jgehrcke:
Currently documented as "Get a list of benchmarks." -- that is confusing me still today when I read that. 

This seems to fetch benchmark _results_, in my mental model.

In the API spec, the schema of the response body should be well specified. Probably a list of BenchmarkResult objects.

What's supposed to happen when I GET /api/benchmarks/ without specifying a filter? I believe this is accidentally something like a rather complete database dump by now; not sustainable. 

Indeed ,`$ curl -v https://conbench.ursa.dev/api/benchmarks/` returns a 504 Gateway Timeout.

It appears that in the foundation of the API we did not really consider pagination so far.

Maybe we want / have to to require a filter here? What are current consumers of this endpoint; and what feature set do they really need?

jgehrcke:
We're using UUIDs as primary keys in the Conbench database, for e.g. storing benchmark results: 
https://github.com/conbench/conbench/blob/61977d0d65a76e75a998f96fda77852dabf66048/conbench/entities/benchmark_result.py#L32

A disadvantage of this: can't order by primary key to get the insert order. "get the N most recently inserted benchmark results, fast" becomes more challenging than it should be. Sorting by timestamp (user-given property) requires reading through all rows.

"get the last 500 benchmark results sorted by user-given timestamp" seems to be the undocumented default behavior of GET /api/benchmarks/:
https://github.com/conbench/conbench/blob/61977d0d65a76e75a998f96fda77852dabf66048/conbench/api/benchmarks.py#L153

This does not respond within 60 seconds on `conbench.ursa.dev`.

jgehrcke:
When ordering by `timestamp` (which we already do for most of /api/benchmarks/), we probably want to create an index on that column. We don't have this yet.

alistaire47:
@boshek might have some useful feedback here; there are some connections here to #746.

Pagination here seems obvious (assuming we can sort 😅). As this is really a results endpoint anyway, maybe we should recreate it paginated as `/results` or `/benchmark-results` and deprecate this?

Filters would be really nice too, though we'd need to do a little legwork to figure out what users can filter on...probably things like `tags["name"]`, case (the rest of `tags`), hardware, context, etc. Maybe it's time to define "benchmark", or concrete options in terms of uniqueness one can mean?

boshek:
[...]

Issue #519, opened 2022-12-07
Provide a `make` target for generating a database migration


austin3dickey:
Sometimes, developers need to generate and edit an `alembic` database migration file locally as part of their changes. In order to do this, there needs to be a local running database that's up-to-date in migrations. However, if the file is generated inside a local running container, it won't be picked up by git due to #437. So (perhaps similarly to `make rebuild-expected-api-docs`) it would be nice to have a target that can do this for you using `docker compose`.

jgehrcke:
Yes!

jgehrcke:
recipe:

- run DB in container (background): `docker compose run --detach db`
- have alembic go through all migrations (do not let app create tables): `CREATE_ALL_TABLES=false docker compose run -e CREATE_ALL_TABLES app alembic upgrade head`
- have alembic create a new migration: `CREATE_ALL_TABLES=false docker compose run -e CREATE_ALL_TABLES app alembic revision --autogenerate -m "new"`

to create on host fs:
```
$ CREATE_ALL_TABLES=false docker compose -f docker-compose.yml -f docker-compose.dev.yml run -e CREATE_ALL_TABLES app alembic revision --autogenerate -m "add_bmres_timestamp_index"
```

jgehrcke:
I think we can close this as 'done'.

PR #788, opened 2023-03-01
db schema: add index for BenchmarkResult.timestamp

jgehrcke:
A bit of a follow-up after https://github.com/conbench/conbench/issues/786.

@austin3dickey can you review this carefully? Thank you! :)

The method I used for generating the migration is shown in https://github.com/conbench/conbench/issues/519.

I am putting this PR up in a rush from a train; haven't looked at the code much and didn't think much about it yet. :)


conbench/api/benchmarks.py
```diff
@@ -34,7 +34,7 @@ def _get(self, benchmark_id):
     def get(self, benchmark_id):
         """
         ---
-        description: Get a benchmark.
+        description: Get a benchmark result.
```
austin3dickey, line 37: Looks like you'll have to `make rebuild-expected-api-docs`.

migrations/versions/fe3b1041ed15_add_bmres_timestamp_index.py
```diff
@@ -0,0 +1,28 @@
...
+Revises: 21519d1e3ddb
+Create Date: 2023-03-01 15:41:25.181769
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'fe3b1041ed15'
```
austin3dickey: Ah, classic, I think you'll have to `make lint` too :)

migrations/versions/fe3b1041ed15_add_bmres_timestamp_index.py
```diff
@@ -0,0 +1,28 @@
...
+down_revision = '21519d1e3ddb'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_index('benchmark_result_timestamp_index', 'benchmark_result', ['timestamp'], unique=False)
```
austin3dickey: So the main problem I foresee is that the Arrow `benchmark_result` table has 30 million rows. Creating the index will probably take a while, and will lock out writes in the meantime. This will probably cause a few benchmark builds to fail with 500s from Conbench when trying to POST.

There is the option to use `postgresql_concurrently=True` here, which will create the index concurrently. That will allow for no locks but "takes significantly longer to complete"[[docs](https://www.postgresql.org/docs/current/sql-createindex.html)]. I have no idea what this order of magnitude of time will be. [This SO answer](https://dba.stackexchange.com/a/249784) says that we'll be able to monitor the progress at least.

austin3dickey (commented):
Looks good! If you're willing to kick off and monitor this deployment, I'm happy for you to merge whenever.

jgehrcke:
```
Run make lint-ci
flake8
./migrations/versions/fe3b1041ed15_add_bmres_timestamp_index.py:9:1: F401 'sqlalchemy as sa' imported but unused
make: *** [Makefile:92: lint-ci] Error 1
```
The _auto-generated_ Python module does not pass linting. That may be considered a linting method problem. I will simply patch the module in this case :shrug:, but wanted to drop a note that this current linting isn't necessarily the gold standard yet.

migrations/versions/fe3b1041ed15_add_bmres_timestamp_index.py
jgehrcke: Thanks for the thought!

I have pushed a commit to instruct postgres to create the index concurrently. In our case here I don't think it will take much longer than the non-concurrent build because I think in our case the mutation transactions are small / short. The concurrent build can theoretically take foreverish because

>  must wait for all existing transactions that could potentially modify or use the index to terminate

This becomes a challenge in DBs with continuous write pressure which I assume/suspect we do not have (the DB I hope spends the majority of its lifetime _not_ writing).

-----

The rest of my comment is not too relevant anymore for this PR because we now try the concurrent index build. But because timescale considerations are always interesting to discuss and a powerful tool to make decisions, I'd love to continue the thought experiment:

> will probably take a while, and will lock out writes in the meantime.

It would be cool if we had an idea about the ballpark. One minute? 10 minutes? 1 hour?

> with 500s from Conbench when trying to POST.

We should have a pretty clear idea about how long recommended client tooling (tooling that _we_ build) attempts to retry, before it drops data and moves on. In similar systems I have always advocated for and also built an explicit, tested design choice: "we want to survive a micro outage / maintenance window of N minutes length; that is our promise". Such a deliberate goal makes it easy to develop/test, and it also makes it easy to communicate guarantees to users.

Given knowledge about those timescales we could make a rather credible estimation: would the non-concurrent index build lead to 'data loss' or not?

For example, if the index build is expected to take ~10 minutes, and client tooling is known to retry for up to one hour, we would have a pretty clear perspective.

jgehrcke:
The way Alembic tries to perform the migration doesn't work with concurrent index creation:
```
    self.dialect.do_execute(
  File "/usr/local/lib/python3.11/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.InternalError: (psycopg2.errors.ActiveSqlTransaction) CREATE INDEX CONCURRENTLY cannot run inside a transaction block

[SQL: CREATE INDEX CONCURRENTLY benchmark_result_timestamp_index ON benchmark_result (timestamp)]
```
The error message is great.

Mike showed an approach to solve this here:
https://bitbucket.org/zzzeek/sqlalchemy_old/issues/3887/index-with-postgresql_concurrently-true

> you need to set psycopg2 AUTOCOMMIT so there's no transaction block

But how can one do this in the context of Alembic? 

This issue is about that very question: https://github.com/sqlalchemy/alembic/issues/123

That linked ticket motivated feature work in Alembic, see this commit: https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/1463/

Indeed, now I found the feature documentation: https://alembic.sqlalchemy.org/en/latest/api/runtime.html#alembic.runtime.migration.MigrationContext.autocommit_block


Here is how to use the `autocommit_block` (it's a context manager):

```python
def upgrade():
    with op.get_context().autocommit_block():
        op.execute("ALTER TYPE mood ADD VALUE 'soso'")
```

documented with 

> This special directive is intended to support the occasional database DDL or system operation that specifically has to be run outside of any kind of transaction block. The PostgreSQL database platform is the most common target for this style of operation, as many of its DDL operations must be run outside of transaction blocks, even though the database overall supports transactional DDL.

Once again huge thanks to @zzzeek for wonderfully clear docs.

jgehrcke:
Landing this now; hoping that index creation will after all be smooth, and result in more advantages than disadvantages.

jgehrcke:
A quick indirect test for this was:
```
time curl https://conbench.ursa.dev/api/benchmarks/
```
Before the change, this timed out after 1 minutes with a 504 Gateway Timeout. After the change this returned within ~1 second. :rocket:

Issue #799, opened 2023-03-06
history concept: add types for individual samples in code, do not go through a serialization/deserialization cycle


jgehrcke:
I've been working in this part of the code every now and then, and each time a slowdown was that the 'history' argument to the `time_series_plot()` function is intransparent:
https://github.com/conbench/conbench/blob/aa95b6d52ff6cb6697b7ab8d04c43d6cf3612860/conbench/app/_plots.py#L416

When following the code path, one does find eventually that `history` is the _deserialized_ API response from this endpoint: https://github.com/conbench/conbench/blob/aa95b6d52ff6cb6697b7ab8d04c43d6cf3612860/conbench/app/_plots.py#L49

That is, it is a list of dictionaries.

Those dictionaries are _serialized_ into JSON here: https://github.com/conbench/conbench/blob/aa95b6d52ff6cb6697b7ab8d04c43d6cf3612860/conbench/api/history.py#L41

The dictionary construction (before JSON serialization) happens here: https://github.com/conbench/conbench/blob/aa95b6d52ff6cb6697b7ab8d04c43d6cf3612860/conbench/entities/history.py#L73

The serialization/deserialization cycle not only does needless work, but it also does type dilution; it adds one more way to communicate essentially the same information within the conbench webapp code base.

Action items:
- remove the serialization/deserialization indirection.
- when querying the history from the database, define a nice type for the individual sample, something like `TypeHistorySample`. Potentially having attributes that themselves deserve custom types. Then directly pass an argument of `list[TypeHistorySample]` around, after all ending up being passed to `time_series_plot()`.

jgehrcke:
Done via #811.

Issue #803, opened 2023-03-06
try to make sense of `complete_iterations` in benchmark_result.py


jgehrcke:
There is difficult-to-read behavior with respect to "failed iterations". That behavior is not really specified via text.

It's unclear if this default is important:
https://github.com/conbench/conbench/blob/aa95b6d52ff6cb6697b7ab8d04c43d6cf3612860/conbench/entities/benchmark_result.py#L75

It's unclear if this really is a good condition:
https://github.com/conbench/conbench/blob/aa95b6d52ff6cb6697b7ab8d04c43d6cf3612860/conbench/entities/benchmark_result.py#L89

(maybe we should always calculate those when at least N samples are there -- regardless of one or more missing data points)

This one seems like it might contain a bug, considering the default `False` from above:
```python
        # If there was no explicit error *and* the iterations aren't complete, we should add an error
        if (
            benchmark_result_data.get("error", None) is None
            and complete_iterations is False
        ):
```
([here](https://github.com/conbench/conbench/blob/aa95b6d52ff6cb6697b7ab8d04c43d6cf3612860/conbench/entities/benchmark_result.py#L122)).

It's certainly unclear how this relates to an emtpy result list provided (should add this to specification: does at least one non-null value need to be provided or not?)

This is an interesting noop:
```
        if "error" in data or not complete_iterations:
            return benchmark_result

        return benchmark_result
```
([here](https://github.com/conbench/conbench/blob/aa95b6d52ff6cb6697b7ab8d04c43d6cf3612860/conbench/entities/benchmark_result.py#L200)).

I've been reading code in this function many times. The meta point of this ticket is: it really takes too long to read this code :). We'll get there. I think we should review the specification, fill specification gaps as they become apparent, and then rewrite this code segment.

jonkeane:
Some history that is relevant here: #416 which resolved #395 

There's definitely more complexity here than there should be, but the specification | behavior of benchmark results with respect to completeness + iterations is:

A benchmark result should ideally have multiple iterations (>3, even!), and there is data returned for each of those iterations.

If there are errors during the running of benchmarks, we try and collect as much information as possible. There are a few ways this could happen:

[...]

Issue #802, opened 2023-03-06
stddev: set to NaN instead of 0 when not applicable


jgehrcke:
Here, when the standard deviation is not calculated as of too-little-datapoints, we set it to 0:
https://github.com/conbench/conbench/blob/aa95b6d52ff6cb6697b7ab8d04c43d6cf3612860/conbench/entities/benchmark_result.py#L95

I think we should set it to NaN instead, to properly distinguish it from zero-ish stddevs.

Note: it's mathematically sound to calculate a standard deviation for a sample size of 2 -- but in practice we may only want to do so given a larger sample size.

jonkeane:
Yeah, agreed, we should use `NaN` here if we can

jgehrcke:
> Thanks for the feedback! I'd say let's build "calculate these statistics only if three or more samples have been provided" and then adjust consumers where necessary.

https://github.com/conbench/conbench/pull/811#discussion_r1129144143

jgehrcke:
This was done in https://github.com/conbench/conbench/pull/1127.

PR #811, opened 2023-03-07
app: introduce HistorySample type, change api/history output schema

jgehrcke:
This is for https://github.com/conbench/conbench/issues/799.

This took a couple of hours on Sunday, and another few hours yesterday to work towards passing tests (I say this because this was way slower than I hoped, and a lot of this has to do with how challenging it is to navigate existing code base and tests -- a patch like this one here is aiming at improving the situation).

I'd love to have review and we can (and probably will) change things again. But I'd also love to land this ~now so that I can move on to other tasks. Letting this sit will result in annoying merge conflicts.

I would still want to change a lot compared to the current state of this patch. Thoughts I'd love to follow:

- `HistorySample` might not be the best name. Maybe `HistoryItem`? Each item is a benchmark result, and as we know each benchmark result can have "multiple samples", from iterations. It makes sense to slowly converge towards established terms on this front.

- I picked `dataclass` for defining the type `HistorySample`. That's nice already. However, during dev I noticed that per-attribute docstrings would be nice (those show upon mouse-over in the IDE, and this is useful, in addition to type information). Maybe pydantic would provide that?

- Adding a better type for arg for _source() for the single-sample points. Now I opted for creatomg dummy objects that contain a lot of information that's not consumed. That is not nice, but I wanted to stop working on this patch for now. 

- For the individual HistorySample I have added a new attribute `zscorestats` that carries the details derived by the "recent history z-score regression detection method". I think fundamentally this introduces a nice concept, especially in view of the future state where we might have more than one such method. It's however also clear to me that at least the name `zscorestats` may not be ideal, and maybe also the entire structure. So: I think this is a clear improvement, but at this point I actually love that we did _not_ specify the history endpoint output too much (https://github.com/conbench/conbench/issues/589). This PR changes the output format to what (I think) the better, but I also anticipate we will change it at least one more time before we want to start settling on the schema.


conbench/api/_examples.py
```diff
@@ -354,24 +354,31 @@ def _api_context_entity(context_id, links=True):
...
+            "zscorestats": {
+                "begins_distribution_change": False,
+                "residual": 0.0,
+                "rolling_mean": 0.036369,
+                "rolling_mean_excluding_this_commit": 0.036369,
+                "rolling_stddev": 0.0,
+                "segment_id": 0.0,
+            },
+        }
+    ]
```
jgehrcke, line 381: new: always emit JSON array at the outest level, even if it's just one item in it (easier for clients to integrate)

conbench/api/benchmarks.py
```diff
@@ -182,6 +182,8 @@ def post(self):
         tags:
           - Benchmarks
         """
+        # Here it sould be easy to make `data` have a precise type (that mypy
+        # can use) based on the schema that we validate against.
```
jgehrcke, line 186: this comment is about https://github.com/conbench/conbench/issues/796

conbench/api/history.py
```diff
@@ -37,8 +26,19 @@ def get(self, benchmark_id):
         tags:
           - History
         """
-        history = self._get(benchmark_id)
-        return self.serializer.many.dump(history)
+        # TODO: think about the case where samples if of zero length. Can this
+        # happen? If it can happen: which response would we want to emit to the
+        # HTTP client? An empty array, or something more convenient?
```
jgehrcke, line 31: @austin3dickey when you think about specifying the history api endpoint: zero length result allowed or not?

conbench/api/history.py
```diff
@@ -37,8 +26,19 @@ def get(self, benchmark_id):
...
+        try:
+            samples = get_history_for_benchmark(benchmark_result_id=benchmark_id)
```
jgehrcke, line 33: In this PR I was sneaking in a clarification from "benchmark ID" towards "benchmark result ID". 

It should be an ongoing effort to make the code base clearer in that regard.

conbench/api/history.py
```diff
@@ -37,8 +26,19 @@ def get(self, benchmark_id):
...
+        except NotFound:
+            self.abort_404_not_found()
+
+        # if performance is a concern then https://pypi.org/project/orjson/
+        # promises to be among the fastest for serializing python dataclass
+        # instances into JSON. Note: wrap this into an array if there is just 1
+        # sample, for consistency (clients can expect an array).
+        return f.jsonify([s._dict_for_api_json() for s in samples])
```
jgehrcke, line 41: Now, the IDE knows what `s` is:
image

conbench/app/_plots.py
```diff
@@ -35,22 +38,19 @@ def get_outliers(self, benchmarks):
         return outliers, outlier_ids, outlier_names
 
     def get_history_plot(self, benchmark, run, i=0):
-        history = self._get_history(benchmark)
-        if history:
+        samples = get_history_for_benchmark(benchmark_result_id=benchmark["id"])
```
jgehrcke, line 41: I changed the name of this function and the name of the first argument to make this more readable than `history = self._get_history(benchmark)`.

Seeing this now, I consider changing this in the future to something like

```python
history_items = get_history_for_benchmark_result(result_id)
```

conbench/app/_plots.py
```diff
@@ -35,22 +38,19 @@ def get_outliers(self, benchmarks):
         return outliers, outlier_ids, outlier_names
 
     def get_history_plot(self, benchmark, run, i=0):
-        history = self._get_history(benchmark)
-        if history:
```
jgehrcke, line 39: `if history` had unclear purpose especially not knowing what `history` even was (it was too hard to infer the intent). I changed this to `if len(samples) > 0` below. It's about details like this to make code easier to read.

conbench/app/_plots.py
```diff
@@ -35,22 +38,19 @@ def get_outliers(self, benchmarks):
...
                 )
             )
-        return None
 
-    def _get_history(self, benchmark):
-        response = self.api_get("api.history", benchmark_id=benchmark["id"])
-        if response.status_code != 200:
-            self.flash("Error getting history.")
-            return []
-        return response.json
```
jgehrcke, line 53: this removes the API serialization/deserialization indirection described in https://github.com/conbench/conbench/issues/796.

conbench/app/_plots.py
```diff
@@ -150,12 +150,18 @@ def simple_bar_plot(benchmarks, height=400, width=400, vbar_width=0.7):
     return p
 
 
-def _should_format(data, unit):
+# Using `Optional[float]` here only because of BenchmarkResult.mean being
+# nullable as of today.
+def _should_format(floats: list[Optional[float]], unit):
```
jgehrcke, line 155: this function is tough. cost of change was already too big. the amount of time I spent on this + tests probably exceeded writing a well-defined replacement from scratch.

conbench/app/_plots.py
```diff
@@ -186,21 +192,25 @@ def _source(
...
-    key = "distribution_mean" if distribution_mean else "mean"
-
     unit_fmt = formatter_for_unit(unit)
 
     # TODO: These commit message prefixes end up in the on-hover tooltip.
     # Change this to use short commit hashes. The long commit message prefix
     # does not unambiguously specify the commit. Ideally link to the commit, in
     # the tooltip?
-    commit_messages = [d["message"] for d in data]
+    commit_messages = [s.commit_msg for s in samples]
```
jgehrcke, line 201: "message" became "commit_msg". Tiny improvements like this make code more readable.

conbench/app/_plots.py
```diff
@@ -209,25 +219,35 @@ def _source(
...
+            points.append(float(x.zscorestats.rolling_mean) - alert)
             values_with_unit.append(unit_fmt(points[-1], unit))
 
     elif alert_max:
-        for x in data:
-            alert = 5 * float(x["distribution_stdev"])
-            points.append(float(x["distribution_mean"]) + alert)
+        for x in samples:
+            alert = 5 * float(x.zscorestats.rolling_stddev)
+            points.append(float(x.zscorestats.rolling_mean) + alert)
```
jgehrcke, line 230: when replacing these I carefully looked at https://github.com/conbench/conbench/blob/aa95b6d52ff6cb6697b7ab8d04c43d6cf3612860/conbench/entities/history.py#L44
I hope I didn't make a mistake here, pls review @austin3dickey 

```python
        return {
            "benchmark_id": history.id,
            "case_id": history.case_id,
            "context_id": history.context_id,
            "mean": _to_float(history.mean),
            "data": data,
            "times": times,
            "unit": history.unit,
            "begins_distribution_change": history.begins_distribution_change,
            "hardware_hash": history.hash,
            "sha": history.sha,
            "repository": history.repository,
            # Note(JP): this is the commit message
            "message": history.message,
            # This is the Commit timestamp. Expose Result timestamp, too?
            "timestamp": history.timestamp.isoformat(),
            "run_name": history.name,
            "distribution_mean": _to_float(history.rolling_mean),
            "distribution_stdev": _to_float(history.rolling_stddev) or 0.0,
        }
```

conbench/app/_plots.py
```diff
@@ -244,20 +264,27 @@ def _source(
         # List of (truncated) commit messages.
         "commit_messages": commit_messages,
         # List of short commit hashes with hashtag prefix
-        "commit_hashes_short": ["#" + d["sha"][:8] for d in data],
-        "relative_benchmark_urls": [f'/benchmarks/{d["benchmark_id"]}' for d in data],
+        "commit_hashes_short": ["#" + s.commit_hash[:8] for s in samples],
```
jgehrcke, line 267: `d["sha"][:8]` now is `s.commit_hash[:8]` :balloon:

conbench/app/_plots.py
```diff
@@ -244,20 +264,27 @@ def _source(
...
-        for d in data:
-            samples = d["data"]
-            strings.append(", ".join(unit_fmt(s, unit) for s in samples))
+        # Note(JP): here it becomes obvious that we may need better naming to
+        # distinguish samples from samples :-). In the next line, `samples` is
+        # a list of entities that each may contain more than one per-iteration
+        # sample.
+        for s in samples:
+            # Terminology: subsample? itersample? rawsample?
+            subsamples = s.data
```
jgehrcke, line 286: naming naming naming :) we will get there

conbench/app/_plots.py
```diff
@@ -286,7 +313,7 @@ def _inspect_for_multisample(items) -> tuple[bool, Optional[int]]:
 
     # Get number of samples for each benchmark.
     try:
-        samplecounts = [len(i["data"]) for i in items]
+        samplecounts = [len(i.data) for i in items]
```
jgehrcke, line 316: The major benefit of this patch is not easily seen by looking at the github diff view. It's really things like this:

image

The IDE now knows what `i.data` is. So relieving.

conbench/app/_plots.py
```diff
@@ -413,59 +440,79 @@ def gen_js_callback_click_on_glyph_show_run_details(repo_string):
...
+def time_series_plot(
+    samples: List[HistorySample], benchmark, run, height=380, width=1100
+):
+    # log.info("Time series plot for:\n%s", json.dumps(samples, indent=2))
 
-    unit = history[0]["unit"]
-    with_dist = [h for h in history if h["distribution_mean"]]
-    formatted, axis_unit = _should_format(history, unit)
+    unit = samples[0].unit
+    with_dist = [s for s in samples if s.zscorestats.rolling_mean]
```
jgehrcke, line 449: I like that `h["distribution_mean"]` now is `s.zscorestats.rolling_mean`. As said before, we can totally change the name of `zscorestats` or the overall structure, but "distribution mean" was lacking context, which is why I was confused by it for weeks. The context really is a specific analysis method, and the existence of this method should be reflected in code.

conbench/app/_plots.py
```diff
@@ -413,59 +440,79 @@ def gen_js_callback_click_on_glyph_show_run_details(repo_string):
...
+    formatted, axis_unit = _should_format([s.mean for s in samples], unit)
     dist_change_indexes = [
-        ix for ix, result in enumerate(history) if result["begins_distribution_change"]
+        ix
+        for ix, sample in enumerate(samples)
+        if sample.zscorestats.begins_distribution_change
     ]
 
-    # Note(JP): `history` is an ordered list of dicts, each dict has a `mean`
+    # Note(JP): `samples` is an ordered list of dicts, each dict has a `mean`
```
jgehrcke, line 457: need to update the comment

conbench/app/_plots.py
```diff
@@ -413,59 +440,79 @@ def gen_js_callback_click_on_glyph_show_run_details(repo_string):
...
+                    [s.commit_timestamp.isoformat() for s in samples]
+                ),
                 dist_change_indexes,
             ),
             # TODO: best-case is not always min, e.g. when data has a unit like
-            # bandwidth.
-            y=_insert_nans([min(x["data"]) for x in history], dist_change_indexes),
+            # bandwidth. Note(JP): min() must not have None passed as arg, that
+            # is why I have changed `s.data` to only contain type float
+            # elements (including math.nan, instead of None).
```
jgehrcke, line 473: This change created quite a bit of follow-up work across various  places in the code. But I think it was worth it. Simplification.

conbench/app/_plots.py
```diff
@@ -413,59 +440,79 @@ def gen_js_callback_click_on_glyph_show_run_details(repo_string):
...
-            {
-                "mean": benchmark["stats"]["min"],
-                "message": run["commit"]["message"],
-                "timestamp": run["commit"]["timestamp"],
-                "sha": run["commit"]["sha"],
-                "benchmark_id": benchmark["id"],
-                "repository": run["commit"]["repository"],
-            }
-        ],
+        [dummy_hs_min],
```
jgehrcke, line 515: Now this is the part where I took a quick and dirty way out of the situation. `source_current_bm_mean` and `source_current_bm_min` are one-sample curves and the actually relevant input is `benchmark["stats"]["mean"]` and `benchmark["stats"]["min"]`, respectively. For these one-sample situations I believe we may want to construct the ColumnarDataSource objects directly, and not go through _source().

conbench/entities/_entity.py
```diff
@@ -20,6 +20,8 @@ class NotFound(NoResultFound):
 
 
 def generate_uuid():
+    # Consider using xid or UUID7 or something comparable so that primary
+    # key reflects insertion order.
```
jgehrcke, line 24: this is about https://github.com/conbench/conbench/issues/789

conbench/entities/benchmark_result.py
```diff
@@ -34,8 +35,12 @@ class BenchmarkResult(Base, EntityMixin):
     info_id = NotNull(s.String(50), s.ForeignKey("info.id"))
     context_id = NotNull(s.String(50), s.ForeignKey("context.id"))
     run_id = NotNull(s.Text, s.ForeignKey("run.id"))
-    data = Nullable(postgresql.ARRAY(s.Numeric), default=[])
-    times = Nullable(postgresql.ARRAY(s.Numeric), default=[])
+
+    # These can be emtpy lists. An item in the list is of type float, which
+    # includes `math.nan` as valid value.
+    data: list[float] = Nullable(postgresql.ARRAY(s.Numeric), default=[])
+    times: list[float] = Nullable(postgresql.ARRAY(s.Numeric), default=[])
```
jgehrcke, line 42: Important decisions in here. Maybe not all of them are good / correct.

The current type annotation should maybe be `Optional[list[float]]` because the Column is marked as `Nullable`. But I wonder: why would we want to store null/None _instead of_ an emtpy list? What's the semantic difference between both situations? Maybe we only need the empty list. In that case we remove the `Nullable`, and the type is correct.

tracking that topic here: https://github.com/conbench/conbench/issues/812

A helpful quote from the SQLAlchemy docs:

> [Numeric](https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.Numeric) returns Python decimal.Decimal objects by default, based on the default value of True for the [Numeric.asdecimal](https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.Numeric.params.asdecimal) parameter. If this parameter is set to False, returned values are coerced to Python float objects.

Btw, I am pretty sure that `s.Numeric` the way we use it here allows for saving `NaN` as the individual array item. However, I did not quickly find confirmation for that in the SQLAlchemy docs.

Behavior that matters:
```
>>> Decimal(None)
TypeError: conversion from NoneType to Decimal is not supported
>>> Decimal('NaN')
Decimal('NaN')
```

conbench/entities/benchmark_result.py
```diff
@@ -64,6 +69,13 @@ class BenchmarkResult(Base, EntityMixin):
     change_annotations = Nullable(postgresql.JSONB)
 
     @staticmethod
+    # We should work towards having a precise type annotation for `data`. It's
+    # the result of a (marshmallow) schema-validated JSON deserialization, and
+    # therefore the type information _should_ be available already. However,  I
+    # have not found a quick way to add a type annotation based on a
+    # marshmallow schema. Maybe this would be a strong reason to move to using
+    # pydantic -- I believe when defining a schema with pydantic, the
```
jgehrcke, line 77: this topic is tracked here: https://github.com/conbench/conbench/issues/796

conbench/entities/benchmark_result.py
```diff
@@ -72,26 +84,59 @@ def create(data):
 
         # defaults
         benchmark_result_data = {}
-        complete_iterations = False
+
+        # Note(JP): if no data was provided, should this stay `True`?
+        # We will see. https://github.com/conbench/conbench/issues/803.
+        any_errored_iteration = True
```
jgehrcke, line 90: I renamed `complete_iterations` to `any_errored_iteration` (boolean inverse). As an outsider, the new variable name is much clearer to me.

conbench/entities/benchmark_result.py
```diff
@@ -72,26 +84,59 @@ def create(data):
...
+            # deserializaiton/validation against the `BenchmarkResultCreate`
+            # schema below. This needs better naming. Most importantly, we
+            # should make use of typing information derived from the schema.
+            benchmark_result_data = data["stats"].copy()
+
+            # Note(JP): I think we may want to emit a Bad Request response when
+            # benchmark_result_data['data'] does not contain any non-null
+            # value. An empty list would then also be rejected right away. `if
+            # benchmark_result_data.get("data")` is probably supposed to catch
+            # the case where an empty list was provided.
```
jgehrcke, line 103: tracking this here: https://github.com/conbench/conbench/issues/813

conbench/entities/benchmark_result.py
```diff
@@ -72,26 +84,59 @@ def create(data):
...
 
             # calculate any missing stats if data available
-            if data["stats"].get("data"):
-                dat = [to_float(x) for x in benchmark_result_data["data"]]
```
jgehrcke, line 82: Before, it was practically impossible to understand what `dat` can become. 

I therefore added a code comment and started using the unambiguous built-in `float()`, `None`, and `math.nan`. 

The IDE still does not know what `x` can be. This will be an important future improvement.

conbench/entities/benchmark_result.py
```diff
@@ -72,26 +84,59 @@ def create(data):
...
+                # maybe we want to calculate that always if there are at least
+                # N (3?) data points.
+                any_errored_iteration = any([x is math.nan for x in dat])
+
+                if not any_errored_iteration:
+                    # Note(JP): do we really want to populate all of the below
+                    # except for stddev if the sample size is 1? In that case,
+                    # all of max, min, mean, q1 share the same value. I'd say
+                    # it makes more sense to calculate these statistics only if
+                    # three or more samples have been provided.
```
jgehrcke, line 133: related: https://github.com/conbench/conbench/issues/802

conbench/entities/benchmark_result.py
```diff
@@ -119,10 +164,10 @@ def create(data):
         if has_error:
             benchmark_result_data["error"] = data["error"]
 
-        # If there was no explicit error *and* the iterations aren't complete, we should add an error
+        # If there was no explicit error *and* at least one iteration failed aren't complete, we should add an error
         if (
             benchmark_result_data.get("error", None) is None
-            and complete_iterations is False
+            and any_errored_iteration is True
```
jgehrcke, line 170: I really don't quite know what we're doing here and why :) https://github.com/conbench/conbench/issues/803

Let's re-think specification, put things into clear text form and then adjust code accordingly. It's OK to break with existing behavior if that will be the result.

conbench/entities/benchmark_result.py
```diff
@@ -236,14 +278,38 @@ class BenchmarkResultCreate(marshmallow.Schema):
...
+                """
+                A list of benchmark results (e.g. durations, throughput). This
+                will be used as the main + only metric for regression and
+                improvement. The values should be ordered in the order the
+                iterations were executed (the first element is the first
+                iteration, the second element is the second iteration, etc.).
+                If an iteration did not complete but others did and you want to
+                send partial data, mark each iteration that didn't complete as
+                `null`."""
+            )
```
jgehrcke, line 295: This is the same text, re-wrapped.

conbench/entities/benchmark_result.py
```diff
@@ -236,14 +278,38 @@ class BenchmarkResultCreate(marshmallow.Schema):
...
+                """
+                A list of benchmark durations. If `data` is a duration measure,
+                this should be a duplicate of that object. The values should be
+                ordered in the order the iterations were executed (the first
+                element is the first iteration, the second element is the
+                second iteration, etc.). If an iteration did not complete but
+                others did and you want to send partial data, mark each
+                iteration that didn't complete as `null`.
+                """
+            )
```
jgehrcke, line 312: same text, rewrapped

conbench/entities/commit.py
```diff
@@ -566,9 +567,14 @@ def _parse_commit(commit):
...
-            # tz-aware (Zulu time, UTC).
+            # Note(JP): `commit_author["date"]` represents the time when the
+            # commit was authored (when it was originally made on the developer
+            # machine). This time never changes. git also knows the concept of
+            # the commit date which is updated every time the commit is
+            # modified; for example when rebasing or cherry-picking. We don't
+            # consider this here. `commit_author["date"]` is expected to be an
+            # ISO 8601 timestring as returned by the GitHub HTTP API and that
+            # is tz-aware (Zulu time, UTC).
             "date": util.tznaive_iso8601_to_tzaware_dt(commit_author["date"]),
```
jgehrcke, line 580: When reading `commit_author["date"]` then it's really not obvious what kind of date this might be :).  (date of birth of the author? :cake:) At least somewhere in the Conbench code base we need to clarify the two concepts of "commit timestamp" that git knows about. I opted for quickly doing that here.

conbench/entities/history.py
```diff
@@ -97,12 +87,73 @@ class HistorySerializer:
     many = _Serializer(many=True)
```
jgehrcke, line 87: I think most of this is dead code by now.

conbench/entities/history.py
```diff
@@ -97,12 +87,73 @@ class HistorySerializer:
...
+    unit: str
+    hardware_hash: str
+    repository: str
+    commit_hash: str
+    commit_msg: str
+    # tz-naive timestamp of commit authoring time (to be interpreted in UTC
+    # timezone).
+    commit_timestamp: datetime.datetime
+    run_name: str
+    zscorestats: HistorySampleZscoreStats
```
jgehrcke, line 123: This is how I started this PR; with these two dataclasses.

conbench/entities/history.py
```diff
@@ -97,12 +87,73 @@ class HistorySerializer:
...
+
+    def _dict_for_api_json(self) -> dict:
+        d = dataclasses.asdict(self)
+        # if performance is a concern then https://pypi.org/project/orjson/
+        # promises to be among the fastest for serializing python dataclass
+        # instances into JSON.
+        d["commit_timestamp"] = self.commit_timestamp.isoformat()
```
jgehrcke, line 130: timestamps/datetime objects can be represented in JSON in a variety of custom ways -- here we opt for a specific stringified notation, same as before -- can review this in the future.

conbench/entities/history.py
```diff
@@ -143,7 +194,61 @@ def get_history(case_id: str, context_id: str, hardware_hash: str, repo: str) ->
         history_df, include_current_commit_in_rolling_stats=False
     )
 
-    return list(history_df.itertuples())
+    # return list(history_df.itertuples())
+
+    samples: list[HistorySample] = []
+    # Iterate over rows of pandas dataframe; get each row as namedtuple.
+    for sample in history_df.itertuples():
```
jgehrcke, line 201: What's non-obvious from the code above is the _names_ of the namedtuple attributes. 

They are largely derived from the attribute names in the `Session.query()` call above! In there, we for example see `BenchmarkResult.id`. The resulting dataframe then has a column with name `id`.

conbench/entities/history.py
```diff
@@ -143,7 +194,61 @@ def get_history(case_id: str, context_id: str, hardware_hash: str, repo: str) ->
...
+        # Note(JP): the Commit.timestamp is nullable, i.e. not all Commit
+        # entities in the DB have a timestamp (authoring time) attached.
+        # However, in this function I believe there is an invariant that the
+        # query only returns Commits that we have this metadata for (what's the
+        # precise reason for this invariant? I think there is one). Codify this
+        # invariant with an assertion.
```
jgehrcke, line 207: @austin3dickey what's your take on this? 

Do we currently have commits in the database that have `timestamp` set to `None`?

If yes: is there a chance that one of the history items here is associated with one of those, or not (why?)?

Do we want / need the ability to keep track of a commit in the database that has _no_ authoring date?

Let's add the findings to https://github.com/conbench/conbench/issues/814.

conbench/entities/history.py
```diff
@@ -143,7 +194,61 @@ def get_history(case_id: str, context_id: str, hardware_hash: str, repo: str) ->
...
+        # potentially empty. `data` and `times` contain more than one value if
+        # this was a multi-sample benchmark.
+
+        data = []
+        if sample.data is not None:
+            data = [float(d) if d is not None else math.nan for d in sample.data]
+
+        times = []
+        if sample.times is not None:
+            times = [float(t) if t is not None else math.nan for t in sample.times]
```
jgehrcke, line 230: reading this code right now I am not even sure if `d` and `t` can be `None`. If this comes straight from the database then both `sample.data` and `sample.times` are either `None` or `postgresql.ARRAY(s.Numeric)` (see above). 

In that case, this should be fine:
```
[float(d) for d in sample.data]
```
because `d` is guaranteed to be a `decimal.Decimal()` and the following is fine:
```python
>>> float(Decimal('NaN'))
nan
```

conbench/entities/history.py
```diff
@@ -163,7 +268,7 @@ def set_z_scores(benchmark_results: List[BenchmarkResult]):
     attribute will be None on that BenchmarkResult.
     """
     # For most invocations of this function, there are very few unique run_ids among the
-    # benchmark_results. Sort them by run_id and run an optimized query for each group.
+    # benchmark_results. Sort them by run_id and run aoptimized query for each group.
```
jgehrcke, line 271: oh.

jgehrcke:
ci/lint failed with:
```
mypy conbench
conbench/entities/commit.py:120: error: Item "datetime" of "Optional[datetime]" has no attribute "label"  [union-attr]
conbench/entities/commit.py:120: error: Item "None" of "Optional[datetime]" has no attribute "label"  [union-attr]
conbench/entities/commit.py:126: error: Unsupported operand types for >= ("datetime" and "None")  [operator]
conbench/entities/commit.py:126: note: Left operand is of type "Optional[datetime]"
conbench/entities/commit.py:133: error: Item "datetime" of "Optional[datetime]" has no attribute "label"  [union-attr]
conbench/entities/commit.py:133: error: Item "None" of "Optional[datetime]" has no attribute "label"  [union-attr]
conbench/entities/commit.py:140: error: Unsupported operand types for >= ("datetime" and "None")  [operator]
conbench/entities/commit.py:140: note: Left operand is of type "Optional[datetime]"
conbench/entities/commit.py:331: error: Item "datetime" of "Optional[datetime]" has no attribute "isnot"  [union-attr]
conbench/entities/commit.py:331: error: Item "None" of "Optional[datetime]" has no attribute "isnot"  [union-attr]
conbench/entities/commit.py:334: error: Item "datetime" of "Optional[datetime]" has no attribute "desc"  [union-attr]
conbench/entities/commit.py:334: error: Item "None" of "Optional[datetime]" has no attribute "desc"  [union-attr]
conbench/entities/commit.py:360: error: Argument "until" to "get_commits_to_branch" of "GitHub" has incompatible type "Optional[datetime]"; expected "datetime"  [arg-type]
Found 11 errors in 1 file (checked 106 source files)
make: *** [Makefile:96: lint-ci] Error 1
```
working on a commit.

conbench/app/_plots.py
jgehrcke: Re-reading this I don't even know how I changed this. Did I? :) Well.

jgehrcke:
I'd love to get your eyes onto this @alistaire47 @jonkeane @austin3dickey -- one reason is of course to get feedback, but another reason is that I think we need to work on a mutual _detailed_ understanding of how challenging it currently is to work in the codebase, as of all those tiny or bigger ambiguities. I hope that we can develop a mutual understanding about the type and amount of work we have to do to _continuously_ to set ourselves up for success/sustainability/development speed. (we already do that kind of work, but maybe with too little visibility, too little explicitness). This patch and the code comments / issue comments do not really provide the big picture, but I think it's a useful set of fragments that provide an impression about the challenges we face.

jgehrcke:
Bear with me please, landing this! Super open to iterating, changing.

conbench/app/_plots.py
jonkeane: Thank you for this! This is a huge improvement IMO

conbench/app/_plots.py
jonkeane: Love this, yes. I'll think some more on zscorestats and if there's a better name (something like distribution_distance_stats maybe? zscore is nice in that it doesn't use a relatively common word like distribution, but what this is fundamentally about is taking a distribution of points and then seeing if the current thing is close to that distribution or far from that distribution). But regardless, this is much much clearer as is IMO

conbench/entities/benchmark_result.py
jonkeane: If I'm following correctly (this is about if the list can have _any_ `null` elements, yeah? Like `[0.01, null, 0.015]`? Or only cases like `[null]`?), this is ultimately to support folks uploading results even if one (or more) iterations failed. 

The docs for that are at https://github.com/conbench/conbench/blob/aa95b6d52ff6cb6697b7ab8d04c43d6cf3612860/conbench/entities/benchmark_result.py#L239 (these could and must be surface better than where they are currently, but it's what we've got given our current swagger setup).

conbench/entities/benchmark_result.py
jonkeane: Thanks for flagging things like this, makes skipping over so much easier

conbench/api/_examples.py
jonkeane: I love this change, though we might need to think about updating our clients in various places before we merge this. I can think of a handful of places we use this now within the conbench ecosystem, as well as some of the our reporting infrastructure: https://github.com/conbench/conbenchcoms for example. It shouldn't be too hard to update all of those, but we'll need to coordinate that to make sure we're ready for that to land when it does.

jonkeane: Same is true for `sha` -> `commit_hash` I love it, it's the right answer, but it will likely require some work from others (mostly thinking of @boshek but there are others too) to prepare for this change to go live

conbench/app/_plots.py
```diff
@@ -186,21 +192,25 @@ def _source(
...
 
     # The `timestamp` property corresponds to the UTC-local commit time
     # (example value: 2022-03-03T19:48:06). That is, each string is ISO 8601
     # notation w/o timezone information. Transform those into tz-aware datetime
     # objects.
-    datetimes = util.tznaive_iso8601_to_tzaware_dt([x["timestamp"] for x in data])
+    #
+    # Note that we do too much forth and back here; define `commit_timestamp`
+    # properly on `HistorySample`  (to be tzaware and in UTC), then we can
+    # simplify.
```
jonkeane, line 210: ❤️

conbench/entities/benchmark_result.py
jonkeane: I love this, absolutely we should do this (happy to have it done now if it makes what you're working on easier, or we can make a ticket to do it later). 

Part of the trouble here ends up being that we use the mean to do many of our subsequent calculations and visualizations: I suspect (like you mention in #802 we probably should "just" set this to calculate these if there are 2 or more (I agree having a nudge towards higher iterations is good, and maybe we can take that stand now, but it would be weird to document ~"you get no stats if you have 2 or less, and therefore we can't do anything with regressions (since we use the mean to calculate that, and it's no calculated). But if it's 1, then we do actually just use that value so it's fine, but you really should be sending 3" Probably best to just have 1 iteration, that's the measure that's used, and then anything more than that we calculate stats like standard

conbench/entities/benchmark_result.py
jonkeane: Cross linking: I added a bunch of text at https://github.com/conbench/conbench/issues/803#issuecomment-1458203148

conbench/entities/benchmark_result.py
```diff
@@ -236,14 +278,38 @@ class BenchmarkResultCreate(marshmallow.Schema):
         marshmallow.fields.Decimal(allow_none=True),
         required=True,
         metadata={
-            "description": "A list of benchmark results (e.g. durations, throughput). This will be used as the main + only metric for regression and improvement. The values should be ordered in the order the iterations were executed (the first element is the first iteration, the second element is the second iteration, etc.). If an iteration did not complete but others did and you want to send partial data, mark each iteration that didn't complete as `null`."
+            # Note(JP): we must specify: is it allowed to send an empty list?
+            # Is it allowed to send a list with only `null` values? Should
+            # there be the requirement that at least one non-null value must be
+            # present?
```
jonkeane, line 284: > is it allowed to send an empty list?

IMO no this is bad, we should not do this or support this. But maybe there's a usecase? I'm open to being wrong, but this seems pathological IMHO

> Is it allowed to send a list with only `null` values?

Yes, if all iterations failed, one can send `[null, null, null, null]`

> Should there be the requirement that at least one non-null value must be present?

IMO no. That should be "fine" (though of course we treat this benchmark result as a total failure then. But I could imagine a world where we have two different kinds of benchmarkResults (a successful one and a failure one) where we might split in the API if we have an all-failure benchmark. That seems overly complicated just now, but that might make it easier to see what's up?

conbench/entities/commit.py
jonkeane: Oooooph yeah that is super confusing! Thanks for the comment

jonkeane:
Some thoughts on naming: 

> HistorySample might not be the best name. Maybe HistoryItem? Each item is a benchmark result, and as we know each benchmark result can have "multiple samples", from iterations. It makes sense to slowly converge towards established terms on this front.

I like the clearness of this name, though do worry slightly that "sample" might be interpreted too literally and folks will think that it is a (random) subset of the data that they are getting, rather than all of the (relevant) recent past data. What about something like `HistoryTail`? 

> For the individual HistorySample I have added a new attribute zscorestats that carries the details derived by the "recent history z-score regression detection method". I think fundamentally this introduces a nice concept, especially in view of the future state where we might have more than one such method. It's however also clear to me that at least the name zscorestats may not be ideal, and maybe also the entire structure. So: I think this is a clear improvement, but at this point I actually love that we did not specify the history endpoint output too much (https://github.com/conbench/conbench/issues/589). This PR changes the output format to what (I think) the better, but I also anticipate we will change it at least one more time before we want to start settling on the schema.

I mentioned this above, but it's worth a larger (not in the midst of code discussion) "I'll think some more on zscorestats and if there's a better name (something like distribution_distance_stats maybe? zscore is nice in that it doesn't use a relatively common word like distribution, but what this is fundamentally about is taking a distribution of points and then seeing if the current thing is close to that distribution or far from that distribution). But regardless, this is much much clearer as is IMO"

jgehrcke:
> though we might need to think about updating our clients in various places before we merge this.

Let's please collect links here to code that consumes the history API.

Let's please list those places that we anticipate to be broken. Let's create lightweight tickets to respond. This can be sorted out quickly, if there is something to sort out. Thank you!

I searched the Conbench repository carefully before (`grep -nR 'history' | grep -v 'conbench/tests' | grep -v '\.git' | grep -v mypy | grep -v 'conbench/'`).

Searching GitHub in `voltrondata-labs` I found no result for `api/history`.

Searching in the Conbench org I find https://github.com/conbench/conbenchcoms/blob/d73b6911ef647ab69031337330140d02343aca4f/tests/testthat/resp-class/localhost/api/history/hist-id.json#L4

which seems to be already rather out-of-touch with current Conbench. Interestingly, it has the outer array despite just containing a single item:

```
    expect_type(
      the_list_history,
      "list"
    )
```

(here, the test was always out-of-sync with the API implementation). I don't see where it consumes / expects precise attribute names. Maybe it does.

> https://github.com/conbench/conbenchcoms
> Communicate with the conbench API in R

I see a separate repo. I see a repo that does not try to consume API specification via code. Since it consumes underspecified endpoints, users of this code should expect to see breakage at any time. That's the world I come from. This perspective helps to move forward with reasonable velocity. 

A world of underspecified API and loosely coupled underdocumented API clients is only glue to walk in. If I stir up enough chaos and potentially even annoyance/bad emotions with a change like this then I think I actually nudge us into the right direction.

@jonkeane @boshek I have strong opinions about how to get into a state of not breaking consumers. I am reasonably sure that I come from a software engineering culture where this was a more sacred principle than it should currently be in our Conbench universe. We cannot defend ambiguity, because it always slows down and inhibits decision-making and iteration. Let's maybe talk in person.

I made the effect of this patch prominent in the title, but should have given a clearer heads-up via Slack.

Btw, here is a ticket where we can discuss the relevance of specifying the behavior of the history API endpoint: https://github.com/conbench/conbench/issues/589

conbench/api/_examples.py
jgehrcke: Replied below: https://github.com/conbench/conbench/pull/811#issuecomment-1459771722

Most importantly, I'd appreciate if we can simply list the places where we expect breakage, and I am happy to learn about them and contribute patches. Thanks.

conbench/app/_plots.py
jgehrcke: > I'll think some more on zscorestats and if there's a better name

Thanks!

conbench/entities/benchmark_result.py
jgehrcke: > (this is about if the list can have any null elements, yeah? Like [0.01, null, 0.015]? Or only cases like [null]?),

no it's really about 

> why would we want to store null/None instead of an emtpy list? 

because the database schema as-is allows for storing null instead of a list/array. I don't think this should be allowed.

> What's the semantic difference between both situations? 

(`null` vs. `[]`).

Thanks for touching on the other topic; for posterity most of this is discussed here: https://github.com/conbench/conbench/issues/813

conbench/entities/benchmark_result.py
jgehrcke: Thanks for the feedback! I'd say let's build "calculate these statistics only if three or more samples have been provided" and then adjust consumers where necessary.

conbench/entities/benchmark_result.py
jgehrcke: Thanks for the feedback here @jonkeane 

For connecting the dots, most of this discussion is now in https://github.com/conbench/conbench/issues/813. I think we are converging to something rather useful.

Thank you for that, this is fun!

jgehrcke:
Thanks for all the feedback here. This was a rabbit hole, and it created more rabbit holes. But I think that's actually super valuable; thanks for helping me with that.

jgehrcke:
> I like the clearness of this name, though do worry slightly that "sample" might be interpreted too literally and folks will think that it is a (random) subset of the data that they are getting, rather than all of the (relevant) recent past data.

Hear you.

> What about something like HistoryTail?

Hm. We think about a time series where each item corresponds to a benchmark result. We need a good name for each individual item. "Tail" to me sounds like more than one item.

> "I'll think some more on zscorestats and if there's a better name (something like distribution_distance_stats maybe? zscore is nice in that it doesn't use a relatively common word like distribution, but what this is fundamentally about is taking a distribution of points and then seeing if the current thing is close to that distribution or far from that distribution).

Thanks for the feedback. Yeah.. I used `zscore` here precisely for this aspect of memorability/brand recognition :tm: :D

Previously when I didn't think much and felt the need to attach a name/word/sentence, then I used:

>  z-score-based method that considers the distribution of results within "recent history".

> we largely discussed this in the context of the z-score-recent-history regression detection method.

> Recent history z-score method

Will also think about this more. Let's continue here: https://github.com/conbench/conbench/issues/830

conbench/api/history.py
austin3dickey: Wouldn't we at least want to always return one sample corresponding to the given `benchmark_id`?

conbench/app/_plots.py
austin3dickey, line 155: I'm still a little worried about this 3.9 style of type hints (i.e. `list[X]` instead of `List[X]`). I know `conbench.app._plots` will probably not be imported by the `conbench` CLI, but it's hard to know for sure since we're not running tests using python 3.8 right now. See #616

conbench/app/_plots.py
austin3dickey: I think you got it right, but the `_to_float()` was doing a little extra work to ensure we returned a float or None (but not NaN). It's probably fine how you have it. `bokeh` will probably figure it out

austin3dickey: Oh, I got to the code down below where you kept `_to_float()` in the function before returning it from the API. That's great.

conbench/app/_plots.py
jgehrcke: Thanks for connecting dots. Did this once again unintentionally. I will of course try to keep this in mind, but we have to implement a structural change if we want to prevent this from happening.

conbench/api/history.py
jgehrcke: > Wouldn't we at least want to always return one sample corresponding to the given benchmark_id?

Ah :) I didn't even think that "far" lol. You are saying: if benchmark_id is known that we can emit at least one sample; in all other cases we can have decent error cases (benchmark ID not known, benchmark known but doesn't contain samples (err state?), ...); in those cases we return an error response (and do not need to show an empty array).

austin3dickey: Yes, exactly! :)

conbench/app/_plots.py
jgehrcke: https://github.com/conbench/conbench/pull/956/files#r1155354917

Issue #809, opened 2023-03-07
POST benchmark result: status code 500 response as of duplicate key value violates unique constraint "commit_index"


jgehrcke:
follow-up from #804.

```
[2023-03-04 05:32:02,666] ERROR in app: Exception on /api/benchmarks/ [POST]
...
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "commit_index"
...
DETAIL:  Key (sha, repository)=(6b6cdaf7dcd03779193b8891349cdaa8d9fc56f4, https://github.com/<redacted>) already exists.
```

jgehrcke:
more context:
```
File "/app/conbench/api/benchmarks.py", line 186, in post
benchmark_result = BenchmarkResult.create(data)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/app/conbench/entities/benchmark_result.py", line 168, in create
Run.create(
File "/app/conbench/entities/run.py", line 98, in create
commit = Commit.create_github_context(sha, repository, gh_commit_dict)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/app/conbench/entities/commit.py", line 176, in create_github_context
return Commit.create(
^^^^^^^^^^^^^^
File "/app/conbench/entities/_entity.py", line 93, in create
```

![image](https://user-images.githubusercontent.com/265630/223394448-eb526c3b-e037-4b94-8846-d59591858dea.png)

It's so hard to properly copy/paste from the datadog log viewer

jgehrcke:
Will need to go back to specification, but it seems like we simply want to silently ignore when the commit is already in the database. It's a bit surprising that we don't do that so far. Maybe we introduced this as a regression, recently? Let's cover this with tests.

jonkeane:
Yeah, if I'm reading this code correctly (which it doesn't look too unobvious, honestly!) line 81+83 should catch if the commit already exists in the DB and then skip over trying to create a new one, but clearly that didn't work here.


https://github.com/conbench/conbench/blob/408cba4db5ac914212b0be9cfd1bd0ceeccc7ff0/conbench/entities/run.py#L80-L113

jgehrcke:
> line 81+83 should catch if the commit already exists in the DB and then skip over trying to create a new one

Ah, a typical check-then-try race condition :)

[...]

PR #816, opened 2023-03-07
run.py: expect IntegrityError during Commit insert

jgehrcke:
See commit msg. Addresses #809.

### DIFF
--- a/conbench/entities/run.py
+++ b/conbench/entities/run.py
@@ -1,4 +1,5 @@
 import logging
+import time
 from datetime import timezone
 from typing import Optional
 
@@ -61,7 +62,7 @@ class Run(Base, EntityMixin):
 
     @staticmethod
     def create(data):
-        # create if not exists
+        # create Hardware entity it it does not yet exist in database.
         hardware_type, field_name = (
             (Machine, "machine_info")
             if "machine_info" in data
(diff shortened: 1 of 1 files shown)


conbench/entities/run.py
```diff
@@ -209,6 +179,117 @@ def get_baseline_id(self):
...
+                    "Could not backfill default branch commits. Ignoring error "
+                    "during backfill_default_branch_commits():  %s",
+                    exc,
+                )
+
+        # If at least a commit hash and/or repository are provided, insert
+        # that. Note(JP): wouldn't it be better if we hat the constraint that a
+        # commit entity in the database has _at least_ repository specifier and
+        # commit hash? What is the value of a commit object that only refers to
+        # a repository (and not to a commit hash)?
```
jgehrcke: @jonkeane @austin3dickey would love to have quick opinion, thanks!
let's put findings here: https://github.com/conbench/conbench/issues/817

conbench/entities/run.py
```diff
@@ -209,6 +179,117 @@ def get_baseline_id(self):
...
+                # the HTTP request processing (we're right now in the middle of
+                # processing an HTTP request with new benchmark run data).
+                log.info(
+                    "Could not backfill default branch commits. Ignoring error "
+                    "during backfill_default_branch_commits():  %s",
+                    exc,
+                )
+
+        # If at least a commit hash and/or repository are provided, insert
+        # that. Note(JP): wouldn't it be better if we hat the constraint that a
```
jonkeane:
```suggestion
        # that. Note(JP): wouldn't it be better if we had the constraint that a
```

conbench/entities/run.py
jonkeane: I'm fine with that and can't think of bad things that would happen if it didn't. The original motivation for the looseness here was to also allow us to store results on other softwares (some where we might not have a specific hash that we know we're running). We haven't done that so far, and I suspect if (when?) we do, we'll want to create a separate table for "external benchmark" that has different constraints so we can deal with that there (or we can loosen this back if we get there.

conbench/entities/run.py
```diff
@@ -209,6 +179,117 @@ def get_baseline_id(self):
...
+    commits) from GitHub HTTP API if possible. Exceptions during this process
+    are logged and otherwise swallowed.
+
+    Return Commit.id (DB primary key) of existing Commit entity or of newly
+    created one. Expect database collision upon insert (in this case the ID for
+    the existing commit entity is returned).
+
+    Has slightly ~unpredictable run duration as of interaction with GitHub HTTP
+    API.
+    """
```
jgehrcke, line 200: That's roughly the level of docstringing we should have in the code base. Does not need to be perfect, but between fellow developers we have to communicate the intent, in particular about error handling.

conbench/entities/run.py
jgehrcke: After #817 I have now removed the comment and updated the code to be clear about the fact that here, both `sha` and `repository` are non-None.

jgehrcke:
```
        elif sha:
            # As of input schema validation this means that both, commit has
            # and repository specifier are set. Also the database schema as of
            # the time of writing this comment requires both commit hash and
            # repo specifier to be non-null. Encode this assumption here.
>           assert sha and repository
E           AssertionError
```

Oh wow. I am glad I did put this assertion in.
```
    def test_create_allow_just_repository(self, client):
        self.authenticate(client)
        data = copy.deepcopy(self.valid_payload)
        data["run_id"] = _uuid()
        data["github"]["commit"] = ""
        data["github"]["repository"] = ARROW_REPO
    
        response = client.post("/api/benchmarks/", json=data)
        new_id = response.json["id"]
        benchmark_result = BenchmarkResult.one(id=new_id)
        assert benchmark_result.run.commit.sha == ""
>       assert benchmark_result.run.commit.repository == ARROW_REPO
E       AssertionError: assert '' == 'https://github.com/apache/arrow'
E         - https://github.com/apache/arrow
```

conbench/entities/run.py
jgehrcke: :) :tophat:

Issue #805, opened 2023-03-07
benchadapt: consistently use stdlib logging (do not mix stdout, stderr -- bad message time order)


jgehrcke:
follow-up from #804.

> It's a good idea to not mix stdout/stderr for the app's log output because then the relative timing between messages sometimes becomes unclear (when querying logs from a log aggregator).

one problem is here: https://github.com/conbench/conbench/blob/aa95b6d52ff6cb6697b7ab8d04c43d6cf3612860/benchadapt/python/benchadapt/adapters/_adapter.py#L177

alistaire47:
Ah yeah, I think that ended up as prints because logging was going weird somewhere. Really all that should get moved into benchclients directly and the logging integrated with that logging.

jgehrcke:
> Ah yeah, I think that ended up as prints because logging was going weird somewhere.

Would love to understand more about what went weird where :)

Well, the problem is not necessarily the `print()` itself, but the fact that we write to `stdout` instead of `stderr`. Each stream is independently buffered, so exactly how they interleave is unpredictable, messing with relative order of messages.

One could also write to `stderr` when not using `logging`: `sys.stdout.write()` or `print(..., file=sys.stderr)`.

As far as I understand, `benchadapt` is a Python _library_. We still do not have that term in its README (I would love so, it's the canonical description, if that's what it is). As a Python library, one can sanely use `logging` when one does not modify the so-called root logger. That's what `logging.basicConfig()` is good for. I'll try this out.

jgehrcke:
I think we can close this as 'done'

PR #820, opened 2023-03-07
benchadapt: log to stderr instead of stdout

jgehrcke:
For issue #805.

### DIFF
--- a/benchadapt/python/benchadapt/adapters/_adapter.py
+++ b/benchadapt/python/benchadapt/adapters/_adapter.py
@@ -1,14 +1,27 @@
 import abc
 import datetime
+import logging
 import subprocess
 import uuid
 from typing import Any, Dict, List
 
 import requests
 from benchclients.conbench import ConbenchClient
-from benchclients.logging import fatal_and_log, log
+from benchclients.logging import fatal_and_log
 
 from ..result import BenchmarkResult
+
+log = logging.getLogger("benchadapt.adapters")
+
+# `basicConfig()` does nothing if the root logger already has handlers
+# configured (that is, if this benchadapt library is imported in the context
+# of a program that already has a root logger with handlers set up, then the
+# following call is a noop.
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s.%(msecs)03d %(levelname)s: %(message)s",
+    datefmt="%y%m%d-%H:%M:%S",
+)
 
 
 class BenchmarkAdapter(abc.ABC):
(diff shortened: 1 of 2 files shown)


benchclients/python/benchclients/logging.py
```diff
@@ -1,9 +1,12 @@
 import logging
 
+log = logging.getLogger(__name__)
+
 logging.basicConfig(
-    format="%(levelname)s [%(asctime)s] %(message)s", level=logging.INFO
```
jgehrcke, line 4: compared to the 'old' format this adds fractions of a second with millisecond resolution: `%(msecs)03d`

benchadapt/python/benchadapt/adapters/_adapter.py
```diff
@@ -1,15 +1,28 @@
...
+
+# `basicConfig()` does nothing if the root logger already has handlers
+# configured (that is, if this benchadapt library is imported in the context
+# of a program that already has a root logger with handlers set up, then the
+# following call is a noop.
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s.%(msecs)03d %(levelname)s: %(message)s",
+    datefmt="%y%m%d-%H:%M:%S",
+)
```
jgehrcke, line 24: This takes effect if the benchadapt library used from within any program that is not `benchclients`.

benchclients/python/benchclients/logging.py
```diff
@@ -1,9 +1,12 @@
...
+    level=logging.INFO,
+    format="%(asctime)s.%(msecs)03d %(levelname)s: %(message)s",
+    datefmt="%y%m%d-%H:%M:%S",
```
jgehrcke, line 8: This takes effect if the benchadapt library is used from within `benchclients`.

jgehrcke: Is benchclients a CLI? Or another set of libraries?

alistaire47: Libraries. The only one that's a CLI is benchconnect

benchadapt/python/benchadapt/adapters/_adapter.py
```diff
@@ -190,7 +203,7 @@ def post_results(self) -> list:
             res_list.append(res)
 
         if error:
-            print(f"POST at {error_time} timed out twice:")
+            log.error(f"POST at {error_time} timed out twice:")
             raise error
```
alistaire47, line 207: If there's value in this retry-on-timeout code, it should get migrated to benchclients so all requests get more robust handling, not just these particular [important!] ones. Made #824 for this

benchclients/python/benchclients/logging.py
alistaire47: ...and datalogistik, but I think that's still in a separate repo for now

alistaire47 (approved):
Looks good. I think in a lot of contexts users will want the benchclients log set at DEBUG so all request bodies get logged, but it's fine to let users opt-in to that.

benchclients/python/benchclients/logging.py
jgehrcke: Thanks!

jgehrcke:
>  I think in a lot of contexts users will want the benchclients log set at DEBUG so all request bodies get logged,

not sure, are you implying that you think that behavior changed in this regard via this PR?

Issue #819, opened 2023-03-07
docs: Swagger isn't surfacing the docs in the way we need to.


jonkeane:
We have lots of details in our marshmallow-based API fields which bubble up into swagger (which is great!) but the way that swagger displays them (especially for the various object schemas at the bottom) makes it unlikely folks will find what they are looking for. 

What can we do to improve this?

jgehrcke:
Is this about information that is currently not at all visible in the `api/docs` UI or (only) difficult to see?

Edit: judging by https://github.com/conbench/conbench/issues/818#issuecomment-1458468884 this is about info that's already there; I should have had another look! :)

But yeah, in the future I think we should have a nicer way to expose the schemas.

jonkeane:
Yeah, it's _there_ but generally hidden behind (multiple!) 🔽 s. 

The other thing I personally find more frustrating than it should be: you can't deep-link into those schemas. So if I wanted to share with you details about the data fields, I need to send the main URL and tell you to click all the places 😦 

These might be easily solvable with a different swagger theme (if that's a thing — I haven't looked too much). Though it's not _totally_ off that those object schemas get lower priority than we want in most API docs like this. So maybe that's an indication swagger isn't good for us here?

jgehrcke:
>  you can't deep-link into those schemas. So if I wanted to share with you details about the data fields, I need to send the main URL and tell you to click all the places frowning

Oh yes, that's an important one!

alistaire47:
I've used [redoc](https://github.com/Redocly/redoc) since it comes with fastapi, and it gives schemas higher priority: https://redocly.github.io/redoc/#tag/pet/operation/addPet I think it should mostly just work on the same openapi metadata as swagger, though I haven't tried

jgehrcke:
Closing this as 'done' for now, now that we have a redoc-based presentation. We have a lot of work to do on this front, but we should try to track specific goals with specific tickets.

PR #826, opened 2023-03-08
Introduce Redoc for displaying HTTP API specification

jgehrcke:
Motivated by https://github.com/conbench/conbench/issues/819.

Run `make run-app-dev`. Point your browser to http://127.0.0.1:5000/api/redoc.

Update:

This is ready for review. Some notes:
- I have worked towards displaying the object models in the sidebar (see below). Confirmed that linking directly to models works, e.g.  `http://127.0.0.1:5000/api/redoc#tag/Models/BenchmarkResultCreate`
- There are many details to be figured out. For now, I'd love to keep the swagger UI at `/api/docs`. We can consider `/api/redoc` to be an experiment that we can iterate on together. If we can agree on this, then I think this PR is ready to be merged. If we want a more complete replacement from the start then I think there's quite a bit more work to do.


jgehrcke:
Sidebar:
image

Does not show models yet. Needs investigation.

Main content is aware of models:
image

`Submit a BenchmarkResult within a specific Run. If` suggests that we may want to pay more attention to differentiating heading from body.

jgehrcke:
Redoc documentation isn't super precise. Current HTML is roughly is based on https://redocly.com/docs/redoc/deployment/html/, but already required a little bit of debugging and some "aahaa" effects (behavior out of sync with docs). 

In particular I wanted to use the https://redocly.com/docs/redoc/deployment/html/#the-redoc-object approach in order to have access to the config parameters, documented here: https://redocly.com/docs/api-reference-docs/configuration/functionality/

Within config parameters we quickly see statements like "Supported in Redoc CE." -- not everything is supported in their free offering.

jgehrcke:
PetStore shows models in side bar: https://redocly.github.io/redoc/
How?
Edit: Ha, https://stackoverflow.com/questions/68475706/have-schemas-in-redoc-side-me

jgehrcke:
So. In order to show models in the sidebar Redoc decided to use the x-tagGroups mechanism:

> x-tagGroups is used to group tags in the side menu. Before you use x-tagGroups, make sure you add all tags to a group, since a tag that is not in a group, will not be displayed at all!

https://github.com/Redocly/redoc/blob/main/docs/redoc-vendor-extensions.md#x-tagGroups

Interesting: https://github.com/Redocly/redoc/issues/134#issuecomment-525162727

Also:

> In my opinion, if ReDoc was implementing the "list-all-models-automatically" feature, this should be an option nativeScrollbars or theme. It can be false by default.

(https://github.com/Redocly/redoc/issues/134#issuecomment-525351309)


> It looks like on the example, the Models section is created by defining it in the sidebars.yaml with links to the tags in the openapi.yaml.

(https://github.com/Redocly/developer-portal-starter/issues/106#issuecomment-1057396779)

https://github.com/Redocly/redoc/issues/1717 ("An option to render components in a "Models" section of the HTML site")

Similar / same problem, plus a solution: https://github.com/Topsort/openapi/pull/32
(that solution is not necessarily the best, but it can work)

jonkeane:
> Submit a BenchmarkResult within a specific Run. If suggests that we may want to pay more attention to differentiating heading from body.

Absolutely, yeah. That's definitely an oversight (or something vaguely encouraged or not discouraged with swagger) but strong agree that headers separate from body are super important for being able to skim, focus on what's imporant.

Another thing I noticed in redoc is our response codes could use a bit more detail too — they look a little lacking in redoc. In swagger there is some info, for example, a pretty generic description of "created" for 201 (that might just be default from swagger...). Anyway, another thing we'll want to polish if we switch (and I'm glad to see tools like this nudging us to do better 💅 )

jonkeane:
Oh, and forgot to add: it's "just" style (though that's important!) I really like the cleanliness of this + having the samples off to the side instead of just below, etc. It looks much more modern + sleek IMHO. Love this!

jgehrcke:
Two screenshots of the current state, for easier review (though I suggest running `make run-app-dev` from this branch)

### sidebar
image

### view on a specific schema:
image

jgehrcke:
`ci/testsuite` failed with
```
  -                           '## GitHubCreate\n'
  ?                               ^^ ---
  +                           '## BenchmarkResultCreate\n'
  ?                               ^^^^^^^^^^^^^^
                              '<SchemaDefinition '
  -                           'schemaRef="#/components/schemas/GitHubCreate" />\n'
  ?                                                            ^^ ---        ----
  +                           'schemaRef="#/components/schemas/BenchmarkResultCreate" '
  ?                                                            ^^^^^^^^^^^^^^
  +                           '/>\n'
                              '\n'
                              '## ClusterCreate\n'
                              '<SchemaDefinition '
                              'schemaRef="#/components/schemas/ClusterCreate" />\n'
                              '\n'
  +                           '## GitHubCreate\n'
  +                           '<SchemaDefinition '
  +                           'schemaRef="#/components/schemas/GitHubCreate" />\n'
  +                           '\n'
                              '## MachineCreate\n'
                              '<SchemaDefinition '
                              'schemaRef="#/components/schemas/MachineCreate" />\n'
  -                           '\n'
  -                           '## BenchmarkResultCreate\n'
  -                           '<SchemaDefinition '
  -                           'schemaRef="#/components/schemas/BenchmarkResultCreate" '
  -                           '/>\n'
                              '\n'
```

That's because dictionary key iteration order in CPython is now (well, since 3.6) ordered by insertion; and this demonstrates nicely that module import order is not stable. Solution is probably to sort keys alphabetically (trying that).

jonkeane (approved):
Thanks for this! I was going to try and push some of the cleanups we talked about here (basically adding `summary: ...` so places where it makes sense, but ran out of time today. I agree that we should merge this now and try it in parallel. I'll try to come back through and do a pass of cleaning up things like adding summaries

Issue #839, opened 2023-03-10
UI: a few problems, cosmetic issues in: navbar/menu, footer, element spacing, table responsiveness


jgehrcke:
We should, step by step, level up the UI. Here I want to quickly list tiny bugs, weirdnesses, or tiny improvement ideas:

### dropdown in navbar
We have a dropdown in the navbar that is not coordinated with the navbar itself. That's one level too much. In a narrow viewport when the navbar collapses, the dropdown stays a dropdown. So one needs two taps/clicks to get to the useful menu contents:
![image](https://user-images.githubusercontent.com/265630/224365549-7eb511c3-d037-4373-a1ef-945ea03577e5.png)

### logged in identity, api docs
![image](https://user-images.githubusercontent.com/265630/224366011-ed963922-9626-4c69-8af2-8ba5611eba1f.png)
two observations:
- The logged in identity shows as a menu item, is clickable. Click leads nowhere. We should maybe show email instead (the unique constraint), indicate that this is the login state, and probably bring `Logout` closer to this.
- A link to API docs is great. But this has nothing to do with being logged in. Put this maybe into the page footer.

### Conbench footer only on landing page

The footer (with version info) only shows on the landing page. It should be everywhere. It's maybe a good place to put the link to the API docs.

### Run view: various observations

- Add space between those elements:
![image](https://user-images.githubusercontent.com/265630/224367083-d30766f0-06ff-487a-b974-d84fb02fdeda.png)
- Make deletion button red, use clearer label

- maybe show this in red:
![image](https://user-images.githubusercontent.com/265630/224381803-33627e62-b716-430b-a875-2c566a900c6b.png)

- table is not responsive (too wide):
![image](https://user-images.githubusercontent.com/265630/224381998-98c3ee13-8ad0-41d0-84b9-f4dba3ae1a3b.png)

- do not attempt to render table when there are no benchmarks in the run

jgehrcke:
I think all the details are taken care of. Closing as 'done'.

Issue #836, opened 2023-03-10
UI: work towards few/no w3c validator warnings/errors


jgehrcke:
When I built websites for businesses one of my most important tools was the W3C validator. Using it for `https://conbench.ursa.dev/login/` (I picked it because it has little content) currently shows warnings and errors, for example:
- `The align attribute on the div element is obsolete`
- `Bad value for attribute action on element form: Must be non-empty`

(`https://validator.w3.org/nu/?doc=https%3A%2F%2Fconbench.ursa.dev%2Flogin%2F`)

It's good to keep these at a minimum; so that the validator can be used efficiently during dev (high signal to noise ratio).

PR #842, opened 2023-03-10
UI: show footer in all views (+ misc changes)

jgehrcke:
Related: #836, #839

### DIFF
--- a/conbench/app/_endpoint.py
+++ b/conbench/app/_endpoint.py
@@ -8,7 +8,20 @@ import flask.views
 import flask_login
 import werkzeug
 
+from .. import __version__
+from ..buildinfo import BUILD_INFO
+
 log = logging.getLogger(__name__)
+
+
+# Default to importlib_metadata version string.
+VERSION_STRING_FOOTER = __version__
+
+
+# Enrich with short commit hash, if available.
+# Also see https://github.com/conbench/conbench/issues/461
+if BUILD_INFO is not None:
+    VERSION_STRING_FOOTER = f"{__version__}-{BUILD_INFO.commit[:9]}"
 
 
 def authorize_or_terminate(func):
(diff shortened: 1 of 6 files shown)


conbench/templates/app.html
```diff
@@ -102,7 +102,13 @@ <h4 class="modal-title"><span class="glyphicon glyphicon-trash submenu"></span>&
 
         {# application content needs to be provided in the app_content block #}
         {% block app_content %}{% endblock %}
+    <br>
+    <hr>
+    <div style="display:inline-block; float: right; color: #918d8c;">
```
jgehrcke, line 107: Also changed the color a bit (darker grey).

conbench/app/auth.py
```diff
@@ -106,6 +108,11 @@ def post(self):
 
                 return self.redirect("app.index")
             else:
+                log.info(
+                    "login request failed, response details: %s, %s",
+                    response,
+                    response.text,
+                )
```
jgehrcke, line 115: This does otherwise swallow useful error information (e.g., if the API response has status code 400 and details in the body about how the request was bad -- then this would be swallowed).

conbench/app/_endpoint.py
```diff
@@ -125,6 +138,9 @@ def redirect(self, endpoint, **kwargs):
         return f.redirect(f.url_for(endpoint, **kwargs))
 
     def render_template(self, template, **kwargs):
+        # inject/overwrite
+        kwargs["version_string_footer"] = VERSION_STRING_FOOTER
```
jgehrcke, line 142: This was the simplest / most immediate way I thought of to affect all views (they all have in common that they use this `render_template()` method). There might be more elegant ways.

conbench/templates/app.html
```diff
@@ -102,7 +102,13 @@ <h4 class="modal-title"><span class="glyphicon glyphicon-trash submenu"></span>&
...
+        Conbench {{ version_string_footer }} | <a href="/api/docs" target="_blank">api docs</a>
```
jgehrcke, line 108: This adds a link to API docs into the footer -- want to remove it from the navbar (see #839).

jgehrcke:
Screenshot of footer part of ci-generated screenshot, there you can see the new horizontal ruler, the version info, and the api docs link:
image

conbench/templates/app.html
austin3dickey: love this!

austin3dickey (approved):

PR #833, opened 2023-03-09
benchmark result view: tiny bit of docs for distr chng annotation

jgehrcke:
Context: https://github.com/conbench/conbench/issues/691. Motivated by @joosthooz's question today on Slack, and also by https://github.com/conbench/conbench/issues/829. 

My attempt to improve current docs a bit.

Future: use a proper collapse / tooltip mechanism, with established UI/UX concepts for 'how-to-get-help'. (tiny question mark icon, etc).

PR #843, opened 2023-03-10
benchmark result view: update help for distr chng btn

jgehrcke:
Follow-up for https://github.com/conbench/conbench/pull/833 after talking to @joosthooz again. 
For https://github.com/conbench/conbench/pull/833.

One important bit is to clarify that the annotation really is persisted in the database (users might rightfully wonder if that only modifies their plot locally, or if they do global damage -- and then of course it's important to tell them they can revert this again!)


austin3dickey (approved):
beautiful!

PR #627, opened 2023-01-25
ci: introduce mypy

jgehrcke:
This is a first baby step for https://github.com/conbench/conbench/issues/605.

The goal of this was to introduce a `mypy` check to CI that is initially green, and covers most of the conbench web application code (not cli, not tests).

So that we can widen and extend type checking in the future to actually extract value.

Not so surprisingly, around existing type annotations there were some errors which I either quickly addressed or which I decided to explicitly ignore for now.

I backed out of the attempt/inclination to add 'nice types' here and there in _this_ PR. I think that all updated/changed/introduced type annotations in this PR are here only to pragmatically fix mypy errors.

Side note: I took opportunity to rename "requirements-test" to "requirements-dev". I think of this as "dependencies needed for development and CI". I also renamed "requirements-server" to "requirements-webapp". I think "web application" is a rather canonical term for what we build (offering an HTTP API backed by a database, and a UI/frontend). 

A bit of a reminder w.r.t. benefits:

> Static type annotations improve readability. Improved readability is probably the most important benefit of statically typed Python. With static types, every function clearly documents the types of its arguments and return value, which saves a ton of time reading the codebase to figure out the precise types a function expects so that you can call it.

> We can refactor with confidence. Refactoring a Python codebase without breaking things involves a lot of careful manual checking to confirm that you’ve updated all the code that used the old interface. With a statically typed Python codebase, the type checker can often verify this for you automatically.

PR #859, opened 2023-03-13
typing: add type hints for DB columns

austin3dickey:
Now that we have SQLAlchemy 2.0, there is more support for rich nested typing of SQLAlchemy objects. This PR adds type hints for all our DB columns based on the [recommended way](https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html) to type-hint declarative tables.

I think this PR is extremely exciting! I was able to fix some additional typing-related errors that `mypy` found, since it's now possible to statically check the types of not only the raw column objects, but also the query result columns themselves.

VS Code is also showing me a lot more detail about each object. As just one example, it now knows the types of the `commit_ancestry_query` columns:

image

### DIFF
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -2,7 +2,7 @@ import collections
 import copy
 import json
 import logging
-from typing import List, Optional, no_type_check
+from typing import List, Optional, Tuple, no_type_check
 
 import bokeh.events
 import bokeh.models
@@ -152,7 +152,7 @@ def simple_bar_plot(benchmarks, height=400, width=400, vbar_width=0.7):
 
 # Using `Optional[float]` here only because of BenchmarkResult.mean being
 # nullable as of today.
-def _should_format(floats: list[Optional[float]], unit):
+def _should_format(floats: List[Optional[float]], unit):
     unit_fmt = formatter_for_unit(unit)
 
     units_formatted = set()
@@ -182,7 +182,7 @@ def _insert_nans(some_list: list, indexes: List[int]):
 
 @no_type_check
 def _source(
-    samples: list[HistorySample],
+    samples: List[HistorySample],
     unit,
     formatted=False,
     distribution_mean=False,
@@ -299,7 +299,7 @@ def _source(
     return bokeh.models.ColumnDataSource(data=dsdict)
 
 
-def _inspect_for_multisample(items: list[HistorySample]) -> tuple[bool, Optional[int]]:
+def _inspect_for_multisample(items: List[HistorySample]) -> Tuple[bool, Optional[int]]:
     """
     `items`: list of benchmark results as encoded by the history variant of the
     `_Serializer(EntitySerializer)`.
@@ -584,7 +584,7 @@ def time_series_plot(
         toolbar_location="right",
         x_range=(t_start, t_end),
     )
-    p.toolbar.logo = None
+    p.toolbar.logo = None  # type: ignore[attr-defined]
 
     # TapTool is not responding to each click event, but but only triggers when
     # clicking a glyph:
@@ -688,7 +688,7 @@ def time_series_plot(
     dist_change_in_legend = False
     for ix in dist_change_indexes:
         p.add_layout(
-            bokeh.models.Span(
+            bokeh.models.Span(  # type: ignore[attr-defined]
                 location=util.tznaive_iso8601_to_tzaware_dt(
                     # transforming from datetime to string back to datetime is
                     # weird; done as part of refactoring to not change too much
--- a/conbench/entities/_entity.py
+++ b/conbench/entities/_entity.py
@@ -3,16 +3,16 @@ import uuid
 from typing import List
 
 import flask as f
-from sqlalchemy import Column, distinct
+from sqlalchemy import distinct
 from sqlalchemy.dialects.postgresql import insert as postgresql_insert
-from sqlalchemy.orm import declarative_base
+from sqlalchemy.orm import declarative_base, mapped_column
 from sqlalchemy.orm.exc import NoResultFound
 
 from ..db import Session
 
 Base = declarative_base()
-NotNull = functools.partial(Column, nullable=False)
-Nullable = functools.partial(Column, nullable=True)
+NotNull = functools.partial(mapped_column, nullable=False)
+Nullable = functools.partial(mapped_column, nullable=True)
 
 
 class NotFound(NoResultFound):
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -1,5 +1,6 @@
 import math
-from datetime import timezone
+from datetime import datetime, timezone
+from typing import List, Optional
 
 import flask as f
 import marshmallow
@@ -7,7 +8,7 @@ import numpy as np
 import sqlalchemy as s
 from sqlalchemy import CheckConstraint as check
 from sqlalchemy.dialects import postgresql
-from sqlalchemy.orm import relationship
+from sqlalchemy.orm import Mapped, relationship
 
 import conbench.util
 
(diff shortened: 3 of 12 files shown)


conbench/app/_plots.py
```diff
@@ -152,7 +152,7 @@ def simple_bar_plot(benchmarks, height=400, width=400, vbar_width=0.7):
 
 # Using `Optional[float]` here only because of BenchmarkResult.mean being
 # nullable as of today.
-def _should_format(floats: list[Optional[float]], unit):
+def _should_format(floats: List[Optional[float]], unit):
```
austin3dickey, line 155: I took the opportunity to revert to the backwards-compatible typing containers here, just in case (see https://github.com/conbench/conbench/pull/811#discussion_r1131556101). Agreed that we should put a framework in place for ensuring we never accidentally break the legacy `conbench` CLI running in the arrow-benchmarks-ci repo.

conbench/app/_plots.py
```diff
@@ -584,7 +584,7 @@ def time_series_plot(
         toolbar_location="right",
         x_range=(t_start, t_end),
     )
-    p.toolbar.logo = None
+    p.toolbar.logo = None  # type: ignore[attr-defined]
```
austin3dickey, line 587: I don't know why mypy started complaining about this. 

`error: "Instance[Toolbar]" has no attribute "logo"  [attr-defined]`

conbench/app/_plots.py
```diff
@@ -688,7 +688,7 @@ def time_series_plot(
     dist_change_in_legend = False
     for ix in dist_change_indexes:
         p.add_layout(
-            bokeh.models.Span(
+            bokeh.models.Span(  # type: ignore[attr-defined]
```
austin3dickey, line 691: Same here.

`error: Module has no attribute "Span"  [attr-defined]`

conbench/app/_plots.py
jgehrcke: missing annotations within bokeh / insufficient stubs (I think), yeah -- we'll have to endure treating false negatives.

conbench/entities/_entity.py
```diff
@@ -3,16 +3,16 @@
...
+from sqlalchemy.orm import declarative_base, mapped_column
 from sqlalchemy.orm.exc import NoResultFound
 
 from ..db import Session
 
 Base = declarative_base()
-NotNull = functools.partial(Column, nullable=False)
-Nullable = functools.partial(Column, nullable=True)
+NotNull = functools.partial(mapped_column, nullable=False)
+Nullable = functools.partial(mapped_column, nullable=True)
```
jgehrcke, line 15: this is interesting! thanks for thinking through this; reading docs, doing what helps!

conbench/entities/benchmark_result.py
```diff
@@ -30,43 +31,47 @@
...
+    max: Mapped[Optional[float]] = Nullable(s.Numeric, check("max>=0"))
+    mean: Mapped[Optional[float]] = Nullable(s.Numeric, check("mean>=0"))
+    median: Mapped[Optional[float]] = Nullable(s.Numeric, check("median>=0"))
+    stdev: Mapped[Optional[float]] = Nullable(s.Numeric, check("stdev>=0"))
+    q1: Mapped[Optional[float]] = Nullable(s.Numeric, check("q1>=0"))
+    q3: Mapped[Optional[float]] = Nullable(s.Numeric, check("q3>=0"))
+    iqr: Mapped[Optional[float]] = Nullable(s.Numeric, check("iqr>=0"))
+    error: Mapped[Optional[dict]] = Nullable(postgresql.JSONB)
+    validation: Mapped[Optional[dict]] = Nullable(postgresql.JSONB)
+    change_annotations: Mapped[Optional[dict]] = Nullable(postgresql.JSONB)
```
jgehrcke, line 74: :muscle:

conbench/entities/commit.py
```diff
@@ -120,7 +120,7 @@ def commit_ancestry_query(self) -> Query:
         query = Session.query(
             Commit.id.label("ancestor_id"),
             Commit.timestamp.label("ancestor_timestamp"),
-            s.sql.expression.literal(True).label("on_default_branch"),
+            s.sql.expression.literal(True, s.Boolean).label("on_default_branch"),
```
jgehrcke, line 123: I see, there's new candy everywhere now :candy:

jgehrcke (approved):
:rocket: it's a pleasure the see this! Thanks.

conbench/entities/benchmark_result.py
austin3dickey: oh, this reminds me: SQLAlchemy would actually let me do this:
```
change_annotations: Mapped[Optional[dict]] 
```
and it would figure out that it's a nullable JSONB column automatically. But I thought that was a little too "magic". I'd rather be as explicit as we were before.

jgehrcke:
> VS Code is also showing me a lot more detail about each object.

:heart:

conbench/entities/benchmark_result.py
jgehrcke: > it would figure out that it's a nullable JSONB column automatically

So SQLAlchemy would do introspection of type annotations at runtime? :)

What's interesting here (for others reading along, I know Austin knows this) is that `postgresql.JSONB` allows for an arbitrary JSON document (with nesting, and all JSON types, including array).

To make this line of code `Mapped[Optional[dict]] = Nullable(postgresql.JSONB)` wholesome and 'correct' there must be schema validation / careful JSON document construction happening that is in alignment with `dict`.

And `dict` is seemingly correct right now based on https://github.com/conbench/conbench/blob/5e626d479ed38ef53e02d102f084df4387f331eb/conbench/entities/benchmark_result.py#L233 and https://github.com/conbench/conbench/blob/5e626d479ed38ef53e02d102f084df4387f331eb/conbench/entities/benchmark_result.py#L587.

In the future I think we want to maybe narrow down the schema (and then also type) a little further, do get a handle on nesting.

conbench/app/_plots.py
austin3dickey: So my question was more "why did these errors just start showing up today?" I didn't think I touched anything relevant to these errors.

I dove into it a little more. During CI, we just install linting dependencies. 

https://github.com/conbench/conbench/blob/5e626d479ed38ef53e02d102f084df4387f331eb/.github/workflows/actions.yml#L26-L29

But in my local dev, I have all the dependencies installed. That's why `mypy` is able to complain that, for example, `bokeh.models has no attribute "Span"`.

If we did have the webapp dependencies installed, mypy on `main` has been broken for a while. This PR fixes it. As a precaution, I'm going to update our CI so it installs the webapp dependencies too for linting.

conbench/entities/benchmark_result.py
jgehrcke: @austin3dickey looking closer at https://github.com/conbench/conbench/blob/5e626d479ed38ef53e02d102f084df4387f331eb/conbench/entities/benchmark_result.py#L231 I think that we do not store `None`, but an empty dict if user didn't provide anything during BenchmarkResult creation. Can you double-check if this can ever be `None`? Or is this again 'only' for the scenario of DB schema migration?

austin3dickey: Haha, yes, the main point of my comment was that I chose to opt-out of SQLAlchemy doing introspection of type annotations at runtime. I think that's a little too magic.

You're right that our current JSONB schemas are underdone; they all _should_ represent maps, not arrays. And some of them probably shouldn't allow for nesting.

conbench/app/_plots.py
jgehrcke: > So my question was more "why did these errors just start showing up today?" I didn't think I touched anything relevant to these errors.

Ah, makes sense! Thanks for clarifying, and thanks for taking care of this. Indeed super important.

> As a precaution, I'm going to update our CI so it installs the webapp dependencies too for linting.

Yes, please! Maybe it's worth separating the `mypy` check out then. `ci/lint` was / can be so fast because of very little dependencies an _fast_ tooling.  So far, it completed within ~30 s, and this can be even faster.

The type checking can by definition never be so fast when it (of course) requires all libs installed, and has sometimes even some computational work to do. We can do that separately.

conbench/entities/benchmark_result.py
austin3dickey: >I think that we do not store None, but an empty dict if user didn't provide anything during BenchmarkResult creation. Can you double-check if this can ever be None? Or is this again 'only' for the scenario of DB schema migration?

Right, this should only be nullable due to DB schema migration. I tried to ensure that all future data is stored as empty dictionaries. But I believe at least the Arrow database has many `None`s in this column still, and it might take a while to update those rows.

alistaire47: > You're right that our current JSONB schemas are underdone; they all _should_ represent maps, not arrays. And some of them probably shouldn't allow for nesting.

I'd love to see us lock these down better, both on the db side as well as in the request schemas by which its populated. We do need flexibility in places, but in others we have strong, undocumented opinions about JSON structure which are a nightmare for users trying to adapt results.

alistaire47 (approved):
cool cool; the start of lovely typing efforts to come

jgehrcke:
> I'd love to see us lock these down better, both on the db side as well as in the request schemas by which its populated.

Ack!

conbench/app/_plots.py
jgehrcke: @austin3dickey it's always impressive how quickly one forgets... :) In #627 I did think of this problem space, and that thinking led to the naming of `requirements-dev.txt`, and I added type check dependencies there: https://github.com/conbench/conbench/pull/627/commits/e9a126298c206ac0d4fc4e33d7f1366268ac9627

I suppose the answer is we want CI to have the same environment that we also use locally in our IDEs, but on the other hand we don't want to (need to) have those `types-*` packages in our prod environment. 

We can try to resolve all this at a later point in time.

jgehrcke:
Let's land this!

And then see over time what we can learn about dependency collection separation in the future (https://github.com/conbench/conbench/pull/859#discussion_r1135108859).

Issue #668, opened 2023-02-03
machine_info.py: warnings do not show up during tests


jgehrcke:
https://github.com/conbench/conbench/blob/b524d43468bc4cfce04c7aba59bc2392a595355f/conbench/machine_info.py#L100

This indirection is not of clear value:
```
class GitParseWarning(RuntimeWarning):
    pass
```

This indirection makes it so that these warnings do not show during a pytest session (and possibly never by default, I did not test this).

`RuntimeWarning` does not seem to be the right base to derive from when one wants to emit warnings from applications.

Leaving this as the _default_ https://docs.python.org/3/library/exceptions.html#UserWarning is more useful. This shows up during a pytest session:

```
conbench/tests/benchmark/test_runner.py::test_runner_simple_benchmark
  /app/conbench/machine_info.py:102: UserWarning: error in github_info(): probably not in a git repo; returning {}
    warnings.warn(
```

jgehrcke:
I just saw that these warnings were explicitly silenced during a pytest session:

https://github.com/conbench/conbench/blob/b524d43468bc4cfce04c7aba59bc2392a595355f/conbench/tests/benchmark/test_runner.py#L16

This is like a trap for fellow developers:
- have a test require a feature to work
- make that feature emit a specific warning when it does not work, but otherwise not throw an exception
- then silence that specific warning during tests

I am sure it wasn't meant that way, but this feels like an easteregg :egg:  :D :male_detective:

austin3dickey:
Oh I did this! Here were my thoughts:

1. I want to suppress expected warnings in the pytest summary, to reduce noise
2. I want to see unexpected warnings in the pytest summary, so developers can act on them
3. `github_info()` tries to read certain git information from the local git setup, and emits a warning if it fails to do so for some reason. One instance that it will fail to read the correct information is if it's running on a detached HEAD.
4. Sometimes we run tests on a local computer on a tracked branch, where `github_info()` would succeed. In GitHub Actions, it always runs tests on a detached HEAD, where `github_info()` would fail and warn.
5. Because of (3) and (4), sometimes we'll see these warnings in the test suite and sometimes we won't. This fails point (1). So we have to catch the warnings in the test suite somehow. If we don't catch them (like you tried in #667) we will see expected noise like [this](https://github.com/conbench/conbench/actions/runs/4084987390/jobs/7046718157#step:3:2471) in the pytest summary.
[...]

PR #860, opened 2023-03-13
benchadapt: Remove excess GitHub data warning

### LINKED ISSUE #808: benchadapt: emits many UserWarning: Both CONBENCH_PROJECT_REPOSITORY and ....., review verbosity
After #804.

Seen in `organizations/jenkins/<redacted>/detail/dev/108/pipeline/128/`, many of these warnings (between 10 and 100):
```
/opt/mambaforge/envs/<redacted>/lib/python3.10/site-packages/benchadapt/_machine_info.py:95: UserWarning: Both CONBENCH_PROJECT_REPOSITORY and CONBENCH_PROJECT_COMMIT must be set if `github` is not specified. CONBENCH_PROJECT_PR_NUMBER should be not be set for builds on the default branch.
```

- If the warning is meaningful and important: we should change the pipeline to use better config.
- If the pipeline is actually fine: we should consider removing the warning.

@alistaire47 @joosthooz @jonkeane I'd love to hear your opinions, thanks!

alistaire47:
Closes #808 by removing the warning in `github_info()`. The situation it was to address is already covered by conditions in the `.to_pusblishable_dict()` methods of `BenchmarkRun` and `BenchmarkResult`, so this should just quiet logs when GitHub metadata is specified after dataclass instantiation without any loss of functionality.

### DIFF
--- a/benchadapt/python/benchadapt/_machine_info.py
+++ b/benchadapt/python/benchadapt/_machine_info.py
@@ -91,16 +91,6 @@ def github_info():
         "pr_number": pr_number,
     }
 
-    if not (repo and commit):
-        warnings.warn(
-            (
-                "Both CONBENCH_PROJECT_REPOSITORY and CONBENCH_PROJECT_COMMIT must be "
-                "set if `github` is not specified. CONBENCH_PROJECT_PR_NUMBER should be "
-                "not be set for builds on the default branch."
-                f"\nValues: `{gh_info}`"
-            ),
-        )
-
     return gh_info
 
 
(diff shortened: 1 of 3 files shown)


jgehrcke (approved):
Cool. Haven't thought this through, but  I trust that it's an improvement :rocket:. 

Can you maybe also clarify how this relates to https://github.com/conbench/conbench/issues/668?

alistaire47:
> Can you maybe also clarify how this relates to #668?

That's about the codepath that tries to detect git info from git; this function just gets it from environment variables. This is also in benchadapt's port of those utils used by the legacy runner referred to there. But because it was ported from there, the same situation highlighted in that issue exists in the benchadapt detecting version (which should be rarely used): 

https://github.com/conbench/conbench/blob/67cbe18fe7674a065f08e72d10ea38a966952f3a/benchadapt/python/benchadapt/_machine_info.py#L97-L145

Issue #868, opened 2023-03-14
UI work after landing #851


jgehrcke:
@alistaire47 @austin3dickey and I reviewed staging env after landing #852. We noted:

User management:
- the `delete` button is wrong
- relates to empty columns
- check `add user` form

Other:
- benchmark result listing (`/benchmarks`): trashcan icons missing
- 'compare to baseline' should maybe be a button (thanks @alistaire47)
- there are more table captions that are now below the table, should be replaced by headings

jgehrcke:
Took care of the most important parts in this.

> benchmark result listing (/benchmarks): trashcan icons missing

#964 might be more important potentially implies removal of that view.

Closing this as 'done' for now.

PR #869, opened 2023-03-14
UI: fix user list / management, and hardware listing

jgehrcke:
Addresses the user management issues described in #868.

A number of changes along the way. Screenshots / vid follow.

------

Now with table heading, and right-aligned `add user` button:

Screenshot from 2023-03-14 18-50-15

------

Hardware list analogue to user list:

Screenshot from 2023-03-14 18-53-30

------

Some menus are still visually weird/bad, but management functionality is restored:

https://user-images.githubusercontent.com/265630/225098926-dec767d5-a4a9-46bb-8d5b-83c1a9086f1a.mov


jgehrcke:
Moving ahead with this -- please keep the feedback coming!

Issue #873, opened 2023-03-15
api spec: document current meaning of batches and/or change the UI/UX concept right away


jgehrcke:
That is all we have to date in the API spec:
![image](https://user-images.githubusercontent.com/265630/225232699-1a9d7cdd-2160-49e7-bfcd-25c56c214864.png)

Column [def](https://github.com/conbench/conbench/blob/b84e345f04dfe06f3e6855ede20339a982a35a00/conbench/entities/benchmark_result.py#L58):
```
 batch_id: Mapped[Optional[str]] = Nullable(s.Text)
```

Why is this `s.Text` and not limited in length like so many other IDs?

and then we [do](https://github.com/conbench/conbench/blob/b84e345f04dfe06f3e6855ede20339a982a35a00/conbench/entities/benchmark_result.py#L230
)
```
benchmark_result_data["batch_id"] = data["batch_id"]
```
and then we also do
```
s.Index("benchmark_result_batch_id_index", BenchmarkResult.batch_id)
```
Why is the index important?

----

We make good progress discovering the intended meaning of existing concepts. To that end, I also want to quote

> A note about terminology: A “run” is a set of batches, and a “batch” is a set of benchmark results. For example, all of these Python “file-read” benchmarks are in one batch, and the R “file-read” benchmarks are in another batch, but both sets are part of the same run.

from https://ursalabs.org/blog/announcing-conbench/

We also have this comment from @jonkeane:

https://github.com/conbench/conbench/issues/561#issuecomment-1367982438

There is related detail in https://github.com/conbench/conbench/issues/252 it mentions "manually_batch and mark_new_batch". These appear in code only in CLI context where in runner.py mark_new_batch() means `self._batch_id = uuid.uuid4().hex`.

The API and database only knows of batches in the sense that each BenchmarkResult must store a batch ID. This raises questions, for example: can a Batch ID re-appear across two or more Runs? (is a Batch guaranteed to be Run-local?). Judging by code, I don't think it is.

And maybe most importantly I want to quote @jonkeane's train of thought here:

> Maybe if we loosen up the constraint on the UI side (or provide an entirely different way to look at (arbitrary?) groups of benchmarks together). We could avoid this confusing(!) duplication and use batch for all and only groups of benchmarks that don’t share a name, but generally should be consumed together (e.g. "suites").

My intuition is that in terms of UX _simplification_ the less we can expose this, the better.

jgehrcke:
[...]

PR #875, opened 2023-03-15
ui: per-batch view: clarifications (#873)

jgehrcke:
This is for https://github.com/conbench/conbench/issues/877, i.e. a follow-up after inspection documented here: https://github.com/conbench/conbench/issues/873#issuecomment-1469751855

This is not really meant to look great, and is not tackling _all_ challenges, but is supposed to provide really important conceptual clarifications UI-wise.

The Run -> Batch hierarchy is still misleadingly represented, but I didn't yet want to change too much in this PR. For the 'common' case, this breadcrumb hierarchy allows for navigating back to the Run.

Screenshot from 2023-03-15 11-53-12

### DIFF
--- a/conbench/app/batches.py
+++ b/conbench/app/batches.py
@@ -13,7 +13,7 @@ from ..config import Config
 
 
 class BatchPlot(AppEndpoint, ContextMixin):
-    def page(self, by_group):
+    def page(self, by_group, requested_batch_id):
         plots, raw, i = [], [], 1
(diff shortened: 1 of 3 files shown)


jgehrcke:
Landing this to unwind the PR FIFO queue. Had a really good conversation about this topic in a sync meeting with @alistaire47 and @austin3dickey -- I think they are OK with this PR here as-is. But please, as always, keep feedback coming also after merge.

PR #876, opened 2023-03-15
UI: fix /benchmarks view (benchmark result listing)

jgehrcke:
This repairs brokenness described in #868.

Note that this view cannot easily be reached by just using navigation in the UI :shrug:  :)

Screenshot before: [here](https://user-images.githubusercontent.com/265630/225289026-314b3edc-6a93-4cd7-8f7c-1d8c3bebcb41.png)

After:
Screenshot from 2023-03-15 11-50-31

### DIFF
--- a/conbench/templates/user-list.html
+++ b/conbench/templates/user-list.html
@@ -63,7 +63,7 @@ var table = $('#users').DataTable( {
     // published in https://cdn.datatables.net/plug-ins/1.13.3/features/.
     // kudos to https://stackoverflow.com/a/29639664/145400
     "conditionalPaging": true,
-    "pageLenth": 50,
+    "pageLength": 50,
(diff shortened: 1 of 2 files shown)


conbench/templates/user-list.html
```diff
@@ -63,7 +63,7 @@
     // published in https://cdn.datatables.net/plug-ins/1.13.3/features/.
     // kudos to https://stackoverflow.com/a/29639664/145400
     "conditionalPaging": true,
-    "pageLenth": 50,
+    "pageLength": 50,
```
jgehrcke, line 66: Tiny bug fix piggybacked on this PR. The outfall was: the table showed 10 entries. So, not too bad. But yeah -- linting would help. Rigorous manual testing is almost impossible.

jgehrcke:
As always, keep feedback coming also after merge. Thanks!

PR #870, opened 2023-03-14
UI: further work after migration to bootstrap 5 (re-create jumbotron, ... )

jgehrcke:
Currently this also includes #869.

This PR:

- re-creates the Jumbotron (that type of component does not exist anymore in Boostrap 5)
- introduces a simple table heading for the landing page `Recent runs` in level 3, followed by a ruler -- like I did before for the user list and hardware list. Just a tiny visual/design tweak that hopefully pleases some eyes for now.
- does visual cleanup work around tables, including landing page 'Recent runs' (see #861)
- tweaks the DataTables options for the 'Recent runs'

Screenshot from local dev:

Screenshot from 2023-03-14 19-51-02

PR #879, opened 2023-03-15
ui: link github project in footer, landing page: remove jumbotron

jgehrcke:
In UI design it's all about space, making good use of space and spacing. The jumbotron was nice. 

But with the link to the project in the footer I don't think we need this huge branding piece anymore I think.

This also cleans up the spacing/padding a bit via p-2 -> p-0.

And brings back the entry count selection menu to the top, it fits again I think, and is simply more accessible.

Screenshot from 2023-03-15 14-51-29

Side thought, this triggers thinking about multi-repo support vs single-repo support.
I think all rows in the `repository` column have the same value here in this case. 
We probably want to remove this column and then show the table once per repo (if there are multiple repos known).

### DIFF
--- a/conbench/templates/index.html
+++ b/conbench/templates/index.html
@@ -3,7 +3,7 @@
 {% block app_content %}
 
 <div class="p-0"><h3>Recent runs</h3></div>
-<hr class="border border-danger border-2 opacity-50">
+<hr class="border border-danger border-1 opacity-50">
 
 <!-- https://getbootstrap.com/docs/5.2/content/tables/#responsive-tables -->
 <div class="table-responsive">
--- a/conbench/templates/index.html
+++ b/conbench/templates/index.html
@@ -2,24 +2,7 @@
 
 {% block app_content %}
 
-<div class="p-3 mb-4 bg-light rounded-3">
-  <div class="container-fluid py-2">
-    <div class="row mb-3">
-      <div class="col-md-5">
-        <h1 class="display-5 fw-bold">Con·bench</h1>
-        <p class="col-md-8 fs-4">/kənˈben(t)SH/ <br /><i>noun</i></p>
-      </div>
-      <div class="col-md-7">
-        <div class="float-end">
-          <img class="media-object" style="padding: 5px" src="{{ url_for('static', filename='bar-graph.png') }}" height="180">
-        </div>
-      </div>
-    </div>
-  </div>
-</div>
-
-
-<div class="p-2"><h3>Recent runs</h3></div>
+<div class="p-0"><h3>Recent runs</h3></div>
 <hr class="border border-danger border-2 opacity-50">
 
 <!-- https://getbootstrap.com/docs/5.2/content/tables/#responsive-tables -->
@@ -113,7 +96,7 @@
       "order": [[0, 'desc']],
       // https://datatables.net/reference/option/dom
       // put `l` to the bottom
-      "dom": "frtipl",
+      "dom": "lfrtip",
       //"lengthMenu": [ 20, 25, 50, 75, 100 ],
       // Make "Commit" (commit hash) column (4) more narrow, hide 'sorting'
       // toggle. Sorting by commit hash is not needed, in particular because
--- a/conbench/templates/index.html
+++ b/conbench/templates/index.html
@@ -2,6 +2,9 @@
 
 {% block app_content %}
 
+
+<h1 class="display-3 text-center" style="margin-bottom: 40px">/kənˈben(t)SH/</h1>
+
 <div class="p-0"><h3>Recent runs</h3></div>
 <hr class="border border-danger border-1 opacity-50">
 
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -93,6 +93,7 @@
 
     {% endwith %}
     {# application content needs to be provided in the app_content block #}
+    <div style="margin-top: 30px"></div>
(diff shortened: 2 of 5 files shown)


jgehrcke:
`ci/cb-on-minikube` failed with an instability, now tracked here: https://github.com/conbench/conbench/issues/881.

conbench/templates/index.html
```diff
@@ -2,24 +2,7 @@
...
-        <p class="col-md-8 fs-4">/kənˈben(t)SH/ <br /><i>noun</i></p>
-      </div>
-      <div class="col-md-7">
-        <div class="float-end">
-          <img class="media-object" style="padding: 5px" src="{{ url_for('static', filename='bar-graph.png') }}" height="180">
-        </div>
-      </div>
-    </div>
-  </div>
-</div>
```
austin3dickey, line 19: I like the link in the footer! But I do miss the jumbotron. 

To your point, efficient use of space is good. But with this change, I feel a little overwhelmed by the amount of text immediately presented on the landing page. There's something to be said about balanced "blankness" being a good use of space too, especially on the main page of the app.

I prefer keeping the jumbotron. But I showed this change to my wife, who's never seen Conbench before, and she agreed with you that "less scrolling is better." So perhaps I'm in the minority here 🙂 Though we still may want to think about what else we could put at the top so that the table looks more balanced (maybe something more useful than just a logo).

jgehrcke:
I can see how this one might be controversial!

I really like the idea of using the landing page space a little more wisely.

Maybe we can re-introduce some branding-related visual candy later? (we should!)

lol, [Conbench, what changed? AI-generated... ](https://user-images.githubusercontent.com/265630/225377051-d83e3a70-4b13-4ae1-b126-f8b43eff79db.png)... anyway :)

I like that the work of migrating the jumbotron [was done](#870) independently; we can always re-introduce it maybe in an `about` page, if we'd like to? CC @jonkeane I want to mention you here explicitly; let's talk where to put the `/kənˈben(t)SH/` niceness.

conbench/templates/index.html
alistaire47: One way I get utility from the jumbotron was knowing instantly I'm at a home page—a lot of our pages are really just tables, and the jumbotron makes that particular one distinctive. But there are probably more useful ways to do that than with a logo, and we should figure out what some are.

jgehrcke: >  There's something to be said about balanced "blankness" being a good use of space too, especially on the main page of the app.

Yes, more vertical whitespace would be more eye-pleasing than putting `Recent run` right under the navbar.

jgehrcke: >  and she agreed with you that "less scrolling is better."

haha, yes!

From a very puristic, functional point of view the jumbotron is a UX 'failure' :-). I know so much. 

> Though we still may want to think about what else we could put at the top so that the table looks more balanced (maybe something more useful than just a logo).

Absolutely! UX-wise, we achieve perfection when no element can be taken away. Removing the jumbotron is just a quick idea towards breathing, towards making space for better ideas. The jumbotron is non-interactive, so this maybe is thought-provoking -- in _this_ geometrical region of the interface I think we totally want to offer interactivity.

jgehrcke: > I like the link in the footer! 

OK! :) :foot: 

>  But I do miss the jumbotron.

Now I consider just adding it back for now. Hm.
Or maybe just adding a bit of whitespace?

I realize I have a little bit of this minimization/removal quest going on -- which elements in the UI are here for what? If something isn't really needed, the inclination is to remove it (for now!).

Let's be in flux, in flow -- remove, remove, and then maybe add later again -- without overthinking too much for now.

Higher-level plan: let's be mindful of which navigation/interaction elements we _need_, and which workflows we want to enable. We _can_ at the same time also pay attention to aesthetics and design. But for a UI designer what's a great entry point is to see a well thought-through UX already. We can (and should!) bring in someone (I have someone on the back of my mind :-)) once we are confident that the UX is reasonably well thought-out.

jgehrcke:
On the one hand I said above that we shouldn't overthink the removal, but on the other hand, I like the feedback here and thought a bit about it. What do we think about the following for now:

Screenshot from 2023-03-15 20-57-59

Has a Japanese touch to it; don't even know why I think that and whether or not this is appropriate :crossed_swords:

conbench/templates/index.html
jgehrcke: > One way I get utility from the jumbotron was knowing instantly I'm at a home page—a lot of our pages are really just tables, and the jumbotron makes that particular one distinctive. But there are probably more useful ways to do that than with a logo, and we should figure out what some are.

Thanks for that feedback, too @alistaire47 -- I only now really saw that. Yeah, totally agree that in a "final UI state" where UX _and_ design have both been figured out there absolute needs to be some kind of landing page experience that also is a recognizable brand experience. We need _something_ on that front, in the long run.

I have temporarily 'fixed' that with adding back the `/kənˈben(t)SH/` :). In the longer run I totally agree with

> there are probably more useful ways to do that than with a logo, and we should figure out what some are.

and as put in a different comment I trust/hope that we will also get some UI design guidance then.

jgehrcke:
Would have loved to get one more round of feedback on the last state -- but it's OK of course. I think I'd like to land this soon today, and keep building with that in mind.

austin3dickey:
Haha, I like this. Thanks for reducing its space!

Issue #774, opened 2023-02-24
benchalerts: allow for conditional pipeline steps


austin3dickey:
Once #770 is completed, we should add a feature that would allow for steps to be run conditionally, based on the output of previous steps. This should enable #756 as well as the use case described here: https://github.com/conbench/conbench/issues/769#issuecomment-1443375149

PR #882, opened 2023-03-15
benchalerts: add step to post a PR comment about a GitHub Check

### LINKED ISSUE #769: benchalerts: post messages to GitHub PRs
Adding a GitHub Check with a markdown summary to a GitHub commit is nice, but it doesn't notify anyone (especially on default-branch commits), so it's not very actionable. Users of `benchalerts` should be able to opt in to the App posting a message on the relevant PR that links to the new Check.

austin3dickey:
This PR adds a `benchalerts` alert pipeline step to comment on a PR that a GitHub Check has been posted. This is useful when we run benchmarks on the default branch, if the relevant parties (e.g. the author of the most-recently-merged PR) are not necessarily monitoring the Checks that are posted to the default branch commits.

Closes #769.

[Here](https://github.com/conbench/benchalerts/pull/5#issuecomment-1470761104) is an example of the type of comment this step might post.

### DIFF
--- a/benchalerts/benchalerts/message_formatting.py
+++ b/benchalerts/benchalerts/message_formatting.py
@@ -108,3 +108,16 @@ def github_check_details(full_comparison: FullComparisonInfo) -> Optional[str]:
         """
     )
     return details
+
+
+def pr_comment_link_to_check(summary: str, check_link: str) -> str:
+    """Generate a GitHub PR comment that summarizes and links to a GitHub Check."""
+    return _clean(
+        f"""
+        ## ⚡️ Benchmark results ⚡️
+
+        {summary}.
+
+        See the full report [here]({check_link}).
+        """
+    )
--- a/benchalerts/benchalerts/pipeline_steps/github.py
+++ b/benchalerts/benchalerts/pipeline_steps/github.py
@@ -7,7 +7,12 @@ from benchclients.logging import fatal_and_log, log
 from ..alert_pipeline import AlertPipelineErrorHandler, AlertPipelineStep
 from ..conbench_dataclasses import FullComparisonInfo
 from ..integrations.github import CheckStatus, GitHubRepoClient, StatusState
-from ..message_formatting import _clean, github_check_details, github_check_summary
+from ..message_formatting import (
+    _clean,
+    github_check_details,
+    github_check_summary,
+    pr_comment_link_to_check,
+)
 
 
 class GitHubCheckStep(AlertPipelineStep):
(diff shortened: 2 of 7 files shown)


benchalerts/benchalerts/pipeline_steps/github.py
```diff
@@ -230,6 +235,82 @@ def _default_status_description(full_comparison: FullComparisonInfo) -> str:
...
+        ``github_client``.
+    ``GITHUB_API_TOKEN``
+        A GitHub Personal Access Token with the ``repo:status`` permission. Only
+        required if not authenticating with a GitHub App, and if ``repo`` is provided
+        instead of ``github_client``.
+    """
+
+    def __init__(
+        self,
+        pr_number: int,
```
austin3dickey, line 286: I was thinking about how we might "automatically" figure this out.

The GitHub API does have a ["get PRs associated with a given commit"](https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#list-pull-requests-associated-with-a-commit) endpoint, which does work on repositories like `apache/arrow`. For a certain default-branch commit (which is given in the alert pipeline input), we would be able to use this endpoint to comment on the PR that caused the commit. That's nice.

However it doesn't work for repositories like `facebookincubator/velox`, because they have a bot [close their PRs](https://github.com/facebookincubator/velox/pull/4297#event-8758907106) instead of going through the GitHub merge process. So that endpoint always returns an empty list. But (check out that link) I noticed the commit is still associated with that "closing" event in the UI. Given that information (and [this](https://stackoverflow.com/a/49327668) sort-of-helpful SO answer), I was able to construct a workaround GitHub GraphQL API query that maps each Velox PR to its merge-commit:
```
{
  repository(name: "velox", owner: "facebookincubator") {
    pullRequests(last: 10, states: [CLOSED]) {
      edges {
        node {
          number
          timelineItems(last: 100, itemTypes: CLOSED_EVENT) {
            edges {
              node {
                ... on ClosedEvent {
                  closer {
                    ... on Commit {
                      oid
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```
Try it out in the [GraphQL explorer](https://docs.github.com/en/graphql/overview/explorer). Given a merge-commit, we could use this query (with careful limits) to find the respective pull request. However, this seems like a lot of complexity to carry around for an uncommon use case. But I'm not sure of another way to automatically find this mapping. I suppose we could parse the commit message, but that seems brittle and wrong.

So after all that, I decided to have the person implementing `benchalerts` figure out how to provide the pull number. I'll try to ensure this is possible for all the Conbench instances we know about. And maybe we can add in "smart" functionality later.

benchalerts/benchalerts/pipeline_steps/github.py
```diff
@@ -230,6 +235,82 @@ def _default_status_description(full_comparison: FullComparisonInfo) -> str:
...
+                check_link=check_details["html_url"],
+            ),
+            pull_number=self.pr_number,
+        )
+        return res
+
+    @staticmethod
+    def _default_comment(summary: str, check_link: str) -> str:
+        """Construct a PR comment that summarizes and links to a GitHub Check."""
+        return pr_comment_link_to_check(summary, check_link)
```
alistaire47, line 311: Is this indirection helpful? It would be simpler to put the body of `pr_comment_link_to_check()` here or call the function directly on L300 unless I'm missing something.

benchalerts/benchalerts/pipeline_steps/github.py
alistaire47: Maybe we should've stored PR number on `Commit` instead of the branch...

alistaire47 (approved):

benchalerts/benchalerts/message_formatting.py
```diff
@@ -108,3 +108,16 @@ def github_check_details(full_comparison: FullComparisonInfo) -> Optional[str]:
         """
     )
     return details
+
+
+def pr_comment_link_to_check(summary: str, check_link: str) -> str:
+    """Generate a GitHub PR comment that summarizes and links to a GitHub Check."""
+    return _clean(
+        f"""
+        ## ⚡️ Benchmark results ⚡️
```
jgehrcke, line 117: The :zap: appears to be quite dramatic! :)

jgehrcke (approved):
:rocket: Lovely! To be honest, I have only looked at the surface, and in this case this really isn't enough to provide qualified feedback. Keep going!

benchalerts/benchalerts/pipeline_steps/github.py
austin3dickey: It is not very helpful today. In my head it serves two future purposes:
- Putting all string formatting code in `message_formatting.py` might reduce the impact of future formatting changes, and make it easier to abstract out common code between different flavors of markdown; e.g. github vs slack. (Today we just have the github flavor in there.)
- Putting this in a method called `_default_comment()` might allow users to override this comment in the future. I'm not sure yet what that interface would look like. I'll probably expand on this in https://github.com/conbench/conbench/issues/774.

But agreed it looks weird right now.

Issue #794, opened 2023-03-02
ci instability: prometheus-k8s-0 stuck in Pending (timeout waiting for condition on pods/prometheus-k8s-0)


austin3dickey:
I just noticed [this build on main](https://github.com/conbench/conbench/actions/runs/4315994831/jobs/7531165136) failed with
```
+ kubectl wait --timeout=90s --for=condition=Ready pods prometheus-k8s-0 -n monitoring
error: timed out waiting for the condition on pods/prometheus-k8s-0
```

jgehrcke:
Thank you for tracking that here @austin3dickey and for linking to the build log. That's exactly the right method towards understanding frequency as well as symptom specifics. Adjust the title a tiny bit to add precision. Pasting a little more of the build log here for posterity:

```
+ kubectl get pods -A
NAMESPACE       NAME                                        READY   STATUS              RESTARTS        AGE
default         acid-minimal-cluster-0                      1/1     Running             0               4m38s
default         conbench-deployment-7fd779cf89-bplxn        0/1     ContainerCreating   0               1s
default         postgres-operator-74bcb97f87-gn79s          1/1     Running             1 (4m45s ago)   4m52s
ingress-nginx   ingress-nginx-admission-create-8kx5f        0/1     Completed           0               5m6s
ingress-nginx   ingress-nginx-admission-patch-fwc2n         0/1     Completed           0               5m6s
+ sleep 1
ingress-nginx   ingress-nginx-controller-6cfc444f5f-qfx75   1/1     Running             0               5m6s
kube-system     coredns-57575c5f89-dsml6                    1/1     Running             0               5m50s
kube-system     coredns-57575c5f89-v2j82                    1/1     Running             0               5m50s
kube-system     etcd-mk-conbench                            1/1     Running             0               6m5s
kube-system     kube-apiserver-mk-conbench                  1/1     Running             0               6m2s
kube-system     kube-controller-manager-mk-conbench         1/1     Running             0               6m2s
kube-system     kube-proxy-p2b5c                            1/1     Running             0               5m50s
kube-system     kube-scheduler-mk-conbench                  1/1     Running             0               6m2s
kube-system     storage-provisioner                         1/1     Running             0               5m48s
monitoring      alertmanager-main-0                         2/2     Running             1 (3m2s ago)    3m7s
monitoring      alertmanager-main-1                         2/2     Running             1 (3m1s ago)    3m7s
[...]

PR #883, opened 2023-03-15
landing page: prevent redraw/layout switch during datatables processing

jgehrcke:
For #884.

This layout switch can be conceptually tamed by not showing the table until DataTables is done with its logic.

This is a strategy also discussed in https://datatables.net/forums/discussion/61332/eliminate-page-flashing-redrawing.

And in case someone wonders: yes, we want to require JavaScript for the Conbench UI :-)

### DIFF
--- a/conbench/templates/index.html
+++ b/conbench/templates/index.html
@@ -24,7 +24,7 @@
 
 <!-- https://getbootstrap.com/docs/5.2/content/tables/#responsive-tables -->
 <div class="table-responsive">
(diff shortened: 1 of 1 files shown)


jgehrcke:
I kept note of the `ci / cb-on-minikube` CI instability in the corresponding ticket: https://github.com/conbench/conbench/issues/794#issuecomment-1471643118

jgehrcke:
Moving ahead with this!

Issue #888, opened 2023-03-17
repo URLs: implicit requirement for github.com? prepare for other prefixes


jgehrcke:
This uses `repository_to_name()` and `repository_to_url()` from current `commit.py`, and shows a problem that I think can happen within Conbench as of today:

```python
>>> repository_to_url('https://gitlab.com/rofl/lol')
'https://github.com/https://gitlab.com/rofl/lol'
```

I believe we have an implicit assumption that user-given repository information -- provided as URL or not -- always relates to github. I think that is not always verified in the convenient places (input, before inserting into DB) -- and that may lead to URL construction bugs when this assumption about input is bad.

Note, I tested this in an isolated REPL after copy/pasting the functions in:

```
$ python
Python 3.11.2 (main, Feb  8 2023, 00:00:00) [GCC 12.2.1 20221121 (Red Hat 12.2.1-4)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> def repository_to_name(repository):
...     if not repository:
...         return ""
...     name = repository
...     if "github.com/" in repository:
...         name = repository.split("github.com/")[1]
...     elif "git@github.com:" in repository:
...         name = repository.split("git@github.com:")[1]
...     return name
... 
>>> 
>>> def repository_to_url(repository):
...     name = repository_to_name(repository)
...     return f"https://github.com/{name.lower()}" if name else ""
... 
>>> 
```

jgehrcke:
It looks like this can indeed happen when user-given input is an empty string, which seems to pass marshmallow schema validation. Tested that: https://github.com/marshmallow-code/marshmallow/issues/76#issuecomment-1473348472

Issue #887, opened 2023-03-16
commit: URL construction based on user-given data should not generally lower()


jgehrcke:
Side discussion from #818. 

> we might want to continue storing URLs lowercase

This is about https://github.com/conbench/conbench/blob/782abc9e2c1147866eb251acb612192d8fded5d3/conbench/entities/commit.py#L265.

Hmm. Danger! :fire:  I mean, I don't know about the fallout, but URLs in general are case-sensitive.

github.com documentation says

> The repository you're trying to fetch must exist on GitHub.com, **and the URL is case-sensitive**.

(https://docs.github.com/en/repositories/creating-and-managing-repositories/troubleshooting-cloning-errors)

> Typos happen, and repository names are case-sensitive. If you try to clone git@github.com:user/repo.git, but the repository is really named User/Repo you will receive this error.
https://stackoverflow.com/q/6650488/145400

> GitHub recently made a change that now allows mixed-case repository names.

(from https://metacpan.org/pod/App::GitHub::FindRepository)

There is value in auto-correcting user-given input _if_ we know the correction is, well, correct :)

austin3dickey:
Yeah, it probably makes sense not to `.lower()` the repository identifier. I think this exists in the codebase as a hack from when we lowercased a private repository and chose not to migrate the database. We probably didn't realize that repo names are case-sensitive when interacting with GitHub.

jgehrcke:
btw, for the record to quickly show what this is about:
```
>>> repository_to_url('peter/PaN')
'https://github.com/peter/pan'
```

jgehrcke:
This was resolved via https://github.com/conbench/conbench/pull/1134.

Issue #817, opened 2023-03-07
db schema: Commit object: require hash and repository specifier?


jgehrcke:
I see that we have code that may create a Commit object in the database that for example has no commit hash set, but a repository specifier: 

https://github.com/conbench/conbench/blob/408cba4db5ac914212b0be9cfd1bd0ceeccc7ff0/conbench/entities/run.py#L110

```python

            elif sha or repository:
                commit = Commit.create_unknown_context(sha, repository)
            else:
                commit = Commit.create_no_context()
```

Let's simplify here if possible, and impose constraints.

jgehrcke:
Update: the `sha` and `repository` columns are `NotNull()`-wrapped: 
https://github.com/conbench/conbench/blob/408cba4db5ac914212b0be9cfd1bd0ceeccc7ff0/conbench/entities/commit.py#L34

IIUC that means that the situation I described above cannot happen (that a Commit gets inserted into the DB specifying only one of both).

In the `GitHubCreate` schema both are also required, so the `elif sha or repository` implies that both are set.

jgehrcke:
Turns out we _do_ allow for zero-length string in the commit hash, when the repository is set. See https://github.com/conbench/conbench/pull/816#issuecomment-1458629359.

Quoting @jonkeane:

> The original motivation for the looseness here was to also allow us to store results on other softwares (some where we might not have a specific hash that we know we're running)

Yes in this case we should not create a `Commit` object associated with the benchmark result I think.

jgehrcke:
There is a test today that confirms that inserting a Commit object with empty string hash (only repo specifier) works: https://github.com/conbench/conbench/blob/020aabc67baed285dfb3ea0d179839bf3d8d7a67/conbench/tests/api/test_benchmarks.py#L818

Let's review this.

jgehrcke:
This relates to https://github.com/conbench/conbench/issues/814. Both discussions will resolve once we have a clearer picture about intent.

jonkeane:
I did a bit of git blame, and https://github.com/conbench/conbench/pull/147 is the source of that test. Unfortunately there's not much (well any!) description in the PR. But judging by the date + what I remember from around that time, what I said above is the source (we wanted to store arbitrary other results where we (might not) know the commit info). 

[...]

PR #889, opened 2023-03-17
commit in db: "no context" for partial data, guarantee that "unknown context" has repo+hash, repo URL-related code consolidation

jgehrcke:
This is a PR mainly towards making code a little clearer.

It's a rather intermediate state towards more rigor.

There is small change in internal behavior, making it so that less empty-string values are inserted into the database. It's better to have two states (None, non-empty string) instead of three states.

This tiny bit of type normalization from empty string to None here and there has yielded some insights.

That changes a code path into something that makes a lot of sense, I think: "no context" commit objects instead of "unknown context". "unknown context" now _requires_ having a commit hash and repository URL set. So, this is a good step forward for #817.

Also see

https://github.com/conbench/conbench/issues/888
https://github.com/conbench/conbench/issues/887

### DIFF
--- a/conbench/app/benchmarks.py
+++ b/conbench/app/benchmarks.py
@@ -158,6 +158,8 @@ class RunMixin:
         # display_message really seems to be name of the link
         run["commit"]["display_message"] = commit_message
         if not commit_message:  # None or empty string
+            # Note(JP): I wonder where the "url" ever got set.
+            # Maybe in the _Serializer(EntitySerializer) for Commit?
             if run["commit"].get("url"):  # be real conservative
                 run["commit"]["display_message"] = run["commit"].get("url")
             else:
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -177,7 +177,7 @@ class Commit(Base, EntityMixin):
         # https://github.com/conbench/conbench/issues/817
         assert hash is not None
         assert len(hash)
-        assert repo_url.startswith('http')
+        assert repo_url.startswith("http")
 
         return Commit.create(
             {
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -38,10 +38,15 @@ class Commit(Base, EntityMixin):
     branch: Mapped[Optional[str]] = Nullable(s.String(510))
     fork_point_sha: Mapped[Optional[str]] = Nullable(s.String(50))
     parent: Mapped[Optional[str]] = Nullable(s.String(50))
+
+    # This is meant to be the URL to the repository without trailing slash.
+    # Should be renamed to repo_url
     repository: Mapped[str] = NotNull(s.String(100))
     message: Mapped[str] = NotNull(s.String(250))
     author_name: Mapped[str] = NotNull(s.String(100))
     author_login: Mapped[Optional[str]] = Nullable(s.String(50))
+
+    # I think this is guaranteed to be a URL. Is it?
     author_avatar: Mapped[Optional[str]] = Nullable(s.String(100))
     # Note(JP): tz-naive datetime, git commit author date, in UTC.
     # Edit: adding the type Optional[datetime] is not sufficient because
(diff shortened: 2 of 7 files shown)


jgehrcke:
Would love to have your eyes on this @austin3dickey, also (and in particular) after merging this.

conbench/tests/api/test_benchmarks.py
```diff
@@ -826,7 +826,9 @@ def test_create_allow_just_repository(self, client):
         new_id = response.json["id"]
         benchmark_result = BenchmarkResult.one(id=new_id)
         assert benchmark_result.run.commit.sha == ""
-        assert benchmark_result.run.commit.repository == ARROW_REPO
+
+        # new code path: no context, not unknown context
+        # assert benchmark_result.run.commit.repository == ARROW_REPO
         assert benchmark_result.run.commit.parent is None
```
jgehrcke, line 832: note: 'no context' should mean that the `commit` attr is None or something comparable. It should not have contents.

conbench/tests/api/test_benchmarks.py
```diff
@@ -855,7 +858,8 @@ def test_create_allow_just_sha(self, client):
         response = client.post("/api/benchmarks/", json=data)
         new_id = response.json["id"]
         benchmark_result = BenchmarkResult.one(id=new_id)
-        assert benchmark_result.run.commit.sha == "something something"
+        # new code path: no context, not unknown context
+        # assert benchmark_result.run.commit.sha == "something something"
         assert benchmark_result.run.commit.repository == ""
```
jgehrcke, line 863: same here

conbench/app/benchmarks.py
```diff
@@ -158,6 +158,8 @@ def _augment(self, run):
         # display_message really seems to be name of the link
         run["commit"]["display_message"] = commit_message
         if not commit_message:  # None or empty string
+            # Note(JP): I wonder where the "url" ever got set.
```
austin3dickey, line 161: /shrug I have no idea what commit url is supposed to be. This is probably dead code

conbench/entities/commit.py
```diff
@@ -38,10 +38,15 @@ class Commit(Base, EntityMixin):
...
     parent: Mapped[Optional[str]] = Nullable(s.String(50))
+
+    # This is meant to be the URL to the repository without trailing slash.
+    # Should be renamed to repo_url
     repository: Mapped[str] = NotNull(s.String(100))
     message: Mapped[str] = NotNull(s.String(250))
     author_name: Mapped[str] = NotNull(s.String(100))
     author_login: Mapped[Optional[str]] = Nullable(s.String(50))
+
+    # I think this is guaranteed to be a URL. Is it?
```
austin3dickey, line 49: I think so too

conbench/entities/commit.py
```diff
@@ -147,6 +152,9 @@ def commit_ancestry_query(self) -> Query:
 
     @staticmethod
     def create_no_context():
+        # Special commit row, a singleton that call results can relate to (in
+        # DB, via forgeign key) that have _no_ commit information set. But: is
+        # that needed? Why have that relation at all then?
```
austin3dickey, line 157: This is probably for the case where someone wants to use Conbench without supplying any `git` information. That use case is very limited in features but theoretically still possible.

jgehrcke: Right, question is if we need a special commit row to represent this case (and that the result can be db-linked with, via foreign key), or if that foreign key relationship is allowed to not exist. Not sure if I use the right terminology here.

conbench/app/benchmarks.py
jgehrcke: By now I think I know this comes from here: https://github.com/conbench/conbench/blob/fa5663866a753d8c2e22ec66d5477aca8931ca2c/conbench/entities/commit.py#L241

i.e. this challenging to answer because of an unnecessary indirection, the API layer re-serialization plus augmentation, analogue to what's described in https://github.com/conbench/conbench/issues/890#issue-1629055001.

conbench/entities/commit.py
austin3dickey: Oh I think I see what you're saying. Right now, we have this:
https://github.com/conbench/conbench/blob/f646b2b0a4c5ab3364bf41a8a35a8fe541e97b38/conbench/entities/run.py#L62
The question is, can we make this foreign key nullable, and use a NULL value when commit information doesn't exist, instead of creating a "dummy" commit with no information? Yes, I think that's probably better practice.

jgehrcke: Thanks for confirming!

jgehrcke: changing this here: https://github.com/conbench/conbench/pull/1134

Issue #890, opened 2023-03-17
landing page: skip API layer re-serialization of all runs


jgehrcke:
Today:

```
DB query -> run objects -> JSON-serialization -> JSON-deserialization -> augmentation ->template injection
```

Go from DB query to HTML template interpolation right away.

Future:

```
DB query -> run objects -> augmentation -> template injection
```

This is good for maintainability, readability, performance, and more.

Related: #799

jgehrcke:
Done in https://github.com/conbench/conbench/pull/891.

Issue #886, opened 2023-03-16
UI: landing page: show one `recent runs` table per repo, remove repo column


jgehrcke:
Strongly relates to https://github.com/conbench/conbench/issues/885.

When thinking about how to display information then an interesting scenario is when there is a table that has a column where _all_ rows share the same value. This is what's happening with the repository column in the `Recent runs` table for most/all? our Conbench deployments.

jgehrcke:
closing as 'done'.

Issue #787, opened 2023-03-01
Think about what "listing benchmarks" should mean


jgehrcke:
Motivated by https://github.com/conbench/conbench/issues/772 and by https://github.com/conbench/conbench/issues/560. 

This needs API work and design work, before entering UI work space.

This is not a no-brainer to add because of ambiguity of existing design and concepts, as for example expressed in these tickets:
https://github.com/conbench/conbench/issues/401
https://github.com/conbench/conbench/issues/561
https://github.com/conbench/conbench/issues/786

We have a great opportunity here (and also need) to remove ambiguity and enhance clarity to define properly what it means to "list benchmarks".

This is supposed to be a bit of a thought dump ticket.

jgehrcke:
I want to try building on the thought I expressed in #561:

> I think even with ambiguity about what the term "benchmark" really means to others, we have to take some kind of stance, in order to make documentation and UI simple. For great UX, too much abstraction and flexibility may hurt. I think we may want to be strict about the idea that an individual benchmark has to have an expressive name which is then constant across case permutations.

Therefore, I like assigning special meaning to the benchmark _name_ when thinking about benchmark identity.

Benchmark identity: pieces of information required to describe a specific benchmark (given current Conbench ecosystem).

A Conbench instance has (implicit) multi-repository support. It seems like the combination of a benchmark name and a code repository specifier (URL? org/repo notation?) can uniquely define a specific benchmark concept. A special case is where a benchmark result is not associated with a commit (and therefore not with a specific code repository). Then we only have the benchmark name to indicate the benchmark's identity.

(complication in the future: when a benchmark concept stays the same but the benchmark code changes then Conbench should allow for that -- ASV seems to have the concept of benchmark version built-in; identified through code hashing (very opinionated, not really flexible)).

So, when I ask Conbench to "list benchmarks in the UI" I think

- the first (visual) categorization should be by code repository `R`.
- within `R`, the categorization should be by benchmark name `N` (idea for future: bm version).
- within a specific `R-N`, a listing of known case permutations `P_n` is interesting: `P_1, P_2, P_3, ...` (idea for future work: what's also interesting: metadata about benchmark author/owner, a description for the benchmark concept, and a link to where its code lives)
- within a specific `R-N-P_n`, we can then show/link to individual benchmark results.

jonkeane:
I really like where you're going with this. I love the simplicity + breaking down into exactly what's needed to identify + target the right concept in this view.

[...]

PR #891, opened 2023-03-17
UI: multi-repo embracement; recent runs table per-repo, skip API reserialization roundtrip

jgehrcke:
This PR touches on a variety of topics.

Mainly, this is for #886 and therefore relevant work towards https://github.com/conbench/conbench/issues/787.

It's also for https://github.com/conbench/conbench/issues/890.

### DIFF
--- a/conbench/app/index.py
+++ b/conbench/app/index.py
@@ -1,12 +1,5 @@
 from collections import defaultdict
-
-
 from dataclasses import dataclass
-
-
-from typing import Optional
-
-import flask as f
 
 from sqlalchemy import select
 
--- a/conbench/app/index.py
+++ b/conbench/app/index.py
@@ -1,5 +1,9 @@
+import logging
+
 from collections import defaultdict
 from dataclasses import dataclass
+from urllib.parse import urlparse
+
 
 from sqlalchemy import select
 
@@ -9,6 +13,8 @@ from ..app.benchmarks import RunMixin
 from ..config import Config
 from ..db import Session
 from ..entities.run import Run
+
+log = logging.getLogger(__name__)
 
 
 class Index(AppEndpoint, RunMixin):
@@ -34,7 +40,7 @@ class Index(AppEndpoint, RunMixin):
         # consistency between benchmark results in the run is currently not
         # granted: https://github.com/conbench/conbench/issues/864
 
-        repo_runs_map = defaultdict(list)
+        reponame_runs_map = defaultdict(list)
 
         for r in runs:
             rd = RunForDisplay(
@@ -42,18 +48,40 @@ class Index(AppEndpoint, RunMixin):
                 commit_message_short=short_commit_msg(r.commit.message),
                 run=r,
             )
-            repo_runs_map[r.commit.repository].append(rd)
 
-        return self.page(repo_runs_map)
+            rname = repo_url_to_display_name(r.commit.repo_url)
+            reponame_runs_map[rname].append(rd)
+
+        return self.page(reponame_runs_map)
+
+
+def repo_url_to_display_name(url: str) -> str:
+    try:
+        result = urlparse(url)
+    except ValueError as exc:
+        log.warning("repo_url failed urlparse(): %s, %s", url, exc)
+        # In this case, don't care about cosmetics: display the 'raw' data.
+        return url
+
+    if result.path == "":
+        # In this case, don't care about cosmetics: display the 'raw' data.
+        return url
+
+    # A common case is that there now is a leading slash. Remove that. Note
+    # that `strip()` also operates on the trailing end. I think there shouldn't
+    # be a trailing slash, but if it's there, remove it, too.
+    return result.path.strip("/")
 
 
 @dataclass
 class RunForDisplay:
     ctime_for_table: str
     commit_message_short: str
-    # Also expose the raw Run object
+    # Expose the raw Run object (but this needs to be used with a lot of
+    # care, in the template -- for VSCode supporting Python variable types and
+    # auot-completion in a jinja2 template see
+    # https://github.com/microsoft/pylance-release/discussions/4090)
     run: Run
-    # commit_url: Optional[str]
 
 
 def short_commit_msg(msg: str):
--- a/conbench/app/index.py
+++ b/conbench/app/index.py
@@ -1,9 +1,7 @@
 import logging
-
 from collections import defaultdict
 from dataclasses import dataclass
 from urllib.parse import urlparse
-
 
 from sqlalchemy import select
 
--- a/conbench/app/index.py
+++ b/conbench/app/index.py
@@ -34,7 +34,7 @@ class Index(AppEndpoint, RunMixin):
             select(Run).order_by(Run.timestamp.desc()).limit(1000)
         ).all()
 
-        # Note(JP): group runs by associated commit.repository value Note that
+        # Note(JP): group runs by associated commit.repository value. Note that
         # consistency between benchmark results in the run is currently not
         # granted: https://github.com/conbench/conbench/issues/864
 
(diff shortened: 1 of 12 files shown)


conbench/app/index.py
```diff
@@ -1,25 +1,76 @@
...
 
     @authorize_or_terminate
     def get(self):
-        runs = self.get_display_runs()
-        return self.page(runs)
+        # Following
+        # https://docs.sqlalchemy.org/en/20/orm/queryguide/select.html#selecting-orm-entities
+        runs = Session.scalars(
+            select(Run).order_by(Run.timestamp.desc()).limit(1000)
+        ).all()
```
jgehrcke, line 35: `runs = self.get_display_runs()` is needless complexity (as of going through HTTP API layer re-serialization, as described in #890). The inferred type of `runs` here was `Any`... it's a dictionary reflecting the _augmented_ JSON-deserialized version of what the HTTP/JSON API would return for a Run. These layers of indirection make it hard to reason about the code. 

The diff here removes the indirections. And therefore will resolve https://github.com/conbench/conbench/issues/890.

CC @austin3dickey 

btw using this SQLAlchemy 2.0 way of building up the query we get nice intellisense features, and proper type inference:
```
        runs = Session.scalars(
            select(Run).order_by(Run.timestamp.desc()).limit(1000)
        ).all()
```
Screenshot from 2023-03-16 13-04-39

conbench/app/index.py
```diff
@@ -1,25 +1,76 @@
...
+        # consistency between benchmark results in the run is currently not
+        # granted: https://github.com/conbench/conbench/issues/864
+
+        repo_runs_map = defaultdict(list)
+
+        for r in runs:
+            rd = RunForDisplay(
+                ctime_for_table=r.timestamp.strftime("%Y-%m-%d %H:%M:%S UTC"),
+                commit_message_short=short_commit_msg(r.commit.message),
+                run=r,
```
jgehrcke, line 47: this is still experimentation in the WIP state of this PR. Basically I think we want to pass the raw `Run` object into the template, but in addition to that we also need some variables tailored for displaying in HTML. The latter is what the "augment" family of functions did before. I'd love to do this closer to where the HTML template is rendered.

conbench/entities/_entity.py
```diff
@@ -58,6 +58,19 @@ def all(cls, limit=None, order_by=None, filter_args=None, **kwargs):
...
+        # extensions) become long term legacy objects, replaced by the direct
+        # usage of the select() [...] Because the vast majority of an ORM
+        # application is expected to make use of Query objects as well as that
+        # the Query interface being available does not impact the new
+        # interface, the object will stay around in 2.0 but will no longer be
+        # part of documentation nor will it be supported for the most part. The
+        # select() construct now suits both the Core and ORM use cases, which
+        # when invoked via the Session.execute() method will return
+        # ORM-oriented results, that is, ORM objects if that’s what was
+        # requested.""
```
jgehrcke, line 73: @austin3dickey :)

conbench/entities/commit.py
```diff
@@ -301,6 +363,18 @@ def repository_to_url(repository: str) -> str:
         # be an empty string, I think.
         return ""
 
+    # Note(JP): `name` may still be a URL. In that case, return this.
```
jgehrcke, line 371: instead of producing `https://github.com/https://hello.kitty`

conbench/templates/app.html
```diff
@@ -212,12 +211,12 @@ <h4 class="modal-title"><i class="bi bi-trash3-fill"></i></h4>
     Based on the indexes of the columns, input boxes are created. The search is triggered on
     key press.
 INPUTS:
- data_table_id: string
+ data_table_elem: DOM element for the <table>
  column_indexes: array
  */
-function free_text_column_search(data_table_id, column_indexes) {
+function free_text_column_search(data_table_elem, column_indexes) {
```
jgehrcke, line 200: CC @sayantikabanik by passing in a DOM element instead of a CSS selector string we can use these functions more flexibly (when a CSS selector string resolves to _multiple_ DOM elements).

conbench/app/index.py
austin3dickey: Isn't that nice?! In #859 I considered [changing](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-orm-usage) all the queries to use the new  2.0-style `Session.scalars()` (et cetera) interfaces. At the time I chose not to for a few reasons:
- it would take a little time
- the docs say there are no plans to deprecate the `Session.query()` interface
- the new interface looks a little more complicated to read on the surface (I understand it's more clear and explicit but it just looks a little... uglier?)
- we get the same intellisense features and type inference from the old interface :)

But I'm definitely not against using this interface for new code moving forward!

conbench/entities/_entity.py
austin3dickey: Oops should have posted this comment here: https://github.com/conbench/conbench/pull/891/files#r1140385420

Agreed that someday we should change it, but it seems to me like there are higher priorites in the meantime.

conbench/app/index.py
jgehrcke: > At the time I chose not to for a few reasons: it would take a little time ...

The right choice, let's do it gradually!

conbench/app/index.py
```diff
@@ -1,25 +1,102 @@
...
-    def page(self, runs):
+    def page(self, repo_runs_map):
         return self.render_template(
             "index.html",
             application=Config.APPLICATION_NAME,
             title="Home",
-            runs=runs,
-            search_value=f.request.args.get("search"),
+            repo_runs_map=repo_runs_map,
+            # Note(JP): search_value is not consumed in template
```
jgehrcke, line 25: comment is out of date

conbench/app/index.py
```diff
@@ -1,25 +1,102 @@
...
-        return self.page(runs)
+        # Following
+        # https://docs.sqlalchemy.org/en/20/orm/queryguide/select.html#selecting-orm-entities
+        runs = Session.scalars(
+            select(Run).order_by(Run.timestamp.desc()).limit(1000)
+        ).all()
+
+        # Note(JP): group runs by associated commit.repository value Note that
+        # consistency between benchmark results in the run is currently not
+        # granted: https://github.com/conbench/conbench/issues/864
```
austin3dickey, line 39: I don't think we need to worry about this too much; I commented on that issue. Looks like you aren't worrying about it here beyond the comment. 🙂

conbench/templates/run.html
```diff
@@ -182,7 +182,7 @@ <h4>Benchmarks in run</h4>
 } );
 
 enable_search_query_string('#benchmarks');
```
austin3dickey: Didn't you comment this out?

conbench/tests/populate_local_conbench.py
```diff
@@ -579,5 +608,64 @@ def sample_slowdown(s):
     """
 
 
+"""
+50 commits of conbench/conbench walking backwards from 782abc9e. Obtained with
+this command:
+
+git log --pretty=%P -n 50 782abc9e2c1147866eb251acb612192d8fded5d
+"""
+CONBENCH_COMMIT_HASH_LINES_50 = """
```
austin3dickey, line 623: Beautiful, thanks for adding new test data!

austin3dickey (approved):
Looks great!

conbench/templates/run.html
jgehrcke: Thanks for noticing! Pushing another commit.

jgehrcke:
As discussed in our sync meeting on Friday, I have added alphabetical repository enumeration order for the per-repo 'recent runs table' listing on the landing page.

jgehrcke:
CI passed finally, going to land this now. Thanks for the valuable feedback everyone!

Issue #892, opened 2023-03-17
DB schema: bump length limit for repository URL property in Commit table


jgehrcke:
In #818 we found that

> The 100 makes me think that it's not URL. Is 100 too limited btw?
> Agreed 100 is too short for URLs

This ticket tracks increasing the limit.

jgehrcke:
Done in #893.

I think the new limit (300) allows for sane use cases. It's still an arbitrary limit though. In general, URLs are of course allowed to be much longer. For popular code repository management solutions and reasonable repository naming choices though it's rather unlikely that a URL pointing to a repository is ever longer than that. We will see. :)

PR #893, opened 2023-03-17
db schema: bump length limit for repository URL property on Commit table

jgehrcke:
For #892.

Created the migration with the new makefile target -- nice.


jgehrcke:
All failures in ci/testsuite are as of `Exception: Hourly GitHub HTTP API quota exhausted` -- tracked here: #895

conbench/entities/commit.py
```diff
@@ -54,6 +58,7 @@ class Commit(Base, EntityMixin):
     timestamp: Mapped[Optional[datetime]] = Nullable(s.DateTime(timezone=False))
 
     def get_parent_commit(self):
+        # Hm -- should this not be done with a foreign key relationship?
```
austin3dickey, line 61: I think this used to be a lot more important while traversing the git graph, but it's not really used anymore. So I don't think we need to try to optimize it. We might even get rid of this method.

austin3dickey (approved):

Issue #878, opened 2023-03-15
UI: initiative to replace some/all "benchmark" with "benchmark result" whenever appropriate


jgehrcke:
This terminology challenge needs to be addressed through docs and UI. This ticket tracks doing that in the UI.

URL paths for views/API endpoints are harder to change, so this is not entailed in this ticket.

From #561:

> The first-class citizen is the case (though semi-confusingly we refer to this as "case" "benchmark" and "benchmark[ ]result" in a few places!).

>  (e.g. rename benchmark_result to case_result)

While I like the idea, we don't yet need to pull through with it. First achieve clarity about concepts. Then use names _consistently_. Then things are already nicer. Then, a rename is just a rename, without thinking too hard.

jgehrcke:
Some work was done here:
- https://github.com/conbench/conbench/pull/876
- https://github.com/conbench/conbench/pull/875

Issue #867, opened 2023-03-14
history plot: do not plot if there are less than e.g. three data points


jgehrcke:
@alistaire47 found this example:

![image](https://user-images.githubusercontent.com/265630/225038648-f1a8fe70-32a9-4e84-b893-ad67ae3d1eec.png)

jgehrcke:
I think this has been addressed via https://github.com/conbench/conbench/pull/896.

PR #896, opened 2023-03-17
history plot: show info/note when not enough data points (instead of empty plot)

jgehrcke:
This addresses #867.

For quick local dev I didn't create an example with just one data point, but I bumped the limit to create this screenshot:

Screenshot from 2023-03-17 15-09-01

Edit: for https://github.com/conbench/conbench/issues/878 this also renamed benchmark-entity.html to benchmark-result.html. Did the `git mv` in its own commit so that ideally git blame knows what happened.


conbench/app/_plots.py
```diff
@@ -18,6 +19,13 @@
 log = logging.getLogger(__name__)
 
 
+@dataclasses.dataclass
+class BokehPlotJSONOrError:
+    # mutually exclusive
+    jsondoc: Optional[str]
+    reason_why_no_plot: Optional[str]
```
jgehrcke, line 26: tiny types like this are not set in stone, can and should always be iterated upon -- but they really really help navigating code quickly :)

austin3dickey (approved):
lovely!

jgehrcke:
Oh:
```
conbench/app/_plots.py:62: error: Argument 2 to "json_item" has incompatible type "str"; expected "Optional[ID]"  [arg-type]
```

PR #903, opened 2023-03-17
benchalerts: reformat comments

### LINKED ISSUE #900: benchalerts: change default PR comment heading
@jgehrcke recently mentioned that this PR comment header might be too off-brand:

## ⚡️ Benchmark results ⚡️

Instead, We might want to switch to something a little smoother like this:

### Benchmark results

austin3dickey:
This is a small PR to reformat the PR comments that `benchalerts` posts. Closes #900.

It posts the first 20 lines of the benchmark report to the comment, instead of just linking to the comment.

### DIFF
--- a/benchalerts/tests/integration_tests/test_alert_pipeline.py
+++ b/benchalerts/tests/integration_tests/test_alert_pipeline.py
@@ -107,11 +107,11 @@ def test_alert_pipeline(monkeypatch: pytest.MonkeyPatch, github_auth: str):
         assert outputs["GitHubCheckStep"]["conclusion"] == "failure"
         if not os.getenv("CI"):
             assert outputs["GitHubPRCommentAboutCheckStep"]["body"].startswith(
-                """## ⚡️ Benchmark results ⚡️
-
-Found 1 regression(s).
-
-See the full report [here]("""
+                "A [full benchmark report](https://github.com/conbench/benchalerts/runs/"
+            )
+            assert (
+                "Contender commit `c76715c9` had 1 performance regression(s) compared to its baseline runs."
+                in outputs["GitHubPRCommentAboutCheckStep"]["body"]
             )
 
     # sleep to see the updated statuses on the PR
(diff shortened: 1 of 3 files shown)


austin3dickey:
Here's an example of the new comment. https://github.com/conbench/benchalerts/pull/5#issuecomment-1474389515

image

benchalerts/tests/integration_tests/test_alert_pipeline.py
```diff
@@ -107,11 +107,11 @@ def test_alert_pipeline(monkeypatch: pytest.MonkeyPatch, github_auth: str):
...
-                """## ⚡️ Benchmark results ⚡️
-
-Found 1 regression(s).
-
-See the full report [here]("""
+                "A [full benchmark report](https://github.com/conbench/benchalerts/runs/"
+            )
+            assert (
+                "Contender commit `c76715c9` had 1 performance regression(s) compared to its baseline runs."
+                in outputs["GitHubPRCommentAboutCheckStep"]["body"]
```
jgehrcke, line 114: It's not the best place to comment on this. But for the future, I think we maybe want to add precision to "1 performance regression(s) compared to its baseline runs." 

It's a performance regression detected with the help of the rolling window z-score method (maybe that's all that needs to be said?)

That method has a specific way to look at history. I think you maybe tried to reflect that with the plural "runs" in "its baseline runs". So far, I think across Conbench, "baseline run" is used often in singular, to describe that one previous run that the contender can/should best be compared to. Here, as of the plural one might wonder: how can a contender run suddenly have more than one baseline run?

In this message there is also the conceptual terminology confusion between "contender commit" and "baseline run".

jgehrcke (approved):
:rocket:

benchalerts/tests/integration_tests/test_alert_pipeline.py
austin3dickey: Yes, this is a great callout. I introduced the "baseline runs" language last year before I truly understood how the "regression" logic coming from the Conbench API was working. I think what you said is more precise, because these comments are coming from the rolling window z-score method, not the "compare to one previous run" method.

Issue #785, opened 2023-02-28
api: POST to /api/runs returns 500 respo upon run primary key collision (instead of 409 response)


austin3dickey:
Recently, @alistaire47 tried to POST a run with an ID that already existed in the database. Symptom:
```
DEBUG [2023-02-28 05:44:09,647] POST https://conbench.ursa.dev//api/runs/ {"name": "Test Build for branch=edward/direct-running-testing and commit=35b961db6ddf05c34df6f1d6f999f5525404631c", "id": "1c0cc72cafd240a6af7277a2f9c1e589", "reason": "test", "info": {}, "machine_info": {"name": "arm64-m6g-linux-compute", "os_name": "Linux", "os_version": "4.14.248-189.473.amzn2.aarch64-aarch64-with-glibc2.17", "architecture_name": "aarch64", "kernel_name": "4.14.248-189.473.amzn2.aarch64", "memory_bytes": "64424509440", "cpu_model_name": "Neoverse-N1", "cpu_core_count": "16", "cpu_thread_count": "16", "cpu_l1d_cache_bytes": "65536", "cpu_l1i_cache_bytes": "65536", "cpu_l2_cache_bytes": "1048576", "cpu_l3_cache_bytes": "33554432", "cpu_frequency_max_hz": "0", "gpu_count": "0", "gpu_product_names": []}, "github": {"repository": "https://github.com/apache/arrow", "commit": "0e752c8655685a0108c3eed3de27e7981db688c7", "pr_number": null}}
ERROR [2023-02-28 05:44:09,873] Response content: {"code": 500, "name": "Internal Server Error"}
```
Server logs:
```
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "run_pkey"
DETAIL:  Key (id)=(1c0cc72cafd240a6af7277a2f9c1e589) already exists.
```
This should instead have returned a 409 Conflict or 422 Unprocessable Entity status code, with an error message detailing the uniqueness index that was violated and how.

This probably could be extended to other POST endpoints that have uniqueness constraints.

jgehrcke:
409 is a great choice (I have always done that when I thought wayyy to much about using HTTP as nicely as possible :man_shrugging:). Also, in the error response body there should be structured error information with a nice, precise error message.

And of course this should be covered by tests.

[...]

PR #905, opened 2023-03-20
api: emit 409 response upon run ID conflict

jgehrcke:
For issue #785.

### DIFF
--- a/conbench/api/runs.py
+++ b/conbench/api/runs.py
@@ -4,7 +4,7 @@ import flask_login
 from ..api import rule
 from ..api._docs import spec
 from ..api._endpoint import ApiEndpoint, maybe_login_required
-from ..entities._entity import NotFound
+from ..entities._entity import NotFound, EntityExists
 from ..entities.benchmark_result import BenchmarkResult
 from ..entities.commit import Commit
 from ..entities.run import Run, RunFacadeSchema, RunSerializer
@@ -136,7 +136,12 @@ class RunListAPI(ApiEndpoint):
           - Runs
         """
         data = self.validate(self.schema.create)
-        run = Run.create(data)
+
+        try:
+            run = Run.create(data)
+        except EntityExists as exc:
+            return f.jsonify(error=409, description=str(exc)), 409
+
         return self.response_201_created(self.serializer.one.dump(run))
 
 
--- a/conbench/api/runs.py
+++ b/conbench/api/runs.py
@@ -4,7 +4,7 @@ import flask_login
 from ..api import rule
 from ..api._docs import spec
 from ..api._endpoint import ApiEndpoint, maybe_login_required
-from ..entities._entity import NotFound, EntityExists
+from ..entities._entity import EntityExists, NotFound
 from ..entities.benchmark_result import BenchmarkResult
 from ..entities.commit import Commit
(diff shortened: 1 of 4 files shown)


conbench/entities/run.py
```diff
@@ -135,7 +142,14 @@ def create(data) -> "Run":
...
 
         run = Run(**data, commit_id=commit_id, hardware_id=hardware.id)
-        run.save()
+        try:
+            run.save()
+        except s.exc.IntegrityError as exc:
+            if "unique constraint" in str(exc) and "run_pkey" in str(exc):
+                raise EntityExists(
+                    f"conflict: a Run with ID {data['id']} already exists"
+                )
```
austin3dickey, line 151: I think you might need to still raise the original exception in an `else` clause here, just in case there's an `s.exc.IntegrityError` that isn't related to a `run_pkey` unique constraint

conbench/entities/_entity.py
```diff
@@ -19,6 +19,18 @@ class NotFound(NoResultFound):
     pass
 
 
+class EntityExists(Exception):
```
austin3dickey, line 22: Nice!

austin3dickey (commented):
Lookin good!

conbench/api/runs.py
```diff
@@ -136,7 +136,12 @@ def post(self):
           - Runs
         """
         data = self.validate(self.schema.create)
-        run = Run.create(data)
+
+        try:
+            run = Run.create(data)
+        except EntityExists as exc:
+            return f.jsonify(error=409, description=str(exc)), 409
```
jonkeane, line 143: Love the try first, ask permission second

conbench/entities/_entity.py
```diff
@@ -19,6 +19,18 @@ class NotFound(NoResultFound):
...
+    """
+    Custom exception representing a database conflict. Meant to be caught from
+    within HTTP request handlers, and to be translated into an HTTP response
+    with status code 409. The `msg` in `raise EntityExists(msg)` is meant to be
+    exposed directly to the HTTP client (that is, construct it mindfully: do
+    not expose meaningless detail).
+    """
```
jonkeane, line 29: Does this bubble up into the API docs? If no, should we (or did I miss it somewhere...)?

jonkeane (approved):
Thanks for this! One small ask to make sure we bubble this up into our docs, but after that (+ of course responding to @austin3dickey 's comments, this is ready to go)

conbench/entities/run.py
jgehrcke: Thank you for catching that; this was indeed absolutely my intention (the main reason for doing all this narrowing down is to re-raise all other exceptions). Will add a commit.

conbench/entities/_entity.py
jgehrcke: This docstring here is developer documentation, for how to use `EntityExists` in the webapp's Python code.

We soon want to add a generic statement to our HTTP API specification that for all POST requests for entity management a to-be-expected HTTP response is one with status code 409, typically indicating an insertion/creation conflict (but the exact nature of the error is clarified in the body of the response). We will do that. This is the same category as documenting a 404 response for "entity not found" (but also here the response body is a decisive part of API spec).

I believe we have more places in code where conflict is not yet properly handled.

PR #907, opened 2023-03-20
UI: show "unknown ID: ..." instead of "error getting benchmark" (fix #906)

### LINKED ISSUE #906: UI shows "error getting benchmark" when an invalid ID was provided.
Seen via CI screenshotting:

image

The cause here was an invalid ID. Improve that error message.

jgehrcke:
This is for addressing #906.

### DIFF
--- a/conbench/templates/app.html
+++ b/conbench/templates/app.html
@@ -86,7 +86,7 @@
             <!-- Categories: success (green), info (blue), warning (yellow), danger (red) -->
             {% for category, message in messages %}
             <div class="alert alert-{{ category }}" role="alert">
-                <i class="bi bi-lightning"></i> {{ message }}
+                <p class="fs-5"><i class="bi bi-exclamation-diamond-fill"></i> {{ message }}</p>
             </div>
             {% endfor %}
         {% endif %}
--- a/conbench/app/benchmarks.py
+++ b/conbench/app/benchmarks.py
@@ -1,3 +1,5 @@
+import logging
+
 import bokeh
 import flask as f
 import flask_login
@@ -11,6 +13,9 @@ from ..app._endpoint import AppEndpoint, authorize_or_terminate
 from ..app._plots import TimeSeriesPlotMixin
 from ..app._util import augment, display_message, display_time
 from ..config import Config
+
+
+log = logging.getLogger(__name__)
 
 
 class UpdateForm(flask_wtf.FlaskForm):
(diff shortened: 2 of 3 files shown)


conbench/templates/app.html
```diff
@@ -86,7 +86,7 @@ <h4 class="modal-title"><i class="bi bi-trash3-fill"></i></h4>
             <!-- Categories: success (green), info (blue), warning (yellow), danger (red) -->
             {% for category, message in messages %}
             <div class="alert alert-{{ category }}" role="alert">
-                <i class="bi bi-lightning"></i> {{ message }}
+                <p class="fs-5"><i class="bi bi-exclamation-diamond-fill"></i> {{ message }}</p>
```
jgehrcke, line 89: two changes:
- this icon: https://icons.getbootstrap.com/icons/exclamation-diamond-fill/
- this font size: https://getbootstrap.com/docs/5.2/utilities/text/#font-size

austin3dickey (approved):

jonkeane (approved):
Am I reading the CSS | HTML right that this modal(ish) dialog will now persist?

jgehrcke:
> Am I reading the CSS | HTML right that this modal(ish) dialog will now persist?

I think the answer is "no". If I understood the question :). The structure for this modal is always there, but typically it's empty. For populating it with content we use the Flask "flash message" paradigm which shows a message in response to the next request, but doesn't show it anymore after that.

jonkeane:
> > Am I reading the CSS | HTML right that this modal(ish) dialog will now persist?
> 
> I think the answer is "no". If I understood the question :). The structure for this modal is always there, but typically it's empty. For populating it with content we use the Flask "flash message" paradigm which shows a message in response to the next request, but doesn't show it anymore after that.

_nods_ got it, it's still an improvement IMO, so let's ship it and see how it looks. Tangentially | separately the quickness with which some of these error dialogs go away has been a bit confusing — but we can address that separately.

jgehrcke:
> nods got it, it's still an improvement IMO, so let's ship it and see how it looks. Tangentially | separately the quickness with which some of these error dialogs go away has been a bit confusing — but we can address that separately.

Oh, is that about making the modal slowly fade out, controlled by JavaScript? I removed that before I think in the larger PR for the migration towards Bootstrap 5. That is, that info/danger modal content persists until one loads a new page or hits F5 or so. W/o further user interaction, it now stays forever.

jonkeane:
> Oh, is that about making the modal slowly fade out, controlled by JavaScript? I removed that before I think in the larger PR for the migration towards Bootstrap 5. That is, that info/danger modal content persists until one loads a new page or hits F5 or so. W/o further user interaction, it now stays forever.

AAAAH ok, ok got it! Yeah I haven't bumped into it recently, so I must have missed that. Thank you! 🙏

Issue #911, opened 2023-03-20
UI: recent runs table(s): show result count per run


jgehrcke:
Especially in view of the fact that runs are created independently of benchmark results, I think it would be nice to see how many results have been submitted (so far) for a specific run.

jgehrcke:
Done in https://github.com/conbench/conbench/pull/912.

PR #912, opened 2023-03-20
UI: show per-run result count in runs table (db: load results immediately with run)

jgehrcke:
This is for #911, maybe a provocative proposal. It's a small code change, so I thought I bring up a specific proposal for review feedback:

Screenshot from 2023-03-20 17-09-34

Would love to hear opinions @jonkeane @austin3dickey @alistaire47.

### DIFF
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -35,7 +35,20 @@ class BenchmarkResult(Base, EntityMixin):
     case_id: Mapped[str] = NotNull(s.String(50), s.ForeignKey("case.id"))
     info_id: Mapped[str] = NotNull(s.String(50), s.ForeignKey("info.id"))
     context_id: Mapped[str] = NotNull(s.String(50), s.ForeignKey("context.id"))
+
+    # Follow
+    # https://docs.sqlalchemy.org/en/20/orm/basic_relationships.html#one-to-many
+    # There is a one-to-many relationship between Run (one) and BenchmarkResult
+    # (0, 1, many). Result is child. Run is parent. Child has column with
+    # foreign key, pointing to parent.
     run_id: Mapped[str] = NotNull(s.Text, s.ForeignKey("run.id"))
+    # Note that `immediate` means that " items should be loaded as the parents
+    # are loaded, using a separate SELECT statement, or identity map fetch for
+    # simple many-to-one references." Do this until we see that this is a perf
+    # problem. Use case: count results per run (that use case can be solved by
+    # adding a cache/counter into the DB, also see
+    # https://stackoverflow.com/q/45407402/145400)
+    run: Mapped[Run] = relationship("Run", lazy="immediate", back_populates="results")
 
     # These can be emtpy lists. An item in the list is of type float, which
     # includes `math.nan` as valid value.
@@ -52,7 +65,7 @@ class BenchmarkResult(Base, EntityMixin):
     # this info should probably be called something like context-info it's details about the context that are optional | we believe won't impact performance
     info: Mapped[Info] = relationship("Info", lazy="joined")
     context: Mapped[Context] = relationship("Context", lazy="joined")
-    run: Mapped[Run] = relationship("Run", lazy="select")
+
     unit: Mapped[Optional[str]] = Nullable(s.Text)
     time_unit: Mapped[Optional[str]] = Nullable(s.Text)
     batch_id: Mapped[Optional[str]] = Nullable(s.Text)
--- a/conbench/templates/index.html
+++ b/conbench/templates/index.html
@@ -28,7 +28,7 @@
     </tr>
     <tr>
         <!-- add tooltip, saying that this is DB insertion time -->
-        <th scope="col" style="width: 180px">insertion time</th>
+        <th scope="col" style="width: 180px">insertion time (result count)</th>
         <th scope="col" class="reason-col">reason</th>
         <th scope="col" style="width: 150px">hardware</th>
         <th scope="col" style="width: 170px">author</th>
@@ -41,7 +41,7 @@
   {% for r in runobjs %}
   <tr>
     <td style="white-space: nowrap;">
-        <a href="{{ url_for('app.run', run_id=r.run.id) }}">{{ r.ctime_for_table }}</a>
+        <a href="{{ url_for('app.run', run_id=r.run.id) }}">{{ r.ctime_for_table }}</a>  ({{ r.result_count }})
         {% if r.run.has_errors or r.run.error_info %}
             <i class="bi bi-exclamation-circle text-danger"
               data-toggle="tooltip"
--- a/conbench/templates/index.html
+++ b/conbench/templates/index.html
@@ -28,7 +28,8 @@
     </tr>
     <tr>
         <!-- add tooltip, saying that this is DB insertion time -->
-        <th scope="col" style="width: 180px">insertion time (result count)</th>
+        <th scope="col" style="width: 180px">insertion time</th>
+        <th scope="col">results</th>
         <th scope="col" class="reason-col">reason</th>
         <th scope="col" style="width: 150px">hardware</th>
         <th scope="col" style="width: 170px">author</th>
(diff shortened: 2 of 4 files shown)


conbench/entities/benchmark_result.py
```diff
@@ -35,7 +35,20 @@ class BenchmarkResult(Base, EntityMixin):
...
+    # (0, 1, many). Result is child. Run is parent. Child has column with
+    # foreign key, pointing to parent.
     run_id: Mapped[str] = NotNull(s.Text, s.ForeignKey("run.id"))
+    # Note that `immediate` means that " items should be loaded as the parents
+    # are loaded, using a separate SELECT statement, or identity map fetch for
+    # simple many-to-one references." Do this until we see that this is a perf
+    # problem. Use case: count results per run (that use case can be solved by
+    # adding a cache/counter into the DB, also see
+    # https://stackoverflow.com/q/45407402/145400)
+    run: Mapped[Run] = relationship("Run", lazy="immediate", back_populates="results")
```
jgehrcke, line 51: This is an interesting change in DB query behavior -- my intuition is that this doesn't change much the timing of the query. It might increase the amount of data returned by the DB by a lot (~factor 10), but probably this still takes about the same time. Of course, to be sure, one should measure.

To be clear: with this change, when selecting one or more Runs from the DB, the corresponding benchmark results are asked for right away (in the same query, I think/hope).

conbench/templates/index.html
```diff
@@ -28,7 +28,7 @@ <h4>Recent runs for {{reponame}}</h4>
     </tr>
     <tr>
         <!-- add tooltip, saying that this is DB insertion time -->
-        <th scope="col" style="width: 180px">insertion time</th>
+        <th scope="col" style="width: 180px">insertion time (result count)</th>
```
jonkeane: What do you think about adding this as a separate column? I know we want to generally keep those to a minimum, but if it's a separate column it becomes filter-able and sort-able. IMO if we're going to expose this kind of info, we should be able to filter | sort by it.

jgehrcke: > What do you think about adding this as a separate column? I know we want to generally keep those to a minimum, but if it's a separate column it becomes filter-able and sort-able. IMO if we're going to expose this kind of info, we should be able to filter | sort by it.

I was hoping for that feedback :rocket:  I agree, my favorite choice was to add a new column, but I thought I keep it more minimal first and simply ask for feedback.

conbench/entities/benchmark_result.py
austin3dickey: Yes, this will be interesting to measure. Let's be sure to deploy to the Arrow Conbench instance soon after merging this PR, monitor it, and revert if it isn't working well.

austin3dickey: Separately: if this works out to not degrade performance, we could probably refactor certain queries in this codebase that search for "other runs that have any of this run's case/context pairs" (e.g. in [get_baseline_run()](https://github.com/conbench/conbench/blob/409537e0ffa9ec2abb1e40b942d5cf7f6cb02f4d/conbench/entities/run.py#L223) and [set_z_scores()](https://github.com/conbench/conbench/blob/409537e0ffa9ec2abb1e40b942d5cf7f6cb02f4d/conbench/entities/history.py#L368))

austin3dickey (approved):
Looking good! Agreed with Jon that a second column would look better and have more utility.

conbench/entities/benchmark_result.py
austin3dickey: Another thing: did you try to generate a DB migration? Everything you touched in this PR seems to be confined to SQLAlchemy sugar/instructions, but I suppose there's a chance it touches the DB schema in some way.

jgehrcke: > Another thing: did you try to generate a DB migration?

Yes, tried! Was empty, right!

jgehrcke:
In a CB deployment I have just seen an example where a single run is associated with 3305 results. Hm, seeing that I think that maybe the immediate query is a bit extreme? With a custom-made SQL statement we could probably get the count w/o asking for all the data, that might be a middle ground. I don't know if O(1000) results per run is an outlier or the norm. Hmm.. Or we could right away do the counter-as-column, and in the DB migration to the initial population based on current data.

jonkeane:
> I don't know if O(1000) results per run is an outlier or the norm

Definitely the norm (or should be!) for when one wants to be using something like conbench. And is one of the core design usecases we have.

With O(10) or even O(100), it's usually easy enough to do eyeball comparisons. But when you get a large suite of benchmarks (like we do with Arrow) is when you need something to keep track of them + bubble up if something has gone wrong.

jgehrcke:
> Definitely the norm (or should be!) 

Thanks for the feedback on this one @jonkeane -- let's also connect the dots here with the order-of-magnitude-thinking in https://github.com/conbench/conbench/issues/787#issuecomment-1451811886.

jonkeane:
> Thanks for the feedback on this one @jonkeane -- let's also connect the dots here with the order-of-magnitude-thinking in https://github.com/conbench/conbench/issues/787#issuecomment-1451811886.

Thanks for back linking here. I just want to make sure I'm on the same page, in that comment you said:

> It's OK to have up to O(100) case permutations per benchmark, but maybe it's not OK to have O(1000) of them.
> It's OK to have up to O(1000) individual conceptual benchmarks, but maybe not O(10000).

O(1000) results is definitely within this set, yeah? Since this describes up to O(1000) conceptual benchmarks with O(100) case permutations, that's O(100 000) benchmark results per commit. Of course, we don't usually have only a single run with _all_ the benchmarks in it for each commit, so we can usually divide that number up a little bit, but even if we have O(10) runs, that's O(10 000) benchmark results per run.

Or am I missing something?

jgehrcke:
> O(1000) results is definitely within this set, yeah?

Hmmm. I didn't deeply think about if this is in alignment or conflict with those previous statements. I only quickly detected that it's a thought of the same category.

Here, we talk about the expected/planned order of magnitude of _benchmark results per run_. Quick assessment is that this is not listed in the previous lust, it's probably an independent, additional item in the list. (it's likely not truly independent, but yeah...).

> I just want to make sure I'm on the same page

 my main point is that you might be a page ahead of me! :scroll:

jgehrcke:
> Of course, we don't usually have only a single run with all the benchmarks in it for each commit, so we can usually divide that number up a little bit, but even if we have O(10) runs, that's O(10 000) benchmark results per run.

Now thinking through that. Now that I am on your page... haha. So, you are precisely saying that `benchmark results per run` is not independent of the previously made statements. That's good thinking. 

>  that's O(10 000) benchmark results per run.

I think we don't want that, as of the arguments you gave before (and we probably find more arguments).

That probably means: if the other numbers are meaningful/OK/not-horribly-wrong, we want to give new advice about how to group benchmark results into runs, before submission. This might really be a productive line of thinking.

jonkeane:
Awesome, yes I think we are on the same page too. I will also entirely admit, I hadn't thought about that other comment + those numbers until you linked it. But then when I went and looked I was like "Oh, hey this is totally in line with those!"

> if the other numbers are meaningfull/OK/not-horribly-wrong, we want to give new advice about how to group benchmark results into runs, before submission. This might really be a productive line of thinking.

Yes, absolutely. We organically did this in Arrow, and I imagine any large enough suite of benchmarks will as well. But having it written down is good. This grouping isn't _just_ even about response performance when querying the results either: if you split across hardwares (either so you can parallelize or so that you can test across different configurations), you need to do those in different runs anyway (since each run has a single hardware it's run on).

jgehrcke:
This is now updated, with a new column for showing the result count. I have left the `column_search_implementation()` unaltered (it applies to this column exactly as it applies to other columns).

Screenshot from ci-generated screenshot:
image
btw, I kind of like this monospaced font in CI :)  (even though it is an accident)

jgehrcke:
Will land this now, as of the positive feedback so far -- but please keep the feedback coming.

jonkeane:
> btw, I kind of like this monospaced font in CI :) (even though it is an accident)

Ha I was thinking the same thing (tho I assumed it was your local browser | env that was contributing 🙈 )

PR #913, opened 2023-03-20
GitHub HTTP API client: support token pool, prepare for comma-separated tokens in GITHUB_API_TOKEN

jgehrcke:
This is for addressing #917.

### DIFF
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -786,7 +786,7 @@ class GitHubHTTPApiClient:
         while time.monotonic() < deadline:
             attempt += 1
 
-            result = self._get_response_retrycore(url)
+            result = self._get_response_retry_guts(url)
 
             if result is not None:
                 return result
@@ -811,7 +811,7 @@ class GitHubHTTPApiClient:
             f"_get_response(): deadline exceeded, giving up after {time.monotonic()-t0:.3f} s"
         )
 
-    def _get_response_retrycore(self, url) -> Optional[dict]:
+    def _get_response_retry_guts(self, url) -> Optional[dict]:
         """
         Return deserialized JSON-structure or raise an exception or return
         `None` which indicates a retryable error.
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -577,8 +577,11 @@ class GitHubHTTPApiClient:
         for tc in token_candidates:
             # Remove leading and trailing whitespace.
             t = tc.strip()
-            if len(t) < 5 or len(t) > 120:
-                log.info("bad token length, ignore: %s", len(t))
+            # Sanity-check the length. The fine-grained personal tokens
+            # seemingly have a length smaller 100, but don't try to be precise
+            # here.
+            if len(t) < 5 or len(t) > 130:
+                log.info("unexpected token length, ignore: %s", len(t))
             else:
                 tokens_to_use.append(t)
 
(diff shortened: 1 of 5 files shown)


conbench/entities/commit.py
```diff
@@ -532,6 +534,62 @@ def __init__(self):
...
+            return
+
+        if len(tokens_to_use) == 1:
+            self._current_auth_token = tokens_to_use[0]
+            log.info("configured one GitHub HTTP API auth token")
+            return
+
+        self._token_pool = itertools.cycle(tokens_to_use)
+        self._current_auth_token = next(self._token_pool)
+        log.info("current auth token starts with: %s", self._current_auth_token[:3])
```
jgehrcke: note: use _rotate_auth_token() here instead

conbench/entities/commit.py
```diff
@@ -820,6 +877,10 @@ def _get_response_retrycore(self, url) -> Optional[dict]:
             )
 
             if reqquota == 0:
+                if self._rotate_auth_token():
+                    log.info("rotated auth token, retry request")
+                    return None
```
jgehrcke: Note to self: need to catch case where a full rotation was performed

conbench/entities/commit.py
```diff
@@ -532,6 +534,62 @@ def __init__(self):
...
+        self._current_auth_token = None
+        if data is None:
+            log.info("GITHUB_API_TOKEN env not set")
+            return
+
+        log.info("GITHUB_API_TOKEN env was set, length of data: %s", len(data))
+
+        token_candidates = [data]
+        if "," in data:
+            token_candidates = data.split(",")
```
austin3dickey: nit: `split()` will always return a list even if there's no comma!
```suggestion
        token_candidates = data.split(",")
```

conbench/entities/commit.py
```diff
@@ -532,6 +534,62 @@ def __init__(self):
...
+        """
+        Return True if token was rotated, False otherwise.
+        """
+        if self._token_pool is None:
+            # convention: pool size greater than 1, rotation takes effect.
+            log.debug("tried rotating auth token, but no pool is configured")
+            return False
+
+        self._current_auth_token = next(self._token_pool)
+        log.info("current auth token starts with: %s", self._current_auth_token[:3])
```
austin3dickey: nit: I think personal access tokens always start with `ghp_`. Maybe log the last three chars instead?

conbench/entities/commit.py
jgehrcke: done

conbench/entities/commit.py
jgehrcke: done

conbench/entities/commit.py
jgehrcke: Yes! and there also is the prefix `github_pat_` for the new fine-grained tokens. I opted for still showing prefix, just a tiny bit longer (with length depending on token type).

jgehrcke:
Log lines from CI to show parts of this in action:
```
[conbench.entities.commit] INFO: configured GitHub HTTP API authentication token pool: ghp_BV..., ghp_e1..., ghp_i7...
[conbench.entities.commit] INFO: current auth token has length 40 and starts with: ghp_BV
```

jgehrcke:
Token rotation wasn't hit yet in CI (according to logs), but it will probably be hit soon when I land this now.

(more review feedback welcome, and I want to really provoke the rotation code path now to see if this still contains a bug -- I even didn't hit this yet locally, misusing CI for that!).

jgehrcke:
In first build on `main` after merging this, still no rotation performed. Will report back when I see this.

migrations/versions/d3515ecea53d_backfill_all_commits.py
```diff
@@ -43,8 +42,11 @@ def upgrade():
             continue
 
         name = repository_to_name(repo)
-        default_branch = github.get_default_branch(name)
-        all_commits = github.get_commits_to_branch(
+
+        # This deliberately (mis)uses the per-process singleton GitHub HTTP API
+        # client object from commit.py
```
austin3dickey, line 47: Yeah, I shouldn't have done this, but it seemed much easier than rolling my own code to do these things at the time

jgehrcke: That's completely fine! :)

PR #916, opened 2023-03-20
compare API: make thresholds float args

### LINKED ISSUE #915: compare API: threshold arguments should be float, not int
Currently, in the `/api/compare` endpoint, we try to coerce the threshold arguments to `int`:
https://github.com/conbench/conbench/blob/f193895123de5e9a5266a1ad0ae3f4b1b941303a/conbench/api/compare.py#L84-L90
If a float argument is passed, since these arguments are query parameters, they come in as strings. For example, `"5.7"`. This results in a 500 INTERNAL_SERVER_ERROR due to this exception:
```
ValueError: invalid literal for int() with base 10: '5.7'
```
I don't think there's any reason to limit the threshold arguments to integers only. We should expect float arguments to work.

austin3dickey:
Fixes #915. See that issue for more.

### DIFF
--- a/conbench/tests/api/test_compare.py
+++ b/conbench/tests/api/test_compare.py
@@ -1,3 +1,5 @@
+import pytest
+
 from ...api._examples import (
     _api_compare_benchmark_result,
     _api_compare_entity,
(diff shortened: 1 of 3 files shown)


austin3dickey:
I'm going to merge this to unblock future work but feedback is always appreciated!

jonkeane (commented):
Oh weird, these definitely should have been floats all along! Thanks for catching this.

jgehrcke (commented):
Thanks!

conbench/tests/api/test_compare.py
```diff
@@ -336,7 +345,10 @@ def test_compare(self, client):
             batch_id=batch_id,
         )
         new_ids = [e.id for e in new_entities]
-        response = client.get(f"/api/compare/batches/{compare.id}/")
+        query_string = {"threshold_z": threshold_z} if threshold_z else None
+        response = client.get(
+            f"/api/compare/batches/{compare.id}/", query_string=query_string
+        )
```
jgehrcke, line 351: Just a quick comment on HTTP testing style: When I read test code that involves making HTTP requests then right after getting the response I like having something in the lines of:
`assert response.status_code == expected_code, response.text`

This would dump the response contents for the unexpected status code.

The logic that's commented on with ` # cheating by comparing batch to same batch...` only matters when the status code is as expected.

The late ` self.assert_200_ok(response, None, contains=expected[0])` is too indirect/complex for my taste :) I have been trying to do baby steps into 'my preferred style' when I was touching test code here and there. This is no call to action now, just sharing what I do.

austin3dickey: That makes sense to me. I see the draw of the `self.assert_` methods to reduce boilerplate, but also understand it hides layers of complexity.

Issue #801, opened 2023-03-06
client HTTP tooling: set deliberate timeout constants (connect()/recv()/total, ...)


jgehrcke:
Goes hand-in-hand with https://github.com/conbench/conbench/issues/800.

Towards good UX and overall system responsiveness it's critical to handle errors as early as possible. Otherwise some entity waits unnecessarily long (might be a human), or a window of opportunity closes; resulting in an outcome that could have been prevented.

Default timeout constants are often getting in the way (by definition, there are no silver bullet timeout constants). Infinite wait is often the worst decision (but many client libs implement that by default).

There is a lot of value in setting a
i) deliberate, short TCP connect() timeout (same is true for DNS resolution)
ii) deliberate, long-ish overall HTTP response timeout (measured from the point in time the request was written to the TCP connection)

(i) often allows for quick auto-healing (e.g., by triggering a different networking path through some round-robin load balancing infra).
ii) should use knowledge about the expected response generation time on the receiving end, and maybe go by a factor of ~5 over that. If an HTTP response is expected to be generated within 0.5 s then a timeout of 5-10 seconds makes much more sense than waiting for one minute. A timeout constant should be paired with a reasonable back-off time (which should use application knowledge, too).

A properly configured HTTP reverse proxy chain can allow the API server to _stop_ an execution unit upon a "client went away" error (TCP connection shutdown). That can also be further propagated through the system layers and after all make the database client emit an explicit cancel-query-instruction (libpq provides the functions `PQgetCancel()` and `PQcancel()`) or otherwise result in early termination of a database query if the DB client's connection goes away (quote from [here](https://dba.stackexchange.com/a/81424)):

> If the client vanishes due to network connection loss the query will generally run until it's retrieved enough rows to fill its network send buffer, then stop and get stuck until the TCP connection drops, at which point it'll abort.

When thinking about choosing timeout constants, several aspects matter. This is never about perfection, but about nudging the probabilities to our favor. One aspect is that it's not a good idea to stop waiting _shortly_ before a sane/good answer would have arrived. That implies throwing away valuable work (most of the work having been done in the database, for example). One should choose timeout constants so that it is unlikely to initiate a loop that creates work, ignores the result, re-creates the same work, ignores the result, ... . One goal of a timeout constraint is to detect and back out of an _erroneous_ state with rather high probability.

In addition to (i) and (ii) above, two more types of timeout constraints in the context of HTTP that are absolutely worth putting into place (with application-specific values):

[...]

PR #920, opened 2023-03-21
db: define connect timeout

jgehrcke:
Related to #801.

For example, allows for quicker retrying upon networking hiccup in the context of docker-compose.

### DIFF
--- a/conbench/db.py
+++ b/conbench/db.py
@@ -38,7 +38,11 @@ def configure_engine(url):
         # tickets and discussions:
         # https://github.com/conbench/conbench/issues/599
         # https://github.com/conbench/conbench/pull/690
-        connect_args={"options": "-c timezone=utc -c statement_timeout=60s"},
+        # https://docs.sqlalchemy.org/en/20/core/engines.html#use-the-connect-args-dictionary-parameter
+        connect_args={
+            "options": "-c timezone=utc -c statement_timeout=60s",
+            "connect_timeout": 3,
+        },
     )
     log.info("bind engine to session")
     session_maker.configure(bind=engine)
@@ -132,6 +136,7 @@ def create_all():
         else:
             raise
 
+    log.info("create_all(engine) returned. dispose()")
     engine.dispose()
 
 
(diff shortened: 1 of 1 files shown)


conbench/db.py
```diff
@@ -132,6 +136,7 @@ def create_all():
         else:
             raise
 
+    log.info("create_all(engine) returned. dispose()")
```
jgehrcke, line 141: Of the many many `make run-app-dev` I execute locally, sometimes the stack startup hangs in `create_all()`. This log statement helps in the future to better understand where the hang is; but for now I think I have an educated guess that it was a long-hanging connect() -- the timeout plus retrying will help resolve this faster.

conbench/db.py
```diff
@@ -38,7 +38,11 @@ def configure_engine(url):
         # tickets and discussions:
         # https://github.com/conbench/conbench/issues/599
         # https://github.com/conbench/conbench/pull/690
-        connect_args={"options": "-c timezone=utc -c statement_timeout=60s"},
+        # https://docs.sqlalchemy.org/en/20/core/engines.html#use-the-connect-args-dictionary-parameter
+        connect_args={
+            "options": "-c timezone=utc -c statement_timeout=60s",
+            "connect_timeout": 3,
```
jonkeane: Is this 3 seconds? Or some other unit? Not a big deal, just curious (I did look in the URL up above, but couldn't find what I was looking for either with a quick scan)

jgehrcke: Yes, unit is seconds!

For the record, the `connect_timeout` arg is documented here: https://www.postgresql.org/docs/12/libpq-connect.html

> Maximum wait for connection, in seconds (write as a decimal integer, e.g., 10). Zero, negative, or not specified means wait indefinitely. The minimum allowed timeout is 2 seconds, therefore a value of 1 is interpreted as 2. This timeout applies separately to each host name or IP address. For example, if you specify two hosts and connect_timeout is 5, each host will time out if no connection is made within 5 seconds, so the total time spent waiting for a connection might be up to 10 seconds.


What is maybe not obvious that the Python package psycopg2 is a wrapper around libpq, and libpq can be in client or server mode.

austin3dickey: Would you mind adding a small comment? Agreed with Jon that it's hard to find.

```suggestion
            "connect_timeout": 3,  # seconds
```

austin3dickey (approved):
Looks great!

conbench/db.py
jgehrcke: Yes! Adding a commit.

Issue #922, opened 2023-03-21
kube-prometheus stack: inject custom grafana admin credentials


jgehrcke:
For local dev, anonymous access is nice:
https://github.com/conbench/conbench/blob/016fc318b50ae90549c231658ef443836fc9f932/k8s/kube-prometheus/conbench-flavor.jsonnet#L38

For more serious usage, build a pragmatic way to set credentials via k8s secrets/config map.

jgehrcke:
Potentially useful command from https://github.com/prometheus-operator/kube-prometheus/discussions/1578
```
GRAFANA_POD=$(kubectl get pods --selector=app=grafana -n monitoring | tail -n 1 | awk '{print $1}')
kubectl exec --stdin --namespace="monitoring" --tty $GRAFANA_POD -- /bin/bash -c "grafana-cli admin reset-admin-password $GRAFANA_ADMIN_PASSWORD"
```

jgehrcke:
Changing selector, that works:
```
GRAFANA_POD=$(kubectl get pods --selector=app.kubernetes.io/name=grafana -n monitoring | tail -n 1 | awk '{print $1}')
```
and then this worked:
```
$ kubectl exec --stdin --namespace="monitoring" --tty $GRAFANA_POD -- /bin/bash -c "grafana-cli admin reset-admin-password $GRAFANA_ADMIN_PASSWORD"
INFO [03-21|15:58:31] Starting Grafana                         logger=settings version= commit= branch= compiled=1970-01-01T00:00:00Z
INFO [03-21|15:58:31] Config loaded from                       logger=settings file=/usr/share/grafana/conf/defaults.ini
INFO [03-21|15:58:31] Config overridden from Environment variable logger=settings var="GF_PATHS_DATA=/var/lib/grafana"
INFO [03-21|15:58:31] Config overridden from Environment variable logger=settings var="GF_PATHS_LOGS=/var/log/grafana"
INFO [03-21|15:58:31] Config overridden from Environment variable logger=settings var="GF_PATHS_PLUGINS=/var/lib/grafana/plugins"
INFO [03-21|15:58:31] Config overridden from Environment variable logger=settings var="GF_PATHS_PROVISIONING=/etc/grafana/provisioning"
INFO [03-21|15:58:31] Path Home                                logger=settings path=/usr/share/grafana
INFO [03-21|15:58:31] Path Data                                logger=settings path=/var/lib/grafana
[...]

PR #933, opened 2023-03-22
ci: modify JSONNET before deploy to minikube

jgehrcke:
This makes it so that `make jsonnet-kube-prom-manifests` does _not_ by default entail stack modifications that are in fact only of value / use when deploying on minikube. For example, do not by default use `"auth.anonymous": { enabled: true }`. This is related to #922.


jgehrcke:
Going ahead with this one.

Issue #935, opened 2023-03-22
POST BenchmarkResult: missing `name` key results in HTTP response with code 500


jgehrcke:
I saw https://github.com/conbench/conbench/blob/5154d8bdf540dc4d6600df4712e94bd27b9d6f6c/conbench/entities/benchmark_result.py#L194

and tested
```
>>> {1:2}.pop("foo")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'foo'
```

and then confirmed that 

```
conbench-app-1  | 172.22.0.1 - - [22/Mar/2023:13:54:59 +0000] "POST /api/benchmarks/ HTTP/1.1" 500 0 "-" "-" (0.002664 s)
conbench-app-1  | [2023-03-22 13:54:59 +0000] [8] [ERROR] Error handling request /api/benchmarks/
...
conbench-app-1  |   File "/app/conbench/entities/benchmark_result.py", line 194, in create
conbench-app-1  |     name = tags.pop("name")
conbench-app-1  |            ^^^^^^^^^^^^^^^^
conbench-app-1  | KeyError: 'name'
```

action items:
- document this in api spec
- emit 400 bad request resp with proper err msg
- cover with tests

jgehrcke:
> document this in api spec

Did that here: https://github.com/conbench/conbench/commit/9237373f99b2aa04906496e0393eabc0ec1ecd45

"The special key "name" must be provided with a string value: it indicates the name of the conceptual benchmark that was performed for obtaining the result at hand"

> emit 400 bad request resp with proper err msg

https://github.com/conbench/conbench/pull/1126

> cover with tests

still a TODO

jgehrcke:
> cover with tests

Oh, I did that here:

https://github.com/conbench/conbench/commit/fdaa36d263cb28ccf5697c286a5b11aa1b633512

So, we can close this issue as 'done'.

PR #936, opened 2023-03-22
api: bmresult POST: document and test that name is required

jgehrcke:
This is a quick follow-up for https://github.com/conbench/conbench/issues/935.

### DIFF
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -191,12 +191,16 @@ class BenchmarkResult(Base, EntityMixin):
                 "status": "Partial result: not all iterations completed"
             }
 
+        # Note(JP): this implicitly introduces a requirement for `name` to be
+        # set: https://github.com/conbench/conbench/issues/935
+        # Also note: this pulls the `name=xx` key/value pair out of tags.
         name = tags.pop("name")
 
         # create if not exists
         c = {"name": name, "tags": tags}
         case = Case.first(**c)
         if not case:
+            # Note(JP): there is a race condition here, expect this to collide
             case = Case.create(c)
 
         # create if not exists
@@ -576,7 +580,31 @@ class _BenchmarkFacadeSchemaCreate(marshmallow.Schema):
     tags = marshmallow.fields.Dict(
         required=True,
         metadata={
-            "description": "Details that define the individual benchmark case that is being run (e.g. name, query type, data source, parameters). These are details about a benchmark that define different cases, for example: for a file reading benchmark, some tags might be: the data source being read, the compression that file is written in, the output format, etc."
+            "description": conbench.util.dedent_rejoin(
+                """
+                The set of key/value pairs that defines a benchmark case
+                permutation. Top-level keys must be strings. Allowed value
+                types are ... (to be specified).
+
+                The special key "name" must be provided with a string value: it
+                indicates the name of the conceptual benchmark that was
+                performed for obtaining the result at hand. All case
+                permutations of a conceptual benchmark by definition have this
+                name in common.
+
+                We advise that each unique case (as defined by the complete set
+                of key/value pairs) indeed corresponds to unique benchmarking
+                behavior. That is, typically, these key/value pairs directly
+                correspond to input parameters to your benchmarking method.
+
+                Example: a conceptual benchmark with name "foo-write-file"
+                might have meaningful case permutations involving tag names
+                such as `compression_method` (values: `gzip`, `lzma`, ...),
+                `file_format` (values: `csv`, `hdf5`, ...), `dataset_name`
+                (values: `foo`, `bar`, ...). For each conceptual benchmark, it
+                is OK to have no or many case permutations.
+                """
+            )
         },
     )
     optional_benchmark_info = marshmallow.fields.Dict(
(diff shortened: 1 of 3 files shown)


conbench/entities/benchmark_result.py
```diff
@@ -576,7 +580,31 @@ class _BenchmarkFacadeSchemaCreate(marshmallow.Schema):
...
+                of key/value pairs) indeed corresponds to unique benchmarking
+                behavior. That is, typically, these key/value pairs directly
+                correspond to input parameters to your benchmarking method.
+
+                Example: a conceptual benchmark with name "foo-write-file"
+                might have meaningful case permutations involving tag names
+                such as `compression_method` (values: `gzip`, `lzma`, ...),
+                `file_format` (values: `csv`, `hdf5`, ...), `dataset_name`
+                (values: `foo`, `bar`, ...). For each conceptual benchmark, it
+                is OK to have no or many case permutations.
```
jgehrcke, line 605: These paragraphs are meant to be provocative! Would love to get your feedback.

Based on https://github.com/conbench/conbench/issues/787 and https://github.com/conbench/conbench/pull/875 (and other discussion threads) I think we're already converging to something.

conbench/tests/api/test_benchmarks.py
```diff
@@ -936,6 +936,16 @@ def test_create_benchmark_context_missing(self, client):
...
 
+    def test_create_benchmark_name_missing(self, client):
+        self.authenticate(client)
+        payload = _fixtures.VALID_PAYLOAD.copy()
+        del payload["tags"]["name"]
+        resp = client.post(self.url, json=payload)
+        # TODO: https://github.com/conbench/conbench/issues/935
+        # This here just quickly checks that there is a failure at all,
+        # that 'name' is required.
+        assert resp.status_code == 500
```
jgehrcke: This quickly confirms existence of https://github.com/conbench/conbench/issues/935, and at least locks in the 'name is required!' condition for now. 

I don't expect this test code to live for long.

conbench/entities/benchmark_result.py
austin3dickey: This makes sense to me based on what we've talked about!

conbench/tests/api/_expected_docs.py
```diff
@@ -1027,7 +1027,7 @@
...
+                        "description": 'The set of key/value pairs that defines a benchmark case permutation. Top-level keys must be strings. Allowed value types are ... (to be specified).  The special key "name" must be provided with a string value: it indicates the name of the conceptual benchmark that was performed for obtaining the result at hand. All case permutations of a conceptual benchmark by definition have this name in common.  We advise that each unique case (as defined by the complete set of key/value pairs) indeed corresponds to unique benchmarking behavior. That is, typically, these key/value pairs directly correspond to input parameters to your benchmarking method.  Example: a conceptual benchmark with name "foo-write-file" might have meaningful case permutations involving tag names such as `compression_method` (values: `gzip`, `lzma`, ...), `file_format` (values: `csv`, `hdf5`, ...), `dataset_name` (values: `foo`, `bar`, ...). For each conceptual benchmark, it is OK to have no or many case permutations.',
```
austin3dickey, line 1030: Does `dedent_rejoin` not support multiple paragraphs, or is that an OpenAPI spec thing? Would this look cleaner with your paragraphs preserved?

austin3dickey (approved):
nice!

conbench/tests/api/_expected_docs.py
jgehrcke: > Would this look cleaner with your paragraphs preserved?

Yes, it would! Will look into that separately. Thanks for sharing the opinion, I thought the same.

conbench/entities/benchmark_result.py
jonkeane: Thanks for this provocative language + pushing us to be precised.

> it is OK to have no or many case permutations.

Functionally speaking this is true (so long as you have `{name: "foo"}` you're fine[^1]. But practically speaking, there always ends up being at least one parameter and thus at least one case (if you have a parameter that has one value you have a single case, yeah?). So I wonder if we might rephrase this to be "It is ok to have few or many case permutations" and possibly also discuss that there's always at least one case (or else you have an conceptual benchmark with no actual instantiations)?

This hopefully will also nudge folks into thinking a little bit about their case permutations up front. One thing that we've struggled a bit with is as a benchmark evolves over time it usually grows more parameters. That's ok and natural, but currently conbench treats each one as if it were totally separate for history and comparison purposes, so there's a bit of friction on that front. We can't (and shouldn't) solve that here, but something to note that if we can get folks to thing "what are the parameters I might use" that's not a bad thing.


[^1]: Though I suspect that there totally might be code that doesn't handle there being no other tags in that dict and will do funny things if that's all you have.

jgehrcke: > there's always at least one case (or else you have an conceptual benchmark with no actual instantiations)?

Yes! This must be true (as in: other thinking needs to change so that this _is_ true): one should be able to submit a benchmark result that has only the `name` set -- it's a special case, but it is a case.

I didn't see this feedback before merging (hmm.. sorry!). Will adjust the text. I think this is what I meant with "no case permutations", but in fact this should be called _one_ case, a special, valid case.

For implementation, we should totally cover that with tests, because it might be an interesting edge case: `name` is set, but no other keys set within `tags`.

PR #937, opened 2023-03-22
UI, internals: more of "benchmark name" and "benchmark case permutation", start cleaning up hacks.py

jgehrcke:
This addresses:
- #943
- #944 
- #400
- #942 
- #940, partially at least

I think this is getting interesting. Simplification, more or less with confidence. First, along terminology within code but especially in the UI (user-facing terminology).

This should make the 'case permutation' idea more explicit, and easier to understand for users. This is certainly a greatly helpful direction for me, wrapping my head around concepts.

### DIFF
--- a/conbench/api/compare.py
+++ b/conbench/api/compare.py
@@ -10,7 +10,7 @@ from ..entities.benchmark_result import BenchmarkResult
 from ..entities.commit import Commit
 from ..entities.compare import CompareBenchmarkResultSerializer
 from ..entities.history import set_z_scores
-from ..hacks import set_display_batch, set_display_name
+from ..hacks import set_display_benchmark_name, set_display_case_permutation
 
 
 def _compare_entity(benchmark_result):
@@ -134,10 +134,10 @@ class CompareEntityEndpoint(ApiEndpoint, CompareMixin):
         baseline_benchmark_result = self._get(baseline_id)
         contender_benchmark_result = self._get(contender_id)
         set_z_scores([baseline_benchmark_result, contender_benchmark_result])
-        set_display_name(baseline_benchmark_result)
-        set_display_name(contender_benchmark_result)
-        set_display_batch(baseline_benchmark_result)
-        set_display_batch(contender_benchmark_result)
+        set_display_case_permutation(baseline_benchmark_result)
+        set_display_case_permutation(contender_benchmark_result)
+        set_display_benchmark_name(baseline_benchmark_result)
+        set_display_benchmark_name(contender_benchmark_result)
 
         baseline = _compare_entity(baseline_benchmark_result)
         contender = _compare_entity(contender_benchmark_result)
@@ -191,12 +191,12 @@ class CompareListEndpoint(ApiEndpoint, CompareMixin):
 
         baseline_items, contender_items = [], []
         for benchmark_result in baselines:
-            set_display_name(benchmark_result)
-            set_display_batch(benchmark_result)
+            set_display_benchmark_name(benchmark_result)
+            set_display_benchmark_name(benchmark_result)
             baseline_items.append(_compare_entity(benchmark_result))
         for benchmark_result in contenders:
-            set_display_name(benchmark_result)
-            set_display_batch(benchmark_result)
+            set_display_benchmark_name(benchmark_result)
+            set_display_benchmark_name(benchmark_result)
             contender_items.append(_compare_entity(benchmark_result))
 
         pairs = _get_pairs(baseline_items, contender_items)
--- a/conbench/api/compare.py
+++ b/conbench/api/compare.py
@@ -13,7 +13,7 @@ from ..entities.history import set_z_scores
 from ..hacks import set_display_benchmark_name, set_display_case_permutation
 
 
-def _compare_entity(benchmark_result):
+def _compare_entity(benchmark_result: BenchmarkResult):
     return {
         "id": benchmark_result.id,
         "batch_id": benchmark_result.batch_id,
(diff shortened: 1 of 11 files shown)


conbench/hacks.py
```diff
@@ -1,49 +1,97 @@
...
-        tags["name"] = name
-    case = get_case(tags, include_dataset=True)
-    if "suite" in tags:
-        case = [name] + case
-    name = ", ".join(case) if case else name
+        kvpairs.append(("dataset", tags["dataset"]))
+
+    # # Rationale?
+    # if "language" in tags:
+    #     kvpairs.append(("language", tags["language"]))
```
jgehrcke: I have removed `language` from `CASE_KEYS_DO_NOT_USE_FOR_NAME` -- should result in the same outcome I think. (first it was removed, then it was added again)

conbench/hacks.py
```diff
@@ -1,49 +1,97 @@
...
+
+    # booleans = [True, False, "true", "false"]
+
+    # case = [(k, f"{k}={v}") if v in booleans else (k, v) for k, v in case]
+
+    # return [f"{k}={v}" if isinstance(v, (int, float)) else str(v) for k, v in case]
+
+    return [f"{k}={v}" for k, v in kvpairs]
```
jgehrcke: This always adds the key name to the string. That should address #400.

conbench/hacks.py
```diff
@@ -1,49 +1,97 @@
...
+    tags = bmresult["tags"] if is_api else bmresult.case.tags
+
+    # Note(JP): this tries to pull in the benchmark name into the case
+    # permutation k/v pairs. Skip this for now. Unique combination must be
+    # name+case_perm, so adding the name into the case_perm again should not be
+    # conceptually needed.
+
+    # name = tags.get("name") if is_api else bmresult.case.name
+    # if "name" not in tags:
+    #     tags["name"] = name
```
jgehrcke: This added `name` back into the output. But I think it might be cleaner to not have it here on the output side of things. It always _is_ contained in the output, as higher-level property (the benchmark name, previously "batch name").

conbench/hacks.py
```diff
@@ -1,49 +1,97 @@
-def get_case(tags, include_dataset=False):
-    case = [
+from typing import Dict, List
+
+CASE_KEYS_DO_NOT_USE_FOR_NAME = ("id", "name", "suite", "source", "dataset")
+
+
+def get_case_kvpair_strings(tags: Dict[str, str], include_dataset=False) -> List[str]:
```
jgehrcke: I realize that `Dict[str, str]` might not be true, I think the values can be arbitrary JSON types, i.e. nesting is allowed?

jonkeane: I think that is technically allowed, but I can't think of a place where we do that or use that. So maybe we can tighten that restriction?

jgehrcke: > So maybe we can tighten that restriction?

I do hope so! Created https://github.com/conbench/conbench/issues/940

conbench/api/compare.py
```diff
@@ -10,10 +10,10 @@
 from ..entities.commit import Commit
 from ..entities.compare import CompareBenchmarkResultSerializer
 from ..entities.history import set_z_scores
-from ..hacks import set_display_batch, set_display_name
+from ..hacks import set_display_benchmark_name, set_display_case_permutation
```
jgehrcke, line 13: Even this is a bit of a relief, to me at least.

conbench/templates/benchmark-result.html
```diff
@@ -126,9 +127,7 @@
       {% for k,v in benchmark.tags.items() %}
       <li class="list-group-item" style="overflow-y: auto;">
         <b>{{ k }}</b>
-        {% if v is not none %}
         <div align="right" style="display:inline-block; float: right;">{{ v }}</div>
-        {% endif %}
```
jgehrcke, line 131: This is just to make sure we show what's in the database. a `param=None` as of today contributes defining a specific case permutation.

conbench/templates/run.html
```diff
@@ -67,8 +67,8 @@ <h4>Benchmarks in run</h4>
                 <tr>
                     <th scope="col">Date</th>
                     <th scope="col" style="width: 80px">Lang</th>
-                    <th scope="col" style="width: 80px">Batch</th>
-                    <th scope="col">Benchmark</th>
+                    <th scope="col" style="width: 80px">Benchmark name</th>
+                    <th scope="col">Case permutation</th>
```
jgehrcke, line 71: This is also an important change.

conbench/tests/api/test_compare.py
```diff
@@ -9,7 +9,7 @@
 from ...tests.api import _asserts, _fixtures
 from ...tests.helpers import _uuid
 
-CASE = "snappy, cpu_count=2, parquet, arrow, nyctaxi_sample"
+CASE = "compression=snappy, cpu_count=2, dataset=nyctaxi_sample, file_type=parquet, input_type=arrow"
```
jgehrcke, line 12: Here we see how a case permutation is stringified now. See #400.

jgehrcke:
Landing this. More feedback is very welcome (thanks for the previous quick feedback @jonkeane).

conbench/hacks.py
alistaire47: The only nested one I know of offhand is [array altrep materialization](https://github.com/voltrondata-labs/arrowbench/blob/ebfab3d5535a1a7ba8680be4d0394296b9b1a5ee/R/bm-array-altrep-materialization.R#L19), which is not currently run in CI. And it may have been a bad idea there.

But we have lots of cases where values are not `str`s.

conbench/tests/api/test_compare.py
austin3dickey: Note to self: fix `benchalerts` tests to deal with this change to the `/api/compare` endpoint (the "benchmark" key contains this stringification), and of course longer term #720 is very important :)

Issue #940, opened 2023-03-23
case tags: re-specify structure, simplify (str/str mapping?)


jgehrcke:
Today: database allows any JSON document as value: https://github.com/conbench/conbench/blob/9237373f99b2aa04906496e0393eabc0ec1ecd45/conbench/entities/case.py#L12

JSON schema validation also kind of allows everything (not sure what marshmallow.fields.Dict() really allows, but probably nesting): https://github.com/conbench/conbench/blob/9237373f99b2aa04906496e0393eabc0ec1ecd45/conbench/entities/benchmark_result.py#L580

Nesting might hurt us. None-values might raise complications. Mixed value types increase complexity for which value?

I think we should make this a str/str mapping. Empty string values? Not allowed. Then we know exactly what a key/value pair looks like.

Looks like @jonkeane agrees: https://github.com/conbench/conbench/pull/937#discussion_r1145196482

jonkeane:
I'm fine with this — I don't know of any examples of someone using something nested right now and better to cut off that possibility (+ needing to deal with it) sooner than later.

> Empty string values? Not allowed. Then we know exactly what a key/value pair looks like.

When you say empty string here are you thinking `""` or `null`s?

jgehrcke:
> When you say empty string here are you thinking "" or nulls?

When I wrote the above I meant to not allow JSON null as well as to not allow empty string "".

I thought _only_ not including a key should semantically mean "not set".

The special value `""` (empty string) often also means "not set", and if we allow that then we have more than one way to represent "not set".

In our discussion I think I just convinced myself that it's ok/good to allow for empty string, but now I am not so convinced anymore. I will give this more thought during texting/implementing.

The one big/important design choice is: that absence of a key name still allows for that labelset to be part of the set of all case permutations, as long as the benchmark _name_ is in common.

jonkeane:
> The one big/important design choice is: that absence of a key name still allows for that labelset to be part of the set of all case permutations, as long as the benchmark name is in common.

💯 agreed on this

jgehrcke:
I think we can close this one here as 'done' for now after having landed #948. Let's go back to this discussion here / open more specific tickets when the need arises.

Issue #862, opened 2023-03-14
timeseries fingerprint: use this to define comparability (re-think term "tags" and "case" -- generic label set, with fingerprint being hash thereof)


jgehrcke:
Borrowing some terms and concepts from the Observability space, enriched with own ideas.

A benchmark might be uniquely identified by a labelset (collection of key-value pairs). That is what we call case/context, I think.

In code, the hash of that labelset can be called "fingerprint" (as Prometheus does [here](https://github.com/prometheus/common/blob/66b493f42b757378fbad83e42b0ed441d425a17b/model/labelset.go#L142)). We also perform hashing on the context k/v pair set. Strong conceptual analogy.

In text, "fingerprint" can be used as a nice, recognizable term for "a unique labelset". For Conbench documentation and UI/UX it might make sense to use the term "fingerprint" as much as possible to explain what e.g. data points shown in a history plot have in common.

Architecturally, this perspective might give us important clues!  In the Prometheus ecosystem, kind of everything related to data storage and data querying is centered around those label sets. We may want to take inspiration from there. Everything is optimized for dynamic time series querying (an important part of the query is the user-given labelset filter).

(this is just a collection of ideas I wanted to keep note of, not yet convinced about everything in here, let's see where this leads)

jonkeane:
I do very much like the idea | use of the term fingerprint here. And it could help as a nice bundler for all of the things that come together (tags + context and since context + hardware). And then we can document that in detail and use the term fingerprint to go from there.

jgehrcke:
I like the term "timeseries fingerprint". However, regardless of the name, the underlying concept is one of the most foundational ones in what we do. Questions came up. So, let's dive into the concept.

We have implicitly used the concept all the time.

We started using the term "timeseries fingerprint" here: https://github.com/conbench/conbench/blob/b03b1ffa0716098143ac0cbe354aa9169205a275/conbench/entities/history.py#L191

We will make more use of it in a DB schema iteration in the context of https://github.com/conbench/conbench/issues/1227.

What is it? "timeseries fingerprint"  is
- a grouping construct.
- a formalization of the 'comparability' question.
- a precise string identifier, mainly meant consumption by machines, but also for humans.

We (try hard to) group benchmark results by **comparability**. The timeseries fingerprint is an attempt to give each group (timeseries) a precise (unique!) and short-ish name (string identifier).

Examples for usage: that identifier can be used for
- efficiently querying the database (implementation detail)
[...]

PR #948, opened 2023-03-24
POST BenchmarkResult handler: validate&normalize tags (key / value pairs for case parameters)

jgehrcke:
EDIT: super interesting discussion ongoing, I have updated the patch. See comments.

This is largely for https://github.com/conbench/conbench/issues/940.


conbench/api/benchmarks.py
```diff
@@ -182,11 +186,67 @@ def post(self) -> None:
...
+                        description=f"tags: bad value type for key `{key}`, only string values are allowed, got `{type(value)}` "
+                    ),
+                    400,
+                )
+
         benchmark_result = BenchmarkResult.create(data)
+
         set_z_scores([benchmark_result])
+
+        log.info("benchmark_result: %s", benchmark_result)
```
jgehrcke: will remove that

conbench/api/benchmarks.py
```diff
@@ -182,14 +186,75 @@ def post(self) -> None:
...
+            if value == "" or value is None:
+                log.warning("silently drop tag key/value pair: `%s`, `%s`", key, value)
+                # Remove key/value pair, check next one. Mutating the
+                # dictionary while we iterate over it? That is why the
+                # iteration happens over a copy of kv pairs.
+                del tags[key]
+                continue
+
+            # Note(JP): this code path should go away after we adjust our
+            # client tooling to not send numeric values anymore.
```
jgehrcke: That is, I'd love to emit a bad request response for those value types, but we don't need to that in _this_ PR, but can delay that further.

conbench/api/benchmarks.py
```diff
@@ -182,14 +186,75 @@ def post(self) -> None:
...
+            if isinstance(value, int) or isinstance(value, float):
+                log.warning(
+                    "stringify numeric case parameter value: `%s`, `%s`", key, value
+                )
```
jgehrcke: these warning-level log messages will allow us to adjust our client tooling over time

conbench/api/benchmarks.py
```diff
@@ -182,14 +186,75 @@ def post(self) -> None:
...
+
         return self.response_201_created(self.serializer.one.dump(benchmark_result))
 
 
+def resp400(description: str) -> f.Response:
+    """
+    Utility for canonical generation of 400 Bad Request response with an
+    error description. Define elsewhere once used elsewhere.
+    """
+    return f.make_response(
```
jgehrcke: I used `flask.make_response()` because it has a well-defined return type. Important for type checking. I know this might overlap with other util functions elsewhere, we will reconcile over time.

conbench/tests/api/test_benchmarks.py
```diff
@@ -946,6 +946,28 @@ def test_create_benchmark_name_missing(self, client):
...
 
+    @pytest.mark.parametrize(
+        "tagdict",
+        [
+            # empty string key
+            {"name": "foo", "": "foo"},
+            # dict type value
+            {"name": "foo", "foo": {"foo": "bar"}},
+            # array type value
+            {"name": "foo", "foo": ["foo"]},
```
jgehrcke: in the future also int and float

conbench/tests/api/test_benchmarks.py
```diff
@@ -946,6 +946,28 @@ def test_create_benchmark_name_missing(self, client):
...
+    )
+    def test_create_benchmark_bad_tags(self, client, tagdict):
+        self.authenticate(client)
+        payload = copy.deepcopy(_fixtures.VALID_PAYLOAD)
+        payload["tags"] = tagdict.copy()
+        resp = client.post(self.url, json=payload)
+        assert resp.status_code == 400
+        assert (
+            "tags: bad value type for key" in resp.text
+            or "tags: zero-length string as key is not allowed" in resp.text
```
jgehrcke: quick, not the most precise, but already valuable

conbench/tests/app/_asserts.py
```diff
@@ -80,6 +80,7 @@ def assert_page(self, r, title):
     def create_benchmark(self, client):
         self.authenticate(client)
         response = client.post("/api/benchmarks/", json=_fixtures.VALID_PAYLOAD)
+        assert response.status_code == 201, response.text
         benchmark_id = response.json["id"]
```
jgehrcke, line 84: Got `KeyError` in `response.json["id"]` -- because of response status code being non-201. Misleading error. Therefore the assertion in status code above now.

conbench/api/benchmarks.py
```diff
@@ -182,14 +186,75 @@ def post(self) -> None:
...
         data = self.validate_benchmark(self.schema.create)
+
+        # This next block inspects and mutates `tags`, with the goal that all
+        # keys are non-empty strings, and all values are non-empty strings.
+        # ideally this should be done as part of schema validation above.
+        tags = data["tags"]
+        for key, value in list(tags.items()):
+            # In JSON, a key is always of type string. We can rely on this
+            # here, but codify this invariant.
+            assert isinstance(key, str)
```
alistaire47: We have tags which are not strings, e.g. [query ID and scale factor for TPC-H](https://github.com/voltrondata-labs/benchmarks/blob/75bc80e1bc75b428734f7a018a7f5460d6094274/benchmarks/tpch_benchmark.py#L8-L9). They can be coerced to strings (if that's desirable?), but if we merge this before doing so we're going to get a lot of rejected results.

alistaire47: Oops, query ID does get formatted as a string. But `memory_map` is a bool: https://github.com/voltrondata-labs/benchmarks/blob/75bc80e1bc75b428734f7a018a7f5460d6094274/benchmarks/tpch_benchmark.py#L26-L27

conbench/tests/api/_expected_docs.py
```diff
@@ -1027,7 +1027,7 @@
...
+                        "description": 'The set of key/value pairs that defines a benchmark case permutation. Top-level keys must be strings. Values must be strings, too. Keys with empty string values are ignored.  The special key "name" must be provided with a string value: it indicates the name of the conceptual benchmark that was performed for obtaining the result at hand. All case permutations of a conceptual benchmark by definition have this name in common.  We advise that each unique case (as defined by the complete set of key/value pairs) indeed corresponds to unique benchmarking behavior. That is, typically, these key/value pairs directly correspond to input parameters to your benchmarking method.  Example: a conceptual benchmark with name "foo-write-file" might have meaningful case permutations involving tag names such as `compression_method` (values: `gzip`, `lzma`, ...), `file_format` (values: `csv`, `hdf5`, ...), `dataset_name` (values: `foo`, `bar`, ...). For each conceptual benchmark, it is OK to have no or many case permutations.',
```
jonkeane:
```suggestion
                        "description": 'The set of key/value pairs that defines a benchmark case permutation. Top-level keys must be strings. Values must be strings, too. Keys with empty string values are ignored.  The special key "name" must be provided with a string value: it indicates the name of the conceptual benchmark that was performed for obtaining the result at hand. All case permutations of a conceptual benchmark by definition have this name in common.  We advise that each unique case (as defined by the complete set of key/value pairs) indeed corresponds to unique benchmarking behavior. That is, typically, these key/value pairs directly correspond to input parameters to your benchmarking method.  Example: a conceptual benchmark with name "foo-write-file" might have meaningful case permutations involving tag names such as `compression_method` (values: `gzip`, `lzma`, ...), `file_format` (values: `csv`, `hdf5`, ...), `dataset_name` (values: `foo`, `bar`, ...). For each conceptual benchmark, it is OK to have no or many tags which would result in one or many case permutations (if you supply no tags, there is necessarily a single mutation with a special property that it has no tags).',
```

Ok to clean this up here? I'm open to other wording suggestions of course, but do think it would be good to be explicit here.

conbench/api/_examples.py
```diff
@@ -127,7 +127,7 @@ def _api_benchmark_entity(
         "tags": {
             "id": case_id,
             "compression": "snappy",
-            "cpu_count": 2,
+            "cpu_count": "2",
```
jonkeane, line 128: Seeing this here has me wondering: will forcing all to be strings end up making some of our downstream visualizations (in the app as well as in other external reporting) more complicated needing to know which of these values can | should be cast (back) to numbers?

Ultimately I suspect that it is a very small number of places where we will end up doing this (`cpu_count` being one), so maybe that's totally fine to deal with downstream. But seeing this here now made me think of the code I would need to write in order to order by `cpu_count` if I were using that in a report. 

I'm not trying to block this direction (I suspect like I say above, the number of places we do that will be small so it's worth it for the benefits of consistency), but it would be good to actively make that choice here :)

austin3dickey: At first I was convinced that it's better to store these in their "native" types. But one of you convinced me that strings may even be better for representing numbers, because it's impossible to accidentally mess with floating-point precision and float-vs-int. We wouldn't want `"cpu_count": 2.0`, for example, to ever show up.

So I think it's a bit of an overreach for the Conbench API to promise "we will store your values in whatever native types you give us, guaranteed without loss." And if that's the case, well, strings are a great way to ensure there's no loss. Downstream applications can handle the casting logic.

alistaire47: yes but `False` vs `false` vs `FALSE` and sorting. JSON types are junk, but no types is also junk

conbench/tests/api/_expected_docs.py
jgehrcke: This is the auto-generated diff corresponding to the actual diff somewhere else in this PR: https://github.com/conbench/conbench/pull/948/files#diff-9d75d1bf6ac23bfc056a410721cf4c9bccd8b302f8e3e83e4b4bf1698d0eaad7R590

conbench/api/_examples.py
jgehrcke: It's a lovely, important discussion.

It's not an easy decision, for sure. I am rather confident, but not toooo confident.

You almost had me "yes OK let's allow for boolean in addition to string". But then I kind of decided against this again.

I think we need to do a little more work discussing pros and cons.

This here is a really important choice with respect to interfaces to humans. I believe we should think about the querying interface, and dynamic case querying, esp. with browser UI-based query construction.

A central question: do we want to require a _typed query_?

- Imagine e.g. putting `=false` into a URL query string -- what should it mean?
- Is it simpler to type `=false` into an input field than to fill in an actual boolean from a drop-down?

These two points above are important. My feeling is that we do not want to rely on a type-aware query language here.

It's a trade-off, as always. Some higher-level thoughts, and comparison with other systems that needed to think about this:

The mental burden of allowing for more than one type is actually rather tremendous whenever we build an interface for humans on the input and output side of things. Do we really want to pay attention to displaying the difference between a `false` and a `"false"` to the user (or should the string value`"false"` be forbidden if the boolean value `false` is allowed? No).

More perspective: these case tags are primarily labels. Their role as metadata storage for an external system is secondary, in my opinion (it is not unimportant, but in a trade-off consideration some aspects by definition need to suffer a bit).

Let's think about why we are different from labels in the Prometheus ecosystem. There might be good arguments!

https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

> Every time series is uniquely identified by its metric name and optional key-value pairs called labels.

The similarity to benchmark name and case key/value pairs cannot be overlooked. I mentioned this [before](https://github.com/conbench/conbench/issues/862).

> Changing any label value, including adding or removing a label, will create a new time series.

Same for us.

> Labels enable Prometheus's dimensional data model: any given combination of labels for the same metric name identifies a particular dimensional instantiation of that metric (for example: all HTTP requests that used the method POST to the /api/tracks handler). The query language allows filtering and aggregation based on these dimensions. 

> A label with an empty label value is considered equivalent to a label that does not exist.

Just sayin.

> Given a metric name and a set of labels, time series are frequently identified using this notation:
> `<metric name>{<label name>=<label value>, ...}`

IMO, stringified notation and stringified querying is rather valuable in practice.

But yeah maybe we want to be different. We should find out why, though.

For completeness, this is the alternative model: https://opentelemetry.io/docs/reference/specification/common/#attribute 

> The attribute key MUST be a non-null and non-empty string.
> The attribute value is either: A primitive type: string, boolean, double precision floating point (IEEE 754-1985) or signed 64 bit integer.

I just don't know how querying works in that ecosystem.

alistaire47: As labels, yes, all strings make sense. But tags don't start as labels, they start as arguments to benchmarks, which need to interpret them as types and work for authors (so switching to `memory_mapped="false"` is suboptimal). We can postprocess those arguments into strings for tags (like what happens for TPC-H query numbers -> IDs), but that indirection does come at a mental cost when tracing activity from benchmark through to frontend. Maybe we're ok with that cost?

Moreso than anything else, though, before enforcing that tags are strings we need PRs to all the benchmark repos to adjust tags such that lots of POSTs don't start failing. And we should clean up the docs which presently state `tags` is `Dict[str, Any]`.

alistaire47: ...and we might want to put together a little set of opinions on how to stringify, e.g. `"2."` vs `"2.0"`, capitalization for booleans, etc. if we want to be able to consistently convert them back to types where we need

conbench/tests/api/_expected_docs.py
jgehrcke: Ah, I now saw what you edited. This is about https://github.com/conbench/conbench/pull/936#discussion_r1145988615 (where I very much agreed, just missed that before merge; sry for that)

conbench/api/_examples.py
jgehrcke: Thanks for more feedback :). 

> Moreso than anything else, though, before enforcing that tags are strings we need PRs to all the benchmark repos to adjust tags such that lots of POSTs don't start failing. 

Of course. One fundamental idea in this PR is to not break (most of) our current tooling. Hope you saw corresponding code comments and code. In my proposal, `int` and `float` are explicitly still accepted with the idea to have a transition phase. `boolean` I forgot about, this belongs into the list clearly. Middle ground therefore would be to have a transition phase in which we disallow only `[]` and `{}`.

I think you're saying that even those are used somewhere in code that we control, yes? Or is your feedback "no, I don't expect breakage of existing jobs when we start rejecting json objects and arrays as value types"?

Really, I don't want to be too strict initially (see code comments); this is about thinking about future state, and then slowly working towards that. So, proposed middle ground buying us time for better thinking is to accept int, float, boolean (but to stringify already before storing in DB, to massage a certain philosophy into our brains).

>  And we should clean up the docs which presently state tags is Dict[str, Any].

Of course. This PR tries to improve textual specification, did you see that? Which part of docs do you refer to with specifically "state tags is Dict[str, Any]"?

> [...] we might want to put together a little set of opinions on how to stringify, e.g. "2." vs "2.0", capitalization for booleans

I always try to point this out elsewhere: it's important to draw imaginary lines between areas of responsibility. These two statements certainly relate to the idea that benchmark authors _can_ implement a custom convention that allows for the information flow  `any -> str/str -> any`, hopefully without ambiguity. We _can_ guide them to do so. But most importantly, we don't have to dictate a way for doing so.

>  come at a mental cost when tracing activity from benchmark through to frontend. Maybe we're ok with that cost? 

I agree about the cost. For a more complete pros/cons discussion we have to take into account the type-aware query consideration in my lengthy comment above.

Today's state, I want to remind us (to corroborate that cleanup will provide value):
- Today we present more than one stringified version of the "tags" in the UI. That stringified version has already dropped type information, in an undocumented way.
- In the API on the output side of things (which can be used for "reporting" we add a new key/value pair to the "tags".

conbench/tests/api/_expected_docs.py
jgehrcke: Cool. So, I added the corresponding change to `benchmark_result.py` in #954 (where we also needed to re-generate the `_expected_docs.py`).

conbench/api/_examples.py
alistaire47: > I think you're saying that even those are used somewhere in code that we control, yes? Or is your feedback "no, I don't expect breakage of existing jobs when we start rejecting json objects and arrays as value types"?

Non-empty arrays are generated by [this benchmark](https://github.com/voltrondata-labs/arrowbench/blob/ebfab3d5535a1a7ba8680be4d0394296b9b1a5ee/R/bm-array-altrep-materialization.R#L19), but since we're not running that presently, we can edit it pretty easily before sending anything to Conbench. So asserting a flat structure shouldn't break anything active.

> > And we should clean up the docs which presently state tags is Dict[str, Any].
> 
> Of course. This PR tries to improve textual specification, did you see that? Which part of docs do you refer to with specifically "state tags is Dict[str, Any]"?

Importantly in benchadapt.BenchmarkResult: https://github.com/conbench/conbench/blob/16d0243833764f13f84978c43177871a89fb67a9/benchadapt/python/benchadapt/result.py#L43-L47 Less importantly in places like type hints in labs/benchmarks: https://github.com/voltrondata-labs/benchmarks/blob/75bc80e1bc75b428734f7a018a7f5460d6094274/benchmarks/_benchmark.py#L175-L177



> > [...] we might want to put together a little set of opinions on how to stringify, e.g. "2." vs "2.0", capitalization for booleans
> 
> I always try to point this out elsewhere: it's important to draw imaginary lines between areas of responsibility. These two statements certainly relate to the idea that benchmark authors _can_ implement a custom convention that allows for the information flow `any -> str/str -> any`, hopefully without ambiguity. We _can_ guide them to do so. But most importantly, we don't have to dictate a way for doing so.

My worry here is if we want types back later, e.g. for sorting or visualization. Maintaining a function like [type.convert()](https://rdrr.io/r/utils/type.convert.html) sounds like hell. If we're sure we're _only_ ever going to use these as labels, this doesn't matter.

jgehrcke: > So asserting a flat structure shouldn't break anything active.

Thanks!!

>  in benchadapt.BenchmarkResult:

> in places like type hints in labs/benchmarks

OK, thanks for sharing. Can do this in a follow-up.

jgehrcke:
I rebased this, solved merge conflicts, and have updated spec and behavior after this great discussion.

See diff for details. Summary: int/float/boolean value types are currently not rejected. JSON array and object would now be rejected with a Bad Request response.

I also added more test coverage.

> If we're sure we're only ever going to use these as labels

The provocative, productive thought is that within Conbench web application boundaries this will be the case.

Some perspective: a good way of looking at this I think is that _this_ (only using these as labels) is the goal. To make Conbench web application internals easier to reason about, to simplify, and to be able to optimize UX for that scenario. To remove that source of complexity from Conbench web application/database internals.

One way that I also like to think about this is via the term "parameters" vs the term "tags".

If we were to call this "parameters" then I think it would imply a little more of the promise of actually storing all details about input parameters.

If a benchmark author absolutely needs to recover all details of the original parameter set then _from_ tags, then they will find a way! But for Conbench web application internals, we keep it easy.

I want to quote directly what @jonkeane said in his initial feedback:

>  But seeing this here now made me think of the code I would need to write in order to order by cpu_count if I were using that in a report.

Of course. My view: as a benchmark author and as the person working on the integration with the Conbench HTTP API, one would read about the constraints and what `tags` should be used for -- and in this specific case one would do `str(cpu_count)` on the input side of things, and one would do `int(tags['cpu_count'])` on the output side of things, without considering this to be ugly. One uses additional knowledge that Conbench does not need to have internally.

conbench/api/results.py
```diff
@@ -182,14 +186,86 @@ def post(self) -> None:
...
+                continue
+
+            # Note(JP): this code path should go away after we adjust our
+            # client tooling to not send numeric values anymore.
+            if isinstance(value, (int, float, bool)):
+                # I think we first want to adjust some client tooling before
+                # enabling this log line:
+                # log.warning("stringify case parameter value: `%s`, `%s`", key, value)
+                # Replace value, proceed with next key.
+                tags[key] = str(value)
```
alistaire47, line 255: @boshek A little heads-up! I'm not sure offhand if any of the reporting relies on non-string tag values, but if so you may need to cast back

austin3dickey (approved):
I'm cool with shipping this as long as our known downstream reports won't break.

boshek:
One tiny comment if I imagine that I am consuming conbench data from the api for reporting. 

> one would do int(tags['cpu_count']) on the output side of things, without considering this to be ugly. One uses additional knowledge that Conbench does not need to have internally.

It makes sense to me to cast types explicitly after collecting from the api rather than expecting conbench to do it for me particularly if it makes conbench internals simpler.

jgehrcke:
> It makes sense to me to cast types explicitly after collecting from the api rather than expecting conbench to do it for me particularly if it makes conbench internals simpler.

:heart: thank you for the feedback on this @boshek 

Also thank you for feedback @austin3dickey.

I think we might have some interesting insights about this approach in the future; our opinions might drift. But yes, I'd love to take the provocative route here; trying out the _simple_ way! Navigating the simplicity-vs-complexity landscape is kind of my favorite thing to do in software engineering :) 

Thanks for being flexible!

jgehrcke:
I have rebased this on current head of main, resolved merge conflicts, and force-pushed.
Let's land this and be excited about what we learn.

PR #952, opened 2023-03-24
benchalerts: fix language in comments and checks

### LINKED ISSUE #919: benchalerts: add more to PR comments
Right now we show the first 20 lines of the body of the report in the PR comment. But we're missing the title of the report and the link to Conbench, both of which would add value.

austin3dickey:
Closes #919.

This PR makes `benchalerts`'s PR comments less verbose and more actionable. It also clarifies some language around using the lookback z-score method, and adds links to all the commit's runs at the bottom of the report.

As usual, you can see the differences in the `expected_md` folder -- but [here](https://github.com/conbench/benchalerts/pull/5#issuecomment-1483311616) are some new-style comments posted from the integration tests too!

### DIFF
--- a/benchalerts/benchalerts/message_formatting.py
+++ b/benchalerts/benchalerts/message_formatting.py
@@ -89,11 +89,10 @@ def github_check_summary(
             There weren't enough matching historic runs in Conbench to make a call on
             whether there were regressions or not.
 
-            To use the lookback z-score method of determining regressions, there needs
-            to be at least one historic run on the default branch which, when compared
-            to one of the runs on the contender commit, is in the same repository, on
-            the same hardware, and has at least one of the same benchmark case and
-            context pairs.
+            To use the lookback z-score method of determining regressions, there need to
+            be at least two historic runs on the default branch which, when compared to
+            one of the runs on the contender commit, are on the same hardware, and have
+            at least one of the same benchmark case and context pairs.
             """
         )
         # exit early
@@ -185,7 +184,7 @@ def pr_comment_link_to_check(
         comment += _clean(
             f"""
             There {were} {len(full_comparison.benchmarks_with_z_regressions)} benchmark
-            result{s} with a performance regression:
+            result{s} indicating a performance regression:
             """
         )
         comment += _list_results(full_comparison.benchmarks_with_z_regressions, limit=2)
--- a/benchalerts/benchalerts/message_formatting.py
+++ b/benchalerts/benchalerts/message_formatting.py
@@ -11,6 +11,27 @@ def _clean(text: str) -> str:
     return textwrap.fill(textwrap.dedent(text), 10000).replace("  ", "\n\n").strip()
 
 
+class _Pluralizer:
+    """Depending on the length of the input list, return the correct pluralization
+    suffixes and plural forms of special English words.
+    """
+
+    def __init__(self, some_list: list) -> None:
+        self.plural = len(some_list) != 1
+
+    @property
+    def were(self):
+        return "were" if self.plural else "was"
+
+    @property
+    def s(self):
+        return "s" if self.plural else ""
+
+
+def _run_bullet(reason: str, time: str, link: str) -> str:
+    return f"- {reason.title()} Run at [{time}]({link})"
+
+
 def _list_results(benchmark_results: List[BenchmarkResultInfo]) -> str:
     """Create a Markdown list of benchmark result information."""
     out = ""
@@ -19,9 +40,11 @@ def _list_results(benchmark_results: List[BenchmarkResultInfo]) -> str:
     for benchmark_result in benchmark_results:
         # Separate each run into a section, with a title for the run
         if benchmark_result.run_id != previous_run_id:
-            out += (
-                f"\n\n- {benchmark_result.run_reason.title()} Run at "
-                f"[{benchmark_result.run_time}]({benchmark_result.run_link})"
+            out += "\n\n"
+            out += _run_bullet(
+                benchmark_result.run_reason,
+                benchmark_result.run_time,
+                benchmark_result.run_link,
             )
             previous_run_id = benchmark_result.run_id
         out += f"\n  - [{benchmark_result.name}]({benchmark_result.link})"
@@ -45,8 +68,8 @@ def github_check_summary(
             ## Benchmarks with errors
 
             These are errors that were caught while running the benchmarks. You can
-            click the link next to each result to go to the Conbench entry for that
-            benchmark, which might have more information about what the error was.
+            click each link to go to the Conbench entry for that benchmark, which might
+            have more information about what the error was.
             """
         )
         summary += _list_results(full_comparison.benchmarks_with_errors)
@@ -55,21 +78,26 @@ def github_check_summary(
 
     if full_comparison.no_baseline_runs:
         summary += _clean(
-            f"""
-            Conbench could not find a baseline run for contender commit `{hash}`. A
-            baseline run needs to be on the default branch in the same repository, with
-            the same hardware, and have at least one of the same benchmark case/context
-            pairs as one of the runs on the contender commit.
+            """
+            There weren't enough matching historic runs in Conbench to make a call on
+            whether there were regressions or not.
+
+            To use the lookback z-score method of determining regressions, there needs
+            to be at least one historic run on the default branch which, when compared
+            to one of the runs on the contender commit, is in the same repository, on
+            the same hardware, and has at least one of the same benchmark case and
+            context pairs.
             """
         )
         # exit early
         return summary
 
+    s = _Pluralizer(full_comparison.benchmarks_with_z_regressions).s
     summary += _clean(
         f"""
         Contender commit `{hash}` had
-        {len(full_comparison.benchmarks_with_z_regressions)} performance regression(s)
-        compared to its baseline runs.
+        {len(full_comparison.benchmarks_with_z_regressions)} performance regression{s}
+        using the lookback z-score method.
         """
     )
     summary += "\n\n"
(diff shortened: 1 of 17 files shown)


jonkeane:
Thanks for this! I'm excited to see this cleaned up and refined. I'm going to start out generally + with the content first — will dig deeper into the code later.

image

For the message: I love that it's shorter for sure — but this might be a bit too far in the short direction. I know this is a tricky balancing act, but getting this right will help us tremendously with uptake:

In this case we have a single regression, but now we just say "there's a regression" and force someone to click through to get a link to that regression. But that's actually more content then if we had just listed the one regression and a link, no? I would love to see this be something like the first 2-3 regressions if there are any and then a `...` and a call to go to the report. We might tune this knob of how long before it `...`s over time — see what folks like. But ideally if there's a tiny number of regressions it should be quick to get to them, and anything after that you've been teased enough to go into the report and see more.

_Something like (if there's a small number of regressions):_

> 🚨 There was a performance regression:
> * [flatMap](http://velox-conbench.voltrondata.run/benchmarks/ff7a1a86df5a4d56b6dbfb006c13c638)
> 
> The [full Conbench report](https://github.com/conbench/benchalerts/runs/12262727829) for commit `c76715c9` has more details. 

_With no regressions something like:_

> 🎉  There were no benchmark regressions
> 
> The [full Conbench report](https://github.com/conbench/benchalerts/runs/12262727829) for commit `c76715c9` has more details. 

_And finally, if there are a large number of regressions, something like:_

> 🚨 There were performance regressions:
> * [flatMap0](http://velox-conbench.voltrondata.run/benchmarks/ff7a1a86df5a4d56b6dbfb006c13c638)
> * [flatMap1](http://velox-conbench.voltrondata.run/benchmarks/ff7a1a86df5a4d56b6dbfb006c13c638)
> * [flatMap2](http://velox-conbench.voltrondata.run/benchmarks/ff7a1a86df5a4d56b6dbfb006c13c638)
> * and 22 more see the [full Conbench report](https://github.com/conbench/benchalerts/runs/12262727829) for these
> 
> The [full Conbench report](https://github.com/conbench/benchalerts/runs/12262727829) for commit `c76715c9` has more details. 

And then the same thing for errors, if there are a small(ish) number of them, they all get listed, if not we `...` and point to the full report. How does that sound?


image

I'm a but surprised by the note on this — that commit looks like it _does_ have a baseline, no? Maybe this is just an artifact of  this being testing? 

For the heading "All benchmark runs" it would be nice to use something like what we have in details below (and see my ask about de-duping this): "All benchmark runs for the commit ..." or "All benchmark runs for this commit"

In the details:

> Conbench has details about 1 run on this commit.

That first line feels a little duplicative of the runs info up above. Could we condense those together? 

> This report was generated using the lookback z-score method with a z-score threshold of 5.0. A regression is defined as a benchmark exhibiting a z-score higher than the threshold in the "bad" direction (e.g. down for iterations per second; up for total time taken).

I like the details here, though I'm not sure we need that second sentence. If we want to talk about being outside the bounds, how about "A regression is defined as a benchmark result with a z-score that exceeds the threshold". We can totally document the "bad" direction business, but IMO that really should be elsewhere and is more noise here than help. I could be wrong, but I'm not certain we have ever had any confusion about what the right direction is and what that means so maybe we can streamline this prose which will make the other things more clear + emphasized.

austin3dickey:
Thanks for the feedback! This is exactly what I was hoping for. I'll go ahead and implement these tweaks.

austin3dickey:
>I'm a but surprised by the note on this — that commit looks like it does have a baseline, no? Maybe this is just an artifact of this being testing?

@jonkeane This "Note" is a bit more nuanced. In this case, the baseline run is actually on the _grandparent_ commit of the commit we're reporting on. We couldn't find an effective baseline run on the direct parent commit, so we had to take a step back in the git graph. That's why we have the "Note" here; in the past, some users were interested in knowing when this kind of thing happens.

I think this "Note" will go away once we do #953 (soon!) because then we won't really care about the baseline; we'll be comparing to all matching historic runs using the lookback z-score method. It shouldn't matter then if we've skipped a few commits in history.

>I like the details here, though I'm not sure we need that second sentence.

Yes, agreed, it's a lot of noise here. I'll remove the definition for now, since it's not the most precise and it's also wordy, and once we have precise docs about the lookback z-score method we can link to those here.

austin3dickey:
Okay, I implemented @jonkeane 's feedback. New comment example is [here](https://github.com/conbench/benchalerts/pull/5#issuecomment-1485672711). Take a look!

jonkeane:
> @jonkeane This "Note" is a bit more nuanced. In this case, the baseline run is actually on the grandparent commit of the commit we're reporting on. We couldn't find an effective baseline run on the direct parent commit, so we had to take a step back in the git graph. That's why we have the "Note" here; in the past, some users were interested in knowing when this kind of thing happens.

I think it's ok | good to message when this happens, but we should be super clear about it. I read the message as extra cruft (even though I do know the various bits here) but having something that explained how many hops or whatever would be helpful here. Though this might be a bit of a stretch goal regardless, maybe we cut scope to ship this without this and we can add it later?

austin3dickey:
>Though this might be a bit of a stretch goal regardless, maybe we cut scope to ship this without this and we can add it later?

Cool, I'll cut the `warn_if_baseline_isnt_parent` Note in a separate PR (because it will touch a lot of (to be fair, gross) code)

alistaire47 (approved):
Lgtm, though tbh the report does not really currently have more details as promised. I'm not sure how much it's worth pulling more in vs just redirecting to Conbench here. If we plan on fleshing out the report this is fine, but if it's likely to remain as-is for a while, maybe we should emphasize the Conbench links as where to click for more details rather than adding another click for the user.

benchalerts/benchalerts/message_formatting.py
```diff
@@ -11,17 +11,47 @@ def _clean(text: str) -> str:
     return textwrap.fill(textwrap.dedent(text), 10000).replace("  ", "\n\n").strip()
 
 
-def _list_results(benchmark_results: List[BenchmarkResultInfo]) -> str:
+class _Pluralizer:
```
jgehrcke, line 14: :)

benchalerts/tests/unit_tests/expected_md/summary_noerrors_nobaselines.md
```diff
@@ -1,3 +1,5 @@
 ## Benchmarks with performance regressions
 
-Conbench could not find a baseline run for contender commit `no_basel`. A baseline run needs to be on the default branch in the same repository, with the same hardware, and have at least one of the same benchmark case/context pairs as one of the runs on the contender commit.
+There weren't enough matching historic runs in Conbench to make a call on whether there were regressions or not.
+
+To use the lookback z-score method of determining regressions, there needs to be at least one historic run on the default branch which, when compared to one of the runs on the contender commit, is in the same repository, on the same hardware, and has at least one of the same benchmark case and context pairs.
```
jgehrcke: > there needs to be at least one historic run

Is that true? :) Do we provide a result when we only have one datapoint in the "lookback" window?

jgehrcke: > is in the same repository,

I like the completeness, but it feels like this shouldn't be part of this list, but it should be kind of obvious (and implied by "there needs to be at least one historic run on the default branch which").

benchalerts/benchalerts/message_formatting.py
```diff
@@ -11,17 +11,47 @@ def _clean(text: str) -> str:
...
+
+def _run_bullet(reason: str, time: str, link: str) -> str:
+    return f"- {reason.title()} Run at [{time}]({link})"
+
+
+def _list_results(
+    benchmark_results: List[BenchmarkResultInfo], limit: Optional[int] = None
+) -> str:
     """Create a Markdown list of benchmark result information."""
     out = ""
```
jgehrcke, line 39: When dynamically assembling Markdown, I have often named the variable `md` :)

benchalerts/benchalerts/message_formatting.py
```diff
@@ -98,32 +145,63 @@ def github_check_details(full_comparison: FullComparisonInfo) -> Optional[str]:
...
+        )
+        comment += "\n\n"
+    elif full_comparison.benchmarks_with_z_regressions:
+        pluralizer = _Pluralizer(full_comparison.benchmarks_with_z_regressions)
+        were = pluralizer.were
+        s = pluralizer.s
+        comment += _clean(
+            f"""
+            There {were} {len(full_comparison.benchmarks_with_z_regressions)} benchmark
+            result{s} with a performance regression:
```
jgehrcke: > with a performance regression:

I prefer something like "indicating a performance regression". That is more scientific writing style: less absolute and implies a little more that we know what we are doing (we apply a certain method that _tries to_ detect a performance regression, and that method is imperfect).

jgehrcke (changes requested):
Looks lovely, thanks! With respect to the wording I just have one 'please change', as commented below (to make the conclusion sound less definite, towards more credibility).

benchalerts/benchalerts/message_formatting.py
jonkeane: 💯

austin3dickey: Oh good catch; I tried to use language like that elsewhere but missed it here.

benchalerts/tests/unit_tests/expected_md/summary_noerrors_nobaselines.md
austin3dickey: Good callout; switching to "at least two."

Issue #949, opened 2023-03-24
benchmark result view: hide case ID


jgehrcke:
Now the case "id" pops up in the benchmark result view, in the list case-defining key/value pairs. This is not a user-given piece of data. Hide it again in those enumerations/stringifications where we want to show the user-given case key/value pairs.

This is the database primary key; it's an implementation detail. 

The case ID can be represented as a string, maybe. Relates to this discussion: https://github.com/conbench/conbench/pull/948#discussion_r1149090197

But: build a new section in the benchmark result view which self-documents the case permutation, and _there_ we can show the ID, because it uniquely represents the case. And we can use that to self-document nicely what all data points in the history plot have in common.

jgehrcke:
This was done in https://github.com/conbench/conbench/pull/954 :rocket:

Issue #865, opened 2023-03-14
ui: benchmark result view: breadcrumb: hierarchy not properly used


jgehrcke:
![image](https://user-images.githubusercontent.com/265630/224996374-dccc48d9-b59c-4919-895c-80e57226795e.png)

`Benchmark / 3ba0f122cb974472b507882c09c1a3f0` are the same level, the separator is misleading.

jgehrcke:
When changing this, also change this to "benchmark result"

jgehrcke:
A little better now, referring to "benchmark result", and showing the ID: 
![image](https://user-images.githubusercontent.com/265630/225242429-f28886d4-a3cb-4570-bf81-880d89f71e00.png)

We can shorten the ID (for display), and _also_ make this a link.

jonkeane:
> We can shorten the ID (for display), and also make this a link.

Also what about the form: 

> Run / Batch / Benchmark result: `2b6c486`

+ `: ` and then like you say shortening the ID + linking

jgehrcke:
> Also what about the form:

I like that! yes!

jgehrcke:
Improved a bit in #954; current look:
![image](https://user-images.githubusercontent.com/265630/228216979-3522282d-0770-47af-a315-ef8a1a29e509.png)

Closing this as 'done' for now.

PR #954, opened 2023-03-27
UI/API: hide case ID/primary key, update history plot title

jgehrcke:
What's currently exposed as `tags.id` is an implementation detail. It is in fact the primary key of a `case` entity in the DB.

It should not be exposed by the API, in particular not as part of `tags`. This is confusing; because it looks like this was added at the input side of things.

Bringing structure into this has the nice side effect of automatically resolving #949.

For "case ID" we want to have a concept, yes. Ideally there is _one_ way to stringify a case so that it unambiguously represents the case, without information loss. This relates to the discussion we have in https://github.com/conbench/conbench/pull/948.

Also see commit msgs.

For paper trail, I'd love to state that this is a breaking change in the HTTP/JSON API surface in the sense that on the output side of things the `tags` JSON object will now not have the `id` property anymore.

Also worked a bit on history plot title, after discussing this a bit last week in Conbenchat:

Screenshot from 2023-03-24 16-52-07

### DIFF
--- a/conbench/templates/benchmark-result.html
+++ b/conbench/templates/benchmark-result.html
@@ -18,11 +18,13 @@
 </nav>
 
 
+
 <br />
 {% if history_plot_info.jsondoc is not none %}
-<div align="center"><p class="fs-4">{{ benchmark.display_bmname }}</p></div>
-<div align="center">{{ benchmark.display_case_perm }}</div>
-<div id="plot-history-0" align="center"></div>
+<h3>History plot for benchmark '{{ benchmark.display_bmname }}'</h3>
+<p class="fst-italic fw-light">For a specific case permutation and hardware/context (see below)</p>
+<!-- <div>{{ benchmark.display_case_perm }}</div> -->
+<div class="mt-4" id="plot-history-0" align="center"></div>
 <div class="row">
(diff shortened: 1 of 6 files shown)


austin3dickey (approved):

jgehrcke:
Thanks for review @austin3dickey -- I am going to land this now that we have passing tests. Of course, feedback is still appreciated. The UI title change I look at as temporary; already some improvement.

conbench/templates/benchmark-result.html
```diff
@@ -10,16 +10,21 @@
     <li class="breadcrumb-item active">
       <a href="{{ url_for('app.batch', batch_id=benchmark.batch_id) }}">Batch</a>
     </li>
-    <li class="breadcrumb-item active aria-current="page">Benchmark result {{ benchmark.id }}</li>
+    <li class="breadcrumb-item active aria-current="page">
+      <a href="{{ url_for('app.benchmark', benchmark_id=benchmark.id) }}">
+      Benchmark result {{ benchmark.id[:7] }}...</a>
+    </li>
```
jgehrcke, line 16: This is for #865 CC @jonkeane 
image

Issue #633, opened 2023-01-27
discuss pros and cons for various 'history plot' methods


jgehrcke:
There are many cool ideas to visualize history for a given benchmark case permutation. Let's systematically collect thoughts about pros and cons.

Starting with two main ideas here (not aiming for completeness, let's add info as we have ideas)

### Measurement result over time, in commit-agnostic view

- each data point corresponds to a benchmark result
- position on time axis: time of the benchmark run
- show _all_ corresponding results in the plot, regardless of commit and branch information.

The view is very easy to understand. Yes, it's a wild mixture of main branch and PR branches, and time axis is not strictly related to code evolution. But each data point may _quickly_ be inspected for corresponding repo/commit/branch information.

More thoughts to be collected.

### Measurement result over time, in code evolution-flavored view

- each data point corresponds to a benchmark result
- position on time axis: time of _commit_ corresponding to benchmark run
- show only those results corresponding to commits that are linearly related (and _maybe_ not even all of those)

This view is maybe best understood with supporting text or visualization, clearly describing what the shown data points have in common (and also describing that potentially many other data points for this same benchmark case permutation are not shown).

I think that's the current history plot. If one wants to write a text about how exactly it's build up one realized that it takes some effort to explain properly. We should try to do so on the history plot page itself (it should be self-documenting).

alistaire47:
Do we want to plot all iterations for a permutation? I think we'd tend towards yes, but in the commit-agnostic view here it may get busy and hard to understand depending on how busy the repo is

jgehrcke:
> Do we want to plot all iterations for a permutation?

Good question. I also think that in certain views we want that, yes. And I agree, one needs to be careful with choosing a visualization method that things do not get too busy.

austin3dickey:
Will come back and add thoughts to this, but #1489 is a symptom that this ticket is important. :)

PR #956, opened 2023-03-27
z-scores: Autodetect outliers and ignore them

### LINKED ISSUE #730: Auto-detect an outlier in the past and exclude it
For an example like: 
image

https://conbench.ursa.dev/benchmarks/6ea6b5c03ef44e529128c90894c3fc0d/

Within a small handful of runs _after_ the spike, we should be able to detect that single run that is well outside of the SDs and auto mark that specific benchmarkResult as a suspected outlier. IMO "just" saying "would this have been marked as a regression and then immediate improvement, if yes, it's an outlier" so we don't need to create or implement any new maths to do this — we might get more complex and iterate later as we see how people feel about this feature and if it matches their real-world expectations, of course — but let's get something simple out now. 

Once we have markings for suspected outlier, we can exclude it entirely from the distribution calculations (both mean and sd; possibly optionally — see below for why making this optional would be nice). 

We will probably need to do _something_ with the UI (though maybe as a broken down follow on?), here are some non-exhaustive thoughts:
For the UI: we might want to make this optional in the display, so that folks could toggle it back and forth (default ignore all suspected outliers) but let folks see what it would look like if they included them. Or (not exclusive or!) we could also let someone (with authz) override the suspect outlier and say "no, you think it is, but it really isn't I want it to always be included"

Some additional context is in #474 where there's a discussion of a manual version of this.

alistaire47:
Closes #730. Integrates with existing z-score code in `_add_rolling_stats_columns_to_df()` with new code that detects outliers and adds a boolean column for them. In the course of this is also automatically detects step changes, though with a slightly different approach than previously; this is not integrated for now, though it is set to do so easily should we wish to in the future.

The basic method here still uses z-scores, but trims off the top and bottom 5% of data (specifically the point-to-point diffs) when calculating mean and standard deviation to calculate them, so outliers and step changes do not affect magnitudes. 

[Trimmed estimators](https://en.wikipedia.org/wiki/Trimmed_estimator) remove the impact of outliers on the statistic, allowing us to use them to create z-scores by which to establish which points are outliers. Without trimming, outliers would blow out the standard deviation and tweak the mean a bit, shrinking their z-scores. Trimming only useful for getting less-biased estimators, though; we still need z-scores to determine outlier status since the percentage of outliers will not always be the percent trimmed. 

10% is not totally random, as in anomaly detection literature, an anomaly is defined as something that occurs less than 10% of the time; more than that and it is a mode. With z-scores, outliers are defined as those points which subsequently revert to non-extreme values; step changes are those that do not.

### DIFF
--- a/conbench/api/_examples.py
+++ b/conbench/api/_examples.py
@@ -374,6 +374,7 @@ def _api_history_entity(benchmark_id, case_id, context_id, run_name):
                 "rolling_mean_excluding_this_commit": 0.036369,
                 "rolling_stddev": 0.0,
                 "segment_id": 0.0,
+                "is_outlier": False,
             },
         }
     ]
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -484,6 +484,7 @@ def time_series_plot(
     # key which is extracted here by default.
     source_mean_over_time = _source(inliers, unit, formatted=formatted)
     source_outlier_mean_over_time = _source(outliers, unit, formatted=formatted)
+    source_unfiltered_mean_over_time = _source(samples, unit, formatted=formatted)
 
     source_min_over_time = bokeh.models.ColumnDataSource(
         data=dict(
@@ -629,7 +630,9 @@ def time_series_plot(
     # of a specific data source this can be decided in the callback when
     # passing a data source to the callback and then inspecting
     # `s1.selected.indices`.
-    p.js_on_event("tap", gen_js_callback_tap_detect_unselect(source_mean_over_time))
+    p.js_on_event(
+        "tap", gen_js_callback_tap_detect_unselect(source_unfiltered_mean_over_time)
+    )
 
     p.xaxis.formatter = get_date_format()
     p.xaxis.major_label_orientation = 1
@@ -643,13 +646,12 @@ def time_series_plot(
         if multisample_count:
             label += f" (n={multisample_count})"
 
-    scatter_outlier_mean_over_time = p.circle(
-        source=source_outlier_mean_over_time,
+    scatter_mean_over_time = p.circle(
+        source=source_unfiltered_mean_over_time,
         legend_label=label,
-        name="outliers",
+        name="samples",
         size=6,
-        line_color="#ccc",
-        fill_color="white",
+        color="#ccc",
         line_width=1,
         selection_color="#76bf5a",  # like bootstrap panel dff0d8, but darker
         selection_line_color="#5da540",  # same green, again darker
@@ -659,12 +661,13 @@ def time_series_plot(
         nonselection_line_alpha=1.0,
     )
 
-    scatter_mean_over_time = p.circle(
-        source=source_mean_over_time,
-        legend_label=label,
-        name="samples",
+    scatter_outlier_mean_over_time = p.circle(
+        source=source_outlier_mean_over_time,
+        legend_label=f"outlier {label}",
+        name="outliers",
         size=6,
-        color="#ccc",
+        line_color="#ccc",
+        fill_color="white",
         line_width=1,
         selection_color="#76bf5a",  # like bootstrap panel dff0d8, but darker
         selection_line_color="#5da540",  # same green, again darker
@@ -766,6 +769,7 @@ def time_series_plot(
 
     hover_renderers = [
         scatter_mean_over_time,
+        scatter_outlier_mean_over_time,
         cur_bench_mean_circle,
     ]
 
@@ -780,18 +784,6 @@ def time_series_plot(
             ],
             formatters={"$x": "datetime"},
             renderers=hover_renderers,
-        )
-    )
-
-    p.add_tools(
-        bokeh.models.HoverTool(
-            tooltips=[
-                ("commit date", "$x{%F}"),
-                ("value", "@values_with_unit"),
-                ("outlier", "True"),
-            ],
-            formatters={"$x": "datetime"},
-            renderers=[scatter_outlier_mean_over_time],
         )
     )
 
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -27,11 +27,11 @@ class BokehPlotJSONOrError:
 
 
 class TimeSeriesPlotMixin:
-    def get_outliers(self, benchmarks):
+    def get_biggest_changes(self, benchmarks):
         benchmarks_by_id = {b["id"]: b for b in benchmarks}
 
-        # top 3 outliers
-        outliers = sorted(
+        # top 3 regressions or improvements
+        biggest_changes = sorted(
             [
                 (abs(float(b["stats"]["z_score"])), b["id"])
                 for b in benchmarks
@@ -40,13 +40,13 @@ class TimeSeriesPlotMixin:
             reverse=True,
         )[:3]
 
-        outliers = [benchmarks_by_id[o[1]] for o in outliers]
-        outlier_ids = [b["id"] for b in outliers]
+        biggest_changes = [benchmarks_by_id[o[1]] for o in biggest_changes]
+        biggest_changes_ids = [b["id"] for b in biggest_changes]
         # Note(JP): TODO: it appears that this is not covered by the test suite
-        outlier_names = [
-            f'{b["display_bmname"]}, {b["display_case_perm"]}' for b in outliers
+        biggest_changes_names = [
+            f'{b["display_bmname"]}, {b["display_case_perm"]}' for b in biggest_changes
         ]
-        return outliers, outlier_ids, outlier_names
+        return biggest_changes, biggest_changes_ids, biggest_changes_names
 
     def get_history_plot(self, benchmark, run, i=0) -> BokehPlotJSONOrError:
         """
@@ -471,6 +471,8 @@ def time_series_plot(
 
     unit = samples[0].unit
     with_dist = [s for s in samples if s.zscorestats.rolling_mean]
+    inliers = [s for s in samples if not s.zscorestats.is_outlier]
+    outliers = [s for s in samples if s.zscorestats.is_outlier]
     formatted, axis_unit = _should_format([s.mean for s in samples], unit)
     dist_change_indexes = [
         ix
@@ -480,14 +482,15 @@ def time_series_plot(
 
     # Note(JP): `samples` is an ordered list of dicts, each dict has a `mean`
     # key which is extracted here by default.
-    source_mean_over_time = _source(samples, unit, formatted=formatted)
+    source_mean_over_time = _source(inliers, unit, formatted=formatted)
+    source_outlier_mean_over_time = _source(outliers, unit, formatted=formatted)
 
     source_min_over_time = bokeh.models.ColumnDataSource(
         data=dict(
             # Insert NaNs to break the line at distribution changes
             x=_insert_nans(
                 util.tznaive_iso8601_to_tzaware_dt(
-                    [s.commit_timestamp.isoformat() for s in samples]
+                    [s.commit_timestamp.isoformat() for s in inliers]
                 ),
                 dist_change_indexes,
             ),
@@ -495,7 +498,7 @@ def time_series_plot(
             # bandwidth. Note(JP): min() must not have None passed as arg, that
             # is why I have changed `s.data` to only contain type float
             # elements (including math.nan, instead of None).
-            y=_insert_nans([min(s.data) for s in samples], dist_change_indexes),
+            y=_insert_nans([min(s.data) for s in inliers], dist_change_indexes),
         )
     )
 
@@ -523,6 +526,7 @@ def time_series_plot(
             rolling_mean=0.0,  # not consumed
             residual=0.0,  # not consumed
             rolling_stddev=0.0,  # not consumed
+            is_outlier=False,  # not consumed
         ),
     )
 
@@ -638,6 +642,22 @@ def time_series_plot(
         label = "benchmark mean"
         if multisample_count:
             label += f" (n={multisample_count})"
+
+    scatter_outlier_mean_over_time = p.circle(
+        source=source_outlier_mean_over_time,
+        legend_label=label,
+        name="outliers",
+        size=6,
+        line_color="#ccc",
+        fill_color="white",
+        line_width=1,
+        selection_color="#76bf5a",  # like bootstrap panel dff0d8, but darker
+        selection_line_color="#5da540",  # same green, again darker
+        # Cannot change the size upon selection
+        # selection_size=10,
+        nonselection_fill_alpha=1.0,
+        nonselection_line_alpha=1.0,
+    )
 
     scatter_mean_over_time = p.circle(
         source=source_mean_over_time,
@@ -760,6 +780,18 @@ def time_series_plot(
             ],
             formatters={"$x": "datetime"},
             renderers=hover_renderers,
+        )
+    )
+
+    p.add_tools(
+        bokeh.models.HoverTool(
+            tooltips=[
+                ("commit date", "$x{%F}"),
+                ("value", "@values_with_unit"),
+                ("outlier", "True"),
+            ],
+            formatters={"$x": "datetime"},
+            renderers=[scatter_outlier_mean_over_time],
         )
     )
 
--- a/conbench/entities/history.py
+++ b/conbench/entities/history.py
@@ -489,7 +489,7 @@ def _add_rolling_stats_columns_to_df(
         for x in df["change_annotations"]
     ]
 
-    # NOTE(EV): If unquoted, this line will integrate manually-specified distribution
+    # NOTE(EV): If uncommented, this line will integrate manually-specified distribution
     # changes with those automatically detected. Before enabling this, we want a way for
     # users to manually remove an automatically-detected step-change.
     #
--- a/conbench/entities/history.py
+++ b/conbench/entities/history.py
@@ -584,7 +584,9 @@ def _calculate_z_score(
     return z_score
 
 
-def zscore_automated_detect(self, df: pd.DataFrame, z_score_threshold=5.0) -> pd.DataFrame:
+def zscore_automated_detect(
+    self, df: pd.DataFrame, z_score_threshold=5.0
+) -> pd.DataFrame:
     """Detect outliers and distribution shifts in historical data
 
     Uses a z-score-based detection algorithm, taking a dataframe of the same
@@ -618,9 +620,7 @@ def zscore_automated_detect(self, df: pd.DataFrame, z_score_threshold=5.0) -> pd
 
     # split / apply
     out_group_df_list = []
-    for _, group_df in tmp_df.groupby(
-        ["case_id", "context_id", "hash", "repository"]
-    ):
+    for _, group_df in tmp_df.groupby(["case_id", "context_id", "hash", "repository"]):
         # clean copy will only get result columns
         out_group_df = copy.deepcopy(group_df)
 
@@ -634,14 +634,7 @@ def zscore_automated_detect(self, df: pd.DataFrame, z_score_threshold=5.0) -> pd
         ) / mean_diff_clipped.std()
 
         group_df["is_shift"] = group_df.z_score.abs() > z_score_threshold
-        group_df["reverts"] = (
-            group_df.is_shift
-            & group_df.is_shift.shift(-1)
-            & (
-                (group_df.z_score + group_df.z_score.shift(1)).abs()
-                > z_score_threshold
-            )
-        )
+        group_df["reverts"] = group_df.is_shift & group_df.is_shift.shift(-1)
         out_group_df["is_step"] = (
             group_df.is_shift
             & ~group_df.reverts
--- a/conbench/entities/history.py
+++ b/conbench/entities/history.py
@@ -472,7 +472,7 @@ def _add_rolling_stats_columns_to_df(
     exclusive on the right side. This is useful if you want to compare each commit to
     the previous commit's rolling stats.
     """
-    df = zscore_automated_detect(df=df)
+    df = _detect_shifts_with_trimmed_estimators(df=df)
 
     # pandas likes the data to be sorted
     df.sort_values(
@@ -588,12 +588,15 @@ def _calculate_z_score(
     return z_score
 
 
-def zscore_automated_detect(df: pd.DataFrame, z_score_threshold=5.0) -> pd.DataFrame:
+def _detect_shifts_with_trimmed_estimators(
+    df: pd.DataFrame, z_score_threshold=5.0
+) -> pd.DataFrame:
     """Detect outliers and distribution shifts in historical data
 
-    Uses a z-score-based detection algorithm, taking a dataframe of the same
-    sort as `_add_rolling_stats_columns_to_df`, and returning that dataframe with
-    two additional columns:
+    Uses a z-score-based detection algorithm that uses trimmed rolling means and standard
+    deviations to avoid influence by outliers. Takes a dataframe of the same sort as
+    `_add_rolling_stats_columns_to_df`, and returning that dataframe with two additional
+    columns:
 
     - `is_step` (bool): Is this point the start of a new segment?
     - `is_outlier` (bool): Is this point an outlier that should be ignored?
--- a/conbench/entities/history.py
+++ b/conbench/entities/history.py
@@ -96,6 +96,7 @@ class HistorySampleZscoreStats:
     rolling_mean: Optional[float]
     residual: float
     rolling_stddev: float
+    is_outlier: bool
 
 
 @dataclasses.dataclass
@@ -216,6 +217,7 @@ def get_history_for_cchr(
             rolling_mean=_to_float(sample.rolling_mean),
             residual=sample.residual,
             rolling_stddev=_to_float(sample.rolling_stddev) or 0.0,
+            is_outlier=sample.is_outlier or False,
         )
 
         # For both, `sample.data` and `sample.times`, expect either None or a
--- a/conbench/entities/history.py
+++ b/conbench/entities/history.py
@@ -4,7 +4,6 @@ import datetime
 import decimal
 import logging
 import math
-from abc import ABC, abstractmethod
 from collections import defaultdict
 from typing import DefaultDict, Dict, List, Optional, Tuple, Union
 
@@ -473,7 +472,7 @@ def _add_rolling_stats_columns_to_df(
     exclusive on the right side. This is useful if you want to compare each commit to
     the previous commit's rolling stats.
     """
-    df = _ZScoreAutomatedDetector().detect(df=df)
+    df = zscore_automated_detect(df=df)
 
     # pandas likes the data to be sorted
     df.sort_values(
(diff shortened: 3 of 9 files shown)


conbench/entities/history.py
```diff
@@ -570,3 +583,82 @@ def _calculate_z_score(
...
+    def detect(self, df: pd.DataFrame) -> pd.Series:
+        """Detect outliers and distribution shifts in historical data
+
+        For a particular detection algorithm, should take a dataframe of the same
+        sort as `_add_rolling_stats_columns_to_df`, and return that dataframe with
+        two additional columns:
+
+        - `is_step` (bool): Is this point the start of a new segment?
+        - `is_outlier` (bool): Is this point an outlier that should be ignored?
+        """
```
alistaire47: I initially added this to try to make it obvious how to add new methods, but unless we actually plan to do so I worry this is just a bunch of useless boilerplate, and the actual class could just be a function. I'm not totally sure future algorithms would have exactly the same inputs and outputs either, depending on how they operate, and I don't want to put too much effort into divining that. Leaning towards ripping this out unless there are contrary opinions.

conbench/entities/history.py
```diff
@@ -570,3 +583,82 @@ def _calculate_z_score(
...
+
+
+class _ZScoreAutomatedDetector(_AutomatedDetector):
+    def detect(self, df: pd.DataFrame, z_score_threshold=5.0) -> pd.DataFrame:
```
alistaire47: `z_score_threshold` here should obviously be parameterized, probably by an env var, but I'm not sure whether we want to use the existing one, since the usage here is slightly different. Open to opinions.

conbench/entities/history.py
```diff
@@ -570,3 +583,82 @@ def _calculate_z_score(
...
+        tmp_df = copy.deepcopy(df)
+
+        # skip computation if no history
+        if df.shape[0] == 0:
+            tmp_df["is_step"] = pd.Series([], dtype=bool)
+            tmp_df["is_outlier"] = pd.Series([], dtype=bool)
+            return tmp_df
```
alistaire47: We apparently call this with no history a _lot_ in our tests

conbench/entities/history.py
```diff
@@ -570,3 +583,82 @@ def _calculate_z_score(
...
+                & (group_df.mean_diff < group_df.mean_diff.quantile(0.95))
+            ]
+            group_df["z_score"] = (
+                group_df.mean_diff - mean_diff_clipped.mean()
+            ) / mean_diff_clipped.std()
+
+            group_df["is_shift"] = group_df.z_score.abs() > z_score_threshold
+            group_df["reverts"] = (
+                group_df.is_shift
+                & group_df.is_shift.shift(-1)
```
alistaire47: This does look one observation into the future because it has to to see if it reverts. We obviously don't know if the last value reverts or not.

conbench/tests/entities/test_history.py
```diff
@@ -11,7 +11,7 @@
     28.477042698148946,
     0.7071067811865444,
     2.1213203435596393,
-    13.435028842544385,
+    None,  # 13.435028842544385,
```
alistaire47: Right now this breaks with either of these values; trying to sort out tests

conbench/entities/history.py
```diff
@@ -495,8 +505,9 @@ def _add_rolling_stats_columns_to_df(
     )
 
     # Add column with rolling mean of the means (only inside of the segment)
-    df["rolling_mean_excluding_this_commit"] = (
-        df.groupby(["case_id", "context_id", "hash", "repository", "segment_id"])
+    df.loc[~df.is_outlier, "rolling_mean_excluding_this_commit"] = (
+        df.loc[~df.is_outlier]
```
alistaire47, line 514: I hate all these `.loc`s, but every other approach I can think of here is also gross. Open to ideas on how to make this prettier.

conbench/entities/history.py
```diff
@@ -339,6 +342,7 @@ def _query_distribution_stats_by_run_id(
             BenchmarkResult.context_id,
             BenchmarkResult.change_annotations,
             BenchmarkResult.mean,
+            BenchmarkResult.timestamp.label("result_timestamp"),
```
alistaire47, line 346: Sorting by this lets us handle multiple results for the same commit correctly, e.g. if a result was an outlier and it got retried and reverted, sorting by result time will allow us to handle that

conbench/entities/history.py
austin3dickey: Agreed, I think this can just be a function.

conbench/entities/history.py
austin3dickey: oof, yeah, it's pretty different from the existing "z-score threshold" concept in this codebase, because it operates on a (clipped) mean-diff distribution instead of a lookback mean distribution. Maybe this parameter could be something like "mean_diff_z_score_threshold" or something less bad than that? ;)

I think it's fine to not introduce the server-wide configuration environment variable until someone asks for it later.

conbench/entities/history.py
```diff
@@ -570,3 +583,82 @@ def _calculate_z_score(
...
+            out_group_df = copy.deepcopy(group_df)
+
+            group_df["mean_diff"] = group_df["mean"].diff()
+            mean_diff_clipped = group_df.mean_diff[
+                (group_df.mean_diff > group_df.mean_diff.quantile(0.05))
+                & (group_df.mean_diff < group_df.mean_diff.quantile(0.95))
+            ]
+            group_df["z_score"] = (
+                group_df.mean_diff - mean_diff_clipped.mean()
+            ) / mean_diff_clipped.std()
```
austin3dickey: Just so we're all clear: this analysis uses an expanding window, correct? So when we look for outliers in mean-diff, we consider the distribution of mean-diffs since the beginning of time.

Notably, this is a different strategy from the lookback z-score method. We had a conversation on why that method needed a max window size [here](https://github.com/conbench/conbench/pull/636#issuecomment-1412422072). Basically, if we didn't have one, we'd risk the distribution having too high a variance (without a LOT of careful pruning).

Here I think I agree with you that that matters less. I don't think the mean-diff distribution necessarily would change a lot over time. In fact, we're kind of relying on a similar property in the lookback z-score method, where we assume the spread of the means doesn't change much over time. (Plus you're clipping outliers so that should help a little more.)

conbench/entities/history.py
```diff
@@ -570,3 +583,82 @@ def _calculate_z_score(
...
+
+            group_df["is_shift"] = group_df.z_score.abs() > z_score_threshold
+            group_df["reverts"] = (
+                group_df.is_shift
+                & group_df.is_shift.shift(-1)
+                & (
+                    (group_df.z_score + group_df.z_score.shift(1)).abs()
+                    > z_score_threshold
```
austin3dickey: What does this criterion do?

conbench/entities/history.py
```diff
@@ -481,6 +488,9 @@ def _add_rolling_stats_columns_to_df(
         for x in df["change_annotations"]
     ]
 
+    # Add in step changes automatically detected
+    df["begins_distribution_change"] = df["begins_distribution_change"] | df["is_step"]
```
austin3dickey: Nice, wow, if this works well we won't need to manually curate the data anymore, and therefore we could probably take away the DISTRIBUTION_COMMITS max window size I talked about elsewhere on this PR. That could be cool.

conbench/entities/history.py
austin3dickey: Yeah, pandas isn't pretty. :)

What does this do for rows that are outliers? Are none of these stats columns populated?

austin3dickey (commented):
Cool stuff. Let me know if you want help with tests. I'd expect this to end up changing the tests' expected z-scores, because it might identify outliers or distribution changes that end up changing the test distributions.

conbench/entities/history.py
alistaire47: Right, they'll just be nan. I think that's what we want for plotting purposes—outliers should just get ignored for rolling things. We could locf (forward-fill) it if we want them populated, but I think that's more confusing if anything.

conbench/entities/history.py
alistaire47: It's not even expanding here, it's just the straight mean of everything minus the clipped points. I think I could do it as expanding or rolling if I inserted nans instead of filtering. That would be more reproducible than this, which depends entirely on what data gets sent. 

Practically speaking, though, none of these are likely to vary much at all, so I'll see if I can make this more reproducible without overcomplicating too much.

conbench/entities/history.py
alistaire47: If the sum of the present and previous z-score is less than the threshold, it reverts, i.e. ends up back within the normal range. I guess this is pretty equivalent to if present < threshold? I think I got caught up in translating some of the thoughts from issue, but I'm not seeing clear benefits to this framing now

conbench/entities/history.py
alistaire47: Yeah, this is a little change with a big impact. It looks good so far, but some more testing before we let this live may be a good idea.

conbench/entities/history.py
alistaire47: removed

conbench/entities/history.py
alistaire47: wait `>`...how is this working 😵‍💫

alistaire47: Ok so I've played around with this some, and realized why I did what I did (though it was still wrong). What I intended, I think, was `((group_df.z_score + group_df.z_score.shift(-1)).abs() < z_score_threshold)`, but that actually makes things slightly worse by turning more outliers into step changes (which they're not) because they don't revert perfectly. In practice, the best results come from just dropping the condition altogether, and just looking for two sequential shifts. I'm not convinced this is robust, but the benefit is it will err towards overclassifying things as outliers instead of step changes, which is less disruptive. But we should test this all more.

jgehrcke:
Thank you for trying to describe the method in words; to me this is the most important part of this PR, before looking at code.

> ignores the top and bottom 5% of data when calculating mean and standard deviation to calculate them, so outliers and step changes do not affect magnitudes. 10% is not totally random

The "10 %" is surprising after the "5 %" in the sentence before that. Should both be "5 %" or both be "10 %"?


> in anomaly detection literature, an anomaly is defined as something that occurs less than 10% of the time; more than that and it is a mode.

Can we link to literature here in this PR, for posterity, etc? I'd love that, would be rather professional :).

> With z-scores, outliers are defined as those points which subsequently revert to non-extreme values; step changes are those that do not.

Is that 'your' definition, for writing the code, or is that also a definition from literature?

I was curious about how we will define 'outlier' here, because obviously we who have worked on serious research problems know that there is no simple rule, and that this can be a difficult, almost philosophical discussion.

The 'immediate reversal' criterion seems to err on the side of caution. The underlying mechanism for an outlier that might look like a regression always is a methodological instability that hits in with a certain probability. That probability is assumed to be _small_. That is, it is rather unlikely for this instability to hit in, say, "twice in a row". However, we know that this still can happen. It looks like this detection method does not try to account for that, which is (I think) a good choice.

conbench/entities/history.py
```diff
@@ -468,6 +472,8 @@ def _add_rolling_stats_columns_to_df(
     exclusive on the right side. This is useful if you want to compare each commit to
     the previous commit's rolling stats.
     """
+    df = zscore_automated_detect(df=df)
```
jgehrcke: Going from top to bottom (I have not looked at the implementation of the function below), I think we can find a more expressive name than `zscore_automated_detect()`. If this is about outlier detection, then maybe put that in the name? If this is about more -- maybe... well :). I know, naming is hard. The current name sounds like this tries to detect the z-score, which is probably not the case.

conbench/entities/history.py
jgehrcke: Two people are wowed, I don't know what this is doing w/o reading further docs. Can we add a come comment so I can have a quicker wow effect, too? :)

conbench/entities/history.py
```diff
@@ -570,3 +582,67 @@ def _calculate_z_score(
         z_score = z_score * -1
 
     return z_score
+
+
+def zscore_automated_detect(df: pd.DataFrame, z_score_threshold=5.0) -> pd.DataFrame:
+    """Detect outliers and distribution shifts in historical data
```
jgehrcke, line 596: :heart: I like this sentence. I didn't know that we wanted to automatically detect distribution shifts, too. Is this required for the outlier detection? Sorry if that should be clear by now, but I'd appreciate clarification.

conbench/entities/history.py
jgehrcke: This is an important topic.

Relates to https://github.com/conbench/conbench/issues/633.

conbench/entities/history.py
jgehrcke: >  if this works well we won't need to manually curate the data anymore

Ah. I think I slowly get this; this is about automatic distribution change detection. I didn't expect this as part of this PR, that's also probably what Austin didn't expect, and therefore the surprise moment.

conbench/entities/history.py
```diff
@@ -570,3 +582,67 @@ def _calculate_z_score(
...
+
+    Uses a z-score-based detection algorithm, taking a dataframe of the same
+    sort as `_add_rolling_stats_columns_to_df`, and returning that dataframe with
+    two additional columns:
+
+    - `is_step` (bool): Is this point the start of a new segment?
+    - `is_outlier` (bool): Is this point an outlier that should be ignored?
```
jgehrcke, line 604: Can you comment on how you think about using the output? 

I can imagine the obvious part where `is_outlier` would be used in `_add_rolling_stats_columns_to_df()`  to exclude those points for all calculations. I think this is what you do above in all those places where I see `df.loc[~df.is_outlier]`.

It appears as if adding `is_step` to the output could be controversial; this could be kind of more than we initially wanted, right?

jgehrcke (commented):
Thanks for putting up the PR! That makes discussion really efficient.

I would love to see some discussion about if we can build valuable outlier detection _without_ automatic distribution change detection _or_ if the outlier detection is strictly coupled to an automatic distribution change detection.

My feeling is that we want to keep the distribution shift detection manual for now; that's really why I think that part of the discussion is important.

alistaire47:
> > ignores the top and bottom 5% of data when calculating mean and standard deviation to calculate them, so outliers and step changes do not affect magnitudes. 10% is not totally random
> 
> The "10 %" is surprising after the "5 %" in the sentence before that. Should both be "5 %" or both be "10 %"?

It's slicing off 5% of data from the top and bottom end for 10% total to get [trimmed estimators](https://en.wikipedia.org/wiki/Trimmed_estimator) akin to the `trim` param of base R's `mean()`: https://rdrr.io/r/base/mean.html for 10% total. Arguably it shouldn't be symmetrical, and we should be slicing off the most extreme 10%, wherever that is, maybe defined by Mahalanobis distance. But that's complicated and the benefits are minimal, so for a first pass I tried to keep it simple.

> > in anomaly detection literature, an anomaly is defined as something that occurs less than 10% of the time; more than that and it is a mode.
> 
> Can we link to literature here in this PR, for posterity, etc? I'd love that, would be rather professional :).

Yeah, I need a good citation for this, but I don't have a good foundational text for anomaly detection; I'm pretty sure I picked that up from a paper proposing a particular method, but I've forgotten which one. You do see 10% used a lot without explanation in that sort of literature, though, e.g. https://arxiv.org/pdf/2106.06115.pdf I don't want to present 10% as too settled, either; this is the sort of thing people argue about, particularly when it comes to assumptions about generative processes.

A future opportunity for improvement here would be to find a way to detect the percentage of outliers. But maybe if you can do this you already have an alternative outlier-detection algorithm.

> > With z-scores, outliers are defined as those points which subsequently revert to non-extreme values; step changes are those that do not.
> 
> Is that 'your' definition, for writing the code, or is that also a definition from literature?
> 
> I was curious about how we will define 'outlier' here, because obviously we who have worked on serious research problems know that there is no simple rule, and that this can be a difficult, almost philosophical discussion.

It's mine, I guess, in this context of timeseries observations—outliers shift and then shift back; steps shift and then do not shift back. Picking more concrete definitions of "outlier" and "step" may be helpful as we firm all this up, but agreed, I don't want to get too far into the weeds there without a practical context to work from, which is hopefully where this PR aims us.

> The 'immediate reversal' criterion seems to err on the side of caution. The underlying mechanism for an outlier that might look like a regression always is a methodological instability that hits in with a certain probability. That probability is assumed to be _small_. That is, it is rather unlikely for this instability to hit in, say, "twice in a row". However, we know that this still can happen. It looks like this detection method does not try to account for that, which is (I think) a good choice.

💯. Yeah, there's definitely opportunity to go further here in managing high erraticity, but it's a lot of corner cases that require a lot of work for minimal benefit. This is definitely an "80% solution" in that sense—it's very possible to do better, but this aims to be a simple, useful start.

> I would love to see some discussion about if we can build valuable outlier detection _without_ automatic distribution change detection _or_ if the outlier detection is strictly coupled to an automatic distribution change detection.

As we talked about this morning, automatic step detection is implicit in this method, but we don't need to integrate it now. I don't think it's necessarily implicit in methods we may want to explore in the future, though.

> My feeling is that we want to keep the distribution shift detection manual for now; that's really why I think that part of the discussion is important.

To document what we discussed, we absolutely want to keep manual distribution shift detection regardless of whether/when we augment it with automated detection; there will always be cases users know are shifts that automated methods will miss (e.g. a too-minor-to-detect shift in one case that's big and obvious in a related one). For now I'll remove the integration of the automated step detection here, but it's here when we're ready.

conbench/entities/history.py
alistaire47: Switched these to rolling using the same `Config.DISTRIBUTION_COMMITS`. In practice I don't think what that value is will matter much here, but rolling generally will make results more reproducible when called with different slices of data. And it doesn't peek at the future.

conbench/entities/history.py
alistaire47: Yeah, didn't love this. Trying `_detect_shifts_with_trimmed_estimators()`—this method really turns more on the trimming than the z-scores

jgehrcke:
Thanks for the great replies, @alistaire47!

conbench/entities/history.py
jonkeane: > e.g. if a result was an outlier and it got retried and reverted, sorting by result time will allow us to handle that

Very true if the last result is the best | the one that reverts. Though if someone sends multiple back without checking between each iteration, the last being best isn't guaranteed. But IMO we should keep this here and figure out how to deal with multiple results of the same case separately

conbench/entities/history.py
```diff
@@ -481,6 +487,13 @@ def _add_rolling_stats_columns_to_df(
         for x in df["change_annotations"]
     ]
 
+    # NOTE(EV): If unquoted, this line will integrate manually-specified distribution
```
jonkeane:
```suggestion
    # NOTE(EV): If uncommented, this line will integrate manually-specified distribution
```

Did you mean this?

conbench/entities/history.py
```diff
@@ -570,3 +586,77 @@ def _calculate_z_score(
         z_score = z_score * -1
 
     return z_score
+
+
+def _detect_shifts_with_trimmed_estimators(
+    df: pd.DataFrame, z_score_threshold=5.0
```
jonkeane, line 594:
```suggestion
    df: pd.DataFrame, z_score_threshold=10.0
```

I know 5 is used a bunch of places, but IMO we should increase this default (cf https://github.com/voltrondata-labs/arrow-benchmarks-ci/pull/100). Is this used anywhere? If it isn't (and we're using whatever the config value is) let's not dig the 5 hole deeper and make this be the default we want (bonus points for going through and undoing that default elsewhere, can be a separate PR)

jonkeane:
> The basic method here still uses z-scores, but trims off the top and bottom 5% of data when calculating mean and standard deviation to calculate them, so outliers and step changes do not affect magnitudes. 

For posterity, could you go into a little bit more detail about the trimming + why it's necessary here?

alistaire47:
> > > ignores the top and bottom 5% of data when calculating mean and standard deviation to calculate them, so outliers and step changes do not affect magnitudes. 10% is not totally random
> > 
> > 
> > The "10 %" is surprising after the "5 %" in the sentence before that. Should both be "5 %" or both be "10 %"?
> 
> It's slicing off 5% of data from the top and bottom end for 10% total to get [trimmed estimators](https://en.wikipedia.org/wiki/Trimmed_estimator) akin to the `trim` param of base R's `mean()`: https://rdrr.io/r/base/mean.html for 10% total. Arguably it shouldn't be symmetrical, and we should be slicing off the most extreme 10%, wherever that is, maybe defined by Mahalanobis distance. But that's complicated and the benefits are minimal, so for a first pass I tried to keep it simple.

Thinking more on this, maybe we should _only_ slice off of the top—outliers _faster_ than the mean in our case would be weird. Our data is not normally distributed.

alistaire47:
Added outliers to plots as open circles for means. Mins for outliers are ignored so they don't get zig-zaggy. Put `outlier: True` in the hover only for outliers (i.e. there's no corresponding `outlier: False` line for inliers). 

image

conbench/app/_plots.py
```diff
@@ -27,11 +27,11 @@ class BokehPlotJSONOrError:
 
 
 class TimeSeriesPlotMixin:
-    def get_outliers(self, benchmarks):
+    def get_biggest_changes(self, benchmarks):
```
alistaire47, line 30: Renamed to avoid confusion from different things with the same name. Happy to rename to something else if y'all prefer

conbench/tests/entities/test_history.py
```diff
@@ -150,3 +158,39 @@ def test_append_z_scores_with_distribution_change():
     set_z_scores(benchmark_results)
     for benchmark_result, expected_z_score in zip(benchmark_results, expected_z_scores):
         assert benchmark_result.z_score == expected_z_score
+
+
+def test_detect_shifts_with_trimmed_estimators():
```
jgehrcke, line 163: :heart:

alistaire47:
Updated plots: hovers back to how they were; added outliers to legend instead. Click for details fixed.

image

alistaire47:
> > The basic method here still uses z-scores, but trims off the top and bottom 5% of data when calculating mean and standard deviation to calculate them, so outliers and step changes do not affect magnitudes.
> 
> For posterity, could you go into a little bit more detail about the trimming + why it's necessary here?

I got into this in my answers to JP, but I added a paragraph to the description so it's clearer up-front.

conbench/entities/history.py
alistaire47: I don't actually agree here; as @austin3dickey said in https://github.com/conbench/conbench/pull/956#discussion_r1149668922 this is not quite the same as the config value:

> oof, yeah, it's pretty different from the existing "z-score threshold" concept in this codebase, because it operates on a (clipped) mean-diff distribution instead of a lookback mean distribution. Maybe this parameter could be something like "mean_diff_z_score_threshold" or something less bad than that? ;)
>
> I think it's fine to not introduce the server-wide configuration environment variable until someone asks for it later.

Because of the trimming, this doesn't have the volatility problems we have elsewhere with 5%. Raising it to 10% will result in ignoring a lot of obvious outliers. If we're going to use the plots to keep an eye on how this is performing, this is exactly the lever we can pull to adjust if it's over- or under-identifying outliers. Making it more conservative now means we won't see the line of what it considers an outlier as much.

conbench/entities/history.py
austin3dickey: Nice!

conbench/app/_plots.py
```diff
@@ -471,6 +471,8 @@ def time_series_plot(
 
     unit = samples[0].unit
     with_dist = [s for s in samples if s.zscorestats.rolling_mean]
+    inliers = [s for s in samples if not s.zscorestats.is_outlier]
```
austin3dickey, line 474: :)

austin3dickey (approved):
I'm ready to ship this and see what happens!

conbench/api/_examples.py
```diff
@@ -374,6 +374,7 @@ def _api_history_entity(benchmark_id, case_id, context_id, run_name):
                 "rolling_mean_excluding_this_commit": 0.036369,
                 "rolling_stddev": 0.0,
                 "segment_id": 0.0,
+                "is_outlier": False,
```
jgehrcke, line 377: :+1:

conbench/app/_plots.py
jgehrcke: cool, let's iterate on names over time.

conbench/app/_plots.py
```diff
@@ -480,22 +482,24 @@ def time_series_plot(
 
     # Note(JP): `samples` is an ordered list of dicts, each dict has a `mean`
```
jgehrcke, line 483: note to self: outdated comment :)

conbench/app/_plots.py
```diff
@@ -471,6 +471,8 @@ def time_series_plot(
 
     unit = samples[0].unit
     with_dist = [s for s in samples if s.zscorestats.rolling_mean]
+    inliers = [s for s in samples if not s.zscorestats.is_outlier]
+    outliers = [s for s in samples if s.zscorestats.is_outlier]
```
jgehrcke, line 475: love the readability of these list comprehensions; that also shows the value of recent refactorings in https://github.com/conbench/conbench/pull/811.

conbench/entities/history.py
```diff
@@ -168,6 +170,7 @@ def get_history_for_cchr(
             BenchmarkResult.data,
             BenchmarkResult.times,
             BenchmarkResult.change_annotations,
+            BenchmarkResult.timestamp.label("result_timestamp"),
```
jgehrcke, line 173: what does `label("result_timestamp")` do?

conbench/entities/history.py
```diff
@@ -481,6 +489,13 @@ def _add_rolling_stats_columns_to_df(
         for x in df["change_annotations"]
     ]
 
+    # NOTE(EV): If uncommented, this line will integrate manually-specified distribution
+    # changes with those automatically detected. Before enabling this, we want a way for
+    # users to manually remove an automatically-detected step-change.
```
jgehrcke, line 494: :+1:

conbench/entities/history.py
```diff
@@ -570,3 +588,77 @@ def _calculate_z_score(
...
+
+    # pandas likes the data to be sorted
+    tmp_df.sort_values(
+        [
+            "case_id",
+            "context_id",
+            "hash",
+            "repository",
+            "timestamp",
+            "result_timestamp",
```
jgehrcke, line 622: I see, here "result_timestamp" pops up again. Maybe we should add a comment what "timestamp" is :P

conbench/entities/history.py
```diff
@@ -570,3 +588,77 @@ def _calculate_z_score(
...
+        group_df["is_shift"] = group_df.z_score.abs() > z_score_threshold
+        group_df["reverts"] = group_df.is_shift & group_df.is_shift.shift(-1)
+        out_group_df["is_step"] = (
+            group_df.is_shift
+            & ~group_df.reverts
+            & ~group_df.reverts.shift(1, fill_value=False)
+        )
+        out_group_df["is_outlier"] = group_df.is_shift & group_df.reverts
+
+        out_group_df_list.append(out_group_df)
```
jgehrcke, line 659: Is there anything notable about this business logic that we should write down as a code comment, now that you have it at the top of your head?

conbench/tests/populate_local_conbench.py
```diff
@@ -464,23 +464,24 @@ def generate_synthetic_benchmark_history(commit_hashes: List[str], repo_url: str
...
+    distr_std = 2.0
     # `lower_bound` is to simulate the real-world effect of there being a
     # theoretical optimum: the fastest benchmark duration time with all sources
     # of noise being silenced.
     lower_bound = 17.5
-    slowdown_offset = distr_mean * 0.1
-    slowdown_lin = distr_mean * 0.1
-    distribution = statistics.NormalDist(mu=20.0, sigma=2)
+    slowdown_offset = 2.0 * distr_std
+    outlier_offset = 20.0 * distr_std
```
jgehrcke, line 473: nice

conbench/tests/populate_local_conbench.py
```diff
@@ -522,13 +523,18 @@ def sample_slowdown(s):
...
-            samples = [sample_slowdown(s) for s in samples]
+            samples = [sample_slowdown(s, offset=slowdown_offset) for s in samples]
+
+        # Simulate outliers with more significant slowdowns
+        if random.random() < 0.08:
+            # Shift all samples a little higher (not by the same amount)
+            samples = [sample_slowdown(s, offset=outlier_offset) for s in samples]
 
         # Simulate for any sample to have been affected by some blip-slowdown
         if random.random() < 0.02:
```
jgehrcke, line 534: note to self: the samples in here are the sub samples elsewhere 
This relates to https://github.com/conbench/conbench/pull/811#discussion_r1127723139

jgehrcke:
Let's get this into `main`.

@jonkeane I am not 100 % sure  if you got to communicating all the relevant thoughts you have on this; can you maybe give this another final round of review comments even after merge, summarizing your thinking?

conbench/entities/history.py
alistaire47: Pretty sure it gets translated to `... AS result_timestamp`, which we need here because the commit timestamp is already called `timestamp`.

conbench/entities/history.py
alistaire47: Would it be clearer to alias `timestamp` to `commit_timestamp`? I'd need to make sure it gets changed everywhere, but I think the scope is limited to `_add_rolling_stats_columns_to_df()`

conbench/entities/history.py
alistaire47: Maybe the limitations and how to tune? This won't handle some corner cases like length-2 outliers, and the z-score threshold (which is not quite the same as the previous one, because the data it operates on is filtered) can be adjusted to catch more or less.

I'll go chuck some stuff in the docstring, since the threshold is parameterized there anyway.

PR #958, opened 2023-03-28
webapp: build psycopg2 from source against modern libpq (fix #957)

### LINKED ISSUE #957: make tests: upon db connect: SCRAM authentication requires libpq version 10 or above
@alistaire47 saw

```
/usr/local/lib/python3.11/site-packages/psycopg2/__init__.py:122: in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
E   sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) SCRAM authentication requires libpq version 10 or above
```
during
```
conbench/db.py:128: in create_all
    Base.metadata.create_all(engine)
```
As of the nature of the error (permanent), the retry loop is of course not resolving this.

What's a little puzzling about this error is that it suggests that we use different versions of libpq across people/machines; but as of our containerized workflow we try to prevent exactly this. The error is probably a red herring.

Btw, we pip-install https://pypi.org/project/psycopg2-binary/ in the app's container image, and that bundles a pre-built version of libpq.

jgehrcke:
This should reliably use psycopg2 with libpq 13 across platforms, because the Python-3.11-slim Docker image that we use as build environment here is based on Debian Bullseye which uses libpq 13.

Previously, on M1 (AArch64) the psycopg2 binary we used was statically linked against libpq 9 (built in a manylinux 2.24 environment). See #957 for details.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ FROM python:3.11-slim
 # curl is needed for docker-compose health checks. `git` is needed by some unit
 # tests as of today.
 RUN apt-get update && apt-get install -y -q --no-install-recommends \
-    curl git build-essential && apt-get clean && rm -rf /var/lib/apt/lists/*
+    curl git build-essential libpq-dev && apt-get clean && rm -rf /var/lib/apt/lists/*
 
 COPY requirements-webapp.txt /tmp/
 COPY requirements-dev.txt /tmp/
@@ -72,3 +72,4 @@ COPY ./buildinfo.json /buildinfo.json
 # Re-active this to get ideas for how the image size can be further reduced.
 #RUN echo "biggest dirs"
 #RUN cd / && du -ha . | sort -r -h | head -n 50 || true
+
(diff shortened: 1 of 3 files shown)


jgehrcke:
This is a pretty interesting change to bring to our deployments. Merging now.

jgehrcke:
from our build log:
```
2023-03-28T07:50:38.9275820Z #10 10.39   Building wheel for psycopg2 (setup.py): started
2023-03-28T07:50:47.0465293Z #10 18.53   Building wheel for psycopg2 (setup.py): finished with status 'done'
2023-03-28T07:50:47.0466224Z #10 18.53   Created wheel for psycopg2: filename=psycopg2-2.9.5-cp311-cp311-linux_x86_64.whl size=490493 sha256=3ef5b15ca0d849cf3dc00c3e4481ef04367c52d0ebe5034e512e048e4527bc66
```
we see that about 10 seconds were spent in the build; that's tolerable (I was expecting a bigger delay)

Issue #961, opened 2023-03-28
Rename BenchmarkCreate schema to BenchmarkResultCreate


jgehrcke:
Analogous to #878 there are places in code and also in the API specification that should see clarification in terminology. We identified this task before; let's dig up and link to corresponding discussion.

Also; today we have 
![image](https://user-images.githubusercontent.com/265630/228238470-3b8c37a5-2c68-414b-9f03-87f9e65827f4.png)
whereas the first one really should be called BenchmarkResultCreate, and the latter defines the structure of the stats property, should be called BenchmarkResultStats. (it's not obvious if we want those to be two different schemas, but that's ok)

jgehrcke:
Did that here: https://github.com/conbench/conbench/pull/962.

PR #962, opened 2023-03-28
Rename schemas, change meaning of BenchmarkResultCreate

jgehrcke:
For #961.


conbench/api/benchmarks.py
```diff
@@ -6,7 +6,7 @@
 from ..api._endpoint import ApiEndpoint, maybe_login_required
 from ..entities._entity import NotFound
 from ..entities.benchmark_result import (
-    BenchmarkFacadeSchema,
+    BenchmarkResultFacadeSchema,
```
jgehrcke: Note that personally I am not sure of the information content in `Facade`. I kept it here for now, although maybe it is not useful in the future. Does this try to indicate the JSON surface?

conbench/entities/benchmark_result.py
```diff
@@ -300,7 +300,7 @@ def update(self, data):
 s.Index("benchmark_result_timestamp_index", BenchmarkResult.timestamp)
 
 
-class BenchmarkResultCreate(marshmallow.Schema):
+class BenchmarkResultStatsSchema(marshmallow.Schema):
```
jgehrcke, line 303: This diff line is the actual and rather significant motivator for this patch. The current state is a little too mind-boggling; and this here also is a necessary change towards implementing a clear distinction between "benchmark" and "benchmark result".

conbench/entities/benchmark_result.py
```diff
@@ -358,57 +358,53 @@ class BenchmarkResultCreate(marshmallow.Schema):
...
         required=False,
         metadata={
-            "description": "The inter-quartile range from `data`, will be calculdated on the server if not present (the preferred method), but can be overridden if sent. Will be marked `null` if any iterations are missing."
+            "description": "The inter-quartile range from `data`, will be calculated on the server if not present (the preferred method), but can be overridden if sent. Will be marked `null` if any iterations are missing."
         },
     )
 
 
-class BenchmarkResultSchema:
-    create = BenchmarkResultCreate()
```
jgehrcke, line 409: What is/was the value of this wrapping technique?

conbench/entities/benchmark_result.py
```diff
@@ -573,7 +569,7 @@ class _BenchmarkFacadeSchemaCreate(marshmallow.Schema):
             )
         },
     )
-    stats = marshmallow.fields.Nested(BenchmarkResultSchema().create, required=False)
+    stats = marshmallow.fields.Nested(BenchmarkResultStatsSchema(), required=False)
```
jgehrcke, line 572: Do we lose anything by taking the indirection out?

conbench/entities/benchmark_result.py
```diff
@@ -686,6 +682,6 @@ class _BenchmarkFacadeSchemaUpdate(marshmallow.Schema):
     )
 
 
-class BenchmarkFacadeSchema:
-    create = _BenchmarkFacadeSchemaCreate()
-    update = _BenchmarkFacadeSchemaUpdate()
+class BenchmarkResultFacadeSchema:
+    create = _BenchmarkResultCreateSchema()
+    update = _BenchmarkResultUpdateSchema()
```
jgehrcke, line 687: Here I again tried to stay as consistent as possible with previous naming, but as I said above, the "facade" term is not one that I am too convinced about yet.

conbench/tests/api/_expected_docs.py
```diff
@@ -978,7 +978,7 @@
             },
         },
         "schemas": {
-            "BenchmarkCreate": {
+            "BenchmarkResultCreate": {
```
jgehrcke, line 981: This is important outfall.

conbench/tests/api/_expected_docs.py
```diff
@@ -1022,7 +1022,7 @@
                         "description": "Reason for the Run (optional, does not need to be unique). Ignored when Run was previously created.",
                         "type": "string",
                     },
-                    "stats": {"$ref": "#/components/schemas/BenchmarkResultCreate"},
+                    "stats": {"$ref": "#/components/schemas/BenchmarkResultStats"},
```
jgehrcke, line 1025: This is a user-facing (well, external dev-facing) clarification and un-confusion.

jonkeane (approved):
I'm fine with all of this — I don't think anyone is relying on these names specifically (and correct me if I'm wrong, but doesn't actually change any keys or names in the API itself — just what we refer to them here that then gets bubbled up into the docs, yeah?)

😅 that swagger not letting one deep-link to these also means there are no broken URLs after this (not that that would be a huge deal in this case, but if it doesn't work, it can't break :P)

jgehrcke:
> just what we refer to them here that then gets bubbled up into the docs, yeah?

Yes! Thanks for the quick review and opinion.

conbench/entities/benchmark_result.py
austin3dickey: I suspect it was reserving a spot for a future `update` schema. Not very important

Issue #964, opened 2023-03-28
UI: repurpose /benchmarks view for benchmark-centric ("conceptual benchmark") overview


jgehrcke:
Context: #787, #772

This view is not linked from the main/landing page:
![image](https://user-images.githubusercontent.com/265630/228260089-9e008bea-4c2f-4d63-864d-9f978f2ea23f.png)

Do we want to remove it? I think we want to use the `/benchmarks` path for showing the conceptual benchmarks, as discussed in #787.

jonkeane:
Yeah, I'm totally fine repurposing it. Would it be possible | easy to copy this view over into. `/benchmarkResults` or `/benchmark-results`? Not totally sure if we need it inevitably or if we have anyone using it — but having an option to switch to that would be nice if it comes up (and slightly easier than resurrecting it out of git history).

And then: are we planning to also do the same in the API? IMO we probably should, though we will need to coordinate with a few folks | places in order to do so, since they currently rely in those endpoints (benchalerts is one place and also https://github.com/conbench/conbenchcoms). Both should be pretty easy + are in decently active development right now

jgehrcke:
> Not totally sure if we need it inevitably or if we have anyone using it — but having an option to switch to that would be nice if it comes up

Let's keep it for now, did that quickly in #966.

> `/benchmark-results`?

This! :)

> And then: are we planning to also do the same in the API?

Wondered the same. Probably yes. Thanks for supporting the idea. I'd love to do that independently; I think there will be a time soonish when it feels like the right thing to do.

jgehrcke:
Update: the 'conceptual benchmark' list/entrypoint is right now under a temporary URL, and will (I think, at leats) after all be served under benchmarks.

austin3dickey:
That's good - I was confused what the "c-" meant until just now, looking at the title of this issue, when I realized it stands for "conceptual" :)

PR #966, opened 2023-03-28
UI: introduce /benchmark-results/<brid>, keep old route, add test, more renames in code

jgehrcke:
This is for #964.

Updated description:

This PR
- introduces a new URL path pattern for accessing an individual benchmark result view: `/benchmark-results/<benchmark-result-id>`
- changes the UI/app/template logic to generate URLs of the new shape
- keeps support for the old pattern `/benchmarks/<benchmark-result-id>`
- adds a test for covering the old pattern
- also contains valuable renames in code, all relating to the difference between "benchmark result" and "benchmark".

Note: we do not yet want to break those URLs to specific benchmark results that we have in the past shared in e.g. Slack. See comment below.

### DIFF
--- a/conbench/app/benchmarks.py
+++ b/conbench/app/benchmarks.py
@@ -343,11 +343,29 @@ class BenchmarkResultList(AppEndpoint, ContextMixin):
 
 rule(
     "/benchmark-results/",
-    view_func=BenchmarkResultList.as_view("benchmarks"),
+    view_func=BenchmarkResultList.as_view("benchmark-results"),
     methods=["GET"],
 )
+
 rule(
     "/benchmark-results/<benchmark_id>/",
+    view_func=BenchmarkResult.as_view("benchmark-result"),
+    methods=["GET", "POST"],
+)
+
+
+# Legacy route, which people have used to communicate a URL e.g. via Slack or
+# email, or in downstream reporting, to point to a specific benchmark result
+# view. Keep this working _in addition to_ the new, more descriptive
+# `/benchmark-results/<benchmark_id>/` path. Keep this working for as long as
+# we can do so with ease. Next step for this outstretched transition might be
+# to build custom logic that would act on the shape of <benchmark_id> and emit
+# a redirect response. Note that this needs a different name argument passed to
+# the as_view() method, otherwise one sees an error like `View function mapping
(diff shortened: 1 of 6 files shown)


jgehrcke:
> Are we OK with that?

Had a good idea in sync meeting; can keep those URLs working (in addition to the new ones), for a transition period (which can be long). We can use the fact that the benchmark result ID is a stringified uuid type 4 (i.e., has known length). And potentially we never really want to re-use `/benchmarks/<something>`. And if we do, we know what `<something>` is.

conbench/app/benchmarks.py
```diff
@@ -337,12 +342,30 @@ def _get_benchmarks(self):
...
-    view_func=BenchmarkList.as_view("benchmarks"),
+    "/benchmark-results/",
+    view_func=BenchmarkResultList.as_view("benchmark-results"),
     methods=["GET"],
 )
+
+rule(
+    "/benchmark-results/<benchmark_id>/",
+    view_func=BenchmarkResult.as_view("benchmark-result"),
+    methods=["GET", "POST"],
```
alistaire47, line 353: Ah, one thing that didn't occur to me this morning is that we're posting all results to `/api/benchmarks/`; we should switch those to `/api/benchmark-results/` once we're ready. There are a number of places because our client-side tooling has a lot of variation, but the most obvious place is in benchadapt: https://github.com/conbench/conbench/blob/ab8c64fbd7a88d83d1c84ee74bf0e5bb0d33db5d/benchadapt/python/benchadapt/adapters/_adapter.py#L187 And we should integration test that, which I think we've got a story for somewhere, but we've never gotten around to actually setting up.

jgehrcke: :+1: noting for posterity that this PR and discussion is about the UI routes, i.e. not touching on any URL starting with `/api`.

Jon also asked about `/api/...` in https://github.com/conbench/conbench/issues/964#issuecomment-1486976661.

Yes, let's also plan to rename the `/api/...` routes with clearer terminology (in which case, I agree, doing the transition rather carefully is even more important).

austin3dickey (approved):

jgehrcke:
Thanks for feedback!

Issue #973, opened 2023-03-29
rename modules/files "benchmark" -> "result"


jgehrcke:
A hyphen in a Python module name (file name) is not advisable. I think the `Result` will be a rather popular entity in the future, and `result.py` is better than `benchmarkresult.py` or `bmresult.py`.

PR #974, opened 2023-03-29
Rename Python modules for clarity (benchmark -> result)

jgehrcke:
For #973.

There is more to rename, especially _in_ the files; but it's important to take this step by step.


jgehrcke:
Landing this :ship:

PR #980, opened 2023-03-30
ci: minikube: test with k8s 1.23

jgehrcke:
See commit message.


jgehrcke:
CI went green. Cool. Want to lock this in, for now at least.

PR #981, opened 2023-03-30
kube-prometheus conbench flavor: allow for monitoring other namespace

jgehrcke:
Also, apply `kube-prom-no-req-no-lim.jsonnet` always for now, because even our prod k8s clusters are rather small with 2 CPU cores per node, and two nodes.

This addresses a privilege issue:
```
kubectl logs pods/prometheus-k8s-0 -n monitoring | less
```

```
ts=2023-03-30T10:33:20.091Z caller=klog.go:108 level=warn component=k8s_client_runtime func=Warningf msg="pkg/mod/k8s.io/client-go@v0.26.0/tools/cache/reflector.go:169: failed to list *v1.Endpoints: endpoints is forbidden: User \"system:serviceaccount:monitoring:prometheus-k8s\" cannot list resource \"endpoints\" in API group \"\" in the namespace \"staging\""
ts=2023-03-30T10:33:20.091Z caller=klog.go:116 level=error component=k8s_client_runtime func=ErrorDepth msg="pkg/mod/k8s.io/client-go@v0.26.0/tools/cache/reflector.go:169: Failed to watch *v1.Endpoints: failed to list *v1.Endpoints: endpoints is forbidden: User \"system:serviceaccount:monitoring:prometheus-k8s\" cannot list resource \"endpoints\" in API group \"\" in the namespace \"staging\""
```

key fragment: `User \"system:serviceaccount:monitoring:prometheus-k8s\" cannot list resource ...`

> This is intentional. kube-prometheus configures Prometheus with the principle of least privilege, meaning permission is granted specifically for each namespace when needed instead of for every namespace, even where it's not needed.

(https://github.com/prometheus-operator/kube-prometheus/issues/492#issuecomment-613321437)

Interesting: https://github.com/prometheus-operator/kube-prometheus/issues/492#issuecomment-1450074518

I think in the meantime this has been nicely documented here: https://github.com/prometheus-operator/kube-prometheus/blob/main/docs/monitoring-other-namespaces.md

> By default the RBAC rules are only enabled for the Default and kube-system namespaces.
> You have to give the list of the namespaces that you want to be able to monitor. This is done in the variable prometheus.roleSpecificNamespaces. You usually set this in your .jsonnet file when building the manifests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -302,7 +302,6 @@ jsonnet-kube-prom-manifests:
 			echo "MUTATE_JSONNET_FILE_FOR_MINIKUBE not set"; \
 		else \
 			echo "MUTATE_JSONNET_FILE_FOR_MINIKUBE set, mutate JSONNET"; \
-			sed -i.bak 's|// (import "kube-prom|(import "kube-prom|g' _kpbuild/cb-kube-prometheus/conbench-flavor.jsonnet; \
 			sed -i.bak 's|// "auth.anonymous"|"auth.anonymous"|g' _kpbuild/cb-kube-prometheus/conbench-flavor.jsonnet; \
 			cat _kpbuild/cb-kube-prometheus/conbench-flavor.jsonnet; \
 		fi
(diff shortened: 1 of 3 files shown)


jgehrcke:
Good. Landing this.

PR #983, opened 2023-03-30
k8s deploy.yaml: adjust Service and add ServiceMonitor

jgehrcke:
This defines a ServiceMonitor object; it refers to the corresponding Service by name; therefore a name is introduced here. For a bit of cleanup, we also introduce a name for the container port.

The ServiceMonitor object is a noop, unless kube-prometheus is also running in the same k8s cluster. If it is, then the ServiceMonitor object defines a monitoring target for the kube-prometheus stack.

This is tested and has been running in CI on minikube for a while. The current change brings this to our deployment pipeline, into the k8s cluster(s) that our Conbench deployment(s) run(s) on.

I have tested that manually in vd-2.

### DIFF
--- a/deploy.yml
+++ b/deploy.yml
@@ -60,15 +60,21 @@ spec:
 apiVersion: v1
 kind: Service
 metadata:
+  name: "conbench-service"
+  labels:
+    app: conbench
   annotations:
     # "ip mode is required for sticky sessions to work with Application Load
     # Balancers"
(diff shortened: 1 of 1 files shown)


jgehrcke:
Short-cutting CI (I cancelled the run); there is no coverage of this change. After merge to main, this is exercised as part of conbench-staging-deploy.

PR #984, opened 2023-03-30
k8s rollingUpdate refinement, resource footprint tweak, misc

jgehrcke:
- `replicas: 2` and `maxSurge: 25%, maxUnavailable: 25%` didn't quite fit
- refine strategy, add comments for how/why
- decrease resource footprints for migration job but also for the conbench deployment itself; this is because VD-2 has little resources; and during the rolling update things spiked towards or above max available resources.

### DIFF
--- a/deploy.yml
+++ b/deploy.yml
@@ -7,11 +7,26 @@ spec:
     matchLabels:
       app: conbench
   replicas: 2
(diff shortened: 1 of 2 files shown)


jgehrcke:
As before: short-cutting CI (I cancelled) -- this takes effect during conbech-staging-deploy on BK.

PR #996, opened 2023-03-31
conbench grafana dashboard: various improvements, new panel

jgehrcke:
Some quick improvements now that this is fed with real-world data from vd-2.
image

### DIFF
--- a/k8s/kube-prometheus/conbench-grafana-dashboard.json
+++ b/k8s/kube-prometheus/conbench-grafana-dashboard.json
@@ -972,7 +972,7 @@
     ]
   },
   "time": {
-    "from": "now-6h",
+    "from": "now-24h",
     "to": "now"
   },
   "timepicker": {},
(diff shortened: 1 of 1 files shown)


k8s/kube-prometheus/conbench-grafana-dashboard.json
```diff
@@ -872,14 +873,102 @@
...
+      },
+      "pluginVersion": "9.3.2",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
+          "editorMode": "code",
+          "expr": "sum by (le) (rate(flask_http_request_duration_seconds_bucket{namespace=\"$namespace\",method=\"GET\",path=~\"/api/.*\",path!~\"/api/ping.*\",status!~\"3..\"}[$timewindow]))",
```
jgehrcke, line 963: spent some time tuning this query

jgehrcke, line 963: spent some time tuning this query

jgehrcke: https://www.robustperception.io/negative-lookahead-assertions-in-promql-selectors/

> So this use case could be handled as metric{foo=~"b..",foo!~"bar"}

gave me the confidence that conditions can be chained

PR #1000, opened 2023-04-01
address performance issue: remove history analysis from `/api/benchmark?run_id=<>` query

### LINKED ISSUE #978: inspect DB query backing  `/api/benchmarks/?run_id=<id1,id2,...>`, profile where this spends time
Look at where HTTP response generation spends its time, and pay attention to the DB query method. This is motivated by a report about mediocre latency in this HTTP route, often yielding 504 Gateway Timeout responses.

jgehrcke:
This addresses a part of #1001.

This fixes #978.

This patch changes the HTTP response generation duration for `/api/benchmark` when used with the `run_id` query parameter by one or two orders of magnitude.

The specific query I showed in https://github.com/conbench/conbench/issues/978 now takes ~1 second in my development environment, instead of ~20 seconds (current HEAD of `main`).

The user-facing change is that we do not perform history analysis anymore (the lookback z-score method is not applied). In the team we seem to have the shared opinion that this should only be performed when demanded / expected by the user.

So far we had underspecified behavior in that regard anyway; the API docs made no promise about this type of analysis being performed or not.

### DIFF
--- a/conbench/api/results.py
+++ b/conbench/api/results.py
@@ -40,9 +40,7 @@ class BenchmarkEntityAPI(ApiEndpoint, BenchmarkValidationMixin):
     def get(self, benchmark_id):
         """
         ---
-        description:
-            Get a benchmark result. This includes on-the-fly processing
-            with the lookback z-score change detection method.
+        description: Get a benchmark result.
         responses:
             "200": "BenchmarkEntity"
             "401": "401"
@@ -117,7 +115,6 @@ class BenchmarkListAPI(ApiEndpoint, BenchmarkValidationMixin):
         """
         ---
         description: Get a list of benchmarks.
-
         responses:
             "200": "BenchmarkList"
             "401": "401"
--- a/conbench/api/results.py
+++ b/conbench/api/results.py
@@ -1,10 +1,13 @@
+import logging
+
 import flask as f
 import flask_login
-
 from sqlalchemy import select
+
 from ..api import rule
 from ..api._docs import spec
 from ..api._endpoint import ApiEndpoint, maybe_login_required
+from ..db import Session
 from ..entities._entity import NotFound
 from ..entities.benchmark_result import (
     BenchmarkResult,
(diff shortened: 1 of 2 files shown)


conbench/api/results.py
```diff
@@ -128,34 +134,44 @@ def get(self):
...
+            # TODO: This needs a limit, and sorting behavior.
+            # arbitrary limit for now.
             benchmark_results = BenchmarkResult.search(
                 filters=[Case.name == name_arg],
                 joins=[Case],
             )
-            # Since there's no limit on the number of BenchmarkResults, we could take
-            # forever calculating z-scores with no caching advantage. So don't do that.
-            for benchmark_result in benchmark_results:
-                benchmark_result.z_score = None
```
jgehrcke, line 139: Instead of this, I found it a little cleaner to (for now) handle AttributeError in the to-dict function.

conbench/entities/benchmark_result.py
```diff
@@ -407,61 +483,7 @@ class BenchmarkResultStatsSchema(marshmallow.Schema):
...
-                    "api.context",
-                    context_id=benchmark_result.context_id,
-                    _external=True,
-                ),
-                "run": f.url_for(
-                    "api.run", run_id=benchmark_result.run_id, _external=True
-                ),
-            },
-        }
+        return benchmark_result.to_dict_for_json_api()
```
jgehrcke, line 486: Slowly introducing more of these `to_dict_for_json_api()` methods on our main entities. This will help defuse the mind-boggling EntitySerializer class hierarchy.

jgehrcke:
Landing this, next patch upcoming. Review appreciated!

conbench/api/results.py
```diff
@@ -128,34 +134,44 @@ def get(self):
         tags:
           - Benchmarks
         """
+        # Note(JP): "case name" is the conceptual benchmark name. Interesting,
+        # so this is like asking "give me results for this benchmark".
         if name_arg := f.request.args.get("name"):
+            # TODO: This needs a limit, and sorting behavior.
```
austin3dickey, line 140: yes!

Issue #993, opened 2023-03-31
k8s: automatically deploy/update kube-prometheus as part of deployment pipeline


jgehrcke:
Elsewhere I have described and tested that the initial deployment as well as the update of our Conbench-flavored kube-prometheus stack is by now a matter of three steps:

### 1) JSONNET-build YAML manifests 

Easy:
```
make jsonnet-kube-prom-manifests
```

### 2) deploy to k8s cluster

Easy:
```
export CONBENCH_REPO_ROOT_DIR="."
pushd "${CONBENCH_REPO_ROOT_DIR}"/_kpbuild/cb-kube-prometheus/
    kubectl apply --server-side -f manifests/setup
    kubectl wait \
        --for condition=Established \
        --all CustomResourceDefinition \
        --namespace=monitoring
    kubectl apply -f manifests/
popd
```

### 3) Learn how to count to three

My plan has always been to integrate this into the deployment pipeline, so that for example Grafana dashboard definition changed immediately take effect after merge to `main`.

jgehrcke:
Typical file system permission complications, see https://github.com/conbench/conbench/pull/1003 and #1004.

Related:
https://github.com/buildkite/docs/pull/426/files
https://github.com/buildkite/agent/issues/407

jgehrcke:
Made important progress in #1013, see https://github.com/conbench/conbench/pull/1013#issuecomment-1493355904.

jgehrcke:
:rocket: great news, for the first time this executed the deploy-or-update script w/o erroring out:

```
++ set +x
vd-2: run k8s/kube-prometheus/[REDACTED]-or-update.sh, and pray that this does not impede the robustness of our [REDACTED] pipeline
+ kubectl create namespace staging --dry-run=client -o yaml
+ kubectl apply -f -
E0403 11:17:37.550372   24931 memcache.go:255] couldn't get resource list for external.metrics.k8s.io/v1beta1: Got empty response for: external.metrics.k8s.io/v1beta1
namespace/staging configured
+ pushd /var/lib/buildkite-agent/builds/buildkite-agent-us-east-2-vd-amd64-cpu-i-07675654cefc1d725-1/voltron-data/conbench-staging-[REDACTED]/_kpbuild/cb-kube-prometheus/
[...]

PR #1003, opened 2023-04-01
bk utils: try running make jsonnet-kube-prom-manifests

jgehrcke:
This is towards #993.


jgehrcke:
Merging this to get the desired feedback.

Issue #977, opened 2023-03-29
UI landing page load time regression: fetch all BenchmarkResults for the last 1000 runs when rendering landing page?


jgehrcke:
An investigation happened as a follow-up from this conversation: https://github.com/conbench/conbench/pull/912#discussion_r1142402132

I got curious about the potential outfall of #912 in view of real-world data like the Arrow Conbench deployment. Working with hat locally (also see #976) I found a rather unambiguous performance regression for the landing page; see comment below.

jgehrcke:
Initial impression with real-world database: rather huge impact. Got landing page load times of `0.24 s, 0.08 s` when using `lazy="select"` (not counting the results-per-run), and got e.g. `82.1 s` with `lazy="select"/"immediate"` (counting results-per-run).

austin3dickey:
It may be worth surgically reverting this change temporarily so we can deploy without worry.

jgehrcke:
> It may be worth surgically reverting this change temporarily so we can deploy without worry.

Yes. I want to have this resolved by Monday EOD, either with a revert or with a better solution.

I disappeared in a rabbit hole and asked some :magic_wand: magician :tophat: rabbits :rabbit: what they think about not only how to _efficiently_ query a count in a one-to-many relationship, but also how to do that in SQLAlchemy, using the new-ish primitives.

This snippet translates into a concise query and returns rather fast (below 1 second), returning both: the runs as well as the per-run result counts (but not the results, as intended):
```python
        runs_with_rescount = Session.execute(
            select(Run, func.count(BenchmarkResult.id))
            .outerjoin(BenchmarkResult)
            .group_by(Run)
            .limit(2000)
        ).all()
```

But then I hit another wall, trying to add `ORDER BY run.timestamp DESC`. I mean, I managed to add it to the query, but it resulted in the query taking infinitely long to execute. Maybe this is not even needed. To be continued! I think this might be a well-defined riddle, waiting to be solved :).

Notably, this part of SQLAlchemy talks about incorrectly-placed ORDER BY in context of a query that uses JOIN: https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html#the-zen-of-joined-eager-loading

> got e.g. 82.1 s with lazy="select"/"immediate" (counting results-per-run).

[...]

Issue #772, opened 2023-02-24
ui/ux: a benchmark list view?


jonkeane:
Currently all of our list views are pretty run centric: You can see a list of runs, and then for a run a list of benchmark(Result)s. But the vast majority of benchmark(Result)s happen over and over (e.g. every run, or at least in at least one run per commit or per hardware). 

Could we make a view that shows a list of benchmarks (possibly with filters for "only show me benchmarks on X hardware" or "only show me benchmarks that were benchmarked for Y commit")? And then allow diving in more deeply there.

One good source of prior art here is ASV: 

Their summary page is an interesting set of plots of history per each benchmark: https://www.astropy.org/astropy-benchmarks/#/
There's also a list view: https://www.astropy.org/astropy-benchmarks/#summarylist?sort=0&dir=asc

I really like the almost sparkline nature of the first view: there are not units or other things to distract (you can click through and get to that, of course). We might even be able to combine these together, have the list view with proper sparklines and a bit more info per-benchmark for that visual look and feel of scanning through and looking for bumps along with some of the details we see in the list view.

And this might be more in the spirit of #330 but their regression view is also benchmark-centric: https://www.astropy.org/astropy-benchmarks/#regressions?sort=3&dir=desc

From a value perspective: a view like this would be really fantastic for people who want a more high-level view (and possibly even a kind of daily-check in to scroll through patterns and see if there's anything funky).

jgehrcke:
It's a really important one, I agree!

A benchmark-centric view is also what I had in mind when I thought about https://github.com/conbench/conbench/issues/560. There I forgot to mention the "list all benchmarks" idea explicitly, but I also imagined a clean overview over all known benchmarks (a listing), which can then be searched/filtered (by e.g. benchmark name, and all the other useful criteria you enumerate here).

This UX concept also helps clarifying the difference between "benchmark" and "benchmark result", if done well.

jonkeane:
_nods_ yeah totally, there's some overlap there — it's possible I should have extended that issue more (when I searched the titles I assumed (wrongly) it was only about searching too. I'm not too worried about ticket structure just yet, I'm sure that these broader idea tickets will get broken down into more concrete tickets as we start working on them

jgehrcke:
After making big leaps forward (in particular https://github.com/conbench/conbench/pull/1086 and https://github.com/conbench/conbench/pull/1085), I am closing this here as 'done'. Let's create more precise follow-up tasks.
[...]

PR #1014, opened 2023-04-02
landing page: fix perf regression #977 (do not display result count for now)

jgehrcke:
Do not display result count for now, but an empty string -- pragmatic relaxation of the situation. My feeling is that we will display a result count again in the next couple of days.

With the Arrow Conbench DB state this again yields
```
± time curl -v http://127.0.0.1:5000/ >/dev/null
real	0m0.142s
```

Without the patch this was > `70 s` in my dev env.

### DIFF
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -42,13 +42,18 @@ class BenchmarkResult(Base, EntityMixin):
     # (0, 1, many). Result is child. Run is parent. Child has column with
     # foreign key, pointing to parent.
     run_id: Mapped[str] = NotNull(s.Text, s.ForeignKey("run.id"))
-    # Note that `immediate` means that " items should be loaded as the parents
-    # are loaded, using a separate SELECT statement, or identity map fetch for
-    # simple many-to-one references." Do this until we see that this is a perf
-    # problem. Use case: count results per run (that use case can be solved by
-    # adding a cache/counter into the DB, also see
-    # https://stackoverflow.com/q/45407402/145400)
-    run: Mapped[Run] = relationship("Run", lazy="immediate", back_populates="results")
+
+    # Between Run (one) and Result (many) there is a one-to-many relationship
+    # that we want to tell SQLAlchemy af ew things about via the following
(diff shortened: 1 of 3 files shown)


conbench/entities/benchmark_result.py
```diff
@@ -42,13 +42,18 @@ class BenchmarkResult(Base, EntityMixin):
...
-    run: Mapped[Run] = relationship("Run", lazy="immediate", back_populates="results")
+
+    # Between Run (one) and Result (many) there is a one-to-many relationship
+    # that we want to tell SQLAlchemy af ew things about via the following
+    # `relationship()` call. Note that `lazy="select"` is a sane default here.
+    # It means that another SELECT statement is issued upon attribute access
+    # time. `select=immediate` would mean that items are to be loaded from the
+    # DBas the parent is loaded, using a separate SELECT statement _per child_
+    # (which did bite us, in particular for those Run entities associated with
+    # O(1000) Result entities, i.e. hundreds of thousands of SELECT queries
```
jgehrcke, line 53: or maybe "just" 1000 additional SELECT queries; in any case the query overhead alone didn't look good

jgehrcke:
Landing this; review of course welcome/desired.

jonkeane:
Thanks for taking the pragmatic approach here. I really like having the result count on this page, but I do wonder if maybe we're striving for a feature here that (while nice!) doesn't drive that much leverage. All things being equal, I would love to have this (back). But there are other things that are more important for us now (#772 for one).

If we can re-enable this with minimal work, let's do it and move on to the features we know folks are requesting. If not, let's leave it and mark it to come back to when we have time.

jgehrcke:
> I do wonder if maybe we're striving for a feature here that (while nice!) doesn't drive that much leverage

Right, if showing a number that nobody asked for is so complicated then it's time to take note of the learnings, but to otherwise abort the mission. :rocket: :chart_with_downwards_trend: :)

> My feeling is that we will display a result count again in the next couple of days.

I want to explicitly acknowledge that re-enabling this should be a side effect of other work or the result of some smart idea that popped up from almost nowhere :bulb: (but not the result of a significant time investment to solve this particular problem).

PR #1015, opened 2023-04-02
ci: refactor: add k8s/kube-prometheus/deploy-or-update.sh

jgehrcke:
The logic in the new file `k8s/kube-prometheus/deploy-or-update.sh` has been used in CI against minikube. This patch is a small change towards making it reusable against our prod k8s clusters. It also adds better commentary. Also, start testing this twice against minikube to test for idempotency -- or at least to validate, that this does not error out when run more than once against the same k8s cluster (which is what we will do in the deploy pipeline).


jgehrcke:
Going ahead with this :ship:

Issue #1012, opened 2023-04-02
public API: add denylist for bots


jgehrcke:
Bots trigger compute-intense queries, at a significant rate. 15 minutes worth of logs:
```
$ csvtool col 4 extract-2023-04-02T10_57_46.522Z.csv | grep -i 'bot' | wc -l
128
```

We can filter based on user agent. That's also what bots/SEP providers _want_ services to do, so that is a sane choice.
```
$ csvtool col 4 extract-2023-04-02T10_57_46.522Z.csv | grep -oi '\w*botw*' | sort | uniq
bot
DataForSeoBot
dotbot
DotBot
petalbot
PetalBot
```

This affects e.g. the Arrow Conbench because it deliberately has its API freely accessible in the Internet (read-only).

Adding a denylist has small/mid impact, but is also low effort.

jgehrcke:
Resolved by https://github.com/conbench/conbench/pull/1017.

Issue #447, opened 2022-11-22
Should individual Conbench instances encourage search engines to scrape them?


jgehrcke:
Follow-up from https://github.com/conbench/conbench/pull/443#issuecomment-1323510414.

> I think the question here may be if individual Conbench instances should be indexed by Google / search engines at all.

I'd say the answer is 'no', but I am happy to learn about other perspectives.

If the answer is 'no', let's work with https://developers.google.com/search/docs/crawling-indexing/block-indexing.

jonkeane:
Agreed that we shouldn't have the individual instances crawled (unless we can come up with some good reason otherwise). 

That said, I think #443 was fine to test the theory we had about those strange links popping up in search and if that alone would resolve them (indicating we needed to do _something else_ on top of `noindex`)

jgehrcke:
> indicating we needed to do something else on top of noindex

My impression was that so far we did _not_ use the `noindex` technique. Is that correct?

jonkeane:
> My impression was that so far we did not use the noindex technique. Is that correct?

Correct, I think we only used the `robots.txt` approach. I thought we had used `noindex` before but I was wrong on that!

ElenaHenderson:
Just want to add my reasons and experience with bots and conbench:
1. I disabled crawling about a year ago because it was making Conbench DB suffer
2. I have done TONS of performance refactoring and DB tuning since then so performance might be fine but we will need to watch out still.
3. I tried to be less strict in robots.txt but not all bots respected "less strict". They do respect `Disallow /`, which is 100% no crawling of any kind.
4. I have tried this and it did not help with all robots:
```
expected_text = """
"User-Agent: *
Disallow: /api/history/
Disallow: /batches/
Disallow: /benchmarks/
Disallow: /compare/
Disallow: /runs/
"""
```
5. I did not try noindex but I wonder if we need to index so conbench.ursa.dev is at the top of Google seearch. I see this right now because robots are disabled:

<img width="617" alt="Screen Shot 2022-11-22 at 9 15 09 AM" src="https://user-images.githubusercontent.com/5522472/203379030-131a6e73-4d11-4f8e-ba0b-cf2a87fd47b6.png">

jgehrcke:
> I did not try noindex 

Thanks for confirming, Elena. Let's try `noindex` to hopefully remove these Conbench instances from the current Google search index, and to prevent new instances from popping up. `noindex` is meant to be used for this purspose. About `noindex`, from the docs linked above:

[...]

PR #1017, opened 2023-04-02
discourage search engines to scrape, reject known bots based on user agent

jgehrcke:
This is a patch that should for now reconcile
- https://github.com/conbench/conbench/issues/1012
- https://github.com/conbench/conbench/issues/447
- https://github.com/conbench/conbench/pull/443#issuecomment-1323510414

### DIFF
--- a/conbench/__init__.py
+++ b/conbench/__init__.py
@@ -42,7 +42,7 @@ def create_application(config):
     app = f.Flask(__name__)
     app.config.from_object(config)
 
-    _init_application(app)
+    _init_flask_application(app)
 
     # Re-configure logging using user-given configuration details.
     log.debug("re-configure logging")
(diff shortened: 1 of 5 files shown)


conbench/__init__.py
```diff
@@ -80,12 +80,13 @@ def create_application(config):
...
 
-def _init_application(application):
+def _init_flask_application(app):
+    import flask
     import flask_swagger_ui
     import werkzeug.exceptions
 
     from .api import api
-    from .app import app
+    from .app import app as blueprint_app
```
jgehrcke, line 89: this was an important rename for clarity. also see commit messages.

conbench/templates/app.html
```diff
@@ -5,6 +5,7 @@
 <head>
 <meta charset="utf-8">
 <meta name="viewport" content="width=device-width, initial-scale=1">
+<meta name="robots" content="noindex">
```
jgehrcke, line 8: once again, this is a powerful technique documented here: https://developers.google.com/search/docs/crawling-indexing/block-indexing

I think this will resolve #443.

jgehrcke:
Should be uncontroversial by now, but please let the feedback and reviews come.

PR #1020, opened 2023-04-03
k8s: deploy.yml: split into chunks, improve comments

jgehrcke:
This is for #1019 and for https://github.com/conbench/conbench/issues/995.

I worked quite a lot on code commentary to reflect knowledge/thoughts that are hard to see through when newly coming to the codebase, i.e. this is good for maintainability, being able to navigate the code base quickly, and making changes with more confidence.

Issue #1019, opened 2023-04-03
k8s: deploy.yml: split into independent pieces, refine commentary


jgehrcke:
Requirement for https://github.com/conbench/conbench/issues/726. The current deploy.yml mixes infrastructure (ALB) with deployment details; making it difficult to re-use the file.

While chunking this up, refine code commentary, especially in utils.sh -- quite a bit of non-obvious behavior.

jgehrcke:
Closing this as 'done' :balloon:. I think this will help quite a bit in the near future.

Issue #726, opened 2023-02-10
consolidate k8s specs between CI and prod deployment


jgehrcke:
Quote from https://github.com/conbench/conbench/pull/662#issue-1568306456:

> This PR introduces [minikube](https://minikube.sigs.k8s.io/docs/) to CI as a foundation to test k8s-specific parts of Conbench. This **allows for testing Conbench-related k8s YAML manifests in pull request CI** (valuable because we embrace k8s as a/the recommended way to run Conbench, and we run it on k8s in prod ourselves).

This is an important idea. Towards getting there we need to do deduplication/consolidation work, step by step.

Goal: `ci/minikube` should use much of the same code and config that we also use for deploying Conbench to production k8s environments. And the other way round.

Examples for where we have duplication with divergence as of now:

#### Conbench config 
https://github.com/conbench/conbench/blob/d7cf049acf4912de40343222f8b931b45a6a87e3/conbench-config.yml
https://github.com/conbench/conbench/blob/d7cf049acf4912de40343222f8b931b45a6a87e3/ci/minikube/conbench-config-for-minikube.yml

#### Conbench Secrets
https://github.com/conbench/conbench/blob/d7cf049acf4912de40343222f8b931b45a6a87e3/conbench-secret.yml
https://github.com/conbench/conbench/blob/d7cf049acf4912de40343222f8b931b45a6a87e3/ci/minikube/test-conbench-on-mk.sh#L89

#### Conbench deployment with readiness checks, replica config, & k8s service 
https://github.com/conbench/conbench/blob/d7cf049acf4912de40343222f8b931b45a6a87e3/ci/minikube/deploy-conbench.template.yml
https://github.com/conbench/conbench/blob/d7cf049acf4912de40343222f8b931b45a6a87e3/deploy.yml

Example for where we have one thing only in one of both: kube-prometheus

-----------

I currently build up the minikube flow step by step and collect expertise while doing so, and my plan is to then slowly make the prod deployment use the bits and pieces from there.

jgehrcke:
Still not done, but got real far.

These patches came out of the work of moving things into vd-2 in a reproducible fashion:
https://github.com/conbench/conbench/pull/980
https://github.com/conbench/conbench/pull/981
https://github.com/conbench/conbench/pull/982
https://github.com/conbench/conbench/pull/983
https://github.com/conbench/conbench/pull/984
https://github.com/conbench/conbench/pull/990
https://github.com/conbench/conbench/pull/991


CC @jonkeane

[...]

PR #1021, opened 2023-04-03
k8s: migration-job.yml: move, add commentary

jgehrcke:
For #726. Analogous to #1019. Also fixing a sed replacement bug introduced with #1020.

### DIFF
--- a/.buildkite/utils.sh
+++ b/.buildkite/utils.sh
@@ -85,7 +85,7 @@ deploy() {
 
   # (Re-)apply deployment. BUILDKITE_COMMIT is the Conbench repo commit.
   cat k8s/conbench-deployment.templ.yml | \
-    sed "s/{{CONBENCH_CONTAINER_IMAGE_SPEC}}/${CONBENCH_CONTAINER_IMAGE_SPEC}/g" | kubectl apply -f -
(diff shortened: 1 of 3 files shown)


jgehrcke:
Merging, to test this.

.buildkite/utils.sh
```diff
@@ -50,26 +50,36 @@ deploy_secrets_and_config() {
...
+  sed "s|{{CONBENCH_WEBAPP_IMAGE_SPEC}}|${IMAGE_SPEC}|g" \
+    < k8s/conbench-database-migration.templ.yml \
+    > _jobspec
+
+  # Delete job first -- why is that important?
+  kubectl delete --ignore-not-found=true -f _jobspec
+  kubectl apply -f _jobspec
+
+  # Note(JP): we give this 24 hours of time. Why? For those heavy migration
+  # jobs that really take so long? Interesting.
```
jgehrcke, line 68: I forgot to add to the comment here that maybe this is not wise. Maybe we want this job to error out early, and maybe we should to one-off migrations that are known / expected to take forever manually.

austin3dickey: I do like the accountability & linking of the logs if we apply all migrations through this job, but I agree it's difficult to handle different sized loads. I wonder if there's any way that the person kicking off a deployment could configure a longer timeout than a default, shorter timeout.

Issue #1005, opened 2023-04-02
System metrics for Arrow Conbench: peek into data, document findings, review cardinality


jgehrcke:
A multi-purpose experiment I am doing is to locally run prometheus agent to scrape Arrow Conbench's metrics endpoint and to remote_write the data into my personal GC's managed Prometheus instance (think of it as local smart buffer that tries to do a best-effort forwarding job ([which should not be needed](https://community.grafana.com/t/can-grafana-cloud-scrape-a-prometheus-endpoint-over-the-internet/83608/2))).

There is a lot to learn from here. Start documenting findings, in this ticket.

Here is how I run the agent locally:
```
$ cat agent-conf-proper.yaml 
server:
  log_level: debug

metrics:
  global:
    scrape_interval: 2m
    scrape_timeout: 60s
  configs:
    - name: arrow-conbench
      scrape_configs:
        - job_name: arrow-conbench
          scheme: https
          static_configs:
            - targets: ['<redacted>']
      remote_write:
        - url: https://prometheus-prod-<redacted>.grafana.net/api/prom/push
          basic_auth:
              <redacted>
```

```
$ docker run -it -v $(pwd)/agent:/etc/agent/data -v  $(pwd)/agent-conf-proper.yaml:/etc/agent/agent.yaml  grafana/agent:v0.32.1
...
```

jgehrcke:
The metrics endpoint sometimes is not responsive; from the agent log (rather often):
```
msg="Scrape failed" err="server returned HTTP status 502 Bad Gateway"
```
That is a rough indicator for the pathological state the system is in, and also speaks in favor of potentially running it in its own HTTP server.

jgehrcke:
Cardinality of just the conbench-emitted metrics is already rather significant. Query `topk(10, count by (job)({job=~".+"}))`  results in a bimodal plot that oscillates between ~5000 and ~10000 for `{job="arrow-conbench"}`.

Already got a suspicion here; the many different unique `path` values blow up the dimensions. Confirmed, the following screenshot just shows a small number of examples, but it's clear that there may be thousands of individual path values:
![Screenshot from 2023-04-02 10-32-21](https://user-images.githubusercontent.com/265630/229342730-cdf169b4-b5c3-4636-9e78-a0333955df74.png)

Assuming that many routes behave the same regardless of the ID provided we can dramatically reduce that. Now tracked here: https://github.com/conbench/conbench/issues/1006

jgehrcke:
[...]

PR #1023, opened 2023-04-03
k8s: inject node&pod name to app via env, add envpodname label to metrics

### LINKED ISSUE #1008: k8s & metrics: inject machine name into containers, expose as metrics label
When scraping a metrics endpoint from the outside behind a load balancer or k8s service object then it is typically served by a different environment on a per-request basis. This results in incoherences when not keeping track of each unique environment. For example, counters appear to not be monotonic:

image

A rate calculated from a non-monotonic counter becomes a bit mad:
image

Solution: feed machine/host info deep _into_ the application, and then expose it via an additional label. In a k8s environment, this might be useful: https://stackoverflow.com/q/56285724/145400

Note: this is _not_ a problem in our kube-prometheus stack which does container-aware scraping from _inside_ the k8s environment.

jgehrcke:
This should resolve #1008.

This also attempts to fix the unrelated  `cat: k8s/conbench-cloud-ingress.templ.yml: No such file or directory`.


conbench/metrics.py
```diff
@@ -98,7 +106,10 @@ def decorate_flask_app_with_metrics(app) -> None:
         # defines the upper inclusive bound for the corresponding histogram
         # bucket. Note that there is an implicit last/upper end bucket here
         # catching all observations up to +inf.
-        buckets=(0.05, 0.1, 0.2, 0.5, 1.0, 3.0, 6.0, 10.0, 15.0, 20.0, 30.0, 50.0),
+        # Update(JP): removed two buckets compared to initial stab, to work on
+        # cardinatlity.
+        buckets=(0.05, 0.1, 0.2, 0.5, 1.0, 5.0, 10.0, 20.0, 40.0, 70.0),
```
jgehrcke, line 111: Related: https://github.com/conbench/conbench/issues/1005

conbench/metrics.py
```diff
@@ -82,6 +82,14 @@ def decorate_flask_app_with_metrics(app) -> None:
 
     This mutates `app` in-place.
     """
+    # If the environment tells us something unique about us when we are
+    # one replica of N replicas, then inject the uniqueness too all emitted
+    # metrics -- see https://github.com/conbench/conbench/issues/1008
+    default_labels = {}
```
jgehrcke, line 88: Not: keep this balanced, as this blows up cardinatlity again.

kube-prometheus is doing this magic already; I was thinking that _one_ additional label is fine though for scrapers that are independent of kube-prometheus (not a first-class use case, but useful to keep it correct/working).

jgehrcke:
Merging this :balloon:. I do hope this was also the last repair patch for the `conbench-staging-deploy` pipeline, but we will see.

jgehrcke:
>  I do hope this was also the last repair patch for the conbench-staging-deploy pipeline, but we will see.

Yesss! `conbench-staging-deploy/builds/359` went green.

PR #1024, opened 2023-04-03
metrics: emit endpoint, not path

### LINKED ISSUE #1006: metrics: emit less individual URL path values
Follow-up from #1005.

Overall, significantly reduce the number of unique path values in the metrics endpoint.

Some points to think about:

- The API-layer indirection is likely to be problematic where an HTTP request to /benchmarks/<id> creates a virtual/internal request to /api/benchmarks/<id> -- another reason to remove that indirection everywhere
- Robot-based API submission routes _providing an ID_: do not track those. 
- Robot-consumed API read routes _providing an ID_: do not track those
- Human-based UI routes that are ID-specific: maybe track those, maybe not

How to do that? Deviate from "default metrics" (https://github.com/rycus86/prometheus_flask_exporter#default-metrics).

jgehrcke:
This resolves #1006 for now. It's a bit of a drastic move, but I think a really important one. We might lose quite a bit of granularity/insight, but in this case I think it's better to work from coarse-grained to fine-grained, gradually, as we learn.

Note: we always have the access log showing specific URL paths + request duration times.

### DIFF
--- a/conbench/metrics.py
+++ b/conbench/metrics.py
@@ -96,6 +96,11 @@ def decorate_flask_app_with_metrics(app) -> None:
     _inspect_prom_multiproc_dir()
     GunicornInternalPrometheusMetrics(
         app=app,
+        # See https://github.com/conbench/conbench/issues/1006
+        # We will have to maybe iterate on those endpoint names,
+        # and maybe, just maybe, add _some_ URL paths back when we understand
+        # that we need them.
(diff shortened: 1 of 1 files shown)


jgehrcke:
CI is green, which in this case means a lot. Landing this :ship:. 

Feedback of course super welcome.

Issue #1001, opened 2023-04-01
API spec & behavior: do not perform costly history analysis when not explicitly requested


jgehrcke:
A major outcome of https://github.com/conbench/conbench/issues/978:

> we might put the complex calculation behind a switch/query parameter?

> we should remove z-scores from the /api/benchmark-result endpoints, and only calculate them when one hits the /api/compare endpoints.

We might still want to think about where we allow for getting to this kind of data, but I want to stress that we should striv ve for documenting this with text as part of the API spec.

austin3dickey:
I believe this would greatly simplify #1029. The `set_z_score()` functionality is complex today, because it finds the "best" baseline commit for every given result, so that it can power the z-scores returned from the `/api/benchmarks` endpoint. That endpoint might be returning a multitude of results that have different baseline commits.

But if we remove z-scores from that endpoint, then whenever we do the lookback z-score analysis (which would then be only within the `/api/compare` namespace) we have a well-defined baseline commit. That would simplify the code a lot. Not to mention the performance speedup of `/api/benchmarks`.

austin3dickey:
Okay, after #1041, all z-score-related keys returned from different usages of `/api/benchmarks` have `null` values and therefore are deprecated. I'll leave this ticket open to actually remove those keys from the response (and deal with any downstream clients that might be relying on them).

austin3dickey:
Let me copy some investigation from https://github.com/conbench/conbench/pull/1041#issuecomment-1497419165 into here:

Since we currently get webapp data via the API, there might be webapp code that starts to break. For example:
https://github.com/conbench/conbench/blob/173b970291d6b33531f45a7f0bb103ff1824442f/conbench/app/_plots.py#L36
That code would throw a KeyError and break plots. I suspect there might be more; we can look at it in #1001. I don't think we use this key in any of the clients in this monorepo but there could be other clients.

PR #1027, opened 2023-04-03
api/results: less lookback z-score on-the-fly comp, more docs about that

jgehrcke:
Another patch for #1001. This time with a bit of API spec/docs.


jgehrcke:
:ship: 

Would appreciate feedback about this one in particular @jonkeane @austin3dickey @alistaire47!

Issue #1034, opened 2023-04-04
kube-prometheus: look into using persistent volume(s) for pusher prom statefulset


jgehrcke:
In https://github.com/conbench/conbench/issues/993#issuecomment-1494157591 I found that Prometheus state was lost (scraped data). I assumed that the `kube-prometheus` stack would automatically create small/default persistent volumes as long as it finds some storage resources to use. But yeah, it can't really pick decent defaults.

Likewise, to confirm (for vd-2):
```
$ kubectl get pvc -n monitoring
No resources found in monitoring namespace.
```

There are great docs about this:

> By default (if the following 'storage.volumeClaimTemplate' isn't created), prometheus will be created with an EmptyDir for the 'prometheus-k8s-db' volume (for the prom tsdb).

(from https://github.com/prometheus-operator/kube-prometheus/blob/7a3879ba49bfe8df5eec03847fe5bcd2d4094c73/examples/prometheus-pvc.jsonnet#L24)

```
$ kubectl get storageclass
NAME            PROVISIONER             RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
gp2 (default)   kubernetes.io/aws-ebs   Delete          WaitForFirstConsumer   false                  176d
```

Also interesting:

```
                // The following 'selector' is only needed if you're using manual storage provisioning (https://github.com/coreos/prometheus-operator/blob/master/Documentation/user-guides/storage.md#manual-storage-provisioning).
                // And note that this is not supported/allowed by AWS - uncommenting the following 'selector' line (when deploying kube-prometheus to a K8s cluster in AWS) will cause the pvc to be stuck in the Pending status and have the following error:
                //  * 'Failed to provision volume with StorageClass "ssd": claim.Spec.Selector is not supported for dynamic provisioning on AWS'
                // selector: { matchLabels: {} },
```

- https://github.com/prometheus-operator/kube-prometheus/blob/7a3879ba49bfe8df5eec03847fe5bcd2d4094c73/examples/prometheus-pvc.jsonnet
- https://github.com/prometheus-operator/prometheus-operator/issues/2666#issuecomment-510465282
- https://github.com/prometheus-operator/kube-prometheus/issues/883#issuecomment-764628321

jgehrcke:
[...]

PR #1037, opened 2023-04-04
kube-prometheus: use pvcs

jgehrcke:
So little code. That does so much. This is for #1034.

Not sure what I find more fascinating, the little `sed` expression in here or that this introduces a `fundisk-elb-backed` storage class in vd-2 from which the Prometheus StatefulSet pods can request and obtain persistent volumes. This is (has to be) ephemeral, but works beautifully:
```
± kubectl get pv
NAME                CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM                                           STORAGECLASS         REASON   AGE
local-pv-a9e4d628   19Gi       RWO            Retain           Bound    monitoring/prometheus-k8s-db-prometheus-k8s-1   fundisk-elb-backed            17m
local-pv-bfe2fe30   9915Mi     RWO            Retain           Bound    monitoring/prometheus-k8s-db-prometheus-k8s-0   fundisk-elb-backed            17m
```

A buffer on steroids, that can truly do some best-effort data-forwarding.

About the sed expression; this is not a substitute command, but a range match or [address range](https://www.gnu.org/software/sed/manual/html_node/Range-Addresses.html).

See
- https://catonmat.net/sed-one-liners-explained-part-two, recipe 67 for intro
- https://catonmat.net/sed-one-liners-explained-part-three, recipe `5. Selective deletion of certain lines`
- combined with the idea that a different regex delimiter can be used (when [escaped](https://backreference.org/2010/02/20/using-different-delimiters-in-sed/index.html#comment-25209) when _not_ using the substitute command)

### DIFF
--- a/k8s/kube-prometheus/conbench-flavor.jsonnet
+++ b/k8s/kube-prometheus/conbench-flavor.jsonnet
@@ -61,6 +61,36 @@ local kp =
     prometheus+: {
       prometheus+: {
         spec+: {
+          // https://github.com/prometheus-operator/kube-prometheus/blob/117ce2f8b51621ccb480e8bdd712ae21e92b744e/examples/prometheus-pvc.jsonnet
+          // vd-2 in mind. Almost ephemeral setup. But we want to retain
+          // data across kube-prometheus updates. Use short retention and tiny
+          // persistent volumes.
+          retention: '6d',
+          // <comment-used-by-ci: pvc-start>
+          storage: {
+            volumeClaimTemplate: {
+              apiVersion: 'v1',
+              kind: 'PersistentVolumeClaim',
+              spec: {
+                accessModes: ['ReadWriteOnce'],
+                // 6 days vs 9Gi? I have no idea, but sounds roughly reasonable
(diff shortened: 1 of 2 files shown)


k8s/kube-prometheus/conbench-flavor.jsonnet
```diff
@@ -61,6 +61,36 @@ local kp =
...
+                // mad. EBS-backed "local storage" to on vd-2 nodes however
+                // works. Manually exposed ext4-formatted partitions to k8s
+                // with a rather beautiful setup using the Local Persistent
+                // Volume CSI Driver, offering a storage class which is for
+                // now called 'fundisk-elb-backed' because fun und such.
+                storageClassName: 'fundisk-elb-backed',
+              },
+            },
+          },
+          // <comment-used-by-ci: pvc-end>
```
jgehrcke, line 93: The point here is that this isn't used on minikube. Yes. templating on top of templating, I know. But I think with code comments it's a good trade-off.

PR #1040, opened 2023-04-04
metrics: remove prefix, better http handler name, also fix unbound variable error

jgehrcke:
This is another follow-up for https://github.com/conbench/conbench/issues/1006, see https://github.com/conbench/conbench/issues/1006#issuecomment-1495859683.

In local dev, an unbound variable error sneaked into the code base in `make conbench-on-minikube`, which is addressed in this PR, too.


jgehrcke:
```
ty[e fox
```
love this :fox_face:

jgehrcke:
OK, this is an important patch that will reset data collection. Landing this :ship:

Issue #1029, opened 2023-04-03
api/compare: lookback z-scores should respect the baseline entity when choosing the end of the distribution


austin3dickey:
Today when returning items from the `/api/compare/*/baseline_id...contender_id` endpoints, we return `contender_z_score` and `contender_z_regression` keys (among others). Those values are generated based on the lookback z-score analysis. Right now, when choosing the distribution of results to compare each contender result to, the lookback z-score analysis is hardcoded to include results belonging to default-branch commits up to and including the most recent default-branch commit in the contender commit's history. (That is, the fork-point commit if the contender is on a PR commit, or the parent commit if the contender is on the default branch.) This effectively ignores the given `baseline_id`.

However, that's not the spirit of the lookback z-score analysis. That analysis should take the baseline into account. When choosing the distribution of results to compare each contender result to, we should include results belonging to *any* commit in the baseline commit's history, up to and including the baseline commit.

If the given baseline just so happens to be (on) the "closest-ancestor default-branch commit" of the contender, then this shouldn't affect the returned results.

Issue #388, opened 2022-10-04
Let me filter by runs (or benchmarks?) with a regression


jonkeane:
When searching runs, it would be nice to be able to filter to only runs that have a regression (possibly with a specific z-score threshold...)

pedroerp:
+1. It would be very helpful to be able to search and browse through the benchmarks with the largest regressions (or variation) for  a particular branch (mostly for main).

jgehrcke:
Very important proposals in here. I agree.

>  largest regressions (or variation)

No critique, just a note to self: it's important to carefully distinguish variability/instability from regression detection.

I think that filtering by variability/variation/instability (need to settle on terms... :-)) is a great helper for finding methodically weak benchmark implementations. More generally, instability detection is an important tool towards consolidating the experimental setup. And this, in turn, is a requirement for _robust_ performance regression detection.

Issue #384, opened 2022-09-29
Allow me to compare different baselines


jonkeane:
We currently have two comparison modes for conbench:

* A default-branch commit to a default-branch commit immediately[^1] before it
* A PR-branch commit to the commit that the branch was started with (sometimes called the branch-point commit or the merge base commit)

But there are a few other comparisons we might want to make:
* A default branch commit to the first preceding commit on default that was benchmarked (this is the same as in footnote 1, and would allow conbench to work with merge-commit workflows instead of only squash-commit workflows)
* A PR-branch commit to the head of the default branch
* A PR-branch commit to a previous commit on that PR branch
* A PR merge commit (where the head of the PR branch is merged with the current head of dev) with the current head of dev

Some of those comparisons have caveats and important flags we should raise in the UI, but for each of them we should be able to make that comparison + provide a (relevant) history|distribution to use when calculating Z-Scores.

[^1]: though in some circumstances this isn't actually the immediate commit before it, but the commit that is closest to it, before it, that has a benchmark

austin3dickey:
A proposal for a nice API to power this feature:

- GET `/api/compare/*/contender_id?target_baseline_commit=parent`
- GET `/api/compare/*/contender_id?target_baseline_commit=head_of_default`
- GET `/api/compare/*/contender_id?target_baseline_commit=fork_from_default`

with accompanying thoughts:

1. GET `/api/compare/*/baseline_id...contender_id` still works
2. You can only use one of these three strings, not an arbitrary ID or something
3. `fork_from_default` chooses the parent commit if the contender is already on the default branch. This one is more like "most recent ancestor commit on the default branch" but I'm trying to find a snappier phrase.
4. If there are no matching benchmark results on the target baseline commit, we'd search backwards in ancestry until finding one, and report which actual baseline commit we used (see #1030)
5. If we're using `/api/compare/benchmarks/contender_benchmark_result_id?target_baseline_commit=*`, and there were multiple matching results on the resulting baseline commit, the lookback z-score analysis is fine but the pairwise analysis needs to choose a baseline result. Maybe we'd pick based on matching run reason and tiebreak by time (like we do when trying to find a baseline run)

[...]

PR #1041, opened 2023-04-04
api/benchmarks: deprecate z-scores

austin3dickey:
This PR is one step to address #1001 by removing the functionality of the `z_score` key returned in `/api/benchmarks` payloads. While it doesn't go as far as removing the keys, it will unblock my work on #1029.

### DIFF
--- a/conbench/api/results.py
+++ b/conbench/api/results.py
@@ -16,7 +16,6 @@ from ..entities.benchmark_result import (
     BenchmarkResultSerializer,
 )
 from ..entities.case import Case
-from ..entities.history import set_z_scores
 
 log = logging.getLogger(__name__)
 
@@ -44,9 +43,7 @@ class BenchmarkEntityAPI(ApiEndpoint, BenchmarkValidationMixin):
         description: |
             Get a specific benchmark result.
 
-            This includes on-the-fly analysis with the lookback z-score
-            change detection method, and the corresponding result is emitted
-            as part of the benchmark result object.
+            The "z_score" key in the response is deprecated and only returns null.
         responses:
             "200": "BenchmarkEntity"
             "401": "401"
@@ -60,7 +57,6 @@ class BenchmarkEntityAPI(ApiEndpoint, BenchmarkValidationMixin):
           - Benchmarks
         """
         benchmark_result = self._get(benchmark_id)
-        set_z_scores([benchmark_result])
         return self.serializer.one.dump(benchmark_result)
 
     @flask_login.login_required
@@ -87,7 +83,6 @@ class BenchmarkEntityAPI(ApiEndpoint, BenchmarkValidationMixin):
         benchmark_result = self._get(benchmark_id)
         data = self.validate_benchmark(self.schema.update)
         benchmark_result.update(data)
-        set_z_scores([benchmark_result])
         return self.serializer.one.dump(benchmark_result)
 
     @flask_login.login_required
@@ -289,8 +284,6 @@ class BenchmarkListAPI(ApiEndpoint, BenchmarkValidationMixin):
         # Note(JP): in the future this will also raise further validation
         # errors that are to be exposed via a 400 response to the HTTP client
         benchmark_result = BenchmarkResult.create(data)
-
-        set_z_scores([benchmark_result])
 
         return self.response_201_created(self.serializer.one.dump(benchmark_result))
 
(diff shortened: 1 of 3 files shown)


austin3dickey:
Merging to unblock; review welcome!

conbench/api/results.py
```diff
@@ -290,8 +285,6 @@ def post(self) -> f.Response:
         # errors that are to be exposed via a 400 response to the HTTP client
         benchmark_result = BenchmarkResult.create(data)
 
-        set_z_scores([benchmark_result])
```
jgehrcke, line 293: Makes a lot of sense to me that this is gone! (same for all PUT/POST code paths here in this PR).

jgehrcke:
>  While it doesn't go as far as removing the keys

:) genuinely asking: are we aware of client tooling that likes the `null` value more than the missing property?

I would suspect no client tool reads the property, happy to learn otherwise!

jgehrcke (commented):
Good patch. Provocative. Raises interesting questions. 

In my tiny worldview the GET `/benchmark/<bmr-id>` endpoint was (the) one way to obtain the per-bmr z-score value as an API consumer, assuming that we need one way; do I didn't quite feel like going so far to remove all-the-ways. But maybe that's good!

Is "give me the z-score value for this benchmark result" even well-defined? Not so sure.

Does this have a valid use case, from an API consumer's point of view? Why would an API consumer want to fetch this (genuine use case brainstorming)?

I think this will in the future be "give me meaningful analysis / in historical context for this benchmark result". 

Another provocative point of view would be that analyses are always implementation details leading up to reports, notifications, alerts, etc. And maybe those can also be fetched with API.

jonkeane:
I think this is fine for the individual benchmark page (though we should keep an ear out if people were using these).

> Is "give me the z-score value for this benchmark result" even well-defined? Not so sure.
> 
> Does this have a valid use case, from an API consumer's point of view? Why would an API consumer want to fetch this (genuine use case brainstorming)?
> 
> I think this will in the future be "give me meaningful analysis / in historical context for this benchmark result".
> 
> Another provocative point of view would be that analyses are always implementation details leading up to reports, notifications, alerts, etc. And maybe those can also be fetched with API.

Great questions | suggestions. I _do_ think that there is a usecase for something like this in a summary | report view like you're mentioning here. 

For our data in conbench, most benchmarks have one comparison that is the most helpful to conduct[^1], and so having an API where that happens automatically, that can be useful (this might even be something like what benchalerts consumes by default, but also other downstream API consumers). We don't need to support this now and I think it's a _very good_ intuition to separate the detailed benchmark endpoint from this "gimme a summary about this benchmark" 

[^1]: There might be a small handful of other comparisons that are relevant here — we should explore with our users which are helpful if we should add more, etc. But we've taken the strong approach of having _one analysis_ we offer by default. Having more different possible comparisons we should treat as a separate problem here.

austin3dickey:
>are we aware of client tooling that likes the null value more than the missing property?

To be pedantic, right now I think if `z_score` is None, we're returning `"z_score": 0.0` from the API. 🙃 which is a good way for errors to sneak in upon this change.

Since we currently get webapp data via the API, there might be webapp code that starts to break. For example:
https://github.com/conbench/conbench/blob/173b970291d6b33531f45a7f0bb103ff1824442f/conbench/app/_plots.py#L36
That code would throw a KeyError and break plots. I suspect there might be more; we can look at it in #1001. I don't think we use this key in any of the clients in this monorepo but there could be other clients.

>In my tiny worldview the GET /benchmark/<bmr-id> endpoint was (the) one way to obtain the per-bmr z-score value as an API consumer, assuming that we need one way; do I didn't quite feel like going so far to remove all-the-ways. But maybe that's good!

Last week this is also what I thought, but was convinced by Jon that the lookback z-score analysis needs to be parameterized by baseline commit as well as contender result. So "give me the z-score value for this benchmark result" is not well-defined. You need a baseline too. (Of course we can default to a baseline in certain cases! See below.) 

The `/api/compare` endpoint (with modifications; see below) is perfect for this! I think that should be the primary way to access analysis results.

>Does this have a valid use case, from an API consumer's point of view?

> there is a usecase for something like this in a summary | report view

Agreed that centralizing more summaries and reporting to the server is a good longer-term initiative. In the meantime, another use case is something like https://github.com/conbench/conbench/issues/388. I do think we could accomplish that one using the compare API too.

>For our data in conbench, most benchmarks have one comparison that is the most helpful to conduct, and so having an API where that happens automatically, that can be useful

Agreed! I think the pattern of `/api/compare/*/contender?target_baseline_commit=Y` proposed in https://github.com/conbench/conbench/issues/384 is a good way to accomplish this.

PR #1043, opened 2023-04-04
legacy runner: Validate `name` == `tags["name"]` if both set

### LINKED ISSUE #1042: legacy runner: Only set `name` tag if unset
Currently the legacy runner overrides the `name` tag with whatever is passed to the `name` parameter regardless if the `name` tag is already set: https://github.com/conbench/conbench/blob/248b8909c30a7127199605d2c22505b332a2e74b/conbench/runner.py#L249 Instead, we should only set it if it's unset.

alistaire47:
Closes #1042 by only setting `tags["name"]` in the legacy runner if there is not already something there. If it's already set, ~that now takes precedence~ it checks if they match and error if not. This should lead to less confusion and surprise when the `name` param differs from `tags["name"]` (they shouldn't, but people do things) and metadata changes before posting.

### DIFF
--- a/conbench/runner.py
+++ b/conbench/runner.py
@@ -246,7 +246,8 @@ class Conbench(Connection, MixinPython, MixinR):
             output,
         ) = self._init(kwargs)
 
-        tags["name"] = name
+        if (tags.get("name") is None) or (tags["name"] == ""):
+            tags["name"] = name
 
         batch_id = options.get("batch_id")
         if not batch_id:
--- a/conbench/runner.py
+++ b/conbench/runner.py
@@ -246,8 +246,12 @@ class Conbench(Connection, MixinPython, MixinR):
             output,
         ) = self._init(kwargs)
 
-        if (tags.get("name") is None) or (tags["name"] == ""):
-            tags["name"] = name
+        if (tags.get("name") is not None) and (name == tags["name"]):
+            raise ValueError(
+                f"`name` and `tags[\"name\"] are both supplied and do not match: {name=} != {tags['name']=}"
+            )
+
+        tags["name"] = name
 
         batch_id = options.get("batch_id")
         if not batch_id:


conbench/runner.py
```diff
@@ -246,7 +246,8 @@ def record(self, result, name, error=None, publish=True, **kwargs):
             output,
         ) = self._init(kwargs)
 
-        tags["name"] = name
+        if (tags.get("name") is None) or (tags["name"] == ""):
+            tags["name"] = name
```
jonkeane: So we want to error here? Or is this something that is used elsewhere (passing the name and not putting it in the tags?) Or maybe error IFF there's a name in tags and a name in kwargs?

alistaire47: The kwarg is used by labs/benchmarks. If we error when both are populated, we'll break the case that spurred this issue. We could error if they're both populated and differ.

jonkeane: > We could error if they're both populated and differ.

_nods_ yeah that might be best

alistaire47: updated

jonkeane (approved):
Thanks!

conbench/runner.py
```diff
@@ -246,6 +246,11 @@ def record(self, result, name, error=None, publish=True, **kwargs):
             output,
         ) = self._init(kwargs)
 
+        if (tags.get("name") is not None) and (name == tags["name"]):
```
jgehrcke, line 249: Oh: `name == tags["name"]` and not `name != tags["name"]`, typo or brain fart on my side?

As a user of this code I'd wonder what the `name` argument here is supposed to be. I mean `tags.get("name")` I get it now because I am starting to refine the specification around this. But the other input parameter `name` is to me at least not explained / specified (not in the docstring here, and where else would it be explained?). Is this also the benchmark name, really? If yes: why did / does this code even ever try to provide _two_ ways to set the benchmark name?

jgehrcke: I thought about this a tiny longer and while doing so I took notes in code and did this small patch: https://github.com/conbench/conbench/pull/1048

Issue #992, opened 2023-03-31
api: baseline runs should be on the default branch


austin3dickey:
Currently when we `get_baseline_run()` it usually chooses the parent commit, even if the parent commit was a PR commit. But the baseline run should always be on the default branch for effective pointwise analysis.

Looking through the code, I suspect we can simplify many queries upon making this change. That's also because of the changes we've made to the rolling z-score calculation code.

jgehrcke:
> baseline runs should be on the default branch

I like the clarity of that statement! That's great, towards effectively communicating internally and externally was expected/correct behavior is.

I wonder if/why the definition of "baseline run" might have shifted recently. Did we always have the correct definition, and the code deviated accidentally? Were we not entirely clear about the definition in our heads, and in writing?

I would appreciate if you can add your perspective to this @austin3dickey 

A discussion like this helps developing a paper trail around decision-making, which is invaluable for deriving and iterating on design and knowledge.

jonkeane:
💯 on doing a bit of a retro to figure out what happened (also happy to do something in person if that helps). #384 might be relevant here as well

austin3dickey:
I dove into the blame and found that https://github.com/conbench/conbench/pull/507 is what originally caused this "regression." Regression is in quotation marks because before that PR, we were enforcing that baseline runs should always be on the default branch by filtering the query to `Run.reason == "commit"`. 🙃 That's not the same as being on the default branch, but practically it was, back in those days.

Notably -- in that PR, I [didn't have to](https://github.com/conbench/conbench/pull/507/files#diff-bcc9f9347c1c76d32472bac607097505072ee0d0c0b1c87135e007e040fee2b3) change any tests that were directly related to ensuring baselines were always on the default branch. So I must have misunderstood the intention when doing that PR, and no tests failed, so I thought everything was fine.

I'll definitely add a test in this patch!

jonkeane:
Thanks for all of that details + candid assessment of what happened

PR #1046, opened 2023-04-04
api/compare: respect baseline entity when doing lookback z-score analysis

### LINKED ISSUE #1029: api/compare: lookback z-scores should respect the baseline entity when choosing the end of the distribution

austin3dickey:
Closes #1029. 

This PR ensures that the z-scores returned from the `/api/compare` endpoints respect the baseline entity that was requested.

It also deprecates populating `baseline_z_score` in those responses, because that's unnecessary overhead (and it's unclear what the baseline of the baseline is).

### DIFF
--- a/conbench/api/compare.py
+++ b/conbench/api/compare.py
@@ -135,7 +135,13 @@ class CompareEntityEndpoint(ApiEndpoint, CompareMixin):
 
         baseline_benchmark_result = self._get(baseline_id)
         contender_benchmark_result = self._get(contender_id)
-        set_z_scores([baseline_benchmark_result, contender_benchmark_result])
+        set_z_scores(
+            benchmark_results=[contender_benchmark_result],
+            baseline_commit=baseline_benchmark_result.run.commit,
+        )
+        # TODO: remove baseline-z-score-related keys from this endpoint
+        baseline_benchmark_result.z_score = None
+
         set_display_case_permutation(baseline_benchmark_result)
         set_display_case_permutation(contender_benchmark_result)
         set_display_benchmark_name(baseline_benchmark_result)
@@ -189,7 +195,12 @@ class CompareListEndpoint(ApiEndpoint, CompareMixin):
 
         baselines = self._get(baseline_id)
         contenders = self._get(contender_id)
-        set_z_scores(baselines + contenders)
+        set_z_scores(
+            benchmark_results=contenders, baseline_commit=baselines[0].run.commit
+        )
+        # TODO: remove baseline-z-score-related keys from this endpoint
+        for baseline in baselines:
+            baseline.z_score = None
 
         baseline_items, contender_items = [], []
         for benchmark_result in baselines:
--- a/conbench/entities/history.py
+++ b/conbench/entities/history.py
@@ -4,8 +4,7 @@ import datetime
 import decimal
 import logging
 import math
-from collections import defaultdict
-from typing import DefaultDict, Dict, List, Optional, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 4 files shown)


austin3dickey:
~Marked as draft because I just need to write a test against the actual /api/compare endpoints.~

edit: done

jonkeane:
> It also deprecates populating baseline_z_score in those responses, because that's unnecessary overhead (and it's unclear what the baseline of the baseline is).

Thanks for calling this out — yeah I think this is totally ok. There's _some very minor possible_ uses for this, but I suspect it is mostly there because at the time it was easy to pull (it was stored in the database, and came along with the baseline so why not). But you're totally spot on that it's a bit funny — currently it is (or was) the default comparison for the baseline with it's default baseline) but like we've said in a few places: it's a bit too magical to have that happen like that, if someone is curious about that they should mark the baseline as the contender commit and go back form there.

conbench/tests/api/test_compare.py
```diff
@@ -315,13 +315,38 @@ def test_compare_unknown_compare_ids(self, client):
...
+        [
+            # result on the fork point commit
+            (5, -2.1857497724635935),
+            # result on the parent commit
+            (6, -2.8246140882627757),
+            # result on the head commit of the default branch
+            (8, 0.12045310863100521),
+        ],
+    )
+    def test_compare_different_baselines(
```
austin3dickey, line 329: Okay, ready for review. This test is pretty much the meat of this PR. On `main`, the "parent" and "head of default" cases here would fail because the z-score returned looks identical to the "fork point" case z-score.

conbench/api/compare.py
```diff
@@ -135,7 +135,13 @@ def get(self, compare_ids):
 
         baseline_benchmark_result = self._get(baseline_id)
         contender_benchmark_result = self._get(contender_id)
-        set_z_scores([baseline_benchmark_result, contender_benchmark_result])
+        set_z_scores(
+            benchmark_results=[contender_benchmark_result],
+            baseline_commit=baseline_benchmark_result.run.commit,
+        )
+        # TODO: remove baseline-z-score-related keys from this endpoint
+        baseline_benchmark_result.z_score = None
```
jgehrcke, line 143: Would appreciate if we can in a follow-up PR (while we have this at the top of our head) work on the specification of this endpoint. Where we currently have `description: Compare benchmark results.` 

We need to start having a concise, precise description of what this is doing and what the input parameters are good for.

conbench/api/compare.py
```diff
@@ -189,7 +195,12 @@ def get(self, compare_ids):
 
         baselines = self._get(baseline_id)
         contenders = self._get(contender_id)
-        set_z_scores(baselines + contenders)
+        set_z_scores(
+            benchmark_results=contenders, baseline_commit=baselines[0].run.commit
+        )
```
jgehrcke, line 200: :+1:

conbench/entities/history.py
```diff
@@ -255,58 +254,63 @@ def get_history_for_cchr(
...
-def set_z_scores(benchmark_results: List[BenchmarkResult]):
+def set_z_scores(benchmark_results: List[BenchmarkResult], baseline_commit: Commit):
     """Set the "z_score" attribute on each given BenchmarkResult, comparing it to the
-    distribution of BenchmarkResults that share the same case/context/hardware/repo,
-    ending with the most recent BenchmarkResult in the given BenchmarkResult's commit
-    ancestry on the default branch. (We choose a commit from the default branch as the
-    end commit, because we probably don't want "bad" commits early on in a PR to set
-    artificially low z-scores later in the PR.)
+    distribution of BenchmarkResults that share the same case/context/hardware/repo, in
+    the git ancestry of the baseline_commit (inclusive).
```
jgehrcke, line 260: > in the git ancestry of the baseline_commit (inclusive).

:heart:

conbench/entities/history.py
```diff
@@ -255,58 +254,63 @@ def get_history_for_cchr(
...
+
+    The given benchmark_results must all have the same run_id.
```
jgehrcke, line 262: It's good to call out invariants like this in the docstring. I see that below we make that count with the typical set() approach and otherwise raise a ValueError. Good!

conbench/entities/history.py
```diff
@@ -255,58 +254,63 @@ def get_history_for_cchr(
...
     This function does not affect the database whatsoever. It only populates an
     attribute on the given python objects. This is typically called right before
     returning a jsonify-ed response from the API, so the "z_score" attribute shouldn't
     already be set, but if it is already set, it will be silently overwritten.
 
-    Should not raise anything. If we can't find a z-score for some reason, the z_score
-    attribute will be None on that BenchmarkResult.
+    Should not raise anything, except a ValueError if there are mixed run_ids provided.
+    If we can't find a z-score for some reason, the z_score attribute will be None on
+    that BenchmarkResult.
```
jgehrcke, line 271: future: can think about encoding the reason as a string

conbench/entities/history.py
```diff
@@ -255,58 +254,63 @@ def get_history_for_cchr(
...
-    sorted_by_run_id: DefaultDict[str, List[BenchmarkResult]] = defaultdict(list)
+    run_ids = set(result.run_id for result in benchmark_results)
+    if len(run_ids) != 1:
+        raise ValueError(f"Encountered mixed run_ids in set_z_scores(): {run_ids}")
+    run_id = run_ids.pop()
+
+    if len(benchmark_results) == 1:
+        # performance optimization
+        case_id = benchmark_results[0].case_id
+        context_id = benchmark_results[0].context_id
```
jgehrcke, line 281: note to self: interesting to "know" here if these attr reads result in dynamic db lookups or were previously JOINed.

conbench/entities/history.py
```diff
@@ -255,58 +254,63 @@ def get_history_for_cchr(
...
+    else:
+        case_id = None
+        context_id = None
+
+    distribution_stats = _query_distribution_stats_by_run_id(
```
jgehrcke, line 286: Future: let's think about the name `_query_distribution_stats_by_run_id()` and/or add a code comment. It takes too long for me to figure out what this is doing :).

What `_query_distribution_stats_by_run_id()` actually does. But when reading the code plain as-is, I am confused --- what would the distribution for a Run be, considering that a Run largely is "just" a group (of potentially super unrelated) BenchmarkResults?

Does this operate on all BenchmarkResults associated with that Run? If yes: in that set of BenchmarkResults there might be some that are _not_ in the `benchmark_results` argument provided to this function, yes?

conbench/entities/history.py
```diff
@@ -255,58 +254,63 @@ def get_history_for_cchr(
...
-    run_id: str, case_id: Optional[str], context_id: Optional[str]
+    run_id: str,
+    baseline_commit: Commit,
+    case_id: Optional[str],
+    context_id: Optional[str],
 ) -> Dict[Tuple[str, str], Tuple[Optional[float], Optional[float]]]:
-    """Given a run_id, return stats of the distribution of BenchmarkResults in the run's
-    commit ancestry, by case/context. Returns a dict that looks like
+    """Given a run id and baseline commit, return stats of the distribution of
+    BenchmarkResults matching the run's hardware, in the baseline_commit's ancestry, by
```
jgehrcke, line 312: > return stats of the distribution of  BenchmarkResults matching the run's hardware

Of _all_ `BenchmarkResults` in the database, or just those that are associated with that Run?

Edit:

I see that below we have `.join(Run, Run.id == BenchmarkResult.run_id)`.  The second argument generates the `ON` clause: https://docs.sqlalchemy.org/en/20/core/selectable.html#sqlalchemy.sql.expression.Select.join

The ON clause is roughly like a filter, it is discussed exhaustively (vs. WHERE) here: https://stackoverflow.com/questions/354070/sql-join-where-clause-vs-on-clause

Quick summary of my mental model: in set_z_scores() we require all BMRs to be associated with a single run. In set_z_scores() we call _query_distribution_stats_by_run_id() and provide that run as input. That's input asymmetry. 

Maybe set_z_scores() should have a single run as input, or we need a function analogue to _query_distribution_stats_by_run_id() that operates on a set of `BenchmarkResult`. For the case where `set_z_scores()` is being called for a set of BenchmarkResults that are only a _subset_ of those associated with a single run.

conbench/tests/api/test_compare.py
```diff
@@ -315,13 +315,38 @@ def test_compare_unknown_compare_ids(self, client):
...
+        self, client, baseline_result_id, expected_z_score
+    ):
+        self.authenticate(client)
+        _, benchmark_results = _fixtures.gen_fake_data()
+        contender = benchmark_results[7]
+        baseline = benchmark_results[baseline_result_id]
```
jgehrcke, line 335: I find interesting in this code, the baseline and contender are both "BenchmarkResult"s. For our choice of terminology: do we need to say "baseline result" and "contender result" instead, or is it established enough by now that we mean an individual benchmark result when we say e.g. "contennder"?

conbench/tests/api/test_compare.py
jgehrcke: Thank you for doing this, and for pointing out how this test would fail on main.

jgehrcke (approved):
Great work! I must say I have only done rather localized review instead of a big picture review, and also didn't try to understand all details.This is a great step forward, in particular because of the additional test coverage and the higher degree of code clarity :rocket: I'd like to land this now. If some of the feedback inspires you to do tiny bit of follow-up work that's of course cool, but all my feedback is I think just of general nature.

conbench/tests/api/test_compare.py
```diff
@@ -315,13 +315,38 @@ def test_compare_unknown_compare_ids(self, client):
...
+            # result on the head commit of the default branch
+            (8, 0.12045310863100521),
+        ],
+    )
+    def test_compare_different_baselines(
+        self, client, baseline_result_id, expected_z_score
+    ):
+        self.authenticate(client)
+        _, benchmark_results = _fixtures.gen_fake_data()
+        contender = benchmark_results[7]
```
jgehrcke, line 334: Is the contender on a non-default branch?

I think it would be valuable if you could clarify the relationship of this test with #992! Is #992 resolved now as a side effect of this patch?

jgehrcke:
Maybe above this was a little buried, but I'd love to see your point of view: how does this patch relate to #992?

conbench/api/compare.py
austin3dickey: Yes! I'm going to be changing the compare endpoint a lot over the next few days, so I'll make sure to document the final state.

conbench/entities/history.py
austin3dickey: In this case, `case_id` is stored as a string column in the `benchmark_result` table, so accessing it here doesn't cost any DB lookups. It was already populated when we loaded the benchmark result entity.

I think if we accessed the `benchmark_results[0].case` object, _that_ would be lazy-loaded from the DB.

conbench/entities/history.py
austin3dickey: Ah shoot, I was going to change the name of this function in this PR, and forgot. Thanks for calling this out! 

Run ID is much less important to this function now than it used to be. It used to absolutely determine the distribution. Now the distribution is mostly determined by the baseline commit (and results are filtered to the run's hardware).

conbench/entities/history.py
austin3dickey: I'll clarify this in an upcoming PR (see my above comment)

conbench/tests/api/test_compare.py
austin3dickey: Yep I should have used "contender_result" (because you can have other contender entities) but I didn't want the f-string to be too long. 😂

conbench/tests/api/test_compare.py
austin3dickey: #992 is not resolved, and it becomes much more important to resolve as a result of this patch! Because now, workflows like `benchalerts` that do the following might break even worse:
1. get a contender run id
2. hit `/api/runs/{contender_run_id}` to get its baseline run id
3. hit `/api/compare/runs/{baseline_run_id}...{contender_run_id}` to get comparisons

#992 means that for contender runs on a PR branch with multiple runs on multiple commits, step 2 could return another PR run on the parent commit as the baseline run. Before this patch, it didn't matter, because step 3 always returned the intended comparisons (with the baseline commit being the fork-from-default). But now, step 3 will return a comparison that includes the PR runs in the baseline distribution. That could pollute the z-scores.

PR #1048, opened 2023-04-05
conbench/runner.py: fix 'name' treatment, clarify

jgehrcke:
Tiny follow-up: https://github.com/conbench/conbench/pull/1043#discussion_r1158183990, and some clarifications in the surrounding code context.

### DIFF
--- a/conbench/runner.py
+++ b/conbench/runner.py
@@ -233,8 +233,25 @@ class Conbench(Connection, MixinPython, MixinR):
     def machine_info(self):
         return machine_info(self.config.host_name)
 
-    def record(self, result, name, error=None, publish=True, **kwargs):
-        """Record and publish an external benchmark result."""
+    def record(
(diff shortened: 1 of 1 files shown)


jgehrcke:
Feedback appreciated, :ship:!

conbench/runner.py
```diff
@@ -246,11 +264,14 @@ def record(self, result, name, error=None, publish=True, **kwargs):
             output,
         ) = self._init(kwargs)
 
-        if (tags.get("name") is not None) and (name == tags["name"]):
+        if (tags.get("name") is not None) and (name != tags["name"]):
```
austin3dickey, line 267: Thank you! Agreed this was the intent

jonkeane: yes, this is the right intent — should we add a test to cover this?

austin3dickey: I can add a test when I bump the version. Coming soon.

PR #1049, opened 2023-04-05
minor: docstring/comment/name clarifications

austin3dickey:
This is a small PR in response to @jgehrcke's feedback on #1046.

### DIFF
--- a/conbench/entities/history.py
+++ b/conbench/entities/history.py
@@ -254,12 +254,14 @@ def get_history_for_cchr(
     return samples
 
 
-def set_z_scores(benchmark_results: List[BenchmarkResult], baseline_commit: Commit):
-    """Set the "z_score" attribute on each given BenchmarkResult, comparing it to the
-    distribution of BenchmarkResults that share the same case/context/hardware/repo, in
-    the git ancestry of the baseline_commit (inclusive).
+def set_z_scores(
+    contender_benchmark_results: List[BenchmarkResult], baseline_commit: Commit
+):
+    """Set the "z_score" attribute on each contender BenchmarkResult, comparing it to
+    the baseline distribution of BenchmarkResults that share the same
+    case/context/hardware/repo, in the git ancestry of the baseline_commit (inclusive).
 
-    The given benchmark_results must all have the same run_id.
+    The given contender_benchmark_results must all have the same run_id.
 
     This function does not affect the database whatsoever. It only populates an
     attribute on the given python objects. This is typically called right before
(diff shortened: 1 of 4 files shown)


conbench/entities/history.py
```diff
@@ -302,30 +306,40 @@ def set_z_scores(benchmark_results: List[BenchmarkResult], baseline_commit: Comm
...
+    contender_run_id: str,
     baseline_commit: Commit,
     case_id: Optional[str],
     context_id: Optional[str],
 ) -> Dict[Tuple[str, str], Tuple[Optional[float], Optional[float]]]:
-    """Given a run id and baseline commit, return stats of the distribution of
-    BenchmarkResults matching the run's hardware, in the baseline_commit's ancestry, by
-    case/context. Returns a dict that looks like
+    """Query and calculate rolling stats of the distribution of all BenchmarkResults
+    that:
```
austin3dickey, line 316: @jgehrcke Let me know if this name and docstring clarifies your questions better.

conbench/entities/history.py
```diff
@@ -270,27 +272,29 @@ def set_z_scores(benchmark_results: List[BenchmarkResult], baseline_commit: Comm
...
+        case_id = contender_benchmark_results[0].case_id
+        context_id = contender_benchmark_results[0].context_id
     else:
         case_id = None
         context_id = None
 
-    distribution_stats = _query_distribution_stats_by_run_id(
-        run_id=run_id,
+    distribution_stats = _query_and_calculate_distribution_stats(
+        contender_run_id=contender_run_id,
```
jgehrcke, line 291: :) I want to go back to this point:

> Maybe set_z_scores() should have a single run as input, or we need a function analogue to _query_distribution_stats_by_run_id() that operates on a set of BenchmarkResult. For the case where set_z_scores() is being called for a set of BenchmarkResults that are only a subset of those associated with a single run.

Imagine we call `set_z_scores()` for two benchmark results which are part of a Run that is associated with 2000 benchmark results.

Then I think the current `_query_and_calculate_distribution_stats()` does the work for 2000 benchmark results, although it would be enough to just do it for two of them.

That is, I am still a little confused about what I think is a bit of an input asymmetry:

- input for  `set_z_scores()` is a set `S1` of benchmark results. These have a `run_id` `R1` in common.
- input for `_query_and_calculate_distribution_stats()` is `R1`. `R1` is associated with the set `S2` of benchmark results.

S1 is not necesarily the same as S2. S1 is always a subset of S2, but S2 can be much bigger than S1.

I might be completely off here, but wanted to re-word more clearly what I meant before!

At the heart this point is super closely related to the question what a `Run`-based grouping is useful for! :)

jgehrcke (approved):
These are all valuable clarifications! Thank you!

I have left one more fundamental 'question' here -- but we can discuss that and respond to that independent of this PR.

Let's land this!

conbench/entities/history.py
austin3dickey: Ahh, I see what you're saying. Got it.

Today, `set_z_scores()` is called: 
- on 1 BenchmarkResult
https://github.com/conbench/conbench/blob/173b970291d6b33531f45a7f0bb103ff1824442f/conbench/api/compare.py#L138-L141
- on all the results in one run 
- on all the results in one batch
https://github.com/conbench/conbench/blob/173b970291d6b33531f45a7f0bb103ff1824442f/conbench/api/compare.py#L198-L200

You're talking about the batch use case (as well as any future use cases where multiple given contender results are a strict subset of all the results on one run -- though I can't imagine what those use cases might practically be).

Agreed that there's an input asymmetry here, and that we could squeeze even more performance optimization out of the batch use case if we first-class it with a different query filter. Right now, I think calling `/api/compare/batches/` is unpopular, so I decided not to first-class that case, and instead shoved it all into one function. But I agree that could be cleaned up, potentially with different functions for these different use cases.

conbench/entities/history.py
```diff
@@ -302,30 +306,40 @@ def set_z_scores(benchmark_results: List[BenchmarkResult], baseline_commit: Comm
         )
 
 
-def _query_distribution_stats_by_run_id(
-    run_id: str,
+def _query_and_calculate_distribution_stats(
+    contender_run_id: str,
     baseline_commit: Commit,
     case_id: Optional[str],
     context_id: Optional[str],
```
jgehrcke, line 313: Note from sync meeting: case_id and context_id are where we narrow this analysis down to "just matching benchmark results in the contender_run".

austin3dickey: Another note from the meeting: for code readability, it would probably make sense to pass along the `contender_benchmark_results` instead of `contender_run_id` and `case_id` and `context_id`. But "filtering the distribution query to a potentially-large python list of case/context pairs" seems hard to do in SQLAlchemy. That's why I didn't do it before.

What if we didn't filter case/context at all in the distribution query, and instead did it in pandas after execution? That could probably speed up the query since we'd remove the complicated join!

austin3dickey: Created #1051 for this

conbench/entities/history.py
jgehrcke: Came here again after our in-person discussion yesterday. Thanks for the concise write-up!

>  I decided not to first-class that case, and instead shoved it all into one function. But I agree that could be cleaned up, potentially with different functions for these different use cases.

Thank you for adding this; makes total sense! Also yes it's important to know when even if a function allows for a certain code path, that maybe the caller(s) are _known_ to never exercise it.

I want to kind of reinforce that my questioning is not too much about performance or tidiness of code initially, but more about our data model and which level of complexity is fundamentally required for answering the questions that we need to answer -- i.e. this was a lot about "what's the value of the Run abstraction in this history query information flow?" 

The intuition that has been building up is that we can do virtual grouping constructs in the future, and the only 'real' thing that we ever need is the individual benchmark result. This discussion here more or less supports this, or at least doesn't conflict deeply.

Our discussion has helped greatly on that front, thank you!

PR #1054, opened 2023-04-05
api/runs: ensure baseline run is always on default branch

### LINKED ISSUE #992: api: baseline runs should be on the default branch

austin3dickey:
Closes #992. After this patch, all baseline runs returned from `/api/runs` should be on the default branch.

### DIFF
--- a/conbench/entities/run.py
+++ b/conbench/entities/run.py
@@ -162,17 +162,16 @@ class Run(Base, EntityMixin):
 
         return run
 
-    def get_baseline_run(
+    def get_default_baseline_run(
         self,
         commit_limit: int = 20,
-        on_default_branch: bool = False,
         case_id: Optional[str] = None,
         context_id: Optional[str] = None,
     ) -> Optional["Run"]:
         """Return the closest ancestor of this Run, where the ancestor run:
 
         - is in the last ``commit_limit`` commits of this Run's commit ancestry
-        - is on the default branch, if on_default_branch is True (else on any branch)
+        - is on the default branch
         - shares this Run's hardware
         - has a BenchmarkResult with the given case_id/context_id, if those are given
         - if they aren't given, has a BenchmarkResult with the case_id/context_id of
@@ -189,17 +188,22 @@ class Run(Base, EntityMixin):
 
         this_commit: Commit = self.commit
         try:
-            ancestor_commits = (
-                this_commit.commit_ancestry_query
-                # don't count this run's commit
-                .filter(Commit.id != this_commit.id)
-                .order_by(s.desc("commit_order"))
-                .limit(commit_limit)
-                .subquery()
-            )
+            ancestor_commits = this_commit.commit_ancestry_query.subquery()
         except CantFindAncestorCommitsError as e:
-            log.debug(f"Couldn't get_baseline_run() because {e}")
+            log.debug(f"Couldn't get_default_baseline_run() because {e}")
             return None
+
+        ancestor_commits = (
+            Session.query(ancestor_commits)
+            .filter(
+                # don't count this run's commit
+                ancestor_commits.c.ancestor_id != this_commit.id,
+                ancestor_commits.c.on_default_branch.is_(True),
+            )
+            .order_by(s.desc("commit_order"))
+            .limit(commit_limit)
+            .subquery()
+        )
 
         closest_run_id_query = (
             Session.query(BenchmarkResult.run_id)
@@ -208,11 +212,6 @@ class Run(Base, EntityMixin):
             .join(ancestor_commits, ancestor_commits.c.ancestor_id == Run.commit_id)
             .filter(Hardware.id == self.hardware_id)
         )
-
-        if on_default_branch:
-            closest_run_id_query = closest_run_id_query.filter(
-                ancestor_commits.c.on_default_branch.is_(True)
-            )
 
         # Filter to the correct case(s)/context(s)
         if case_id and context_id:
@@ -250,8 +249,8 @@ class Run(Base, EntityMixin):
 
         return Run.get(closest_run_id)
 
-    def get_baseline_id(self):
-        run = self.get_baseline_run()
+    def get_default_baseline_id(self):
+        run = self.get_default_baseline_run()
         return run.id if run else None
 
 
@@ -404,7 +403,7 @@ class _Serializer(EntitySerializer):
             },
         }
         if not self.many:
-            baseline_id, baseline_url = run.get_baseline_id(), None
+            baseline_id, baseline_url = run.get_default_baseline_id(), None
             if baseline_id:
                 baseline_url = f.url_for("api.run", run_id=baseline_id, _external=True)
             result["links"]["baseline"] = baseline_url
--- a/conbench/tests/api/test_runs.py
+++ b/conbench/tests/api/test_runs.py
@@ -231,6 +231,27 @@ class TestRunGet(_asserts.GetEnforcer):
         response = client.get(f"/api/runs/{contender_run.id}/")
         self.assert_200_ok(response, _expected_entity(contender_run, baseline_run.id))
 
+    def test_baseline_run_always_on_default_branch(self, client):
+        self.authenticate(client)
+        _, benchmark_results = _fixtures.gen_fake_data()
+
+        successes = 0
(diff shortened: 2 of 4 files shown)


conbench/tests/api/test_runs.py
```diff
@@ -231,6 +231,27 @@ def test_closest_commit_different_machines_should_not_prefer_test_runs_as_baseli
...
+            assert response.status_code == 200, response.status_code
+            baseline_run_link = response.json["links"].get("baseline")
+            if baseline_run_link:
+                baseline_run_response = client.get(
+                    baseline_run_link.replace("http://localhost", "")
+                )
+                assert (
+                    baseline_run_response.status_code == 200
+                ), baseline_run_response.status_code
+                assert baseline_run_response.json["commit"]["branch"] == "default"
```
austin3dickey, line 250: This test does fail on this line on `main`.

conbench/entities/run.py
```diff
@@ -165,14 +165,13 @@ def create(data) -> "Run":
     def get_baseline_run(
```
jonkeane: I know this isn't part of this patch, but I wonder if we should rename this to something like `get_default_baseline_run()` so that it's (extra) clear that this isn't _the only_ baseline that could ever exist for the contender we have in `self`, it's just our best offer for a default baseline

conbench/entities/run.py
```diff
@@ -165,14 +165,13 @@ def create(data) -> "Run":
...
         self,
         commit_limit: int = 20,
-        on_default_branch: bool = False,
         case_id: Optional[str] = None,
         context_id: Optional[str] = None,
     ) -> Optional["Run"]:
         """Return the closest ancestor of this Run, where the ancestor run:
 
         - is in the last ``commit_limit`` commits of this Run's commit ancestry
-        - is on the default branch, if on_default_branch is True (else on any branch)
```
jgehrcke, line 175: :+1:

conbench/tests/api/test_runs.py
```diff
@@ -231,6 +231,27 @@ def test_closest_commit_different_machines_should_not_prefer_test_runs_as_baseli
...
         self.assert_200_ok(response, _expected_entity(contender_run, baseline_run.id))
 
+    def test_baseline_run_always_on_default_branch(self, client):
+        self.authenticate(client)
+        _, benchmark_results = _fixtures.gen_fake_data()
+
+        successes = 0
+        for benchmark_result in benchmark_results:
+            response = client.get(f"/api/runs/{benchmark_result.run_id}/")
+            assert response.status_code == 200, response.status_code
```
jgehrcke: unrelated, but I want to say this here because I keep "fixing" this in our codebase. Let's write these as

```
assert response.status_code == 200, response.text
```
Because when we get an unexpected response what we usually want to see in the output (if we have to pick _one_ thing) is the response body. When the assertion fails: yes the status code isn't the one we expected, but now to debug this we usually want to peek into the body.

> (if we have to pick _one_ thing)

 Emitting code _and_ body is the next-better option.

conbench/tests/api/test_runs.py
```diff
@@ -231,6 +231,27 @@ def test_closest_commit_different_machines_should_not_prefer_test_runs_as_baseli
...
+                    baseline_run_link.replace("http://localhost", "")
+                )
+                assert (
+                    baseline_run_response.status_code == 200
+                ), baseline_run_response.status_code
+                assert baseline_run_response.json["commit"]["branch"] == "default"
+                successes += 1
+
+        assert successes == 10
+
```
jonkeane, line 254: Do we already have tests (possibly up above, I skimmed the names but didn't see any that clearly matched this):

* Contender is on a branch and the fork-point has benchmarks (so can be used)
* Contender is on a branch and the fork-point has no benchmarks (so we use the parent of the fork-point if it has benchmarks)
* Contender is on a branch, the parent of the contender has benchmarks but we ignore those and use the fork-point
* Contender is on default, and we use the parent (if it has benchmarks)
* Contender is on default, and the parent doesn't have benchmarks (so we use the grandparent if that has benchmarks)

conbench/tests/api/test_runs.py
```diff
@@ -231,6 +231,27 @@ def test_closest_commit_different_machines_should_not_prefer_test_runs_as_baseli
...
+        successes = 0
+        for benchmark_result in benchmark_results:
+            response = client.get(f"/api/runs/{benchmark_result.run_id}/")
+            assert response.status_code == 200, response.status_code
+            baseline_run_link = response.json["links"].get("baseline")
+            if baseline_run_link:
+                baseline_run_response = client.get(
+                    baseline_run_link.replace("http://localhost", "")
+                )
```
jgehrcke, line 246: note to self: trying to be/do [HATEOAS](https://en.wikipedia.org/wiki/HATEOAS)  is always a little weird :D It never lives up to its promises.

conbench/tests/api/test_runs.py
jgehrcke: In the "fake data" the branch is literally called "default"?

jgehrcke (approved):
:rocket: Love it! Left remarks / questions, but nothing critical.

conbench/tests/api/test_runs.py
austin3dickey: Yep! https://github.com/conbench/conbench/blob/61fac4e5c12ee4b71a493929662e580084e4186b/conbench/tests/api/_fixtures.py#L287

conbench/tests/api/test_runs.py
austin3dickey: haha great point! I'll fix this

conbench/entities/run.py
austin3dickey: Yes that's a great idea. This method in general will probably become much less necessary after #384 and followup UI work. We'll be able to just hit `/api/compare/runs/run_id?target_baseline_commit=fork_point` and can skip the need for this method.

conbench/tests/api/test_runs.py
austin3dickey: Changed to this 👀 
```py
assert response.status_code == 200, f"bad response: {response.__dict__}"
```

conbench/tests/api/test_runs.py
austin3dickey: Yes, these cases are all covered in [gen_fake_data()](https://github.com/conbench/conbench/blob/61fac4e5c12ee4b71a493929662e580084e4186b/conbench/tests/api/_fixtures.py#L265). This means they're being tested in both [TestRunGet.test_baseline_run_always_on_default_branch](https://github.com/conbench/conbench/blob/e57612363845b55d6d8072661c9101277ae53138/conbench/tests/api/test_runs.py#L234) for "is their baseline on the default branch?" and [test_get_baseline_run](https://github.com/conbench/conbench/blob/e57612363845b55d6d8072661c9101277ae53138/conbench/tests/entities/test_run.py#L19) for the explicit behavior you called out.

- Contender is on a branch and the fork-point has benchmarks (so can be used) -> fake benchmark results 2, 3, 6, 7
- Contender is on a branch and the fork-point has no benchmarks (so we use the parent of the fork-point if it has benchmarks) -> fake benchmark result 4
- Contender is on a branch, the parent of the contender has benchmarks but we ignore those and use the fork-point -> fake benchmark result 7
- Contender is on default, and we use the parent (if it has benchmarks) -> fake benchmark result 1
- Contender is on default, and the parent doesn't have benchmarks (so we use the grandparent if that has benchmarks) -> fake benchmark results 5, 8, 9, 16

(Also, fake benchmark result 0 has no parents, and fake benchmark results 10-15 can't be compared for one reason or another.)

conbench/tests/api/test_runs.py
jgehrcke: :+1:

conbench/tests/api/test_runs.py
jgehrcke: Super good discussion.

Issue #1066, opened 2023-04-06
ci: add pylint or comparable


jgehrcke:
https://github.com/conbench/conbench/issues/1055 could have been easily prevented with a canonical linting solution:

![Screenshot from 2023-04-06 13-05-03](https://user-images.githubusercontent.com/265630/230359524-203c44f1-4e30-4084-b8f3-a22916b5c3c3.png)

I think a largish number of projects use both, flake8 and pylint. flake8 these days is pretty much actually pyflakes.

jgehrcke:
Done via #1069.

Issue #1055, opened 2023-04-05
conbench CLI: logging error in runner record() method


austin3dickey:
In #621 we replaced `print()`s with logging. There was a mistake in this line:
https://github.com/conbench/conbench/blob/61fac4e5c12ee4b71a493929662e580084e4186b/conbench/util.py#L173
Whenever the `Conbench.record()` method is called, this results in the following LoggingError. (The code still continues past the line.)
<details>
 <summary>Traceback</summary>

```
--- Logging error ---
Traceback (most recent call last):
  File "/tmp/conda-01875206-650b-44b0-b980-841363c0ac72/lib/python3.10/logging/__init__.py", line 1100, in emit
    msg = self.format(record)
  File "/tmp/conda-01875206-650b-44b0-b980-841363c0ac72/lib/python3.10/logging/__init__.py", line 943, in format
    return fmt.format(record)
  File "/tmp/conda-01875206-650b-44b0-b980-841363c0ac72/lib/python3.10/logging/__init__.py", line 678, in format
    record.message = record.getMessage()
  File "/tmp/conda-01875206-650b-44b0-b980-841363c0ac72/lib/python3.10/logging/__init__.py", line 368, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting
Call stack:
  File "/tmp/conda-01875206-650b-44b0-b980-841363c0ac72/bin/conbench", line 8, in <module>
    sys.exit(conbench())
  File "/tmp/conda-01875206-650b-44b0-b980-841363c0ac72/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/tmp/conda-01875206-650b-44b0-b980-841363c0ac72/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/tmp/conda-01875206-650b-44b0-b980-841363c0ac72/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/tmp/conda-01875206-650b-44b0-b980-841363c0ac72/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/tmp/conda-01875206-650b-44b0-b980-841363c0ac72/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
[...]

PR #1067, opened 2023-04-06
ci: add pylint (#1066)

jgehrcke:
This adds pylint which for starters does not cover the larger chunk of our codebase.

This configuration however should reveal bug #1055.

That is, I expect CI to fail (the lint step).

Notable:
- Pylint has decent docs: https://pylint.readthedocs.io/en/latest/
- Pylint is about to transition from 2.x to 3.x. Via PyPI as of today we still get 2.x when we do not add a version specifier into the requirements files: https://pypi.org/project/pylint/#history

As we cover more source files we may want to explicitly use 3.x. But for now we should also respect that the 3.x is still in alpha release stage, i.e. using 2.x makes sense for the initial introduction (proven tooling).

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,11 @@
+[MAIN]
+
+ignore-paths=^conbench/app/.*$,
+             ^conbench/api/.*$,
+             ^conbench/entities/.*$,
+             ^conbench/tests/.*$,
+
+
+
+
+
--- a/.pylintrc
+++ b/.pylintrc
@@ -5,7 +5,3 @@ ignore-paths=^conbench/app/.*$,
              ^conbench/entities/.*$,
              ^conbench/tests/.*$,
 
-
-
-
-
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,19 @@ lint:
 	flake8
 	isort .
 	black .
+	pylint --errors-only  conbench
+	mypy conbench
+	mypy benchalerts
+
+
+# Run by CI, these commands should not modify files, but only check compliance.
+# and exit with a non-zero status code upon error.
+.PHONY: lint-ci
+lint-ci:
+	flake8
+	isort --check .
+	black --check --diff .
+	pylint --errors-only  conbench
 	mypy conbench
 	mypy benchalerts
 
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -105,16 +118,6 @@ build-docs:
...
 
 
-# Run by CI, these commands should not modify files, but only check compliance.
-.PHONY: lint-ci
-lint-ci:
-	flake8
-	isort --check .
-	black --check --diff .
-	mypy conbench
-	mypy benchalerts
```
jgehrcke, line 115: Moving these two Makefile targets closer again. Slightly violates the rule of where to put which targets, but having those two close-by simply is more important.

.pylintrc
```diff
@@ -0,0 +1,7 @@
+[MAIN]
+
+ignore-paths=^conbench/app/.*$,
+             ^conbench/api/.*$,
+             ^conbench/entities/.*$,
+             ^conbench/tests/.*$,
```
jgehrcke, line 6: I picked this technique: https://github.com/pylint-dev/pylint/issues/2686#issuecomment-864585778

jgehrcke:
Yes:
```
pylint --errors-only  conbench
************* Module conbench.runner
conbench/runner.py:192:29: E0601: Using variable 'benchmark' before assignment (used-before-assignment)
************* Module conbench.util
conbench/util.py:16:0: E0401: Unable to import 'requests.packages.urllib3.util.retry' (import-error)
conbench/util.py:208:12: E1205: Too many arguments for logging format string (logging-too-many-args)
make: *** [Makefile:76: lint-ci] Error 2
```

Deactivated (outcommented) so we can merge this, and then activate again together with a bugfix.

jgehrcke:
:ship: :shipit: 

@austin3dickey I'd in particular love to have your opinion, but I kind of hope and assume you link this additional railguard. Will go ahead with this, and then do the patch for https://github.com/conbench/conbench/issues/1055 independently.

austin3dickey:
I'm always a fan of linting! :) This one seems to take a little longer than mypy but I like it.

Issue #832, opened 2023-03-09
BenchmarkResult schema: reconcile, simplify `error` vs `validation`


jgehrcke:
Follow-up from https://github.com/conbench/conbench/issues/825#issuecomment-1459903081 and https://github.com/conbench/conbench/issues/813.

`error` and `validation` and `stats` are not orthogonal to each other. All three can as of today be used to represent error information. How should users decide?

I described [elsewhere](https://gehrcke.de/2022/11/interface-design-a-craft-and-an-art/) that

> Maximizing simplicity requires identification of the fundamental degrees of freedom (that then need to be exposed). Removing variables. Identifying invariants, and working with them. This kind of thinking can make an interface justifiably beautiful.

I think this here is a great example. The goal is not to remove flexibility, but to remove ambiguity.

The first goal of this ticket is to have a nice piece of documentation that describes users what to use these fields for, and how. 

The second goal is to have updated schemas / input data validation, which is helpfully in sync with the new documentation/specification.

Quoting @jonkeane about the current state:

> The major (only?) difference between error and validation is that if there’s contents in error, the data from the benchmarks are excluded from results, history, etc.  [...] But for validation that’s informational only.

There are historical reasons for introducing `validation`, but we should question those.

In unit testing there often is `passed, failed, skipped`. But when the test _has_ an outcome, it's binary.

I think it makes sense for a benchmark result to be in either of precisely two states: passed (data is being used), failed (data is not being used).

If users believe they need something inbetween, then my answer is that the definition of "passed" vs. "failed" needs to be worked on.

Simplistic, powerful view:

- when a benchmark result is _failed_ then there must be error information
- when a benchmark result is _passed_ then there must not be error information.

The word "validation" has strong semantics which overlaps with the concept of error. I think it only dilutes the view above.

Maybe sometimes in addition to "passed", for some unknown meta use case that has nothing to do with benchmarking, people want to use the Conbench system as a vehicle to cheaply persist some useful data which happens to have been obtained during benchmarking, as a side-effect. We _could_ provide a free text / JSON field where a benchmarking environment can annotate a benchmark result with some additional info; the use case and semantic meaning should be completely neutral.

My intuition is that we should remove 'validation'.

jgehrcke:
> We could provide a free text / JSON field where a benchmarking environment can annotate a benchmark result with some additional info; the use case and semantic meaning should be completely neutral.

It seems like we have that already: `optional_benchmark_info`, currently documented as 

[...]

Issue #813, opened 2023-03-07
POST benchmark result: require at least one non-null iteration result?


jgehrcke:
The current specification is unclear. Code is, too. Maybe this is covered by tests. 

This is a quick ticket based on https://github.com/conbench/conbench/pull/811/files#diff-9d75d1bf6ac23bfc056a410721cf4c9bccd8b302f8e3e83e4b4bf1698d0eaad7R99.

action item: properly specify this in text form (api spec), confirm that specified behavior is covered by tests, and also make sure that the business logic is readable, simple.

jonkeane:
For posterity, the API docs that talk about this are at https://github.com/conbench/conbench/blob/aa95b6d52ff6cb6697b7ab8d04c43d6cf3612860/conbench/entities/benchmark_result.py#L239 (there's lots to improve about where+how this is surfaced, but I do not want to distract with that here!)

While I would love to live in a world where sending something like `[null, null, null]` as results was illegal (and effectively never happened), we've had experience with conbench users where they did, in fact, want to log if a benchmark entirely failed each time (along with a traceback in the error or optional benchmark info fields so that a developer can quickly see what has gone wrong and address it without needing to go into CI logs to find that traceback in the runner). 

Unless we decide to change this (though we will still need to support that all-iterations-errored case somehow), the API docs should be clearer that `[null, null, null]` is acceptable.

jgehrcke:
Thank you for the feedback @jonkeane. This tech talk is fun.

>  we've had experience with conbench users where they did, in fact, want to log if a benchmark entirely failed each time

Sure. I think in that case one should send the wrapper `_BenchmarkFacadeSchemaCreate` without the `stats` property (which is optional):
```
stats = marshmallow.fields.Nested(BenchmarkResultSchema().create, required=False)
```


I'd love to have one recommended way for communicating the all-iterations-failed outcome. Between the two ways we offer today I think I like the do-not-set-stats-property more than the set-all-items-to-null.

In the outer structure one can use the `error` field to represent the wrongness in a custom way.

Let's also require `error` to be populated when no `stats` property was provided.

When there is at least one non-null iteration result but also at least one `null` result: we should not require `error` to be set, but document that we recommend to populate the `error` property with useful information about how the individual iteration(s) failed.

For the case where there is more than one failed iteration (but at least one non-null result) we have the challenge of having more than one error, but only one `error` struct. In that case we should recommend users to create their own iteration-error-mapping.

jonkeane:
[...]

PR #428, opened 2022-11-07
Update BenchmarkResult for API changes

alistaire47:
Updates `benchadapt.BenchmarkResult` for recent changes to the Conbench API. Specifically:

1. Both `stats` and `error` can now be specified without warning
2. Added field for `validation`
3. Added field for `optional_benchmark_info`

PR #1074, opened 2023-04-07
legacy conbench library: publish stats even if error

### LINKED ISSUE #1073: legacy conbench library: allow submission of both results and error
Today when calling `conbench.record()` in the legacy `conbench` library, we suppress POSTing result data if there was an error:
https://github.com/conbench/conbench/blob/5393917be45cb2703e5997d2c73b5cf131e2935b/conbench/runner.py#L309-L317
But these days, the server can handle both keys. We should allow clients to be able to POST stats if there's also an error.

austin3dickey:
Fixes #1073.

This PR lets the legacy `conbench` library post stats to conbench even if an error was provided.

### DIFF
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -399,7 +399,13 @@ class BenchmarkResultStatsSchema(marshmallow.Schema):
                 iteration, the second element is the second iteration, etc.).
                 If an iteration did not complete but others did and you want to
                 send partial data, mark each iteration that didn't complete as
-                `null`."""
+                `null`.
+
+                You may populate both this field and the "error" field in the top level
+                of the benchmark result payload. In that case, this field measures
+                the metric's values before the error occurred. These values will not be
+                compared to non-errored values in analyses and comparisons.
+                """
             )
         },
     )
@@ -416,6 +422,12 @@ class BenchmarkResultStatsSchema(marshmallow.Schema):
                 second iteration, etc.). If an iteration did not complete but
                 others did and you want to send partial data, mark each
                 iteration that didn't complete as `null`.
+
+                You may populate both this field and the "error" field in the top level
+                of the benchmark result payload. In that case, this field measures
+                how long the benchmark took to run before the error occurred. These
+                values will not be compared to non-errored values in analyses and
+                comparisons.
                 """
             )
         },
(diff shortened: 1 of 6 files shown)


jgehrcke (approved):
:rocket: 

It's probably cool to merge this :)

However, I'd love have more papertrail about how this relates to "what do we really want?"

Maybe the discussion in #832 and in #813 is about this very topic, and I'd love to see how this PR relates to that discussion.

Edit: I'd love for us to use this opportunity to further think about the _desired_ (ideal) API specification in this area. And then to see if this change gets us closer or further away. My feeling is that it even brings us closer, but we should take a tiny moment and go back to the previous discussions, and double-check :)

jonkeane:
Agreed and seconding what @jgehrcke says here, we should link this back to further discussion. 

It also sounds like this will send both times and errors, which is not what is currently documented (though as far as I know unenforced):

https://github.com/conbench/conbench/blob/5393917be45cb2703e5997d2c73b5cf131e2935b/conbench/entities/benchmark_result.py#L385-L422

If it is as simple as updating the docs here to say that you can send both and what that means, we should do that with this change IMO. If we need to do more we can split it out — but let's keep ourselves honest and our docs up to date

alistaire47:
I think this makes the legacy runner inconsistent with benchadapt and its docs. If we're going to handle this situation, we should handle it everywhere regardless of tooling.

austin3dickey:
Ha, I thought this PR was making everything _more_ consistent based on past discussions 🙂 I should have linked a better paper trail and fixed the docs though!

The original PR that allowed the Conbench server to accept both "stats" and "error" keys when posting benchmark results was https://github.com/conbench/conbench/pull/416. It looks like that was more for the use case of "partially-successful" results where 1 to (N-1) iterations failed, but still, it started allowing the server to accept both keys to be posted even if all N failed. 

Critically, we then updated `benchadapt` to also allow for this case in https://github.com/conbench/conbench/pull/428. So now only the legacy `conbench` runner is out-of-date.

I do think this is related to https://github.com/conbench/conbench/issues/813, but most of that discussion looks to be around "what if all submitted stats are empty" and not "what if all stats are populated but there's also an error". It does relate to https://github.com/conbench/conbench/issues/832 too in that maybe this use case was intended to populate the "validation" key, not "error", but since that discussion is leaning toward deprecating "validation", I decided not to touch the legacy runner's handling of "validation" in this PR. We could always make a followup PR. Though I'm getting tired of maintaining this legacy codebase!

I will update the API docs to clarify that this use case is supported.

alistaire47:
Thanks for the history; apologies for the confusion! I'm good with this then.

austin3dickey:
Okay, updated API docs with the _current_ state of how we handle both keys being sent at once (which is: we keep both, and don't compare results-with-errors to results-without-errors). If we think the current state should be changed, I think #813 could be a good place to discuss that, or else we could open a new issue.

Issue #570, opened 2023-01-03
Make `get_baseline_run()` `commit_limit` tunable by server


austin3dickey:
Add a `BASELINE_RUN_COMMIT_LIMIT` config option. See comment for more details.

_Originally posted by @jonkeane in https://github.com/conbench/conbench/pull/507#discussion_r1059467454_

Issue #569, opened 2023-01-03
Iterate get_baseline_run() with larger commit_limits


austin3dickey:
If `get_baseline_run()` doesn't return anything with a given `commit_limit`, try again with a bigger `commit_limit`. See the comment below for considerations.

_Originally posted by @jonkeane in https://github.com/conbench/conbench/pull/507#discussion_r1059467454_

PR #1078, opened 2023-04-10
api/run: return more candidate baseline runs

### LINKED ISSUE #384: Allow me to compare different baselines

austin3dickey:
Closes #384.

This PR fleshes out the baseline run information returned from `/api/runs/<run_id>`, providing three different candidates for the baseline run.


conbench/entities/run.py
```diff
@@ -249,6 +247,30 @@ def get_default_baseline_run(
...
+    def get_candidate_baseline_runs(self) -> Dict[str, Optional[_FoundBaselineRun]]:
+        """Return information about a few different candidate baseline runs, including
+        on the parent commit, fork-point commit, and head-of-default-branch commit.
+
+        See docstring of _get_baseline_run_from_baseline_commit() for how these are
+        found.
+        """
+        return {
+            "parent": self._get_baseline_run_from_baseline_commit(
+                self.commit.get_parent_commit()
```
jgehrcke: This is always _the_ direct, single parent in the commit graph, right?

conbench/entities/run.py
```diff
@@ -249,6 +247,30 @@ def get_default_baseline_run(
...
+            ),
+            "fork_point": self._get_baseline_run_from_baseline_commit(
+                self.commit.get_fork_point_commit()
```
jgehrcke, line 263: Is this always a commit on the default branch?

conbench/entities/run.py
```diff
@@ -249,6 +247,30 @@ def get_default_baseline_run(
...
+            ),
+            "head_of_default": self._get_baseline_run_from_baseline_commit(
+                Session.execute(
+                    s.select(Commit)
+                    .filter(Commit.sha == Commit.fork_point_sha)
+                    .order_by(s.desc(Commit.timestamp))
+                    .limit(1)
```
jgehrcke: `HEAD` is a moving target, and depends on perspective.

This here is the "newest commit on the default branch that Conbench has gotten a result for"?

jgehrcke:
Cool! I have left three questions that I think you immediately know the answer to :-). But when I read the part of the code where I noted those questions down I indeed was not 100 % sure.

conbench/entities/run.py
austin3dickey: Yes!

conbench/entities/run.py
austin3dickey: Yes!

conbench/entities/run.py
austin3dickey: Yes! :) perhaps "head_of_default" is the wrong phrase.

austin3dickey:
>But when I read the part of the code where I noted those questions down I indeed was not 100 % sure.

Thanks for adding the questions! Good to know what I'll need to put in docs. I'll add clearer (dev- and user-facing) documentation later.

conbench/entities/run.py
alistaire47: If the contender is on the default branch, this will return itself, right? I think we talked about returning null instead; do we want to do that here? Or pop off all invalid/silly baselines (e.g. head of default (or whatever we call it) to itself)?

austin3dickey: Yes, `fork_point` should be null when the contender run is on the default branch.

Not sure about `head_of_default` when the contender run is on the head of default. Yeah, I suppose we null out any where the contender ID and baseline ID are the same.

austin3dickey:
Okay, I'm happier with where this is at. I tried it for [this Arrow PR run](https://conbench.ursa.dev/runs/85dbee1cb0de40f995592f4d1963c2da/) and got
```
{'candidate_baseline_runs': {'fork_point': {'baseline_run_id': 'b1989ce7e30040c6bf23bd3ac7f483e2',
                                            'commits_skipped': [],
                                            'error': None},
                             'head_of_default': {'baseline_run_id': '85dbee1cb0de40f995592f4d1963c2da',
                                                 'commits_skipped': ['7376b33c03298f273b9120ad83dd05da3d0c3bef',
                                                                     '7ea2d9867cb90f9d0ca1db9b3bffa2fe8426e8ed',
                                                                     'f174ef349ae6c2367b7dbe3f241efdb765e6f1a1',
                                                                     'd5866ec3fa3cff4dbceb5d7838039485cff101c5'],
                                                 'error': None},
                             'parent': {'baseline_run_id': 'b1989ce7e30040c6bf23bd3ac7f483e2',
                                        'commits_skipped': [],
                                        'error': None}}
```
which seems in line with what I expected! 

It does take 5.3 seconds to "hit the API" (using my local setup) vs 1.8 seconds on the `main` branch, but no one really hits this endpoint in bulk, so I'm not worried about this slowing down anything important.

I just need to fix tests and update API documentation. Review welcome in the meantime.

austin3dickey:
Here's for [this run](https://conbench.ursa.dev/runs/85dbee1cb0de40f995592f4d1963c2da/) on the Arrow default branch:
```
{'candidate_baseline_runs': {'fork_point': {'baseline_run_id': None,
                                            'commits_skipped': None,
                                            'error': 'the contender run is on '
                                                     'the default branch'},
                             'head_of_default': {'baseline_run_id': None,
                                                 'commits_skipped': None,
                                                 'error': 'the contender run '
                                                          'is on the default '
                                                          'branch'},
                             'parent': {'baseline_run_id': '598b44a1a3c94c63a4b17330c82c899e',
                                        'commits_skipped': [],
                                        'error': None}}
```
This one only took 2.8 seconds due to skipping two of the searches.

conbench/entities/run.py
jonkeane: > `fork_point` should be null when the contender run is on the default branch.

I super strongly agree with this. I know it's possible to get in the git API, but it's definitely confusing + will lead to people being confused what we mean and what it's useful for.

For `head_of_default` IMO we should return whatever is the head of default (that has benchmark results) regardless. Even if it's the same ID, that's a decent signal that the thing you're looking at is the latest-and-greatest. It's also totally fine that this value will change over time since it is truly intended to track `HEAD`. So sometimes the run id might be the same for `self` and `head_of_default` but 1 hour later, `head_of_default` could be different. 

>  I suppose we null out any where the contender ID and baseline ID are the same.

I think this is trying to find a rule when we don't need one. `fork_point` is null when you're on the default branch cause there was no fork, we don't need to find another reason than that IMHO

jonkeane:
Thanks for this, it is looking good!

> I'll add clearer (dev- and user-facing) documentation later.

I presume you mean later [but still in this PR], right?

For the 5.3 / 2.8 second response times: I'm curious if you've profiled these to see what's taking so long. Even the main branch response time seems long — but 5.3 or 2.8 both seem surprisingly long for this. We (probably) don't need super fast performance on this like you note (we don't query this in bulk) but we should still see if we can figure out why this is taking longer than very quick — we might need to store this stuff in a better way.

conbench/entities/run.py
austin3dickey: I renamed to `latest_default`. Does that make more sense or is it more confusing?

conbench/entities/run.py
austin3dickey: >For head_of_default IMO we should return whatever is the head of default (that has benchmark results) regardless.

Something that came up while I was working on this: do we still want to return a `head_of_default` baseline for contender runs that are on the default branch already? Like on this picture, if you're asking for candidate baseline runs of contender run 4:
image
Obviously, for a contender on 4, then `parent` baseline is 3 and `fork_point` baseline is `null`. But do we really want to say the `head_of_default` baseline of contender run 4 is actually 5? If someone ends up hitting `/api/compare/5...4` for that comparison analysis, then when calculating the z-scores of 4, 4's data will be used as part of the baseline distribution when calculating z-scores (as well as the contender data point), which seems like a smell to me. Instead, I was thinking we should have the `head_of_default` baseline be `null` for default-branch commits too.

Not super tied to that decision.

>I think [nulling out any where the contender ID and baseline ID are the same] is trying to find a rule when we don't need one.

Yeah, I changed the code so this isn't the case. The current code is more aligned to what you're thinking, I think.

austin3dickey:
>I'm curious if you've profiled these to see what's taking so long

I did some basic profiling, and most of the time is spent waiting for the query to execute. I didn't delve into the query plan this time but suspect it's similar to [this one](https://github.com/conbench/conbench/issues/745#issuecomment-1442363191). Yes, our database definitely deserves a second look for how we can optimize this type of query.

conbench/entities/run.py
```diff
@@ -123,101 +143,150 @@ def create(data) -> "Run":
...
-        If there are multiple matches, prefer a Run with the same reason as this Run,
-        and then find the latest commit, finally tiebreaking by latest Run.timestamp.
+    def _search_for_baseline_run(
+        self, baseline_commit: Optional[Commit], commit_limit: int = 20
+    ) -> _CandidateBaselineSearchResult:
+        """Search for and return information about a baseline run of this (contender)
+        run, where the baseline run:
+
+        - is on the given baseline_commit, or in its git ancestry (up to
+          ``commit_limit`` commits ago)
```
austin3dickey, line 153: related: #569, #570

conbench/entities/run.py
```diff
@@ -123,101 +143,149 @@ def create(data) -> "Run":
...
+        # The latest commit on the default branch that Conbench knows about
+        if self.commit and self.commit.sha == self.commit.fork_point_sha:
+            candidates["latest_default"] = _CandidateBaselineSearchResult(
+                error="the contender run is already on the default branch"
+            )
+        else:
+            # Query the DB for the latest commit
+            query = s.select(Commit).filter(Commit.sha == Commit.fork_point_sha)
+
+            # TODO: how do we filter by repository if there's no commit?
```
austin3dickey, line 275: This relates to @jonkeane 's recent comment: https://github.com/conbench/conbench/pull/1134#issuecomment-1518240762

To support multi-repo Conbenches, we may have to figure out how to submit a run with no commit hash but still a repository, to fully power this `latest_default` candidate baseline run logic. That's not critical right now because it will work as expected for all single-repository Conbench instances (we'll still grab the latest run on the default branch). But if there's a multi-repository Conbench instance, we might grab the latest run on _any_ default branch, which could be confusing.

austin3dickey:
This is ready for a final review; I've cleaned up a lot since your last reviews (thank you for those)!

In particular before I merge, I want feedback on 
- user-facing naming
- [this question](https://github.com/conbench/conbench/pull/1078#discussion_r1174160609) about whether we should return a `latest_default` baseline if the contender is already on the default branch

conbench/entities/run.py
jonkeane: That sounds good to me. It's slightly ambiguous if it's "the latest commit" or "the latest commit with benchmarks" but I doubt we can get out of that without a lot more words

conbench/entities/run.py
jonkeane: If we're going to leave this as is, please make an issue to come back and clean it up (and we should either prioritize it or clean this up to emit something less seemingly random if someone hits it)

austin3dickey: Cool, made https://github.com/conbench/conbench/issues/1165

conbench/entities/run.py
austin3dickey: Changed this in https://github.com/conbench/conbench/pull/1275.

PR #1079, opened 2023-04-11
fmt_unit: minimum sigfig count (fix #683, introduce sigfig dependency)

### LINKED ISSUE #683: history plot in comparison view: min higher than mean
Seen in https://conbench.ursa.dev/compare/benchmarks/a3faf3e6c7f048dab960f24f58618aa0...77aef10139db4f0a8376e1baa93f06b0/:

image

There are a couple of issues here, but two of them stand out:
- the seemingly _constant_ nature of the mean-over-time is suspicious.
- the min-over-time is greater than the mean-over-time is mathematically unsound.

We should investigate this information flow end-to-end and identify what's happening why.

Need to look at everything, e.g. "Which entity calculates the mean? Is it Conbench or is it the submission tooling?".

jgehrcke:
This addresses #683.

I have proposed using sigfig before ([here](https://github.com/conbench/conbench/pull/538)), and this PR now adds it as a dependency. It will come in handy in many cases, in particular when populating UI. Sigfig docs: https://sigfig.readthedocs.io/en/latest/api.html

This patch also adds/tweaks code comments here and there.

### Before:

Screenshot from 2023-04-09 14-43-23

### After:

Screenshot from 2023-04-09 14-45-31


jgehrcke:
The test suite adjustments taught me more about where this function is being used. I didn't really expect this to be used for generating JSON output. 

This was unexpected:
```
E         -   'contender_z_score': 0,
E         +   'contender_z_score': 0.0,
```

We need to think this through much better. For now, I make all  test suite adjustments ignorantly, to get things passing again.

jgehrcke:
Limiting the number of significant figures says something about the relative change that we care about. Here, I have provocatively chosen `4`. This method affects small numbers in the same way as it affects large numbers.

People often look at large numbers differently than they look at small numbers, but that's just a human fallacy I think:
- truncation from `0.001234567` to `0.001234` is expected
- truncation from `123456789` to `123400000` maybe seems unexpected

It's probably fair to say (for example): when limiting the number of significant digits to six, then one can only resolve relative change larger than 1 ppm.

Future:
- maybe we want to set this to 5 or 6 or 7, or maybe we want to never do sigfig-based truncation when it affects digits before the decimal separator.
- we certainly should review our method of returning numbers in JSON documents in the JSON API.

austin3dickey (approved):

jgehrcke:
This patch is part of my work for a "conceptual benchmark list PoC/MVP". I appreciate review (I think we will have quite a bit of discussion around this topic in the future) after merge! :ship:.

alistaire47:
This is interesting, and moving us in a good direction. Two thoughts:

1. To state what's hopefully obvious: The right number of sigfigs depends on the most-stable benchmark—one that takes a non-trivial amount of time (so the resolution is not dominated by floating point error) but a very consistent amount of time, e.g. one with an iteration interquartile (say) range of [1.23456 s, 1.23459 s] would need a resolution of at least 7 sigfigs to be useful. I think we could query this? Though it's some work.
2. I'd love for us to move towards a consistent policy on sigfigs, so these formats do mostly nothing because the numbers were already well-rounded before posting. I don't think this is hard; it's just adding similar transforms to [benchadapt.BenchmarkResult.stats](https://github.com/conbench/conbench/blob/6066bc3e7827d9b89600e4b2819550378904fac0/benchadapt/python/benchadapt/result.py#L129), which should really be a nested dataclass. Maybe it's time to switch to pydantic there.

PR #1080, opened 2023-04-11
app: remove 'dated url' hook

jgehrcke:
See commit message. This removes a source of bugs (complexity, confusion -- cache invalidation :fire:)


jgehrcke:
This patch is part of my work for a "conceptual benchmark list PoC/MVP". I appreciate review (I think we will have quite a bit of discussion around this topic in the future) after merge! :ship:.

PR #1065, opened 2023-04-06
app: gunicorn: central cfg, single-process N threads

jgehrcke:
See issue #1018.

One aspect that's maybe buried in #1018 and related discussion: this cuts the memory footprint of Conbench by a significant amount (let's say factor ~5, for simplicity). This will become relevant in the context of #1036.

This also centralizes more of the gunicorn config into a single place, for less duplication (see diff).

New process hiearchy during `make run-app-dev`:
```
root     2867775  0.0  0.0 720172 10240 ?        Sl   11:43   0:00 /usr/bin/containerd-shim-runc-v2 -namespace moby -id be174a1b07d567862bdb7c630686f73c288eea86b970089c11e5ec4de07f2545 -address /run/containerd/containerd.sock
root     2867802  6.2  0.1  44116 37240 ?        Ss   11:43   0:00  \_ /usr/local/bin/python /usr/local/bin/gunicorn -c conbench/gunicorn-conf.py --reload
root     2867871 54.5  0.4 970916 135624 ?       Sl   11:43   0:02      \_ /usr/local/bin/python /usr/local/bin/gunicorn -c conbench/gunicorn-conf.py --reload
```

This shows one observing gunicorn process, and one observed worker process. Good.

PR #1085, opened 2023-04-11
webapp: add benchmark result tail (BMRT) cache

jgehrcke:
Required for #1086.

Implement a simple job which populates and provides a cache for the N most recent benchmark results (the tail end of what was inserted).

Update periodically.

This query/update logic runs in its own thread. Data is shared with other threads (running HTTP handlers) via canonical CPython dictionaries (which have a thread-safe interface).

Makes use of the fact that since #1065 we run only one gunicorn worker process per webapp container.

The cache will be used by a proof of concept UI section that is all about "listing conceptual benchmarks" (their results, grouped by cases, hardware, context). These pages cannot be built in a snappy way when i) basing them on a meaningful amount of data and ii) serving everything freshly from the DB (at least not trivially with the current database schema).

To be clear; for starters this is supposed to be used only in the context of the new UI section.

This entire idea is based on the premise that for some parts of the UI it is okay to update every few minutes (we should discuss and observe how much 'freshness' there really needs to be).

The query for 2*10^5 items takes ~10 seconds on my machine against (a local copy of) the Arrow DB. This cache uses (quite a bit of) RAM. Start conservatively. On my laptop with 32 GB RAM I could put 10^6 items into that cache easily, but we run this as of today with memory constraints.

This is a tiny starting point. Some thoughts:
- fetch in separate process, then share via e.g. SHM
- use `__slots__` to reduce mom footprint, and/or build lighterweight objs
- only add not-yet-known items to cache, full rebuild less often (with a frequent rebuild we do not think 'hard' about cache invalidation).

Some terminology: benchmark result tail-end (BMRT)

### DIFF
--- a/conbench/api/results.py
+++ b/conbench/api/results.py
@@ -5,6 +5,8 @@ import flask_login
 import orjson
 from sqlalchemy import select
 
+import conbench.job
+
 from ..api import rule
 from ..api._docs import spec
 from ..api._endpoint import ApiEndpoint, maybe_login_required
@@ -17,8 +19,6 @@ from ..entities.benchmark_result import (
 )
 from ..entities.case import Case
 from ._resp import json_response_for_byte_sequence, resp400
-
-import conbench.job
 
 conbench.job.start_jobs()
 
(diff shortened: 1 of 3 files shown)


jgehrcke:
Going ahead with this  :ship: then continuing with #1086. 
Feedback appreciated.

austin3dickey (commented):
Excited to see how this plays out!

PR #1093, opened 2023-04-11
metrics: remove multiproc support, add GAUGE_BMRT_CACHE_LAST_UPDATE_SECONDS

jgehrcke:
This is a nice simplification after enabled by https://github.com/conbench/conbench/pull/1065.

The new metric is a valuable thing to observe (I also think this is a great example for how easy it is to add a custom application metric given the new building blocks we have for that).


jgehrcke:
Going ahead with this cleanup :ship:

austin3dickey (commented):
ooh this is a good one

PR #1094, opened 2023-04-11
api/compare: Rename /api/compare/benchmarks to /api/compare/benchmark-results

### LINKED ISSUE #1090: api/compare: Rename /api/compare/benchmarks to /api/compare/benchmark-results
Splitting out of #1075 to keep PRs tidy and small. Rename /api/compare/benchmarks to /api/compare/benchmark-results. Clean up variable names and such along the way.

alistaire47:
Closes #1090 with lots of `benchmarks` -> `benchmark-results`.

~I went further than the story intended on the app side when on a roll trying to unbreak things; I don't know if we're relying on the stability of those URLs anywhere, so unless anybody can confidently say "no", I need to roll back that change before merging.~

Edit: done; left the internal changes.

### DIFF
--- a/conbench/cli.py
+++ b/conbench/cli.py
@@ -8,8 +8,8 @@ from .util import register_benchmarks
 
 register_benchmarks()
 BENCHMARKS = {}
-for benchmark_result in REGISTRY:
-    BENCHMARKS[benchmark_result.name] = benchmark_result
+for benchmark in REGISTRY:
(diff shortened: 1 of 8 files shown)


conbench/cli.py
alistaire47, line 1: wait, not sure why these got changed; I think the vscode refactoring went a little overboard? I'll revert; for once these actually aren't results

jgehrcke (approved):
:rocket:! Nice.

PR #1102, opened 2023-04-12
BMRT cache: pre-fetch hardware per result object; per-hwctx plot: require at least N data points

jgehrcke:
After observing the new benchmark list UI  and BMRT cache behavior in staging, some immediate follow-ups:

- I found that when there are many results per benchmark name then the loop that iterates over all results and looks up `result.run.hwardware` for each result takes to long. Query runs via join, too, and store hardware info on result objects (pragmatic solution).
- For a huge result count per benchmark name, and also a rather big case/hardware permutation space it absolutely makes sense to only plot for at least N data points.
- when displaying many per-hwctx plots it makes sense to show the most populated plots first

### DIFF
--- a/conbench/app/benchmarks.py
+++ b/conbench/app/benchmarks.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import time
 from collections import defaultdict
 from typing import Dict, List, Tuple, TypedDict
 
@@ -29,21 +30,23 @@ def list_benchmarks() -> str:
 @app.route("/c-benchmarks/<bname>", methods=["GET"])  # type: ignore
 def show_benchmark_cases(bname: str) -> str:
     try:
-        results = _cache_bmrs["by_benchmark_name"][bname]
+        matching_results = _cache_bmrs["by_benchmark_name"][bname]
     except KeyError:
         return f"benchmark name not known: `{bname}`"
 
     # cases = set(bmr.case for bmr in results)
 
     results_by_case = defaultdict(list)
-    for r in results:
+    for r in matching_results:
         results_by_case[r.case].append(r)
 
+    t0 = time.monotonic()
     hardware_count_per_case = {}
     for case in results_by_case.keys():
         # The indirection through `.run` here is an architecture / DB schema
         # smell I think. This might fetch run dynamically from the DB>
-        hardware_count_per_case[case] = len(set([r.run.hardware for r in results]))
+        hardware_count_per_case[case] = len(set([r._hardware for r in results]))
+    log.info("building hardware_count_per_case took %.3f s", time.monotonic() - t0)
 
     context_count_per_case = {}
     for case, results in results_by_case.items():
@@ -59,7 +62,7 @@ def show_benchmark_cases(bname: str) -> str:
         hardware_count_per_case=hardware_count_per_case,
         context_count_per_case=context_count_per_case,
         # cases=cases,
-        benchmark_result_count=len(results),
+        benchmark_result_count=len(matching_results),
         application=Config.APPLICATION_NAME,
         title=Config.APPLICATION_NAME,  # type: ignore
     )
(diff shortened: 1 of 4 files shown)


conbench/app/benchmarks.py
```diff
@@ -59,7 +64,7 @@ def show_benchmark_cases(bname: str) -> str:
         hardware_count_per_case=hardware_count_per_case,
         context_count_per_case=context_count_per_case,
         # cases=cases,
-        benchmark_result_count=len(results),
+        benchmark_result_count=len(matching_results),
```
jgehrcke, line 67: There was a naming confusion bug here, leading to a wrong count being displayed. (for loop iteration var leaked into out scope, typical Python thing)

conbench/app/benchmarks.py
```diff
@@ -29,21 +30,25 @@ def list_benchmarks() -> str:
...
 
+    t0 = time.monotonic()
+
     hardware_count_per_case = {}
-    for case in results_by_case.keys():
+    for case, results in results_by_case.items():
         # The indirection through `.run` here is an architecture / DB schema
         # smell I think. This might fetch run dynamically from the DB>
-        hardware_count_per_case[case] = len(set([r.run.hardware for r in results]))
+        hardware_count_per_case[case] = len(set([r._hardware for r in results]))
```
jgehrcke, line 49: This `results` is now only those results with the current `case` (for loop iteration var)

conbench/job.py
```diff
@@ -80,6 +86,8 @@ def _fetch_and_cache_most_recent_results(n=0.07 * 10**6) -> None:
     by_case_dict: Dict[str, List[BenchmarkResult]] = defaultdict(list)
     for result in results:
         by_id_dict[result.id] = result
+        # Add property _hardware on result obj, from Run.
+        result._hardware = result.run.hardware
```
jgehrcke, line 90: And this is just lazy JP who wants to be _sure_ that the next access to `result.run` does not issue a db query. This needs to be cleaned up.

conbench/job.py
```diff
@@ -58,15 +59,20 @@ class CacheDict(TypedDict):
 
 # Fetching one million items from a sample DB takes ~1 minute on my machine
 # (the `results = Session.scalars(....all())` call takes that long.
-def _fetch_and_cache_most_recent_results(n=0.07 * 10**6) -> None:
+def _fetch_and_cache_most_recent_results(n=0.06 * 10**6) -> None:
```
jgehrcke, line 62: Also, the mem consumption with this naive approach is rather high, dialing this back slightly.

image

(we can monitor this now because of the kube-prometheus stack giving us so much for free)

conbench/job.py
```diff
@@ -58,15 +59,20 @@ class CacheDict(TypedDict):
...
         "BMRT cache: keys in cache: %s",
         len(_cache_bmrs["by_id"]),
     )
     t0 = time.monotonic()
 
+    # Also fetch hardware from associated Run, so that result.run is in cache,
+    # too, and so that result.run.hardware is a quick lookup.
     results = Session.scalars(
-        select(BenchmarkResult).order_by(BenchmarkResult.timestamp.desc()).limit(n)
+        select(BenchmarkResult)
```
austin3dickey, line 72: Maybe a good TODO: The query will take less time and memory if you only select the fields you'll need later; i.e. something like this. It will also be guaranteed to not hit the DB again later if you access attributes of the result rows.
```py
select(
  BenchmarkResult.id.label("result_id"),
  Hardware.hash.label("hardware_hash"),
  ...,
  Case.name.label("benchmark_name"),
  Case.tags.label("case_tags"),
  Case.id.label("case_id"),
)
```
Ah, but then you'd lose the ability to use the new `@property` methods on the BenchmarkResult object. Maybe it's worth it?

jgehrcke: Right! I also noted some immediate tuning ideas down in https://github.com/conbench/conbench/pull/1085 in the description. The good thing is that we _know_ that a whole lot can be done to make this quite a bit better/smarter in terms of resource usage.

PR #1103, opened 2023-04-12
UI: benchmark list: limit entries for per-hwctx table, use orjson for encoding plot info

jgehrcke:
A few more follow-ups based on observations in the staging environment.


jgehrcke:
Landing this :ship:

Issue #1101, opened 2023-04-12
ci instability: ci/libraries, GitHub API authentication fails with exp too far in the future


jgehrcke:
seein in `_ test_alert_pipeline[app]` (https://github.com/conbench/conbench/actions/runs/4676794473/jobs/8283535688?pr=1099):

```
if http_error_msg:
>           raise HTTPError(http_error_msg, response=self)
E           requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://api.github.com/app/installations/29535722/access_tokens

/opt/hostedtoolcache/Python/3.10.10/x64/lib/python3.10/site-packages/requests/models.py:1021: HTTPError
------------------------------ Captured log call -------------------------------
INFO     benchclients.logging:github.py:151 Attempting to authenticate as a GitHub App.
DEBUG    benchclients.logging:base.py:46 GET https://api.github.com/app/installations params=None
DEBUG    urllib3.connectionpool:connectionpool.py:1003 Starting new HTTPS connection (1): api.github.com:443
DEBUG    urllib3.connectionpool:connectionpool.py:456 [https://api.github.com:443](https://api.github.com/) "GET /app/installations HTTP/1.1" 200 1636
DEBUG    benchclients.logging:base.py:57 POST https://api.github.com/app/installations/29535722/access_tokens {}
DEBUG    urllib3.connectionpool:connectionpool.py:456 [https://api.github.com:443](https://api.github.com/) "POST /app/installations/29535722/access_tokens HTTP/1.1" 201 216
INFO     benchclients.logging:github.py:151 Attempting to authenticate as a GitHub App.
DEBUG    benchclients.logging:base.py:46 GET https://api.github.com/app/installations params=None
DEBUG    urllib3.connectionpool:connectionpool.py:1003 Starting new HTTPS connection (1): api.github.com:443
DEBUG    urllib3.connectionpool:connectionpool.py:456 [https://api.github.com:443](https://api.github.com/) "GET /app/installations HTTP/1.1" 200 1636
DEBUG    benchclients.logging:base.py:57 POST https://api.github.com/app/installations/29535722/access_tokens {}
DEBUG    urllib3.connectionpool:connectionpool.py:456 [https://api.github.com:443](https://api.github.com/) "POST /app/installations/29535722/access_tokens HTTP/1.1" 401 121
[...]

PR #1104, opened 2023-04-12
benchalerts: use a closer JWT expiration time

austin3dickey:
To address #1101. This PR lowers the time before the `benchalerts` GitHub App JWT expires, to reduce the likelihood of that issue happening.

We cannot test this change explicitly, but we'll be able to tell if it worked by monitoring our integration tests to ensure this issue doesn't come up again.

### DIFF
--- a/benchalerts/benchalerts/integrations/github.py
+++ b/benchalerts/benchalerts/integrations/github.py
@@ -66,8 +66,13 @@ class GitHubAppClient(BaseClient):
         payload = {
             "iss": app_id,
             "iat": datetime.datetime.utcnow() - datetime.timedelta(minutes=1),
-            "exp": datetime.datetime.utcnow() + datetime.timedelta(minutes=10),
+            # Expiration time must be no more than 10 minutes, but using exactly 10
+            # minutes may be flaky due to system clock disagreements (see
+            # https://github.com/conbench/conbench/issues/1101). We only need this token
+            # briefly anyway.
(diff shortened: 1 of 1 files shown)


benchalerts/benchalerts/integrations/github.py
```diff
@@ -66,8 +66,13 @@ def _encode_jwt(app_id: str, private_key: str) -> str:
...
             "iss": app_id,
             "iat": datetime.datetime.utcnow() - datetime.timedelta(minutes=1),
-            "exp": datetime.datetime.utcnow() + datetime.timedelta(minutes=10),
+            # Expiration time must be no more than 10 minutes, but using exactly 10
+            # minutes may be flaky due to system clock disagreements (see
+            # https://github.com/conbench/conbench/issues/1101). We only need this token
+            # briefly anyway.
+            "exp": datetime.datetime.utcnow() + datetime.timedelta(minutes=3),
         }
+        log.debug("Payload to encode for JWT: %s", str(payload))
```
jgehrcke, line 76: Oh, we sign our own token here.

Which entity is the authenticator (the entity that will cryptographically validate this token)?

>  Expiration time must be no more than 10 minutes

Where does this constraint come from? Sounds like some docs :)

Terminology: exp is an absolute point in time, a deadline. "10 minutes" is what I typically call the lifetime (life time?) :).

austin3dickey: Good questions; I just clarified the code comment.

PR #1106, opened 2023-04-12
benchclients: Log failed requests

### LINKED ISSUE #807: benchadapt: log request body after request failed finally
Follow-up from #804. 

When a request eventually fails (after retrying) it's incredibly helpful to dump the body, all the details. This saves a lot of precious human time. An advantage of the stateless nature of HTTP is that a request/response cycle is in theory easy to debug -- as long as one has all the details about the request, and the response. Makes sense to keep that in mind!

alistaire47:
Closes #807, which I saw and then lost track of, but which I noticed again today.

Logs exactly the same request string as in `DEBUG` mode as `ERROR` when things blow up.

### DIFF
--- a/benchclients/python/benchclients/base.py
+++ b/benchclients/python/benchclients/base.py
@@ -43,9 +43,12 @@ class BaseClient(abc.ABC):
     def get(self, path: str, params: Optional[dict] = None) -> dict:
         """Make a GET request"""
         url = self.base_url + path
-        log.debug(f"GET {url} {params=}")
+
+        req_string = f"GET {url} {params=}"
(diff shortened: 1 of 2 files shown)


jgehrcke (approved):
Thank you! I had a brief look only. In the details, I think in the mid term we want to re-work this with more hard-learned best practices in mind (log only body prefix, and various other aspects), as a quick improvement this is of course great!

austin3dickey:
I will [push this change](https://github.com/conbench/conbench#to-upload-new-version-of-packages-to-pypi) to PyPI now.

PR #1110, opened 2023-04-13
BMRT cache job: start from within gunicorn hook, add tests, improve signal handling, minor cleanup

jgehrcke:
Add a new test module for the "conceptual benchmark list" views.

Initiate the BMRT cache populuation logic in a gunicorn worker process hook (_after_ having initialized the WSGI app).

That means that simply importing the application object or initializing the Flask app itself does not have the side effect of running this job.

This makes it so that running the test suite with the pytest test runner does not initiate the job.

With that, we can remove the `if Config.TESTING:` bandaid again. The result is that the job starts when running `make run-app-dev` locally (desired outcome). I think we now could also remove the `CREATE_ALL_TABLES`-based check, but it's OK to stay for a little longer.

This patch also invokes the original signal handler for SIGINT which on CPython is to raise KeyboardInterrupt.

### DIFF
--- a/conbench/api/index.py
+++ b/conbench/api/index.py
@@ -36,7 +36,6 @@ def docs():
     # dictionary may be different for each application startup. Sort keys
     # alphabetically to get a stable outcome.
     for schemaname in sorted(spec.components.schemas.keys()):
-        print(schemaname)
         mdchunks.append(
             f"## {schemaname}\n"
             f'<SchemaDefinition schemaRef="#/components/schemas/{schemaname}" />\n'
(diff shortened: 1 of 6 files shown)


conbench/api/index.py
```diff
@@ -36,7 +36,6 @@ def docs():
     # dictionary may be different for each application startup. Sort keys
     # alphabetically to get a stable outcome.
     for schemaname in sorted(spec.components.schemas.keys()):
-        print(schemaname)
```
jgehrcke, line 39: this was a debugging leftover from whenever :)

jgehrcke:
Using this locally, landing :ship: :rocket: 
Feedback, questions: appreciated!

Issue #1063, opened 2023-04-06
UI: compare view: change table terminology: case permutation vs. batch vs. benchmark name


jgehrcke:
Current:

![image](https://user-images.githubusercontent.com/265630/230313417-b347bcdb-135c-4856-8a63-6447e897befd.png)

Action item: change terms, analogous to #877.

Wider context: #880, #873.

Issue #1030, opened 2023-04-03
api/compare: return basic information about the *actual* baseline (vs requested)


austin3dickey:
Currently, for the lookback z-score analysis that happens when hitting the `/api/compare/*/baseline_id...contender_id` endpoints, we compare the contender results to a distribution of results ending with the baseline commit. But what if there were no matching results associated with the baseline commit? It would be helpful to `benchalerts` (see #969) to report the "actual" baseline commit used for this analysis; that is, the most-recent commit in history that was used. If possible, we should return the number of "skipped" commits too.

jonkeane:
One thing that's always been a little bit funny about these endpoints is that they are serve two slightly different purposes depending on which one we are talking about: 

* `compare/runs/...` This gives you a list of benchmarkresults that have matches in both runs. There's no real guessing or redirection here. Either the benchmarkresults exist in both runs or they don't  
* `compare/batches/...` Does very similar to runs, just at the batch level (e.g. a small handful of benchmarkresults)
* `compare/benchmarks/...` Compares two benchmark results directly. Typically these match, but you could in principle actually compare across different ones (so long as you construct the URL yourself right now https://conbench.ursa.dev/compare/benchmarks/53cfcb86f6034973aa8baf15662fef73...d2fc1b7cbc794a52afa1547b0434a9cb/). But you get the comparison that you ask for.
* `compare/commits/...` Is the odd one. This doesn't actually give you a comparison, but rather it gives a set of _possible_ comparisons that one could then iterate over e.g. https://conbench.ursa.dev/api/compare/commits/0e1f5c6c2ce7504d86dd10d91ab7d29c2441d4ce...efb0be0d146715663fbfe9162f91da2e8e05b58f/ notably this is also the only one of these that doesn't have a corresponding UI page (it's API only).

Ultimately, there are two activities here: doing a comparison and finding reasonable comparisons to do. All but `compare/commits/...` do the comparison given, where as `compare/commits/...` is finding reasonable comparisons to do and giving you some URLs where those comparisons could be found. We should probably split these two use cases up (or at least document them much better than they are now, since it's confusing!)

The thrust of this issue though is IMO really only something we should do for the finding reasonable comparisons to do use case only. The _doing the comparison_ use case should give you what you ask for even if it's absurd or results in `null` (and only once we have that entirely sorted + are doing it correctly, maaaaaybe we consider sending someone to a disambiguation page or something — but this is not a requirement or feature we need to support right now).

[...]

PR #1114, opened 2023-04-13
/api/compare: Restructure compare output

### LINKED ISSUE #1091: api/compare: Separate comparison dict into baseline, contender, and analysis sub-dicts
Split out of #1075 to keep PRs tidy and small. In `/api/compare/benchmark-results` and `/api/compare/runs/`, clarify the comparison dict by separating keys into `baseline` (raw info), `contender` (raw info), and `analyses` sub-dicts

alistaire47:
Closes #1091 by restructuring compare output into sub-dicts for `baseline`, `contender`, and `analysis`. The key changes here are in [conbench/entities/_comparator.py](https://github.com/conbench/conbench/compare/edward/restructure-compare-output?expand=1#diff-8d24728314c8f2d80a23a786168da64490c3fa01d26cd0b8da69d750527a0962); essentially everything else is updating all the places where we use compare output (which it turns out is a fair few!).

Related to #1063, this also changes `batch` -> `benchmark_name` and `benchmark` -> `case_permutation` in the `BenchmarkResult` object there to clarify what these things are and reduce confusion.

I've spun this up locally and poked around and didn't notice anything awry, but since this does touch the templates, more eyes would be appreciated.


benchalerts/tests/unit_tests/mocked_responses/GET_conbench_compare_runs_some_baseline_some_contender.json
```diff
@@ -2,81 +2,109 @@
...
+                    "file_type": "parquet",
+                    "input_type": "arrow",
+                    "name": "file-read"
+                }
+            },
+            "analysis": {
+                "pairwise_percent_change": "0.000%",
+                "pairwise_percent_threshold": "5.000%",
+                "pairwise_regression_indicated": true,
+                "pairwise_improvement_indicated": false,
```
jgehrcke: Ha! Where can I read more about `pairwise`? :) I think this might be so important and papertrail-worthy that maybe it goes into its own pull request, so that it's not so buried?

Two questions/comments about "pairwise":
- can you relate this briefly two two 'old ideas' that I have no captured in https://github.com/conbench/conbench/issues/1115?
- I think, I'd love to have one more level in the `analysis` object with the method's identifier, so that i) we do not need these `pairwise_` and `lookback_z_` prefixes, and so that each method can nicely specify its own 'analysis result structure'.

conbench/api/compare.py
```diff
@@ -34,10 +34,8 @@ def _result_dict_for_compare_api(benchmark_result: BenchmarkResult):
         "value": benchmark_result.mean,
         "error": benchmark_result.error,
         "unit": benchmark_result.unit,
-        # TODO: change this property name to reflect the idea of 'case permutation'
-        "benchmark": benchmark_result.display_case_perm,
-        # TODO: change this property name to reflect the idea of 'benchmark name'
-        "batch": benchmark_result.display_bmname,
+        "case_permutation": benchmark_result.display_case_perm,
+        "benchmark_name": benchmark_result.display_bmname,
```
jgehrcke, line 38: :heart:

jgehrcke (commented):
Love this. Thank you!

It's so nice to see more of this conceptual cleanup. I have added questions/proposals, and I hope we can consider 

> have one more level in the analysis object with the method's identifier, so that i) we do not need these pairwise_ and lookback_z_ prefixes, and so that each method can nicely specify its own 'analysis result structure'.

I know this might be a bit painful with with once again quite a bit of manual search/replace etc -- can also be done in a follow-up.

benchalerts/tests/unit_tests/mocked_responses/GET_conbench_compare_runs_some_baseline_some_contender.json
jgehrcke: ah, from https://github.com/conbench/conbench/issues/1030#issuecomment-1494902692

> pairwise compare baseline_result and contender_result

> If we're using /api/compare/benchmarks/contender_benchmark_result_id?target_baseline_commit=*, and there were multiple matching results on the resulting baseline commit, the lookback z-score analysis is fine but the pairwise analysis needs to choose a baseline result. Maybe we'd pick based on matching run reason and tiebreak by time (like we do when trying to find a baseline run)

alistaire47: `pairwise` isn't new, it's just naming what the compare endpoint already does comparing baseline and contender directly to calculate percent change. Austin suggested the naming live, and I liked it given it indicates these numbers are in the context of these two results.

I like the idea of nesting analyses; I'll keep going to do that.

alistaire47:
One more task I'd like to do as part of this: remove the distinction between `formatted()` and `compare()` variants in `_comparator.py`, make all the numbers numbers, and move the formatting to the frontend.

alistaire47:
Ok! Updates:

1. Nested the `analysis` dict into `pairwise` and `lookback_z_score` subdicts per JP's suggestion above
2. Per conversation with Austin, removed `raw` param from endpoint and made all numbers numbers instead of strings and moved formatting to the templates. Because this removed the 3-decimal place truncation but we still presumably don't want to show floating point garbage, I introduced rounding to 4 sigfigs like in #1079.
3. benchalerts github integration test failure still happening; not really sure what's happening

benchalerts/tests/unit_tests/mocked_responses/GET_conbench_compare_runs_some_baseline_some_contender.json
jgehrcke: > Austin suggested the naming live, and I liked it given it indicates these numbers are in the context of these two results.

I like that, too! Cool, thanks for clarifying.

jgehrcke:
>  remove the distinction between formatted() and compare() variants in _comparator.py

Thank you, this is really important consolidation, critical for enabling faster movement in the future.

benchalerts/benchalerts/conbench_dataclasses.py
```diff
@@ -62,10 +62,26 @@ def contender_benchmark_result_info(self) -> List[BenchmarkResultInfo]:
             assert self.compare_link
             return [
                 BenchmarkResultInfo(
-                    name=benchmark_result["benchmark"],
-                    link=self.benchmark_result_link(benchmark_result["contender_id"]),
-                    has_error=bool(benchmark_result["contender_error"]),
-                    has_z_regression=bool(benchmark_result["contender_z_regression"]),
+                    name=benchmark_result["contender"]["case_permutation"]
+                    if "analysis" in benchmark_result
```
austin3dickey, line 66: Haha, nice. This allows for consuming both the "old" and "new" API format. We'll try to get rid of the "old" format as soon as we can, and come back to simplifying this.

benchalerts/tests/unit_tests/mocked_responses/GET_conbench_compare_runs_some_baseline_some_contender.json
```diff
@@ -2,81 +2,117 @@
...
-                "cpu_count": 2,
-                "dataset": "nyctaxi_sample",
-                "file_type": "parquet",
-                "input_type": "arrow",
-                "name": "file-read"
+            "baseline": {
+                "benchmark_name": "file-read",
+                "case_permutation": "snappy, nyctaxi_sample, parquet, arrow",
+                "language": "Python",
+                "value": "0.03637 s",
```
austin3dickey: I haven't gone in and debugged the failing `benchalerts` tests yet, but just noticed these numbers are still strings. You may need to re-do these mocks?

(And yes, I will set up actual integration tests with a CI Conbench API soon! 😬 )

conbench/entities/_comparator.py
```diff
@@ -1,18 +1,15 @@
...
 Z_SCORE = 5.0  # z-score threshold
 
 
-def fmt(value):
-    return "{:.3f}".format(value) if value is not None else None
-
-
-def change_fmt(value):
-    return "{:.3%}".format(value)
+def _round(value) -> float:
```
austin3dickey: Could you add a type hint to the arg? Also I think the return value is optional here

conbench/entities/_comparator.py
```diff
@@ -175,71 +172,54 @@ def tags(self):
...
+            "analysis": {
+                "pairwise": {
+                    "percent_change": _round(to_float(self.change * 100)),
+                    "percent_threshold": to_float(self.threshold),
+                    "regression_indicated": self.regression,
+                    "improvement_indicated": self.improvement,
+                },
+                "lookback_z_score": {
+                    "z_threshold": to_float(self.threshold_z),
+                    "z_score": _round(to_float(self.contender.z_score))
```
austin3dickey, line 217: I think there may be places outside this file that make `z_score` 0.0 if it actually should be missing. I don't think we should ever return 0.0 unless it's actually 0.0 (i.e. all ≥2 historic observations have the same value for some reason). Would you mind going through and ensuring that doesn't happen, and maybe adding a test? (It could be a separate PR)

conbench/tests/api/test_compare.py
```diff
@@ -319,11 +314,11 @@ def test_compare_unknown_compare_ids(self, client):
...
         [
             # result on the fork point commit
-            (5, -2.1857497724635935),
+            (5, -2.186),
             # result on the parent commit
-            (6, -2.8246140882627757),
+            (6, -2.825),
             # result on the head commit of the default branch
-            (8, 0.12045310863100521),
+            (8, 0.1205),
```
austin3dickey, line 321: This is nice.

austin3dickey (commented):
Thank you!!

benchalerts/tests/unit_tests/mocked_responses/GET_conbench_compare_runs_some_baseline_some_contender.json
alistaire47: Oops, I edited most of them, but missed this one; fixed now

conbench/entities/_comparator.py
alistaire47: Made #1128 because that change would have to happen outside the scope of `_comparator.py`, which otherwise is where all the behavioral changes of this PR are

PR #1116, opened 2023-04-14
Dashboard: refine panels, queries, add new panel for BMRT cache update time

jgehrcke:
image

I've made adjustments so that this dashboard works in two different Grafana instances (kube-prometheys, GC) and tested it against three different Conbench deployments (arrow, cb, cb-staging).


jgehrcke:
ci/db-migrations failed with an instability: https://github.com/conbench/conbench/issues/1117

jgehrcke:
:ship:

PR #1119, opened 2023-04-14
WIP: BenchmarkResult submission: re-work validation

jgehrcke:
This touches on a number of tickets and topics. This is work in progress (tests are not yet passing, and I still would like to do a bit of code cleanup), and I think I'd like to split this in to multiple commits / patches after all. But I wanted to get this up for an initial round of feedback/reviews before the weekend. Would appreciate some eyes on this, thanks!

Issue #972, opened 2023-03-29
HTTP API URL path migration  "benchmark" -> "benchmark-result"


jgehrcke:
Similar to #971 (done in #966). Needs to be done carefully; but will be a valuable change, towards consistency and decent semantics:
https://github.com/conbench/conbench/issues/964#issuecomment-1486976661

The title communicates the general idea. This implies `/benchmarks/<id>` ->   `/benchmark-results/<id>` and comparable.

Issue #932, opened 2023-03-22
reject benchmark result when commit information is not consistent with existing run information


jgehrcke:
Recently in https://github.com/conbench/conbench/pull/628 we documented that certain pieces of information may be silently ignored when submitted as part of a benchmark result. Among that may be the commit information, upon mismatch between Run and BenchtmarkResult.

That's of course an important topic.

In the meantime I think we converge to the opinion that we should not silently ignore a mismatch. @austin3dickey from https://github.com/conbench/conbench/issues/864:

> I agree that we should instead reject those payloads if the run/commit details don't match

Let's do that.

jgehrcke:
With https://github.com/conbench/conbench/pull/1123 we now at least require commit hash consistency. In practice, this gets us really far. Closing this for now as 'done'.

(we should do further consistency checking, at least for the repository specifier / URL).

PR #1123, opened 2023-04-17
result validation: run/result commit consistency, more test coverage, misc

jgehrcke:
Main motivation of this PR was to address #932.

Three new result validation criteria:
- benchmark result commit hash matches run commit hash (test added)
- benchmark name set (test adjusted)
- error / stats: at least one is req (test added here)

Other changes:
- add HTTP route `api/benchmark-results/<bmrid>`, for #972.
- new SQLAlchemy query idiom with type inference
- better response validation
- reduce log verbosity during pytest session
- more renaming for sanity: "valid payload" -> "valid result payload"

Some details are broken out of #1119.

### DIFF
--- a/conbench/api/results.py
+++ b/conbench/api/results.py
@@ -294,6 +294,8 @@ class BenchmarkListAPI(ApiEndpoint, BenchmarkValidationMixin):
 benchmark_entity_view = BenchmarkEntityAPI.as_view("benchmark")
 benchmark_list_view = BenchmarkListAPI.as_view("benchmarks")
 
+# Phase these out, at some point.
+# https://github.com/conbench/conbench/issues/972
 rule(
     "/benchmarks/",
     view_func=benchmark_list_view,
@@ -301,6 +303,19 @@ rule(
 )
 rule(
     "/benchmarks/<benchmark_result_id>/",
+    view_func=benchmark_entity_view,
+    methods=["GET", "DELETE", "PUT"],
+)
+
+# Towards the more explicit route path naming":
+# https://github.com/conbench/conbench/issues/972
+rule(
+    "/benchmark-results/",
+    view_func=benchmark_list_view,
+    methods=["GET", "POST"],
+)
+rule(
+    "/benchmark=results/<benchmark_result_id>/",
     view_func=benchmark_entity_view,
     methods=["GET", "DELETE", "PUT"],
 )
--- a/conbench/api/results.py
+++ b/conbench/api/results.py
@@ -14,6 +14,7 @@ from ..entities.benchmark_result import (
     BenchmarkResult,
     BenchmarkResultFacadeSchema,
     BenchmarkResultSerializer,
+    BenchmarkResultValidationError,
 )
 from ..entities.case import Case
 from ._resp import json_response_for_byte_sequence, resp400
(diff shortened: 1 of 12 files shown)


jgehrcke:
Test failure in unrelated test as of
```
E             -             'url': 'https://github.com/org/repo/commit/02addad336ba19a654f9c857ede546331be7b631'},
E             ?                                                                                            ^^^^
E             +             'url': 'https://github.com/org/repo/commit/02addad336ba19a654f9c857ede546331be7aaaa'},
E             ?                                                                                            ^^^^
```
probably a side-effect of not really deep-copying a data structure before mutation.

jgehrcke:
Shipping this :ship:  -- feedback desired, thanks!

conbench/db.py
```diff
@@ -20,10 +21,16 @@
 log.info("psycopg2.__libpq_version__: %s", psycopg2.__libpq_version__)
 
 
+# Pick less log verbosity when
+logfunc = log.info
+if "pytest" in sys.modules:
+    logfunc = log.debug
```
austin3dickey, line 27: haha, isn't this what thresholds are for? So we could set the log level to DEBUG in pytest and INFO otherwise.

But I don't mind this, since it's so limited in scope.

conbench/api/results.py
```diff
@@ -302,6 +306,19 @@ def post(self) -> f.Response:
...
     methods=["GET", "DELETE", "PUT"],
 )
+
+# Towards the more explicit route path naming":
+# https://github.com/conbench/conbench/issues/972
+rule(
+    "/benchmark-results/",
+    view_func=benchmark_list_view,
+    methods=["GET", "POST"],
+)
```
alistaire47, line 316: Created #1130 to implement this in client libraries.

Also, should these be plural? They always have been, but most of them operate on a single result. I tend to type it wrong when I guess.

conbench/api/results.py
```diff
@@ -302,6 +306,19 @@ def post(self) -> f.Response:
...
+rule(
+    "/benchmark=results/<benchmark_result_id>/",
+    view_func=benchmark_entity_view,
+    methods=["GET", "DELETE", "PUT"],
+)
```
alistaire47, line 321: Typo; `=` instead of `-`

jgehrcke: Ouch! Thank you! :notebook:

conbench/db.py
jgehrcke: :) haha yeah kind of everything fits.. maybe this is an inverse threshold problem lol, more or less: during the pytest session we by default set the logger to emit INFO level (not showing DEBUG messages).

This patch makes it so that certain log messages that we want to be emitted in "prod" on INFO level are emitted on DEBUG level during the pytest session, i.e. _not_ showing up during the pytest session by default. :fire: 

Maybe that doesn't make sense at all, but maybe it does.

conbench/api/results.py
jgehrcke: Great question! GET w/o providing a specifier is supposed to return a collection, right?

Lovely philosophical rabbit hole discussion: https://stackoverflow.com/questions/6845772/should-i-use-singular-or-plural-name-convention-for-rest-resources :)

But I think I've grown a preference for plural over the years :shrug:

Issue #1122, opened 2023-04-17
metrics: add counter for ingested benchmark results, label for repo


jgehrcke:
An important perspective for 'usage' of a Conbench instance is the rate of ingested benchmark results. Measure this; and add labels for interesting properties (carefully, let's not add too many dimensions).

jgehrcke:
Done
![image](https://user-images.githubusercontent.com/265630/232688833-a228c073-37e0-45cc-a2c3-dcbba52a3879.png)
(I did cut the legend with sensitive repo name)

PR #1124, opened 2023-04-17
metrics: add conbench_benchmark_results_ingested counter

jgehrcke:
This patch is for #1122. It adds a metric of type counter, and allows for measuring the per-repository ingest rate of benchmark results.

### DIFF
--- a/conbench/metrics.py
+++ b/conbench/metrics.py
@@ -175,7 +175,7 @@ def http_handler_name(r: flask.Request) -> str:
 gauge_gh_api_rem_set = {"set": False}
 
 
-def _periodically_set_q_rem() -> None:
+def periodically_set_q_rem() -> None:
     """
     This function immediately returns after having spawned a thread.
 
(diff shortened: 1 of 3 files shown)


conbench/metrics.py
```diff
@@ -159,32 +169,41 @@ def http_handler_name(r: flask.Request) -> str:
...
-    explicitly _not_ the initialization value.)
+    for a while? Prometheus gauges are somewhat difficult to understand with
+    respect to their initialization state. For us, 0 is a special, allowed
+    value and explicitly _not_ the initialization value.)
     """
 
     def func():
         while True:
             time.sleep(10)
-            if gauge_gh_api_rem_set["set"] - 1:
```
jgehrcke, line 177: I think this was a funny bug.

```
>>> False - 1
-1
```

```
>>> True -1 
0
```


Can't lint tooling warn/error here? Hm. Rarely, it seems intentional to subtract an integer from a boolean.

jgehrcke: > It is legal, because bool is a subclass of int

jgehrcke:
Moving on with this; please leave post-merge review. :shipit:

PR #1125, opened 2023-04-17
_entity.py: change query idiom (use type info), introduce robust get_or_create() functions

jgehrcke:
Main ideas broken out of #1119. 

This patch replaces check-then-create sequences (prone to race condition) with their more robust and correct EAFP-based counterparts: try to create, but expect conflict. Upon conflict, fetch the thing.

Also, change the `first()` method in _entity.py to use a newer SQLAlchemy idiom; with that, the IDE knows the type of the returned object of this method.

We can of course look into implementing the `get_X_or_create()` family of functions in a parent class; and we should. But this is a tiny bit of an effort to do right, with nice type annotations. Can/will do later.

### DIFF
--- a/conbench/entities/case.py
+++ b/conbench/entities/case.py
@@ -13,4 +13,23 @@ class Case(Base, EntityMixin):
     tags: Mapped[dict] = NotNull(postgresql.JSONB)
 
 
+def get_case_or_create(case_dict) -> Case:
+    """
+    Try to create case, but expect conflict (work with unique constraint on
+    name/tags).
+
+    Return (newly created, or previously existing) Case object, or raise an
+    exception.
+    """
+    try:
+        return Case.create(case_dict)
+    except s.exc.IntegrityError as exc:
+        if "unique constraint" in str(exc):
+            # It is known that it exists, otherwise conflict is not precise.
+            c = Case.first(**case_dict)
+            assert c is not None
+            return c
+        raise
+
+
 s.Index("case_index", Case.name, Case.tags, unique=True)
--- a/conbench/entities/case.py
+++ b/conbench/entities/case.py
@@ -2,6 +2,8 @@ import sqlalchemy as s
 from sqlalchemy.dialects import postgresql
 from sqlalchemy.orm import Mapped
 
+from conbench.db import Session
+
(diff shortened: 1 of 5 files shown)


conbench/entities/case.py
```diff
@@ -13,4 +15,24 @@ class Case(Base, EntityMixin):
...
+    """
+    try:
+        return Case.create(case_dict)
+    except s.exc.IntegrityError as exc:
+        if "violates unique constraint" not in str(exc):
+            raise
+
+    Session.rollback()
+    c = Case.first(**case_dict)
+    assert c is not None
```
jgehrcke, line 34: The `assert` statement also enables mypy to infer that this is _not_ None when returning.

conbench/entities/case.py
```diff
@@ -13,4 +15,24 @@ class Case(Base, EntityMixin):
...
+    Return (newly created, or previously existing) Case object, or raise an
+    exception.
+    """
+    try:
+        return Case.create(case_dict)
+    except s.exc.IntegrityError as exc:
+        if "violates unique constraint" not in str(exc):
+            raise
+
+    Session.rollback()
```
jgehrcke, line 32: The rollback is necessary because we got here because the previous transaction failed.

In the future, we can apply an automatic rollback strategy:
https://docs.sqlalchemy.org/en/14/orm/session_basics.html#framing-out-a-begin-commit-rollback-block
https://stackoverflow.com/a/52234206/145400

conbench/entities/context.py
```diff
@@ -12,12 +14,34 @@
 )
 
 
+# Maybe EntityMixin should be a generic class
+# https://mypy.readthedocs.io/en/latest/generics.html#defining-subclasses-of-generic-classes
```
jgehrcke, line 18: https://discuss.python.org/t/a-way-to-typehint-a-return-type-in-parent-based-on-return-type-of-a-child/17020

jgehrcke:
Moving on with this; please leave post-merge review. :shipit:

conbench/entities/case.py
austin3dickey: Nice!

PR #1126, opened 2023-04-17
BenchmarkResult validation: consolidate some code paths, add validation units of work

jgehrcke:
Ideas split out from #1119.

The BenchmarkResult.create() method is more or less unmaintainable. Before we add new logic, we need to make the code easier to read and reason about.

Here, we add three new units of work, and I hope their name reflects quite well what they are doing:

- `validate_run_result_consistency()`
- `validate_and_augment_result_tags()`
- `do_iteration_samples_look_like_error()`

This patch also 
- renames `name` -> `benchmark_name`
- clarifies when a benchmark result is considered "errored"

### DIFF
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -160,7 +160,11 @@ class BenchmarkResult(Base, EntityMixin):
         # This is going to be the structure based on which we will do
         # DB insertion.
         benchmark_result_data = {}
-        benchmark_result_data["error"] = result_error_for_db
+        # Only insert `error` key if there is an error object. In JSON, error:
+        # null still means that the benchmark is errored (at least some
+        # business logic treat it as such.)
+        if result_error_for_db is not None:
+            benchmark_result_data["error"] = result_error_for_db
 
         # Temporary: keep name `data` -- it's unfortunate that we have the name
         # `data` here while also having key(s) in the dict(s) that are called
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -238,7 +238,10 @@ class BenchmarkResult(Base, EntityMixin):
         # Also note: this pulls the `name=xx` key/value pair out of tags.
         name = tags.pop("name")
 
-        case = get_case_or_create({"name": name, "tags": tags})
+        # See https://github.com/conbench/conbench/issues/935,
+        # name is guaranteed to be set.
+        benchmark_name = tags.pop("name")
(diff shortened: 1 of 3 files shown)


jgehrcke:
Interesting. Test failures in 'unrelated test', need some debugging:
```
    def test_compare_different_baselines(
        self, client, baseline_result_id, expected_z_score
    ):
        self.authenticate(client)
        _, benchmark_results = _fixtures.gen_fake_data()
        contender_result = benchmark_results[7]  # on a PR branch
        baseline_result = benchmark_results[baseline_result_id]
        response = client.get(
            f"/api/compare/benchmark-results/{baseline_result.id}...{contender_result.id}/"
        )
        assert response.status_code == 200, response.status_code
>       assert response.json["contender_z_score"] == expected_z_score
E       assert None == -2.1857497724635935
```

conbench/entities/benchmark_result.py
```diff
@@ -130,62 +134,54 @@ def create(userres):
...
-        any_errored_iteration = True
-
         if has_stats:
             # Note(JP): this is the result of marshmallow
             # deserializaiton/validation against the `BenchmarkResultCreate`
             # schema below. This needs better naming. Most importantly, we
             # should make use of typing information derived from the schema.
-            benchmark_result_data = data["stats"].copy()
+            # Merge this on top of the above-defined benchmark_result_data.
+            benchmark_result_data = benchmark_result_data | data["stats"].copy()
```
austin3dickey, line 187: haha, I still have to get used to this notation

austin3dickey (approved):

jgehrcke:
> Test failures in 'unrelated test', need some debugging:

The cause was an unexpected `error` key in a data dictionary with value `None`. Absence of key is treated differently than value `None` (makes sense). Sneaky bug.

The change in `error` state changed the outcome of a z-score comparison, which why the symptom was so far off:
```
>       assert response.json["contender_z_score"] == expected_z_score
E       assert None == -2.1857497724635935
```

conbench/entities/benchmark_result.py
jgehrcke: >  I still have to get used to this notation

Me, too! It's only the second time I have used that. But most times before in my Python life when I did `d.update(d2)`, I felt that I don't want to modify an existing dictionary in-place, and there must be a nicer way to merge dictionaries into a new object. I think I'll get used to the pipy pipe | : )

jgehrcke:
OK, all the bumps are seemingly evened out (relying on the test suite here very much!!). Let's land this, and move on to the next patch in this series. :ship:

Issue #533, opened 2022-12-13
input validation: require consistency between `data`, `iterations`, and aggregates (`stdev` etc)


jgehrcke:
That is interesting:

![image](https://user-images.githubusercontent.com/265630/207323398-b55e4903-cb33-4a72-8208-d42e93b2ccf3.png)

On the one hand, this claims a large amount of iterations having been performed. On the other hand, only one sample was provided, resulting in meaningless `max`, `min`, `stdev`, etc. 

What does `iterations=5436` mean in this case, to you? I read this as "we did a certain thing 5000 times, and here's what we got: one number, the total duration". Do I understand right? I think in this case the recipient needs not to care about the fact that something was repeated many times, from a statistical point of view. In this specific case, I feel like the `5436` may be useful information about the benchmark method, but more in the metadata/context of the benchmark.

Maybe require all samples to be submitted when `iterations > 1`, and none of the statistical aggregates (can be calculated from the raw samples).

Or, for compatibility, allow for submitting the statistical aggregates 'manually' derived from `iterations > 1`, but then require `data` to be `null/None`.

jgehrcke:
I see in `BenchmarkResultCreate` that `iterations` is documented with

> Number of iterations that were executed (should be the length of data and times)

So, the "should be" is there. Is there more value than 'harm' to not requiring this (super open to learning about use cases)?

alistaire47:
This is a microbenchmark, not a macrobenchmark, and these get a bit messy. These run for a set amount of time, and measure the iterations and/or throughput in that time. For these, there's often also a `times` vector supplied as well, and each has a unit attached (`unit` and `time_unit`, here `b/s` (bytes per sec) and `i/s` (iterations per sec)). So `data` is here really throughput, and `times` is (despite the name) really speed (iterations/s). The actual number of iterations here is fairly meaningless; it only makes sense in the context of the time for which it was run. Really we probably want that number to be 1 here; what we call iterations in macrobenchmarking are frequently called "repetitions" (i.e. the number of times it was run) in microbenchmarks.

But also we want to handle multiple metrics like this better; we can store a `times` vector, but we're not doing anything with it, and we can't store any other metrics being collected.

jonkeane:
@alistaire47 is spot on about microbenchmarks + the way iterations are (typically) derived there. 

> The actual number of iterations here is fairly meaningless; it only makes sense in the context of the time for which it was run. Really we probably want that number to be 1 here; 

[...]

PR #1127, opened 2023-04-17
POST result: deeper stats validation, make self-calculated aggregates take precedence

jgehrcke:
More ideas from #1119.

In the meantime we have added a new test currently confirming non-ideal behavior: https://github.com/conbench/conbench/pull/1131

This patch changes tests to keep them passing; I did it in this order so that the change in behavior is more evident.


conbench/entities/benchmark_result.py
```diff
@@ -507,7 +459,115 @@ def ui_hardware_short(self) -> str:
...
+        if stats_usergiven["iterations"] > 1:
+            # Note(JP): Do not yet handle this in a special way, but I think
+            # that this here is precisely the one way to report the result of a
+            # microbenchmark with cross-repetition stats: data: length 1 (the
+            # duration of running many repetitions), and iterations: a number
+            # larger than 1, while mean, max, are now set (then these can be
+            # assumed to be derived from _within_ the microbenchmark. Note that
+            # this is a very special condition). The mean time for a single
+            # repetition now is sample/iterations.
+            ...
```
jgehrcke, line 557: While I was here, I thought that this is a special condition. Not changing behavior yet.
@alistaire47's thoughts:  https://github.com/conbench/conbench/pull/1119#discussion_r1167300853

conbench/entities/benchmark_result.py
```diff
@@ -507,7 +459,115 @@ def ui_hardware_short(self) -> str:
...
+
+    if len(samples) == 1:
+        if stats_usergiven["iterations"] == 1:
+            # If user provides aggregates, then it's unclear what they
+            # mean.
+            for k in ("q1", "q2", "mean", "median", "min", "max", "stdev", "iqr"):
+                if stats_usergiven.get(k) is not None:
+                    raise BenchmarkResultValidationError(
+                        f"one data point from one iteration: property `{k}` "
+                        "is unexpected"
```
jgehrcke, line 545: this is new behavior, not yet covered by tests

conbench/entities/benchmark_result.py
```diff
@@ -507,7 +459,115 @@ def ui_hardware_short(self) -> str:
...
+            # this is a very special condition). The mean time for a single
+            # repetition now is sample/iterations.
+            ...
+
+    if len(samples) == 2:
+        # If user provides aggregates, then it's unclear what they mean.
+        for k in ("q1", "q2", "mean", "median", "min", "max", "stdev", "iqr"):
+            if stats_usergiven.get(k) is not None:
+                raise BenchmarkResultValidationError(
+                    f"with two provided data points, the property `{k}` "
```
jgehrcke, line 564: slightly different error message compared to above for 1 sample

conbench/entities/benchmark_result.py
```diff
@@ -507,7 +459,115 @@ def ui_hardware_short(self) -> str:
...
+        aggregates = {
+            "q1": q1,
+            "q3": q3,
+            "mean": np.mean(samples),
+            "median": np.median(samples),
+            "min": np.min(samples),
+            "max": np.max(samples),
+            # https://numpy.org/doc/stable/reference/generated/numpy.std.html
+            # This has N in the divisor (ddof=0)
+            "stdev": np.std(samples),
```
jgehrcke, line 517: Note: that's the same method for calculating the standard deviation as before (see above in red part of diff: `stdev = np.std(dat)`).

But, previously the user-given data took precedence. And most of the example data used in the test suite was built with ddof=1, I think (`N-1` in the divisor).

That's where I believe most of the numerical differences come from that we see in the diff in the test suite.

conbench/entities/benchmark_result.py
```diff
@@ -138,168 +142,116 @@ def create(userres):
...
+        # The dict that is used for DB insertion later, populated below.
+        result_data_for_db: Dict = {}
+
+        if "stats" in userres:
+            # First things first: use the complete user-given `stats` object
+            # for potential DB insertion down below. In `error` state, do not
+            # perform deeper validation of the user-given stats object (the
+            # benchmark result is not used for any kind of analysis, which is
+            # why it's probably ok to store the user-given 'stats' object w/o
+            # deeper validation, maybe the data is helpful for debugging). Note
```
jgehrcke, line 154: This is essentially a philosophical decision, and a related ticket is
https://github.com/conbench/conbench/issues/395

conbench/entities/benchmark_result.py
```diff
@@ -138,168 +142,116 @@ def create(userres):
...
+        else:
+            # process_samples_build_agg() must only be called if
+            # do_iteration_samples_look_like_error() returned False. That's
+            # the case here.
+            result_data_from_stats = validate_and_aggregate_samples(userres["stats"])
+
+            # Per-iteration samples looked good, and we did (potentially)
+            # rebuild aggregates. Merge dict `result_stats_data_for_db` on top
+            # of dict `benchmark_result_data`, overwriting upon conflict.
+            result_data_for_db |= result_data_from_stats  # PEP 584 update
```
jgehrcke, line 192: change in behavior around here: the aggregates (mean, stdev, etc...) now _take precedence_ over user-given data.

See below: a warning is logged for mismatch.

For data analysis in Conbench I'd absolutely like to work with the guarantee that if there are N data points and a stdev value that the stdev value is _consistent_ with the data points. Trusting user-given data in that regard doesn't provide that guarantee.

This is very much related to / about https://github.com/conbench/conbench/issues/533

conbench/entities/benchmark_result.py
```diff
@@ -138,168 +142,116 @@ def create(userres):
...
-                ]
-
-                q1 = q3 = mean = median = min = max = stdev = iqr = None
-
-                # calculate stats only if the data is complete. Note(JP): why?
-                # maybe we want to calculate that always if there are at least
-                # N (3?) data points.
-                any_errored_iteration = any([x is math.nan for x in dat])
-
-                if not any_errored_iteration:
```
jgehrcke, line 211: When code grows far to the bottom _right_  it gets harder and harder to read.

So, this entire code section is now split into some units of work that are maybe a little easier to handle, and so that the very high-level business logic is easier to infer.

conbench/entities/benchmark_result.py
```diff
@@ -507,7 +459,115 @@ def ui_hardware_short(self) -> str:
...
+    This returns a dictionary with key/value pairs meant for DB insertion,
+    top-level for BenchmarkResult.
+    """
+
+    # Encode invariants.
+    samples = stats_usergiven["data"]
+    assert len(samples) > 0
+    for sa in samples:
+        assert sa is not None
+        assert sa != math.nan
```
jgehrcke, line 480: Of course such code must go away. This should be locked in with proper type annotations. This was me thinking "OK, I _really_ work with these assumptions, and I am rather confident that they are met. And I really want to know when they are note met".

conbench/entities/benchmark_result.py
alistaire47: When I implemented server-side stats calculations, I believe @jonkeane was against them taking precedence over user-supplied values, but I'm trying to think of a case where they're not what we want and coming up blank. Most cases I can think of have switched to not sending them anyway.

jgehrcke: Thanks for feedback! This demands careful consideration, indeed. I think this is reasonably easy to see through. We are in a code section with two invariants:

1) at least three data points ("samples")
2) each one is a numeric value

From here, we have three options:

1) do not validate consistency between samples and aggregates
2) validate, and upon error reject reject
3) validate, and upon error "take the more correct outcome"

I think (3) is more or less obviously the best option. It is pretty much the same as ignoring the user-provided aggregates (but not generally, only given the two invariants above, which sharpen the scenario that we talk about).

>  I'm trying to think of a case where they're not what we want and coming up blank

I think this means you also support (3), yes? :)

Yes, a user should be to inject arbitrary aggregates, but not when _also_ providing "many" individual samples. In my view, these two should be mutually exclusive. 

We need to draw a clear line between:

a) here are N samples, _you_ do the aggregates
b) here are the aggregates, and I specifically _don't_ give you the underlying raw samples -- 

I think drawing this line can be easy to do, especially when starting from scratch. Given what we have, it takes a bit of time, changing API behavior, and changing API docs to massage things into the direction where we have these two clearly different interface use cases. Here, we work on (a) and make it well-specified. (b) still needs a bit of work, mainly specification work.

>  @jonkeane was against them taking precedence over user-supplied values

I hope this was about implementing (b).

> Most cases I can think of have switched to not sending them anyway.

Thanks for confiming!

jgehrcke:
Going ahead with this :ship: -- I'd love to get further feedback though.

conbench/entities/benchmark_result.py
```diff
@@ -138,168 +142,116 @@ def create(userres):
...
-            userres["stats"]["data"]
-        ):
-            # User did not set `error`, but the number sequence indicates
-            # that this result is to be considered 'errored'. Set a generic
-            # error message.
-            result_error_for_db = {
+            # We have business logic elsewhere that checks only for presence of
+            # the `error` key (ignores its value, a value of `None` might
+            # elsewhere be interpreted as error -- this did cost me 30 minutes
+            # of debugging).
```
austin3dickey, line 166: Ah. Yes, I'd think that we'd want to treat a user-given `{"error": null}` as "okay to use in the lookback for z-scores." We should think about changing this behavior.

conbench/entities/benchmark_result.py
austin3dickey: I found the aforementioned conversation with @jonkeane: https://github.com/conbench/conbench/pull/369#discussion_r966331943

I do believe that conversation was about implementing (b). Google and Folly typically give you stats without exposing the raw data.

FWIW I'm in agreement that 
>(3) is more or less obviously the best option

given the invariants and that we're talking about (a).

conbench/entities/benchmark_result.py
```diff
@@ -507,7 +459,122 @@ def ui_hardware_short(self) -> str:
...
+            "max": np.max(samples),
+            # https://numpy.org/doc/stable/reference/generated/numpy.std.html
+            # This has N in the divisor (ddof=0)
+            "stdev": np.std(samples),
+            "iqr": q3 - q1,
+        }
+
+        # Now, overwrite with self-derived aggregates:
+        for key, value in aggregates.items():
+            result_data_for_db[key] = sigfig.round(value, sigfigs=5)
```
austin3dickey, line 523: Is there a reason you chose 5 instead of 4?

conbench/tests/entities/test_history.py
```diff
@@ -52,7 +52,7 @@ def _get_head_of_default(commit: Commit) -> Commit:
         13.435028842544385,
         1.0928748862317967,
         -2.1857497724635935,
-        -4.486538431438488,
+        -4.486481486904943,
```
austin3dickey, line 55: Hm. I wonder why this changed z-scores?

...Oh. Our [mocked data](https://github.com/conbench/conbench/blob/e1cab9b8ebffaba1c85e454b1b13de7f14bfcf02/conbench/tests/api/_fixtures.py#L257) is posting self-calculated stats. That's not what I expected it to be doing.

conbench/entities/benchmark_result.py
jgehrcke: Thank you for the feedback @austin3dickey! And I really really appreciate that we're connecting the dots here once again (_so_ valuable!).

conbench/entities/benchmark_result.py
jgehrcke: Good question! I didn't think of a good reason when I did this I think, maybe I did and now I don't remember anymore.

conbench/tests/entities/test_history.py
jgehrcke: Yes, unexpected fallout of the good type. 

Tightening things on one end often/always reveals insights on another end.

When building a reasonably complex system, the well-definedness of interfaces and the introduction of strictish rules really becomes the _one_ way to keep things more or less simple, sane, reasonable, productive.

Until recently, in and around Conbench, kind of everything was allowed everywhere as in "yeah, that's OK, you can do that, we'll be fine". That is/was a recipe for (more or less) disaster. I don't want to talk and judge too much, but I am happy that we already see the the fruits of tightening the screws here and there.

Things like that are why every day right now I invest work into the foundation, it will pay off and it will be all-around noticeable rather soon.

conbench/tests/api/test_results.py
```diff
@@ -1138,7 +1166,5 @@ def test_create_result_no_stdev(self, client, samples):
         if len(samples) >= 3:
             assert resp.json["stats"]["stdev"] > 0
         else:
-            # This currently encodes bug
-            # https://github.com/conbench/conbench/issues/802
-            assert resp.json["stats"]["stdev"] == 0.0
-            # assert resp.json["stats"]["iqr"] == 0
+            for k in ("q1", "q3", "mean", "median", "min", "max", "stdev", "iqr"):
+                assert resp.json["stats"][k] is None
```
jgehrcke, line 1170: Oh wow, this breaks history analysis in a really mean way.

conbench/entities/benchmark_result.py
jgehrcke: For the record: `sa != math.nan` is an incorrect comparison method. Always use `math.isnan()`!

PR #1133, opened 2023-04-18
stats: stdev with Bessel's correction

jgehrcke:
We had always used `np.std()`  [in the code path](https://github.com/conbench/conbench/blob/377dc55bedccaa60b7d217d3c657a750de4ffbf2/conbench/entities/benchmark_result.py#L172) where the standard deviation is calculated for an incoming multi-sample result (POST BenchmarkResult).

I did embrace that in #1127. That puts `N` into the divisor when calculating the variance ("the population variance -> the population standard deviation").

For dummy data generation and testing we have however always used the "sample standard deviation", with N-1 in the divisor (Bessel's correction). Example: https://github.com/conbench/conbench/blob/833dfa0b29c1207d82fb0951f5327c4e970acbd5/conbench/runner.py#L442

Also see [docs for statistics.stdev/variance](https://docs.python.org/3/library/statistics.html#statistics.variance) which clarifies that this is based on 'variance with N-1 degrees of freedom'.

That kind of inconsistency is not helping our sanity, because it leads to slightly different numerical results depending on the code path.

Picking between both approaches doesn't really matter in our case in terms of data analysis. But what matters:

- Picking _one_ method to use consistently, by convention (that can be documented, and tested).
- The sample standard deviation (with Bessel's correction with `N-1` in the divisor) might win more nerd battles (well, because it is more 'conservative' by its nature, it produces larger uncertainty measures).
- For the many [1, 2, 3]-value cases in our test suite it yields a nicer number: 1.0 :)

So, I am now updating code comments to make this a little clearer that this was picked intentionally, and I am also updating tests again.

But I also really want to stress that while "using Bessel's correction" sounds smart it really doesn't help analysis-wise, this is like using an :axe: in a :christmas_tree: :deciduous_tree: :evergreen_tree: -- we do not have normally distributed data anyway, and the samples sizes we talk about are so small usually.

A quote supporting the idea that "using Bessel's correction" isn't generally better in terms of analysis here:

> Warne (2017) advocates using Bessel’s correction only if you have a sufficiently large sample and if you are actually trying to approximate the population mean. If you’re just interested in finding the sample mean, and don’t want to [extrapolate](https://www.statisticshowto.com/calculus-definitions/extrapolation-interpolation/) your findings to the population, just omit the correction.

https://www.statisticshowto.com/bessels-correction/ is a cute resource.


jgehrcke:
Going ahead with this, curious about feedback!

austin3dickey:
I like this. In the spirit of 
>Picking one method to use consistently, by convention (that can be documented, and tested)

we do calculate a different standard deviation, in another place: during the lookback z-score analysis.

https://github.com/conbench/conbench/blob/e1cab9b8ebffaba1c85e454b1b13de7f14bfcf02/conbench/entities/history.py#L580

That pandas `.std()` is [documented](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.std.html) to use Bessel's correction by default. We should probably make that explicit and documented too.

jgehrcke:
Thank you for looking carefully, and for connecting the dots @austin3dickey!

Issue #1095, opened 2023-04-12
UI landing page: TypeError while calling `repo_url_to_display_name() `


jgehrcke:
Seen in staging:

```
  File "/app/conbench/app/_endpoint.py", line 36, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/app/conbench/app/index.py", line 77, in get
    rname = repo_url_to_display_name(r.commit.repo_url)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/conbench/app/index.py", line 103, in repo_url_to_display_name
    return result.path.strip("/")
           ^^^^^^^^^^^^^^^^^^^^^^
TypeError: a bytes-like object is required, not 'str'
[230412-06:22:51.913] [7] [conbench] ERROR: Exception on / [GET]
```

followed by a

```
  File "/usr/local/lib/python3.11/site-packages/flask/app.py", line 1727, in handle_exception
    server_error = self.ensure_sync(handler)(server_error)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/conbench/__init__.py", line 183, in _json_http_errors
    resp.data = f.json.dumps(data)
                ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/flask/json/__init__.py", line 124, in dumps
    return app.json.dumps(obj, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/flask/json/provider.py", line 230, in dumps
    return json.dumps(obj, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
          ^^^^^^^^^^^
  File "/usr/local/lib/python3.11/json/encoder.py", line 200, in encode
    chunks = self.iterencode(o, _one_shot=True)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/json/encoder.py", line 258, in iterencode
    return _iterencode(o, 0)
           ^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/flask/json/provider.py", line 122, in _default
[...]

PR #1134, opened 2023-04-18
Tighten API spec with respect to GitHub commit information, adjust tests

jgehrcke:
This was rough, but it felt like someone really has to wade through that.

I wanted to address https://github.com/conbench/conbench/issues/817 because not tackling this does create a tail of follow-up work/drag, e.g. https://github.com/conbench/conbench/issues/1095.

After https://github.com/conbench/conbench/pull/889 I didn't quite have the energy to go further, but now I took a couple of hours.

What this patch does, at the high level:

* Introduce a nullable many-to-one relationship between `Run` and `Commit`, where a Run can be explicitly associated with _no_ commit. This mechanism now replaces the special "no context commit object". We discussed this previously here: https://github.com/conbench/conbench/pull/889#discussion_r1140184252. CC @austin3dickey. 

* Update the API spec for the "github-flavored commit info" object. The client can tell the Conbench API that "I don't have commit information associated with this result" by simply leaving out the `github` key (the obvious choice, IMO). The client is now _required_ to provide meaningful repository URL and commit hash when it includes the `github` key. 

* Introduce a rather tight type for the "github-flavored commit info" in code. Makes use of that. It is a mystery to me why we ever tried to bother with the various "incomplete commit info" permutations deep in code.

* Repair the `lower()`ing of URLs: https://github.com/conbench/conbench/issues/887

* Remove support for `git@...` notation.

In other words: with this patch, when a Run or BenchmarkResult is submitted _and_ when it provides commit information then it is now guaranteed that the commit information contains repository URL and commit hash, and throughout the code it's much clearer what to expect in terms of no-key-present, empty string, null, url-or-not, etc.

Related:
https://github.com/conbench/conbench/blame/2c269052f4476c5707d5517688645ba98dc5ae08/conbench/tests/api/test_results.py#L755
https://github.com/conbench/conbench/issues/888
https://github.com/conbench/conbench/pull/147
https://github.com/conbench/conbench/pull/137

PRs like this are a good demonstration of the state of the code base -- we're slowing getting the upper hand here, but it's still glue, glue, glue :).


conbench/tests/api/_fixtures.py
```diff
@@ -252,6 +274,10 @@ def benchmark_result(
         data["github"]["repository"] = commit.repository
         data["github"]["branch"] = commit.branch
 
+    # do at least a bit of what the HTTP path would do; this ensures that the
+    # output type is TypeCommitInfoGitHub
+    data["github"] = _github_commit_info_schema.load(data["github"])
```
jgehrcke: :neutral_face:

jgehrcke:
Brutal. :fire: 

Conbench checks user-given commit metadata, tries to fetch parent and branch information from GitHub.

When that fetch-from-github fails, a "no context" commit is stored in the database. There is no awareness about parent commit or branch the commit lives on. There only is a commit hash.

The test suite has a `test_get_commit()` test which inserts two benchmark results into the database, with each result referring to a special commit hash. Both are 'somehow' known to be in a parent/child relationship.

The test then uses the /api/commits Conbench HTTP API endpoint to fetch the 'child' commit of both, and expects it to refer to the parent.

That test started breaking and it took me forever to debug this. Now I understand. Here is what happened:

When we execute tests we use a set of special commit hashes.

Our 'production code' inspects each incoming benchmark result for such a magic hash: https://github.com/conbench/conbench/blob/e1cab9b8ebffaba1c85e454b1b13de7f14bfcf02/conbench/entities/commit.py#L653

It then loads the 'commit details' from file, instead of hitting the GitHub API.

But wait, there is another such magic criterion where pull request information is fetched from the GitHub HTTP API. This is served from file for the special pull request number 123456:

https://github.com/conbench/conbench/blob/e1cab9b8ebffaba1c85e454b1b13de7f14bfcf02/conbench/entities/commit.py#L747

-------------------

I started tightening the type situation around the GitHub commit metadata, already a while ago. The pull request number is being input as _string_ into the function `get_branch_from_pr_number()`.

You can consider this as a 'bad type annotation'. I added this a while ago (correction: was introduced [before](https://github.com/conbench/conbench/pull/417/files#diff-8e7df94d5fa6b0082305db878bc76670c59fc020792fb4f78a494d083faaedeaR240)). So far this wasn't productively type-checked. That's a dilemma when tightening types in hindsight.

As of adding more type annotations today in the same module, I ran into a mypy error that told me that I am passing in pr_number into this function as int, but it expected str (as of the type signature).

So, I started passing in the PR number as string, commit is here: https://github.com/conbench/conbench/commit/81a535f2540c5b80a56f037663ee482965fe5cba

That breaks a small set of test in a brutal way. Obviously, `pr_number == 12345678` never evaluates to `True`, then.

That is, instead of using the mocked response for the pull request metadata, Conbench actually emits an HTTP request to `https://api.github.com/repos/org/repo/pulls/12345678`

The response is a 404 response. In the meantime, we have pretty good logging around this business logic:
```
treat as unknown context: error during get_github_commit_metadata(): Unexpected GitHub HTTP API response for URL https://api.github.com/repos/org/repo/pulls/12345678: <Response [404]>.
```
The "treat as unknown context" means that a Commit record was indeed written to the database, but without parent/child and branch information.

As a result, when inspecting one of the two commits (implicitly DB-inserted during BenchmarkResult creation), it doesn't know about the other.

Easy to understand when broken down like this. Sadly, I did not know which of the many changes on my branch produced the problem. Also, I didn't really know of the mechanism with the pull request number.

conbench/api/compare.py
```diff
@@ -169,6 +169,10 @@ def get(self, compare_ids) -> f.Response:
         baseline_benchmark_result = self._get_results(baseline_id)[0]
         contender_benchmark_result = self._get_results(contender_id)[0]
 
+        if baseline_benchmark_result.run.commit is None:
+            # TODO: emit a useful HTTP response
+            raise Exception("baseline commit unknown")
```
jgehrcke, line 174: Interesting outfall of type checking, mypy telling us that this condition wants to be handled (makes perfect sense).

conbench/api/compare.py
```diff
@@ -232,6 +236,10 @@ def get(self, compare_ids) -> f.Response:
         baseline_results = self._get_results(baseline_id)
         contender_results = self._get_results(contender_id)
 
+        if baseline_results[0].run.commit is None:
+            # TODO: emit a useful HTTP response
+            raise Exception("baseline commit unknown")
```
jgehrcke, line 241: same here: mypy telling us that this condition wants to be handled explicitly (makes  sense, too, I think).

conbench/api/results.py
```diff
@@ -239,7 +239,7 @@ def post(self) -> f.Response:
         # Rely on the idea that the lookup
         # `benchmark_result.run.commit.repo_url` always succeeds
         conbench.metrics.COUNTER_BENCHMARK_RESULTS_INGESTED.labels(
-            repourl=str(benchmark_result.run.commit.repo_url)
+            repourl=benchmark_result.run.associated_commit_repo_url
```
jgehrcke, line 242: Not attached to this, just re-using a new property (code deduplication, that is good)

conbench/app/index.py
```diff
@@ -63,7 +63,9 @@ def get(self) -> str | flask.Response:
         for r in runs:
             rd = RunForDisplay(
                 ctime_for_table=r.timestamp.strftime("%Y-%m-%d %H:%M:%S UTC"),
-                commit_message_short=conbench.util.short_commit_msg(r.commit.message),
+                commit_message_short=conbench.util.short_commit_msg(
+                    r.commit.message if r.commit else ""
+                ),
```
jgehrcke, line 68: another edge case caught by type checking

conbench/app/index.py
```diff
@@ -74,7 +76,7 @@ def get(self) -> str | flask.Response:
                 run=r,
             )
 
-            rname = repo_url_to_display_name(r.commit.repo_url)
+            rname = repo_url_to_display_name(r.associated_commit_repo_url)
```
jgehrcke, line 79: another bug caught by type checking

conbench/entities/benchmark_result.py
```diff
@@ -690,10 +691,12 @@ def validate_run_result_consistency(userres: Any) -> None:
...
     # The property should not be called "github", this confuses me each
-    # time I deal with that.
-    if userres.get("github") is not None:
-        chrun = run.commit.hash
-        chresult = userres["github"]["commit"]
+    # time I deal with that. This is the "github-flavored commit info".
+
+    gh_commit_info: Optional[TypeCommitInfoGitHub] = userres.get("github")
+    if gh_commit_info is not None:
+        chrun = run.commit.hash if run.commit else None
```
jgehrcke, line 700: this also was a bug caught by type checking

conbench/entities/commit.py
```diff
@@ -960,7 +961,8 @@ def _get_response_retry_guts(self, url) -> Optional[dict]:
         # Non-retryable error.
         metrics.COUNTER_GITHUB_HTTP_API_REQUEST_FAILURES.inc()
         raise Exception(
-            f"Unexpected GitHub HTTP API response: {resp}",
+            "Unexpected GitHub HTTP API response for URL "
+            f"{url}: {resp}. Leading bytes of body: '{resp.text[:150]} ...'"
```
jgehrcke, line 965: better debug info

conbench/entities/history.py
```diff
@@ -138,6 +138,14 @@ def get_history_for_benchmark(benchmark_result_id: str):
     # exception.
 
     benchmark_result: BenchmarkResult = BenchmarkResult.one(id=benchmark_result_id)
+
+    if benchmark_result.run.commit is None:
+        # Alternatively, raise an exception here -- allowing to inform the
+        # user that conceptually there will never be history for this
+        # benchmark result, because it's not associated with repo/commit
+        # information
+        return []
```
jgehrcke, line 147: This also was an interesting case revealed by type checking, feedback appreciated!

conbench/entities/history.py
```diff
@@ -184,8 +199,12 @@ def get_history_for_cchr(
             BenchmarkResult.case_id == case_id,
             BenchmarkResult.context_id == context_id,
             BenchmarkResult.error.is_(None),
+            # This `sha == Commit.fork_point_sha` cannot be satisfied for
+            # results with an unknown commit context where commit parent/child
+            # and branch information is not available. Consequence is to not
+            # allow for history endpoint result for 'unknown context'.
             Commit.sha == Commit.fork_point_sha,  # on default branch
-            Commit.repository == repo,
+            Commit.repository == repo_url,
```
jgehrcke, line 207: trying to gradually add more clarity

conbench/entities/run.py
```diff
@@ -66,8 +67,16 @@ class Run(Base, EntityMixin):
...
     error_info: Mapped[Optional[dict]] = Nullable(postgresql.JSONB)
     error_type: Mapped[Optional[str]] = Nullable(s.String(250))
-    commit_id: Mapped[str] = NotNull(s.String(50), s.ForeignKey("commit.id"))
-    commit: Mapped[Commit] = relationship("Commit", lazy="joined")
+
+    # The type annotation makes this a nullable many-to-one relationship.
+    # https://docs.sqlalchemy.org/en/20/orm/basic_relationships.html#nullable-many-to-one
+    # A nullable many-to-one relationship between Commit (one) and potentially
+    # many Runs, but a run can also _not_ point to a commit.
+    commit_id: Mapped[Optional[str]] = mapped_column(s.ForeignKey("commit.id"))
```
jgehrcke, line 75: FYI: SQLAlchemy infers the 'nullable' from the type annotation :rocket:!

conbench/entities/run.py
```diff
@@ -186,6 +147,11 @@ def get_default_baseline_run(
         """
         from ..entities.benchmark_result import BenchmarkResult
 
+        if self.commit is None:
+            # No associated repo/commit information. Raise an exception
+            # instead?
+            return None
```
jgehrcke, line 152: Another occasion where I think type checking revealed a special case that wants to be handled _here_. I don't know if this was before handled deeper into the business logic, but I think here really is a good place to handle this case, towards code readability.

But maybe there was even a bug hiding here?

conbench/entities/run.py
```diff
@@ -272,42 +254,43 @@ def commit_fetch_info_and_create_in_db_if_not_exists(
...
-        if commit is not None:
-            return commit
+        if dbcommit is not None:
+            return dbcommit
 
         # Try to fetch metadata for commit via GitHub HTTP API. Fall back
         # gracefully if that does not work.
-
-        gh_commit_dict = None
+        gh_commit_metadata_dict = None
```
jgehrcke, line 271: So many different commit objects flying around here. I tried to add more clarity to the names. 

`dbcommit` is the SQLAlchemy-based object that represents our own database entry.

conbench/entities/run.py
```diff
@@ -497,6 +515,49 @@ def change_pr_number_empty_string_to_none(self, data, **kwargs):
         },
     )
 
+    @marshmallow.validates_schema
+    def validate_props(self, data, **kwargs):
+        url = data["repository"]
+        if not url.startswith("https://github.com"):
+            raise marshmallow.ValidationError(
+                f"'repository' must be a URL, starting with 'https://github.com', got `{url}`"
```
alistaire47: There's some legacy code where this will break: If users opt-in to github info autodetection (which we discourage), it will pass their git remote (which may start with `git@`) here: https://github.com/conbench/conbench/blob/e1cab9b8ebffaba1c85e454b1b13de7f14bfcf02/conbench/machine_info.py#L127 

plus in benchadapt. If we're not going to accept that, we should probably move the transform-to-URL code to the client-side, or remove the autodetection entirely.

jgehrcke:
Forgot to `git add` the new DB migration module, done now.

conbench/entities/commit.py
```diff
@@ -414,25 +408,27 @@ def get_github_commit(repository: str, pr_number: str, branch: str, sha: str) ->
...
 
     # `github.get_commit()` below may raise an exception if the GitHub
     # GitHub HTTP API failed, e.g. with a 4xx rate limiting response.
-    commit = _github.get_commit(name, sha)
+    commit = _github.get_commit_info(repospec, cinfo["commit_hash"])
 
-    if branch:
-        commit["branch"] = branch
-    elif pr_number:
+    # Seemingly, branch takes precedence over PR number.
```
austin3dickey, line 419: Yes, the reasoning for this was that
- we only want to store `branch`
- if both are provided, we can just use `branch` instead of hitting the GitHub HTTP API for it

conbench/entities/history.py
austin3dickey: Yeah, maybe it's better to raise an exception here and handle it in the UI. It would be nice to list this as the reason why we can't display a plot, for example.

conbench/entities/commit.py
```diff
@@ -84,8 +98,8 @@ def repo_url(self) -> Optional[str]:
         Return a URL (string) or None. The returned string is guaranteed to
         start with 'http' and is guaranteed to not have a trailing slash.
 
-        The `None` case is here because I think the database may contain other
-        (non-URL) value types; and we need to phase those out.
+        The `None` case is here because I think that legacy databases may
```
austin3dickey, line 101: We can (and I think we should) create an `alembic` migration that doesn't change schema but instead just scans through each database to "correct" any rows displaying this old behavior. In that same PR, we could correct these handful of places that are nervous to handle the `repository` column.

conbench/entities/run.py
austin3dickey: For what it's worth, I purposely chose not to do "nullability-by-type-annotations" for other columns: https://github.com/conbench/conbench/pull/859#discussion_r1134517211

But that was more of a spur-of-the-moment decision; I don't really mind if we decide to switch.

conbench/entities/run.py
```diff
@@ -90,66 +99,18 @@ def create(data) -> "Run":
...
-            pr_number=pr_number,
-            branch=branch,
+        user_given_commit_info: Optional[TypeCommitInfoGitHub] = data.pop(
+            "github", None
         )
 
-        run = Run(**data, commit_id=commit_id, hardware_id=hardware.id)
+        commit_for_run = None
+        if user_given_commit_info is not None:
+            log.info(user_given_commit_info)
```
austin3dickey: I think this could add a lot of noise to logs; perhaps remove it?

conbench/entities/run.py
austin3dickey: This case was handled in the try-caught `this_commit.commit_ancestry_query` property, and turned into a no-op with a debug message. Perhaps add the debug message back here?

conbench/entities/run.py
austin3dickey: Though I'm not sure there's anywhere that's using autodetection today, it does seem like a useful feature to keep. I'd rather move the URL-transformation to benchadapt.

Before merging this PR, we also need to think about (and, IMO, implement!) how to supply "no commit info" through benchadapt. Right now if you do `github=None` it opts you in to autodetection. Maybe that needs to be rethought. We can't allow benchadapt to get to a state where it can't supply "no commit info" to Conbench.

conbench/entities/run.py
```diff
@@ -497,6 +515,49 @@ def change_pr_number_empty_string_to_none(self, data, **kwargs):
...
+            raise marshmallow.ValidationError(
+                f"'repository' must be a URL, starting with 'https://github.com', got `{url}`"
+            )
+
+        try:
+            urlparse(url)
+        except ValueError as exc:
+            raise marshmallow.ValidationError(
+                f"'repository' failed URL validation: `{exc}`"
+            )
```
austin3dickey, line 538: maybe exception-chain here?

conbench/entities/run.py
```diff
@@ -532,7 +593,22 @@ class _RunFacadeSchemaCreate(marshmallow.Schema):
...
+    github = marshmallow.fields.Nested(
+        SchemaGitHubCreate(),
+        required=False,
+        metadata={
+            "description": conbench.util.dedent_rejoin(
+                """
+                Use this object to tell Conbench with which commit the
+                Run/BenchmarkResult is associated.
+
+                This field can be left out for testing purposes. Note however
```
austin3dickey, line 612: Another important use case here is when you run benchmarks on a state that doesn't exist on the GitHub-known git graph, like when something runs in CI on a transient merge-commit between the HEAD of a PR and HEAD of the default branch, or when you run benchmarks locally after making changes that you haven't pushed. In those cases, you must leave this field out, or risk polluting your data.

conbench/entities/run.py
```diff
@@ -452,18 +429,41 @@ def commit_hardware_run_map():
...
 
         return data
 
     commit = marshmallow.fields.String(
         required=True,
         metadata={
-            "description": "The 40-character commit hash of the repo being benchmarked"
+            "description": conbench.util.dedent_rejoin(
+                """
+                The commit hash of the benchmarked code.
```
austin3dickey, line 458: Can we still specify that it should be full (40-char) here, instead of a shortened version? There's so much software out there that accepts a shortened version, and we should not accept that

conbench/tests/api/_fixtures.py
```diff
@@ -229,7 +232,26 @@ def benchmark_result(
     empty_results=False,
     reason=None,
 ):
-    """Create BenchmarkResult and write to database."""
+    """Create BenchmarkResult and directly write to database.
+
+    This is done by going straight to the database, surpassing the HTTP
+    interface. This is an architectural smell, diminishing the efficacy of this
```
austin3dickey, line 238: agreed, this isn't great

conbench/tests/entities/test_commit.py
```diff
@@ -129,28 +129,39 @@ def test_repository_to_name():
...
+#     pr_number = 123
 
-    assert get_github_commit(None, None, None, None) == {}
-    assert get_github_commit("", "", "", "") == {}
-    assert get_github_commit(repo, None, None, None) == {}
-    assert get_github_commit(None, pr_number, None, None) == {}
-    assert get_github_commit(None, None, branch, None) == {}
-    assert get_github_commit(None, None, None, sha) == {}
+#     # What was this supposed to confirm? That indeed all detail needs
+#     # to be provided?
```
austin3dickey: I think simply that this function wouldn't fail if given various missing inputs. We can delete this test

austin3dickey (dismissed):
Very nice! Just some small things. The reason I said "changes requested" is I think we should come up with an immediate plan for benchalerts -- how can that client library provide "no commit information" when posting a result or run? I'm going to need that functionality soon.

conbench/entities/run.py
jgehrcke: Thanks for feedback!

> We can't allow benchadapt to get to a state where it can't supply "no commit info" to Conbench

I thought maybe benchadapt is guaranteed to send plausible commit info or it does error out.

OK, looks like we're not so strict: https://github.com/conbench/conbench/blob/e1cab9b8ebffaba1c85e454b1b13de7f14bfcf02/benchadapt/python/benchadapt/result.py#L191

OK, so _there_ we need to hook in and use the 'new' way of communicating "not commit info". OK, I am on that.

jgehrcke:
> we should come up with an immediate plan for benchalerts -- how can that client library provide "no commit information" when posting a result or run? I'm going to need that functionality soon.

OK, did a tiny bit of work for that: https://github.com/conbench/conbench/pull/1135

conbench/entities/run.py
jgehrcke: > There's some legacy code where this will break: If users opt-in to github info autodetection (which we discourage), it will pass their git remote (which may start with git@) here:

Thanks for letting me know.

To short-cut this conversation/topic I am adding a commit here that for now accepts this URL format and transparently re-writes this into the `https://github.com/...` format.

conbench/entities/history.py
jgehrcke: Good. Let's handle this in a follow-up.

conbench/entities/run.py
jgehrcke: Ah, nice. I love when dots get connected.

> I think that's a little too magic.

Yeah, I can agree to that. I was impressed this is possible at all :P 

For the future we can try to use the more explicit style, I like that.

conbench/entities/run.py
jgehrcke: Thanks for adding detail.

> was handled

Did I change this in this PR?

Looking at https://github.com/conbench/conbench/pull/1134/files#diff-8e7df94d5fa6b0082305db878bc76670c59fc020792fb4f78a494d083faaedeaR158 I am not so sure.

Let's leave this as a follow-up item.

conbench/entities/commit.py
jgehrcke: Thanks for adding detail!

conbench/entities/commit.py
jgehrcke: Yeah, I thought about 'fixing' old databases, but what's the disadvantage of not doing that? I think the code is fine with both the old and new, and I like to save work here and there by not trying to polish 'old' data too much :-). It's always a valid question to wonder who even cares about current historical data in current Conbench deployments.

conbench/entities/run.py
jgehrcke: This was a debugging-leftover I missed, thank you!!

conbench/entities/run.py
jgehrcke: I think the stringified error message is the only relevant error detail from the "url parse exception".

conbench/entities/run.py
jgehrcke: > on a state that doesn't exist on the GitHub-known git graph,

Thanks for pointing this out.

Asking naively: what's the purpose of submitting such a result to Conbench? To _later_ add commit information? I trust there is a purpose, would just love to see this in your words/perspective.

conbench/entities/run.py
jgehrcke: Yes! Adding a commit:

Adding this back to the textual specification, and also adding a new validation error.

jgehrcke: Backing out of this. Have been working for 40 minutes to unbreak our tests, unsuccessfully. The test suite relies on being able to use (very) short commit hashes. Pushing my WIP branch to https://github.com/conbench/conbench/pull/1136.

conbench/entities/commit.py
```diff
@@ -391,15 +387,13 @@ def repository_to_url(repository: str) -> str:
             "repository_to_url() about to create invalid URL, name is: %s", name
         )
 
-    # Note(JP): the `lower()` appears to be dangerous. URLs are case-sensitive.
-    # We should trust user-given input in that regard, or at least think this
-    # through a little further. Also see
-    # https://github.com/conbench/conbench/issues/818
-    return f"https://github.com/{name.lower()}"
+    return f"https://github.com/{name}"
```
jgehrcke, line 390: https://github.com/conbench/conbench/issues/887

jgehrcke:
I'd love to land this now, this was a detour already, and I think it provides value. I think it might stir up some chaos downstream, yes, but I am ready to react :).

I carefully went through the feedback above and replied in the relevant places. But here is a summary of follow-up discussions or tasks:
https://github.com/conbench/conbench/pull/1134#discussion_r1171094732
https://github.com/conbench/conbench/pull/1134#discussion_r1171095587
https://github.com/conbench/conbench/pull/1134#discussion_r1171099349
https://github.com/conbench/conbench/pull/1134#discussion_r1171109372
https://github.com/conbench/conbench/pull/1134#discussion_r1171171685

migration? https://github.com/conbench/conbench/pull/1134#discussion_r1171103924

jgehrcke:
Corresponding follow-up patch for benchadapt is here: https://github.com/conbench/conbench/pull/1135

conbench/entities/run.py
austin3dickey: Before this PR, `self.commit` was guaranteed to be a `Commit` instance—never None—because the foreign key was non-nullable. 

If the run truly wasn't associated with a commit, `self.commit` would have been one of those "no-context" commits with the branch/fork_point/timestamp properties empty. So the `commit_ancestry_query` property would have failed here:

https://github.com/conbench/conbench/blob/a2732d3ee415cdcf0b342f73a15e35110739d6a1/conbench/entities/commit.py#L203-L204

conbench/entities/run.py
austin3dickey: An example: we're currently setting up "pre-merge" benchmarks on one of our company's internal projects. These run in CI on pull requests, building the project and running benchmarks on the "commit" that the CI checks out.

In most CI systems, when you run CI on a PR commit, it actually runs on a temporary merge-commit between the HEAD of the PR branch and the HEAD of main, and that temporary commit doesn't actually exist on the GitHub-known git graph. For example, this happens when using GitHub Actions; check out [the docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request). The GITHUB_REF is `refs/pull/:prNumber/merge`, which is briefly explained by [this SO answer](https://stackoverflow.com/a/63595981).

This sort of goes against our [documented strategy](https://conbench.github.io/conbench/pages/benchmarking_and_the_dev_process.html#possible-alternatives-and-their-limitations) of always benchmarking on the explicit PR head, even if the PR branch is forked from an older default-branch commit. We're benchmarking on a state of code that may not exist or be reference-able in the future, and that's bad for reproducibility. So we shouldn't connect it to the canonical git graph.

(@jonkeane has told me about a brief time years ago where we _did_ post the commit hash of the "transient" merge-commit to Conbench, and it was a nightmare to maintain and explain to users. It's simply hard to draw conclusions on transient history.)

So our strategy will be to post these pre-merge runs with no git commit information, since they don't belong on the git graph. We won't be able to view plots for those results, but we will at least have a URL to point to for looking at their raw data and errors. And importantly, we should still be able to hit `/api/compare/runs/some_good_run_id...pre_merge_run_id` to see whether there were regressions compared to some "good" baseline run. That latter use case is why I've been working on issues like #384 lately.

conbench/entities/run.py
austin3dickey: Oof, I forgot about that...

conbench/entities/run.py
jgehrcke: > So our strategy will be to post these pre-merge runs with no git commit information, since they don't belong on the git graph.

ack

> To later add commit information?

just summarizing, the answer is "no"

> we will at least have a URL to point to for looking at their raw data and errors [...] able to hit /api/compare/runs/some_good_run_id...pre_merge_run_id to see whether there were regressions compared to some "good" baseline run. That latter use case is why I've been working on issues like https://github.com/conbench/conbench/issues/384 lately.

OK, two specific 'value' items. Thanks!

"This field can be left out for testing purposes" isn't a good fit then really (in the diff above). As you point out, this is associated with features. Should change this text again.

conbench/entities/run.py
jgehrcke: Thank you for clarifying! Now I understand. 

Do we want to retain the `CantFindAncestorCommitsError` exception type for now? 

> Perhaps add the debug message back here?

Yes, and also we can maybe make the error propagation better and feed all the way to the user (as you also said somewhere in this discussion) the _reason_ for why history analysis couldn't be performed -- the relevant error detail here would be something like "no commit information stored with this benchmark result, cannot inspect commit ancestry".

austin3dickey: >Do we want to retain the CantFindAncestorCommitsError exception type for now?

Yes, I think that might be a good idea, at least for the case when we're trying to examine a partially-filled commit (i.e. one with a commit hash but no metadata; to be filled in later).

>feed all the way to the user (as you also said somewhere in this discussion) the reason for why history analysis couldn't be performed

Agreed!

austin3dickey: >feed all the way to the user (as you also said somewhere in this discussion) the reason for why history analysis couldn't be performed

I am going to do this in https://github.com/conbench/conbench/pull/1078 (at least when trying to get contender baseline run IDs).

jonkeane:
Reviewing this post-hoc: this looks great, and glad to see the clean ups. I'm curious if we are worried about not having repo information for these runs/results that aren't (able to be) associated with a commit. It's only really an issue for situations where we track multiple repos (which we only do in one conbench AFAIK, and not the conench that will be making use of this unable-to-track-on-github situation). But IIRC there is some matching logic that will only match results for benchmarks that reference the same repo, right?

austin3dickey:
>But IIRC there is some matching logic that will only match results for benchmarks that reference the same repo, right?

Just took a look. Everywhere that we filter to a given repository in our DB queries, it's because we're always getting a baseline distribution, which depends on the given target baseline commit, not the contender commit. So I think we're good.
- The compare endpoint doesn't use the contender repo at all, only the baseline repo
- The history endpoint (which just powers plots) doesn't return anything if the given result doesn't have a repo
- We won't be able to get `parent` or `fork_point` baseline runs for a contender run without a repo, but we can still get the `head_of_default` baseline run (since it uses the baseline's repo)

jgehrcke:
Thanks for the additional eyes on this and the follow-up discussion. Much appreciated.

conbench/entities/run.py
jgehrcke: > Should change this text again.

https://github.com/conbench/conbench/pull/1425

PR #1135, opened 2023-04-19
benchadapt: consolidate the "github-flavored commit info" acquisition and communication

jgehrcke:
Motivated by
https://github.com/conbench/conbench/pull/1134#discussion_r1171073366
https://github.com/conbench/conbench/pull/1134#pullrequestreview-1390569743

> how can that client library provide "no commit information" when posting a result or run? I'm going to need that functionality soon.

> OK, so there we need to hook in and use the 'new' way of communicating "not commit info". OK, I am on that.

Did a few more chores while being in this code base. Clearer types. Clearer naming. More debug info. See commit messages!

### DIFF
--- a/benchadapt/python/benchadapt/_machine_info.py
+++ b/benchadapt/python/benchadapt/_machine_info.py
@@ -127,7 +127,7 @@ def detect_commit_info_from_local_git():
     and only supply the "pr_number" (or leave it None if it's a commit to the default
     branch).
     """
-    commit = _exec_command(["git", "rev-parse", "HEAD"])
+    commit = _exec_command(["git", "rev-parse", "HEAD"], ignore_error=False)
     if not commit:
         warnings.warn(
             "error in detect_commit_info_from_local_git(): probably not in a git repo; returning {}",
@@ -149,7 +149,9 @@ def detect_commit_info_from_local_git():
         _, branch_name, _, upstream, *_ = current_branch.split()
         if not upstream.startswith("["):
             remote = "origin"
-            branch = _exec_command(["git", "branch", "--show-current"])
+            branch = _exec_command(
+                ["git", "branch", "--show-current"], ignore_error=False
+            )
             warnings.warn(
                 f"error in detect_commit_info_from_local_git(): untracked branch; returning {branch=}",
                 GitParseWarning,
@@ -158,7 +160,7 @@ def detect_commit_info_from_local_git():
             remote = upstream[1:].split("/")[0]
             branch = branch_name
 
-    remote_url = _exec_command(["git", "remote", "get-url", remote])
+    remote_url = _exec_command(["git", "remote", "get-url", remote], ignore_error=False)
     fork = re.search("(?<=.com[/:])([^/]*?)(?=/)", remote_url).group(0)
 
     # Assumed to be all str values of non-zero length.
@@ -386,17 +388,26 @@ def _has_missing(info, mapping):
     return False
 
 
-def _exec_command(command):
+def _exec_command(command, ignore_error=True):
+    """
+    The ignore_error behavior is for legacy code, it's not advisable to
+    swallow error like this; really hard to debug.
+    """
     result = subprocess.run(command, capture_output=True)
 
     # Some of these failures are expected, i.e. the command not executing isn't
     # supposed to be fatal to this program here. However, it makes sense to log
     # information about how exactly the child process invocation failed.
     if result.returncode != 0:
-        log.info(
-            "child process returned with code %s, stderr prefix:\n %s",
+        log.warning(
+            "command `%s` returned with code %s, stderr prefix:\n %s",
+            command,
             result.returncode,
             result.stderr.decode("utf-8").strip()[:500],
         )
+        if not ignore_error:
+            raise Exception(
+                "command unexpectedly returned with non-zero exit code, see log for details"
+            )
 
     return result.stdout.decode("utf-8").strip()
--- a/benchadapt/python/benchadapt/_machine_info.py
+++ b/benchadapt/python/benchadapt/_machine_info.py
@@ -97,7 +97,7 @@ def github_info():
     return gh_info
 
 
-def detect_github_info():
+def detect_commit_info_from_local_git():
     """Attempts to inspect a locally cloned repository for git information that can be
     posted to the "github" key when creating a run.
 
@@ -109,7 +109,7 @@ def detect_github_info():
     commit = _exec_command(["git", "rev-parse", "HEAD"])
     if not commit:
         warnings.warn(
-            "error in github_info(): probably not in a git repo; returning {}",
+            "error in detect_commit_info_from_local_git(): probably not in a git repo; returning {}",
             GitParseWarning,
         )
         return {}
@@ -119,7 +119,7 @@ def detect_github_info():
         remote = "origin"
         branch = "<DETATCHED BRANCH>"
         warnings.warn(
-            f"error in github_info(): detached HEAD; returning {branch=}",
+            f"error in detect_commit_info_from_local_git(): detached HEAD; returning {branch=}",
             GitParseWarning,
         )
 
@@ -130,7 +130,7 @@ def detect_github_info():
             remote = "origin"
             branch = _exec_command(["git", "branch", "--show-current"])
             warnings.warn(
-                f"error in github_info(): untracked branch; returning {branch=}",
+                f"error in detect_commit_info_from_local_git(): untracked branch; returning {branch=}",
                 GitParseWarning,
             )
         else:
--- a/benchadapt/python/benchadapt/_machine_info.py
+++ b/benchadapt/python/benchadapt/_machine_info.py
@@ -118,59 +118,6 @@ def gh_commit_info_from_env() -> Dict[str, str]:
     return result
 
 
-def detect_commit_info_from_local_git_or_raise():
-    """Attempts to inspect a locally cloned repository for git information that can be
-    posted to the "github" key when creating a run.
-
-    It's recommended to NOT use this function, and instead manually supply the commit
-    and repository (in the form "org/repo"). When doing so, ignore the "branch" key
-    and only supply the "pr_number" (or leave it None if it's a commit to the default
-    branch).
-    """
-    commit = _exec_command(["git", "rev-parse", "HEAD"], ignore_error=False)
-    if not commit:
-        warnings.warn(
-            "error in detect_commit_info_from_local_git(): probably not in a git repo; returning {}",
-            GitParseWarning,
-        )
-        return {}
-
-    branches = _exec_command(["git", "branch", "-vv"])
-    if "* (HEAD detached" in branches:
-        remote = "origin"
-        branch = "<DETATCHED BRANCH>"
-        warnings.warn(
-            f"error in detect_commit_info_from_local_git(): detached HEAD; returning {branch=}",
-            GitParseWarning,
-        )
-
-    else:
-        current_branch = [b for b in branches.split("\n") if b.startswith("*")][0]
-        _, branch_name, _, upstream, *_ = current_branch.split()
-        if not upstream.startswith("["):
-            remote = "origin"
-            branch = _exec_command(
-                ["git", "branch", "--show-current"], ignore_error=False
-            )
-            warnings.warn(
-                f"error in detect_commit_info_from_local_git(): untracked branch; returning {branch=}",
-                GitParseWarning,
-            )
-        else:
-            remote = upstream[1:].split("/")[0]
-            branch = branch_name
-
-    remote_url = _exec_command(["git", "remote", "get-url", remote], ignore_error=False)
-    fork = re.search("(?<=.com[/:])([^/]*?)(?=/)", remote_url).group(0)
-
-    # Assumed to be all str values of non-zero length.
-    return {
-        "commit": commit,
-        "repository": remote_url.rsplit(".git")[0],
-        "branch": f"{fork}:{branch}",
-    }
-
-
 def machine_info(host_name: Optional[str] = None):
     os_name, os_version = platform.platform(terse=True).split("-", maxsplit=1)
 
--- a/benchadapt/python/benchadapt/_machine_info.py
+++ b/benchadapt/python/benchadapt/_machine_info.py
@@ -1,9 +1,7 @@
 import logging
 import os
 import platform
-import re
 import subprocess
-import warnings
 from typing import Dict, Optional
 
 
--- a/benchadapt/python/benchadapt/_machine_info.py
+++ b/benchadapt/python/benchadapt/_machine_info.py
@@ -118,7 +118,7 @@ def gh_commit_info_from_env() -> Dict[str, str]:
     return result
 
 
-def detect_commit_info_from_local_git():
+def detect_commit_info_from_local_git_or_raise():
     """Attempts to inspect a locally cloned repository for git information that can be
     posted to the "github" key when creating a run.
 
--- a/benchadapt/python/benchadapt/_machine_info.py
+++ b/benchadapt/python/benchadapt/_machine_info.py
@@ -161,11 +161,11 @@ def detect_commit_info_from_local_git():
     remote_url = _exec_command(["git", "remote", "get-url", remote])
     fork = re.search("(?<=.com[/:])([^/]*?)(?=/)", remote_url).group(0)
 
+    # Assumed to be all str values of non-zero length.
     return {
         "commit": commit,
         "repository": remote_url.rsplit(".git")[0],
         "branch": f"{fork}:{branch}",
-        "pr_number": None,
     }
 
 
--- a/benchadapt/python/benchadapt/_machine_info.py
+++ b/benchadapt/python/benchadapt/_machine_info.py
@@ -79,7 +79,7 @@ class GitParseWarning(RuntimeWarning):
     pass
 
 
-def github_info():
+def gh_commit_info_from_env() -> Dict[str, str]:
     """Get github metadata from environment variables"""
     repo = os.environ.get("CONBENCH_PROJECT_REPOSITORY")
     pr_number = os.environ.get("CONBENCH_PROJECT_PR_NUMBER") or os.environ.get(
--- a/benchadapt/python/benchadapt/_machine_info.py
+++ b/benchadapt/python/benchadapt/_machine_info.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import platform
 import re
@@ -9,6 +10,8 @@ from typing import Optional
 def _sysctl(stat):
     return ["sysctl", "-n", stat]
 
+
+log = logging.getLogger(__name__)
 
 MUST_BE_INTS = [
     "cpu_core_count",
@@ -364,4 +367,15 @@ def _has_missing(info, mapping):
 
 def _exec_command(command):
     result = subprocess.run(command, capture_output=True)
+
+    # Some of these failures are expected, i.e. the command not executing isn't
+    # supposed to be fatal to this program here. However, it makes sense to log
+    # information about how exactly the child process invocation failed.
+    if result.returncode != 0:
+        log.info(
+            "child process returned with code %s, stderr prefix:\n %s",
+            result.returncode,
+            result.stderr.decode("utf-8").strip()[:500],
+        )
+
     return result.stdout.decode("utf-8").strip()
--- a/benchadapt/python/benchadapt/_machine_info.py
+++ b/benchadapt/python/benchadapt/_machine_info.py
@@ -92,7 +92,7 @@ def gh_commit_info_from_env() -> Dict[str, str]:
     """
 
     varmap = {
-        "CONBENCH_PROJECT_REPOSITORY": "repo",
+        "CONBENCH_PROJECT_REPOSITORY": "repository",
         "CONBENCH_PROJECT_COMMIT": "commit",
         "BENCHMARKABLE_PR_NUMBER": "pr_number",
         "CONBENCH_PROJECT_PR_NUMBER": "pr_number",  # later entry takes precedence
--- a/benchadapt/python/benchadapt/_machine_info.py
+++ b/benchadapt/python/benchadapt/_machine_info.py
@@ -4,7 +4,7 @@ import platform
 import re
 import subprocess
 import warnings
-from typing import Optional
+from typing import Dict, Optional
 
 
 def _sysctl(stat):
@@ -80,21 +80,42 @@ class GitParseWarning(RuntimeWarning):
 
 
 def gh_commit_info_from_env() -> Dict[str, str]:
-    """Get github metadata from environment variables"""
-    repo = os.environ.get("CONBENCH_PROJECT_REPOSITORY")
-    pr_number = os.environ.get("CONBENCH_PROJECT_PR_NUMBER") or os.environ.get(
-        "BENCHMARKABLE_PR_NUMBER"
-    )
-    commit = os.environ.get("CONBENCH_PROJECT_COMMIT")
+    """
+    Attempt to read a specific set of environment variables expected to carry
+    GitHub-flavored information about the commit/checkout state of the
+    benchmarked repository.
 
-    gh_info = {
-        "commit": commit,
-        "repository": repo,
-        # "branch": None,
-        "pr_number": pr_number,
+    The returned dictionary is guaranteed to have both keys and values of type
+    str, all of length greater zero.
+
+    The returned dictionary may be empty.
+    """
+
+    varmap = {
+        "CONBENCH_PROJECT_REPOSITORY": "repo",
+        "CONBENCH_PROJECT_COMMIT": "commit",
+        "BENCHMARKABLE_PR_NUMBER": "pr_number",
+        "CONBENCH_PROJECT_PR_NUMBER": "pr_number",  # later entry takes precedence
     }
 
-    return gh_info
+    result = {}
+
+    for evarname, dictkey in varmap.items():
+        val = os.environ.get(evarname)
+        if val is not None:
+            stripval = val.strip()
+
+            if len(stripval) > 0:
+                result[dictkey] = stripval
+                continue
+
+            log.warning(
+                "ignoring environment variable %s with whitespace value: `%s`",
+                evarname,
+                repr(val),
+            )
+
+    return result
 
 
 def detect_commit_info_from_local_git():
--- a/benchadapt/python/benchadapt/result.py
+++ b/benchadapt/python/benchadapt/result.py
@@ -288,6 +288,11 @@ def validate_or_remove_github_commit_key(res_dict: Dict, strict=False):
         del res_dict["github"]
         return
 
+    # Note(JP): this is basically non-strict schema validation with the
+    # following behavior: "ok, it looks like you wanted to send commit
+    # information, but you didn't do it properly, and now I pretend as if
+    # you really didn't want to do that, but I emit a bit of a warning".
+    # We should make this stricter in the future.
     for checkkey in ("repository", "commit"):
         if checkkey not in res_dict["github"]:
             _warn_or_raise()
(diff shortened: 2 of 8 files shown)


benchadapt/python/benchadapt/result.py
```diff
@@ -171,10 +173,20 @@ def _cluster_info_property(self, value: Dict[str, Any]) -> None:
...
+        After JSON-serialization, that dictionary is expected to validate
+        against the JSON schema that the Conbench API expects on the endpoint
+        for benchmark result submission.
+        """
+
         res_dict = asdict(self)
 
+        # We should discuss why we don't exit with an error here (publish this
+        # although it's not publishable? who consumes the warning? should the
+        # warning be re-worded to be more user-friendly?)
```
jgehrcke, line 219: I think this is really important. If I imagine to build client tooling and seeing a warning like "Result not publishable!" but  then everything "still works" is kind of confusing.

jgehrcke:
`ci / libraries` failure: ah, right not covered by `make tests`, and also no makefile target yet.

benchadapt/python/benchadapt/_machine_info.py
```diff
@@ -76,25 +79,46 @@ class GitParseWarning(RuntimeWarning):
...
+        "CONBENCH_PROJECT_PR_NUMBER": "pr_number",  # later entry takes precedence
     }
 
-    return gh_info
+    result = {}
+
+    for evarname, dictkey in varmap.items():
+        val = os.environ.get(evarname)
+        if val is not None:
+            stripval = val.strip()
```
jgehrcke, line 104: Considering whitespace is important when reading environment variables.

benchadapt/python/benchadapt/_machine_info.py
```diff
@@ -76,25 +79,46 @@ class GitParseWarning(RuntimeWarning):
...
 
+            if len(stripval) > 0:
+                result[dictkey] = stripval
+                continue
 
-def detect_github_info():
+            log.warning(
```
jgehrcke, line 110: I don't know if this logger will find a root logger to bubble messages into. But if that's not yet the case, then these additional log messages will not break anything, and they will find a root logger in the future, when we consolidate logging.

benchadapt/python/benchadapt/_machine_info.py
```diff
@@ -106,7 +130,7 @@ def detect_github_info():
     commit = _exec_command(["git", "rev-parse", "HEAD"])
     if not commit:
         warnings.warn(
-            "error in github_info(): probably not in a git repo; returning {}",
+            "error in detect_commit_info_from_local_git(): probably not in a git repo; returning {}",
```
jgehrcke: the old name was wrong

benchadapt/python/benchadapt/_machine_info.py
```diff
@@ -364,4 +388,15 @@ def _has_missing(info, mapping):
...
+
+    # Some of these failures are expected, i.e. the command not executing isn't
+    # supposed to be fatal to this program here. However, it makes sense to log
+    # information about how exactly the child process invocation failed.
+    if result.returncode != 0:
+        log.info(
+            "child process returned with code %s, stderr prefix:\n %s",
+            result.returncode,
+            result.stderr.decode("utf-8").strip()[:500],
+        )
```
jgehrcke, line 352: dang, swallowing errors is never nice to the peeps around you :)

benchadapt/python/benchadapt/result.py
```diff
@@ -210,6 +214,48 @@ def to_publishable_dict(self):
         return res_dict
 
 
+def validate_or_remove_github_commit_key(res_dict: Dict, strict=False):
+    """
+    Mutate BenchmarkResult dictionary in-place to make its `github` key
+    property be compliant with the Conbench HTTP API.
```
jgehrcke: Maybe this sentence sets a higher goal than the code below fulfills. But then we know that the code below is buggy.

benchadapt/python/benchadapt/run.py
```diff
@@ -133,15 +136,7 @@ def to_publishable_dict(self):
...
-        if not (
-            res_dict["github"].get("repository") and res_dict["github"].get("commit")
-        ):
-            raise ValueError(
-                "Run not publishable! `github.repository` and `github.commit` must be populated. "
-                "You may pass github metadata via CONBENCH_PROJECT_REPOSITORY, CONBENCH_PROJECT_COMMIT, "
-                "and CONBENCH_PR_NUMBER environment variables. "
-                f"\ngithub: {res_dict['github']}"
-            )
+        validate_or_remove_github_commit_key(res_dict, strict=True)
```
jgehrcke, line 133: We had duplicate code here, I thought it's better to de-duplicate, especially now that I added logic.

Of course there were minor differences between the two code sections, they are now abstracted with the `strict` parameter.

benchadapt/python/tests/test_run.py
```diff
@@ -87,8 +87,8 @@ def test_name_defaulting(self, monkeypatch):
         monkeypatch.delenv("CONBENCH_PROJECT_COMMIT", raising=False)
 
         run = BenchmarkRun(reason=run_json["reason"])
+        assert run.github == {}
```
jgehrcke, line 89: This is an internal, not-so-important state.

Do we need to do more testing on the return value of `to_publishable_dict()` in the future?

benchadapt/python/benchadapt/result.py
austin3dickey: Hm, I thought we had a discussion about this somewhere on GitHub but can't find it.

benchadapt/python/benchadapt/result.py
```diff
@@ -135,7 +135,9 @@ class BenchmarkResult:
     machine_info: Dict[str, Any] = field(default_factory=_machine_info.machine_info)
     cluster_info: Dict[str, Any] = None
     context: Dict[str, Any] = field(default_factory=dict)
-    github: Dict[str, Any] = field(default_factory=_machine_info.github_info)
+    github: Dict[str, str] = field(
```
austin3dickey: Could you adjust the docstring for this field based on this PR's behavior?

benchadapt/python/benchadapt/result.py
```diff
@@ -210,6 +214,48 @@ def to_publishable_dict(self):
...
+    # have the client tooling error out instead?
+    if "github" not in res_dict:
+        # API-wise, that's a valid state, signaling "no commit information
+        # present"
+        _warn_or_raise()
+        return
+
+    for checkkey in ("repository", "commit"):
+        if checkkey not in res_dict["github"]:
+            _warn_or_raise()
```
austin3dickey: I tried out a bunch of different inputs on this branch:
```py
>>> BenchmarkResult(error={"a": 1}).to_publishable_dict()["github"]
# /Users/austin/repos/conbench/benchadapt/python/benchadapt/result.py:239: UserWarning: Entity not publishable! `github.repository` and `github.commit` must be populated. You may pass github metadata via CONBENCH_PROJECT_REPOSITORY, CONBENCH_PROJECT_COMMIT, and CONBENCH_PR_NUMBER environment variables.
# github: {}
# KeyError: 'github'

>>> BenchmarkResult(error={"a": 1}, github=None).to_publishable_dict()["github"]
# {'commit': 'c6a0e28d50ecdd14b7e1b06a1e458ce571cd3787', 'repository': 'git@github.com:conbench/conbench', 'branch': 'conbench:jp/benchadapt-gh-info'}

>>> BenchmarkResult(error={"a": 1}, github={"commit": "asd", "repository": "asd"}).to_publishable_dict()["github"]
# {'commit': 'asd', 'repository': 'asd'}

# Add environment variables
>>> os.environ["CONBENCH_PROJECT_REPOSITORY"] = "asd"
>>> os.environ["CONBENCH_PROJECT_COMMIT"] = "asd"
>>> BenchmarkResult(error={"a": 1}).to_publishable_dict()["github"]
# {'repository': 'asd', 'commit': 'asd'}

>>> BenchmarkResult(error={"a": 1}, github=None).to_publishable_dict()["github"]
# {'commit': 'c6a0e28d50ecdd14b7e1b06a1e458ce571cd3787', 'repository': 'git@github.com:conbench/conbench', 'branch': 'conbench:jp/benchadapt-gh-info'}

>>> BenchmarkResult(error={"a": 1}, github={"commit": "asdfg", "repository": "asdfg"}).to_publishable_dict()["github"]
# {'commit': 'asdfg', 'repository': 'asdfg'}
```

I think this is all what I'd expect, though for the first case, I don't think the warning should say the result isn't publishable. It is publishable now! 🙂  But we should probably warn that this result won't have repo/commit context, since that's a rare (but [supported](https://github.com/conbench/conbench/pull/1134#discussion_r1171810962)) use case.

benchadapt/python/benchadapt/result.py
```diff
@@ -210,6 +214,48 @@ def to_publishable_dict(self):
...
+        )
+        if strict:
+            raise ValueError(msg)
+
+        warnings.warn(msg)
+
+    # For now, the decision is to send the result out, signaling to the
+    # Conbench API that this result has no repo/commit context. Maybe we should
+    # have the client tooling error out instead?
+    if "github" not in res_dict:
```
austin3dickey: Based on how this dataclass is constructed, I don't think we can get to this branch. `res_dict` is constructed from the dataclass itself, and I don't think `asdict()` leaves out any attributes of a dataclass.

benchadapt/python/benchadapt/result.py
austin3dickey: Also, I think we should still support the first case even for runs. So this should not error out, like it does right now:
```py
BenchmarkRun().to_publishable_dict()
# ValueError: Entity not publishable! `github.repository` and `github.commit` must be populated. You may pass github metadata via CONBENCH_PROJECT_REPOSITORY, CONBENCH_PROJECT_COMMIT, and CONBENCH_PR_NUMBER environment variables.
# github: {}
```
We'd like to be able to post a no-context commit.

benchadapt/python/tests/test_run.py
austin3dickey: Yes, I'd rather test the output of `to_publishable_dict()` than the attributes on the dataclass.

austin3dickey (commented):
looks good! Thank you for this!

Just one main thing: I'd still like to be able to use `BenchmarkRun().to_publishable_dict()` with no `github` info.

benchadapt/python/benchadapt/result.py
alistaire47: For a no-context commit, do we want to pass `"github": {}` or just not populate it? In parallel situations (e.g. `error`, `stats`) we pop off empty dicts.

Also, this interface for opting-in to no-context commits (not specifying `github` when no env vars are set) is kinda crappy. An explicit `github={}` (even if we want to pop it off) would be nicer.

And frankly I think usually `repo` should still be set for most of these cases even if `commit`, `pr_number`, and `branch` are empty, but that's maybe bigger than the scope of this PR.

benchadapt/python/benchadapt/_machine_info.py
jonkeane: Should we pull this out into a utility function somewhere?

benchadapt/python/benchadapt/result.py
```diff
@@ -210,6 +214,48 @@ def to_publishable_dict(self):
         return res_dict
 
 
+def validate_or_remove_github_commit_key(res_dict: Dict, strict=False):
+    """
+    Mutate BenchmarkResult dictionary in-place to make its `github` key
+    property be compliant with the Conbench HTTP API.
+
+    Not providing the `github` key in result dictionary tells Conbench that
+    this result is commit context-less (recording it in Conbench might be
```
jonkeane:
```suggestion
    this result is commit-context-less (recording it in Conbench might be
```

Super minor, but we're talking about the commit-context here yeah?

jgehrcke: Yes, thanks! :) Hyphenation is a difficult topic sometimes, and I am glad you gave feedback here!

benchadapt/python/benchadapt/result.py
jgehrcke: > For a no-context commit, do we want to pass "github": {} or just not populate it? 

I take the current state of what the HTTP wants (how it's specified) for granted for the moment. We can always review/refine/change that part, but this here is basically adjusting client tooling to the latest HTTP API change. 

Everything becomes easier when the following is true: if a `github` key is provided, then _it is a full specification of a commit (long hash) in a repository in github (specified by URL)_; no exceptions (and honestly I think only then the _name_ of this dictionary starts to make briefly sense; "github" has rather weak self-documenting properties)

Madness happens given all the various possible permutations between
- no dict allowed
- empty dict allowed
- if dict specified:
    - only commit allowed
    - only repo allowed
    - if repo specified:
        - url allowed
        - org/repo notation allowed

(and more) -- this is a _huge_ permutation space, and it did leak deep into our business logic. We didn't try to normalize this upon receiving the HTTP request, we did put most of this complexity even into the database. 

I just want to reduce the complexity for sanity.

jgehrcke: > And frankly I think usually repo should still be set for most of these cases even if commit, pr_number, and branch are empty, but that's maybe bigger than the scope of this PR.

This is probably correct thinking and heavily relates to https://github.com/conbench/conbench/issues/1165, right?

jgehrcke: > I'd still like to be able to use BenchmarkRun().to_publishable_dict() with no github info.

> I think this is all what I'd expect, though for the first case, I don't think the warning should say the result isn't publishable. It is publishable now! slightly_smiling_face But we should probably warn that this result won't have repo/commit context, since that's a rare (but https://github.com/conbench/conbench/pull/1134#discussion_r1171810962) use case.

> So this should not error out, like it does right now:

Thanks for the feedback everyone. Looking at this now again, then adding commits.

benchadapt/python/benchadapt/result.py
jgehrcke: Thanks for trying to find it!

benchadapt/python/benchadapt/result.py
jgehrcke: Oh, I forgot that indeed. Thanks for the reminder!

benchadapt/python/benchadapt/result.py
jgehrcke: Hm good point. I will think about that.

benchadapt/python/benchadapt/result.py
jgehrcke: Yes, I understand now. I was confusing the two objects here:
- the dataclass instance
- the to_publishable_dict() result

benchadapt/python/benchadapt/result.py
jgehrcke: Yes, I understand now. I was confusing the two objects here:
- the dataclass instance
- the to_publishable_dict() result

benchadapt/python/benchadapt/result.py
jgehrcke: > I don't think the warning should say the result isn't publishable. It is publishable now! 

thanks for confirming, yes. I didn't want to go to deep here with adjusting messages, but I think now I do.

jgehrcke: Thanks for the code examples, too. Thinking about this a little more.

When having env vars set and when
- not providing an argument for `github` then the outcome should be populated from env.
- providing `github=None` that's the explicit way to say "no commit information"

Distinguishing both cases can probably be easily achieved with https://docs.python.org/3/library/dataclasses.html#post-init-processing (if that's needed at all).

alistaire47: I tried to avoid `__post_init__`s with these because you can end up with different behavior if you set something on instantiation vs after, which does happen with these, e.g. https://github.com/conbench/conbench/blob/e23b5fee689034c1cd99f92061cce92d604f59a1/benchadapt/python/benchadapt/adapters/_adapter.py#L153-L161

> * providing `github=None` that's the explicit way to say "no commit information"

That's shorthand for using the autodetection (which in retrospect was a regrettable decision; maybe we can unwind?). I think you need `github={}` to really specify "no repo/commit info".

austin3dickey: As far as I'm aware, no one is using `github=None` for autodetection (scraping git locally) today. It's still a feature I want to keep around but maybe we can move that to a different opt-in interface and use `github=None` for "no commit information"? (though if we do want to do something around #1165 in the future, we'll need to think about that interface too)

jgehrcke: Great discussion!

> That's shorthand for using the autodetection

Oh, and with that you mean the "inspect git repo in current working directory", I see. 

I see that is documented, the docstring says:

> Advanced: if you have a locally cloned repo, you may explicitly supply ``None``

Just for the record, dataclass is able to distinguish between "no value passed" and "`None` passed" (I know you know that Ed, just really important fact for this discussion).

We have code that handles the `None` case already [here](https://github.com/conbench/conbench/blob/e23b5fee689034c1cd99f92061cce92d604f59a1/benchadapt/python/benchadapt/result.py#L159): 

```
        if value is None:
            value = _machine_info.detect_commit_info_from_local_git()
```
OK, about those two statements:

> maybe we can unwind?

> no one is using github=None for autodetection (scraping git locally) today

Let's do it! Let's add a new value for that, e.g. "inspect_git_in_cwd" (I would like the explicit nature of that). And I mean, we could also remove that functionality because that's very magic anyway and nobody uses it, but for now I would totally feel comfortable just to change how to "activate" this.

Then we can repurpose `None` as described above and also do not need `__post_init__`

jgehrcke: I want to get unstuck here, trying to put this into some commits and update the branch, so I can call it a day :). :sun_with_face:

austin3dickey: Lovely!

benchadapt/python/benchadapt/_machine_info.py
jgehrcke: Yeah that is a good idea at some point for sure!  Punting on this for this current patch, takes too long already :).

I'd love to have a `lib` concept here in our monorepo where all the different tools can make use the same utility functions.

jgehrcke:
@austin3dickey @alistaire47 as my focus fades for today it might be that I have missed one of the critical points here, but I tried my best for now to process review feedback and implement that changes we discussed most recently. Please have a look! If there's no big booboo then I'd be happy if you can merge this, otherwise of course I will amend tomorrow :)

benchadapt/python/benchadapt/result.py
```diff
@@ -63,18 +63,45 @@ class BenchmarkResult:
...
+
+        If it was run on a non-default branch and a non-PR commit, you may
+        supply the branch name via the ``branch`` set to a value of the format
+        ``org:branch``.
+
+        Deprecated: pass the special value "inspect_git_in_cwd" to try to
+        attempt commit hash / remote URL / branch information from a .git
+        directory in the current working directory via executing git commands
+        as child processes. This raises an exception when any of the commands
+        fails.
```
jgehrcke: Now this docstring is of course interesting, as it should properly describe the new behavior. I hope it does, but I have not reviewed it again, getting tired; hoping for your critical eyes on this now! :)

benchadapt/python/benchadapt/result.py
```diff
@@ -63,18 +63,45 @@ class BenchmarkResult:
         For benchmarks run on a cluster, information about the cluster
     context : Dict[str, Any]
         Should include ``benchmark_language`` and other relevant metadata like compiler flags
-    github : Dict[str, Any]
-        Keys: ``repository`` (in the format ``org/repo``), ``commit``, and ``pr_number``.
-        If this is a benchmark on the default branch, you may leave out ``pr_number``.
-        If it's a non-default-branch & non-PR commit, you may supply the branch name to
-        the optional ``branch`` key in the format ``org:branch``.
+    github : Union[Optional[Dict[str, Any]], Literal["inspect_git_in_cwd"]
```
jgehrcke: new type annotation looks funny.

benchadapt/python/benchadapt/_machine_info.py
```diff
@@ -362,6 +388,26 @@ def _has_missing(info, mapping):
...
+        log.warning(
+            "command `%s` returned with code %s, stderr prefix:\n %s",
+            command,
+            result.returncode,
+            result.stderr.decode("utf-8").strip()[:500],
+        )
+        if not ignore_error:
+            raise Exception(
+                "command unexpectedly returned with non-zero exit code, see log for details"
+            )
```
alistaire47, line 356: `subprocess.run()` will raise nice classed errors for non-zero return codes [if you set `check=True`](https://docs.python.org/3/library/subprocess.html#subprocess.run) or call [`.check_returncode()`](https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess.check_returncode) on `result`

benchadapt/python/benchadapt/result.py
alistaire47: Instead of a literal, could we just have users call the util function that scrapes and returns the dict?

benchadapt/python/benchadapt/result.py
```diff
@@ -63,18 +63,45 @@ class BenchmarkResult:
...
+
+        If the benchmark was run gainst the default branch, you may leave
+        out ``pr_number`` and ``branch``.
+
+        If it was run on a GitHub pull request branch, you should provide the
+        ``pr_number``.
+
+        If it was run on a non-default branch and a non-PR commit, you may
+        supply the branch name via the ``branch`` set to a value of the format
+        ``org:branch``.
```
alistaire47: This should also mention that `"pr_number": None` is the way to specify a commit is on the default branch

benchadapt/python/benchadapt/result.py
```diff
@@ -210,6 +250,58 @@ def to_publishable_dict(self):
         return res_dict
 
 
+def validate_or_remove_github_commit_key(res_dict: Dict, strict=False):
+    """
+    Mutate BenchmarkResult dictionary (result of asdict(self)) in-place to make
+    its `github` key property be compliant with the Conbench HTTP API:
+    - Remove it when it's set to `None` (silently)
+    - Remove it when it doesn't look good (but also emit a warning)
```
alistaire47, line 252: IMO we should just error in this case (i.e. always make `strict=True`) because no commit information is a weird corner-case you should have to deliberately opt-in to; you don't want to end up like that because you configured your env vars shoddily.

That would also let this function be just validation, and the popping could be handled like other attributes in L238-248.

benchadapt/python/benchadapt/_machine_info.py
jgehrcke: I know this is positive/good intent!

I just tried to make this a tiny tiny bit nicer for the next person who maybe needs to work in this. More changes trigger more feedback :)

Moving on from here.

benchadapt/python/benchadapt/result.py
jgehrcke: I liked this idea, started re-writing the docstring again and than realized: no, we should not invest work into this: I am just removing this now. We don't use it anyway. I want to get over with this :).

benchadapt/python/benchadapt/result.py
jgehrcke: Yes. Also, leaving it out, right?

For the record, documented here: https://github.com/conbench/conbench/blob/6c6f3a441a07a0e3a4ce3f6f5ca46e0ff3ca8d3e/conbench/entities/run.py#L568

Btw there is the ambiguity of this being defined as marshmallow.fields.Integer() and here I think we send it as string, may be fine, maybe marshmallow is OK with that.

In this module we should use the same type as here: https://github.com/conbench/conbench/blob/6c6f3a441a07a0e3a4ce3f6f5ca46e0ff3ca8d3e/conbench/entities/commit.py#L43

(not saying the type is perfect, but code-wise we should use the _same_ type definition (monorepo allows us to!).

jgehrcke: Well.

> is the way

The API spec says that "null" is the way, but probably allows both (not-set, set to null).

And that's interesting because in JSON there is a semantic difference between not set and set-to-null.

This depends on marshmallow schema validation.

Need to sharpen the API spec in that regard.

jgehrcke: I just had the question again if it's OK to leave out pr_number or not, and how this is treated. Already forgot that I handled this very recently here:

https://github.com/conbench/conbench/blob/6c6f3a441a07a0e3a4ce3f6f5ca46e0ff3ca8d3e/conbench/entities/run.py#L621

`null` is explicitly the same as not-set; both are fine. Need to document that in the HTTP API spec. Spent too much time trying get an answer to this now.

jgehrcke: I now also remember that the famous test you added recently with `@pytest.mark.parametrize("pr_number", [12345678, "12345678", None, "", "<absent>"])` covers all this. But yeah, this needs to be textually specified, for our own sake.

jgehrcke: The more I think about this right now, and the more I go back and forth between the different places in code where we talk about this and deal with, the clearer it becomes:

There needs to be a _central place_ for textual specification that all other places refer to. I keep typing the same sentences in different places, and over time they diverge.

benchadapt/python/benchadapt/result.py
jgehrcke: I thought the same, see 
```
    # Note(JP): this is basically non-strict schema validation with the
    # following behavior: "ok, it looks like you wanted to send commit
    # information, but you didn't do it properly, and now I pretend as if
    # you really didn't want to do that, but I emit a bit of a warning".
    # We should make this stricter in the future.
```

and this was my attempt to not change everything at once

jgehrcke: now making this this stricter

benchadapt/python/benchadapt/result.py
jgehrcke: I took your feedback serious and added a commit, working a little more on the HTTP/JSON API spec. Also worked on this text here again.

Future: add typedict / dataclass /pydantic datatype for this dict, and perform runtime schema validation.

benchadapt/python/benchadapt/result.py
jgehrcke: added commits: rewrote message once again. no more warnings, only errors. adjusted tests.

jgehrcke:
Thanks for the great feedback everyone, I think we're on the same page about all the things. We will (have to) do much more here. But for now, we should move on :) 

I started this PR with the intent to not send the `github` key anymore, as a tiny improvement. Now we've done more improvements, because basically everything here asks to be improved over time. It's also a bit of a frustrating area to work in because everything's so ambiguous. I am glad that one outfall was the removal of the "read info from git repo via child processes". Surface area that wants to be maintained and discussed; or removed :fire_engine:. 

Removing ambiguity and adding clarity is precisely what we have to do here, and in that sense I am super glad that we seem to agree on that! I really like @alistaire47 that you advocated for more strictness. Less warnings, more actual errors. Good direction.

I want to add that most of the discussion in here would not happen with a _precisely_ specificed JSON/data schema for the "github-flavored commit info", with runtime validation, and automatically generated nice error messages. That's the way to go I think. After all, `Dict[str, str]` does not even barely reflect what this is.

If this patch now introduced new problems/bugs, we'll find them and fix them!

jgehrcke:
> As far as I'm aware, no one is using github=None for autodetection (scraping git locally) today. 

Mistakes happen. This turned out to not be true! 

https://github.com/apache/arrow/blob/b43f4cd4ce28e2ddb7bad89081e141f381b09d5e/ci/scripts/go_bench_adapt.py#L42

```
else:
    github = None  # scrape github info from the local repo
    run_reason = "branch"
```

I should have kept the slightly cleaner `Literal["inspect_git_in_cwd"]` in this patch, then it would be easy now to adjust arrow's CI script. Now it's a tiny bit more work.

Edit: while in fact arrow CI used that feature, it did not do so intentionally. See https://github.com/apache/arrow/pull/35467. The decisions in this PR were fine.

Issue #789, opened 2023-03-02
DB schema: optimize for tail view, keep track of insertion order for entities (with index, ideally via primary key)


jgehrcke:
Generalization of https://github.com/conbench/conbench/issues/786.

Requirement for https://github.com/conbench/conbench/issues/622.

We want to have a solution for "give the last N inserted entities, fast" for all the things (including benchmark results).

Currently that's not easily possible because of a UUID4 type primary key.

I think we should cheaply keep track of insertion order via primary key. Could use simple auto-increment integer (non-fancy default in many systems, would have served us well here, yeah) or we could do something like [UUID7](https://github.com/uuid6/uuid6-ietf-draft#high-level-overview) (might be a little too new, maybe no established tooling support as of today).

I think this task/goal/ticket should be tracked independently of tracking _insertion time_ (https://github.com/conbench/conbench/issues/790).

jgehrcke:
> We want to have a solution for "give the last N inserted entities, fast" for all the things (including benchmark results).

Adding a thought from #786 to here:

> I keep centering around the idea that most users would like to have a tail view, i.e. insertion order should play a central role: https://github.com/conbench/conbench/issues/789 ... Maybe in the future, a central UI concept for snappy DB queries/snappy UI population should involve the idea of showing tabular data (everywhere in the UI) for the last D days (could default to e.g. 10). That relates to https://github.com/conbench/conbench/issues/790.

jgehrcke:
Btw, I liked https://supabase.com/blog/choosing-a-postgres-primary-key quite a bit when I started to read about about this back then in March.

jgehrcke:
Let's just go ahead with using UUID type 7 for the primary key before we stall on this for even longer.

Interesting perspective from https://www.percona.com/blog/store-uuid-optimized-way/

> The size of the UUID table is almost 50% bigger than Ordered UUID table and 30% bigger than the table with BIGINT as PRIMARY KEY. Comparing the Ordered UUID table BIGINT table, the time is taken to insert rows and the size are almost the same. But they may vary slightly based on the index structure.

I am glad they discuss monotonicity to a great extent in the RFC draft: https://www.ietf.org/archive/id/draft-ietf-uuidrev-rfc4122bis-02.html#name-monotonicity-and-counters.

For CPython someone rather brutally dropped a minimal pull request: https://github.com/python/cpython/pull/29824 -- but there isn't even an issue on the tracker discussing the feature/enhancement to the stdlib.

[...]

PR #1138, opened 2023-04-19
db insert: use uuid7 primary key to track insert order

jgehrcke:
This is for issue #789, see discussion over there.

I noted:
- The `s.String(50)` does not match the idea that the prim key is known to be of length 32. Does that mismatch matter?
- For some entitiy (run, at least) we still allow the user to inject the primary key. That needs to change, tracking this here: https://github.com/conbench/conbench/issues/1139

### DIFF
--- a/conbench/entities/_entity.py
+++ b/conbench/entities/_entity.py
@@ -1,5 +1,4 @@
 import functools
-import uuid
 from typing import List
 
 import flask as f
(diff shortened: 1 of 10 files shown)


austin3dickey (approved):
I do think we should consider a database cleanup for this one too, or else we won't be able to take advantage of it

jgehrcke:
> I do think we should consider a database cleanup for this one too, or else we won't be able to take advantage of it

That is true for existing deployments, yes! But I didn't want to stall in analysis paralysis. We work under the assumption that Conbench has a future beyond the 'legacy systems'. And the earlier we take this step here, the more we learn and move into a helpful direction :-). In that sense, I think that https://github.com/conbench/conbench/issues/1139 is for now more important than doing migration work for older systems.

I have the vague vision that at some point in the future we can make a cut and make legacy systems _actual_ legacy systems, and build fresh systems with fresh database state.

Issue #1140, opened 2023-04-20
webapp: HTTP response compression for deployment scenarios without reverse proxy


jgehrcke:
Before doing #1109 it's low effort/big impact to simply [decorate the Flask app](https://github.com/colour-science/flask-compress) via a two-line change, and get gzip compression for the most relevant types of HTTP responses:

One import is needed, and one `compress.init_app(app)`. 

This does not replace  #1109 but is a valuable default behavior for Conbench deployments _not_ running behind a nicely configured HTTP reverse proxy.

jgehrcke:
I want to say that in terms of separation of concerns this here does not only have advantages. For example, I really like seeing the actual payload response size in the WSGI application access log. Say, the size of a JSON document in bytes is really helpful debug information to see -- now, we can only infer that roughly because usually gzip compression is on top.

jgehrcke:
Done in #1141.

Issue #618, opened 2023-01-23
Test CLI against old versions of python


austin3dickey:
>Yeah, we absolutely should run python 3.8 on the CLI. Honestly, we also should test back as far as we can and support as many versions as can that aren't EOLed yet (for the CLI). I've had at least two conbench setups that had extra work involved because python 3.8 wasn't immediately available where we were trying to set it up. We worked around it, but it's a friction we shouldn't subject our users to if we can avoid it.

_Originally posted by @jonkeane in https://github.com/conbench/conbench/issues/616#issuecomment-1400951460_

jgehrcke:
In https://github.com/conbench/conbench/pull/538 I have learned that we are phasing out this 'old' `conbench` CLI. That is, we should keep investment to a minimum. Would love to hear brief opinions.

The core aspect here once again is "What's not tested is broken" :).

> I've had at least two conbench setups that had extra work involved because python 3.8 wasn't immediately available where we were trying to set it up

Towards compatibility and stability (less surprises) we should

- build the CLI with the Python interpreter, via pyinstaller or similar (relying on platform/system Python implies madness!)
- perform automated tests with the resulting binary
- upload that binary if it has passed tests, using a version specifier
- use a specific CLI version wherever it's consumed

This simply removes so many surprises.

If we do these things it again becomes a little less important how much code the CLI and the web application share.

jgehrcke:
To me, this is still a rather definite "won't do". Need to look forward, not backward. If absolutely necessary we build a container or a single-binary CLI.

jonkeane:
I'm fine if we don't prioritize this right now — but if we're looking for smooth + easy uptake we will definitely need to do _something_ here.

> build a container 

In my experience with both velox and arrow this wouldn't have been easy (or possible). For Velox there was a custom container (we probably could add a layer, but that can be insignificant friction), for arrow: there is not container running at all and that was a design decision of the orchestration.

> single-binary CLI.

Would be very curious to see what this looks like

PR #1141, opened 2023-04-20
webapp: gzip-compress HTTP responses (issue #1140)

jgehrcke:
This works:

```
$ curl -sH 'Accept-encoding: gzip' localhost:5000 | gunzip - | head -n3
<!DOCTYPE html>
  <html lang="en">
      <head>
```

```
 $ curl -sH 'Accept-encoding: gzip' localhost:5000/api/benchmarks/ | gunzip - | head -n3
[
  {
    "id": "0644107c9f5f79bf800082e0cb94fe12",
```

### DIFF
--- a/conbench/__init__.py
+++ b/conbench/__init__.py
@@ -19,6 +19,8 @@ import logging
 import os
 from typing import TYPE_CHECKING
 
+import flask_compress
+
 import conbench.logger
 
 try:
@@ -86,6 +88,13 @@ def create_application(config) -> "flask.Flask":
 
     # This mutates `app` in-place.
     conbench.metrics.decorate_flask_app_with_metrics(app)
+
+    # This also mutates the app in-place, sets up
+    # https://github.com/colour-science/flask-compress for deployments that do
+    # not run an HTTP reverse proxy with proper response compression in front
+    # of Conbench. Conbench amy emit large JSON and HTML responses, and simple
+    # gzip compression yields huge value.
+    flask_compress.Compress().init_app(app)
 
     return app
 
--- a/conbench/__init__.py
+++ b/conbench/__init__.py
@@ -19,8 +19,6 @@ import logging
 import os
 from typing import TYPE_CHECKING
 
-import flask_compress
-
 import conbench.logger
 
 try:
@@ -48,6 +46,7 @@ log = logging.getLogger(__name__)
 
 def create_application(config) -> "flask.Flask":
     import flask as f
+    import flask_compress
 
     import conbench.metrics
 
--- a/requirements-webapp.txt
+++ b/requirements-webapp.txt
@@ -7,6 +7,7 @@ dacite
 email-validator
 Flask
 Flask-Bootstrap
+Flask-Compress
 Flask-Login
 flask-swagger-ui
 Flask-WTF


jgehrcke:
The check-docs check is a tiny bit impactul every now and then.
```
WARNING: autodoc: failed to import module 'runner' from module 'conbench'; the following exception was raised:
No module named 'flask_compress'
```

jgehrcke:
Let's get this in! :rocket:

austin3dickey:
>The check-docs check is a tiny bit impactul every now and then.

This was a legitimate problem, though, right? The reason the build failed is that: 
- during the docs build, we `pip install .` (the `conbench` library/CLI for the legacy runner) with no extras, just like the [Arrow Benchmark CI](https://github.com/voltrondata-labs/arrow-benchmarks-ci/blob/118389b7e367dd2318e2d0a399d20350466f298b/buildkite/benchmark/utils.sh#L76) does
- when building the docs, we tried to `import conbench.runner`, and failed because we didn't have a required dependency

This means that if we had merged the PR as of the failing build, and Arrow updated its peg to use that version of the CLI, then Arrow Benchmark CI would have been broken too.

You moved the import inside the function, which is fine! But another solution would have been to add `Flask-Compress` to `requirements-cli.txt` instead of `requirements-webapp.txt`, because adding a dependency to the top of `conbench.__init__.py` is making it a dependency of the CLI too, not just the webapp.

This of course relates to https://github.com/conbench/conbench/issues/617, and also https://github.com/conbench/conbench/issues/618 (because if we tested `conbench.runner` better, we wouldn't have to rely on the `check-docs` step to catch this type of thing), and it also points a finger toward the removal of the legacy CLI sooner than later. 🙂

jgehrcke:
Thanks for looking at this, Austin! 

> This was a legitimate problem, though, right? 

Oh, jeeeeez. Yes. I didn't look deep enough to see that it's _that_ kind of problem space.

We need to tackle this. I take "The check-docs check is a tiny bit impactul every now and then." back, it in fact has just its own way of reminding us of the dead bodies in the cellar, and I didn't connect the dots.

All the more reason to deal with the "legacy CLI" challenge.

> Arrow updated its peg to use that version of the CLI, then Arrow Benchmark CI would have been broken too.

Over there we do have this pinned in the meantime, right?

>  would have been to add Flask-Compress to requirements-cli.txt

Of course that's technically true! I hope we both feel the pain about the misleading semantics of this :P

I have the suspicion that in the meantime we have added dependencies to -webapp and not -cli.

austin3dickey:
>Over there we do have this pinned in the meantime, right?

Yes, it's pinned: 
https://github.com/voltrondata-labs/arrow-benchmarks-ci/blob/118389b7e367dd2318e2d0a399d20350466f298b/buildkite/benchmark/utils.sh#L76

>I have the suspicion that in the meantime we have added dependencies to -webapp and not -cli.

This is probably true, but the docs-check probably failed whenever we did, which caused us to move those deps inside webapp functions :) luckily at least with the current state of docs-check we know that `conbench.runner` is importable without `requirements-webapp`, because `autodoc` needs to import it to build docs. But yes, this is not ideal.

Issue #1143, opened 2023-04-20
DB entity create-or-fetch: query first, then create, handle transaction error & use generics for typing


jgehrcke:
In #1125 I added the EAFP approach for DB-inserting entities: https://github.com/conbench/conbench/commit/f77194ceb3d9f2ec156db60d5d03a7847ef1e9e5

Today I saw that the database emits a unique key violation error in its log for each of these 'attempted insertions'. That's a cosmetic problem, but I looked into it, and I think it's worth it.

https://stackoverflow.com/questions/12385763/suppress-duplicate-key-value-violates-unique-constraint-errors suggests using `WHERE  NOT EXISTS `, another answer suggests `set log_min_messages=fatal;`.

We also have SQLAlchemy's "upsert do nothing" strategy already in the code base. 

But the best arguments are in this discussion: https://groups.google.com/g/sqlalchemy/c/tVc19TUJQu8

> The *other* way to do it is to use INSERT..ON CONFLICT (PostgreSQL, SQLite) or INSERT..ON DUPLICATE KEY (MySQL, MariaDB) statements.  These are popular choices as well but you need to construct them as statement objects.   

I tried that. This also logs the conflicts.

What's important to consider is

>  Session is in a transaction which completes when you call rollback() or commit().  

I really like (as always) Mike's perspective:

> Right, so the "executing two statements at once instead of two statements" thing in this kind of case is a fleeting optimization. That is, its a micro optimization that is easily nullified by the surrounding context. Such as, if the way the app works in reality is that the row already exists 95% of the time, the optimization saves negligible time. Or if it's trivial to just pre-insert the rows in question, or a whole selection of 100 rows can be selected at once and just the ones that aren't present can be INSERTed in one multi-row statement, would provide even better performance.

> Taking a SELECT then an INSERT and making the choice to turn it into a non-ORM, single statement, database-specific call is something you'd do once the app is up and running, and the separate SELECT/INSERT pair has been observed to be a definite bottleneck with no feasible workaround. I.e. a non-premature optimization.

Action items:
- change paradigm again: read, upon error insert, upon error read again
- implement this in parent class, with proper type inference, as indicated here: https://github.com/conbench/conbench/blame/4e1a92f12e8433161febf347c0559e2aeabc0743/conbench/entities/context.py#L11

jgehrcke:
Done via https://github.com/conbench/conbench/pull/1144.

PR #1144, opened 2023-04-20
_entity.py: add type-generic get_or_create() method

jgehrcke:
This addresses #1143.

### DIFF
--- a/conbench/entities/_entity.py
+++ b/conbench/entities/_entity.py
@@ -1,5 +1,5 @@
 import functools
-from typing import List, Dict, TypeVar, Generic, Type
+from typing import Dict, Generic, List, Type, TypeVar
 
 import flask as f
 import sqlalchemy
--- a/conbench/entities/_entity.py
+++ b/conbench/entities/_entity.py
@@ -1,7 +1,8 @@
 import functools
-from typing import List
+from typing import List, Dict, TypeVar, Generic, Type
 
 import flask as f
+import sqlalchemy
 from sqlalchemy import distinct, select
 from sqlalchemy.dialects.postgresql import insert as postgresql_insert
 from sqlalchemy.orm import declarative_base, mapped_column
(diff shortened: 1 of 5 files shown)


conbench/entities/_entity.py
```diff
@@ -63,7 +64,12 @@ def to_float(value):
     return float(value) if value is not None else None
 
 
-class EntityMixin:
+T = TypeVar("T")
+
+
+class EntityMixin(Generic[T]):
```
jgehrcke, line 70: Counterpart, upon subclassing, example: `class Context(Base, EntityMixin["Context"]):`

conbench/entities/_entity.py
```diff
@@ -167,6 +173,40 @@ def delete(self):
         Session.delete(self)
         Session.commit()
 
+    @classmethod
+    def get_or_create(cls: Type[T], props: Dict) -> T:
```
jgehrcke, line 177: Not sure if it's obvious what this does. This screenshot helps:
image

The IDE Knows that the return value is of type `Context` because `Context` derives from `EntityMixin["Context"]`.

conbench/entities/_entity.py
```diff
@@ -167,6 +173,40 @@ def delete(self):
...
+        name/tags).
+
+        Return (newly created, or previously existing) object, or raise an
+        exception.
+        """
+
+        def _fetch_first():
+            return Session.scalars(select(cls).filter_by(**props)).first()
+
+        result = _fetch_first()
```
jgehrcke, line 189: This here is a relevant part of this patch: we try to fetch before we try to create.

conbench/entities/_entity.py
```diff
@@ -167,6 +173,40 @@ def delete(self):
...
+        if result is not None:
+            return result
+
+        obj = cls(**props)
+        Session.add(obj)
```
jgehrcke, line 194: Here, I wasn't quite sure which exceptions are expected to be raised by `add()` vs. `commit()`, and I'd actually like to test this in practice. I think that `add()` might have validation logic, i.e. it might catch programming errors, and I'd like to make the error handler below rather precise.

jgehrcke:
:ship:

conbench/entities/_entity.py
austin3dickey: amazing!

austin3dickey (commented):
Lovely!

PR #1145, opened 2023-04-20
metrics.py: fix: terminate the quota_remaining gauge init thread after first update was received

jgehrcke:
The thread terminator was never invoked because of a boolean booboo.

### DIFF
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -895,8 +895,9 @@ class GitHubHTTPApiClient:
             # Expect the value to always be int-convertible.
             reqquota = int(resp.headers["x-ratelimit-remaining"])
             metrics.GAUGE_GITHUB_HTTP_API_QUOTA_REMAINING.set(reqquota)
-            # Temporary workaround, see metrics._periodically_set_q_rem()
-            metrics.gauge_gh_api_rem_set["set"] = False
+            # Setting this to `True` stops a thread from periodically setting
+            # this to -1, see metrics.periodically_set_q_rem()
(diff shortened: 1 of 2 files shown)


conbench/metrics.py
```diff
@@ -190,13 +190,14 @@ def periodically_set_q_rem() -> None:
     """
 
     def func():
+        log.info("periodically_set_q_rem(): initiate")
         while True:
-            time.sleep(10)
+            time.sleep(1)
```
jgehrcke, line 195: This probably does not matter at all but is a more reasonable order of magnitude to pick.

conbench/entities/commit.py
```diff
@@ -895,8 +895,9 @@ def _get_response_retry_guts(self, url) -> Optional[dict]:
             # Expect the value to always be int-convertible.
             reqquota = int(resp.headers["x-ratelimit-remaining"])
             metrics.GAUGE_GITHUB_HTTP_API_QUOTA_REMAINING.set(reqquota)
-            # Temporary workaround, see metrics._periodically_set_q_rem()
-            metrics.gauge_gh_api_rem_set["set"] = False
+            # Setting this to `True` stops a thread from periodically setting
+            # this to -1, see metrics.periodically_set_q_rem()
+            metrics.gauge_gh_api_rem_set["first_value_seen"] = True
```
jgehrcke, line 900: The bug was: this coordination property just stayed `False`. It was supposed to be updated here to `True`. I think part of the reason was an inexpressive name where I confused myself.

jgehrcke:
Landing this :rocket:

PR #1150, opened 2023-04-21
ConbenchClient: login: add trailing slash

jgehrcke:
Unclear why there's a difference between 
```
>>> requests.post("https://cb-staging/api/login", json={'lol':'rofl'})
<Response [405]>
```

curl:
```
$ curl -L --header "Content-Type: application/json"  -v --request POST  --data '{"username":"xyz","password":"xyz"}'  "https://cb-staging/api/login"
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 3.139.229.187:443...

> POST /api/login HTTP/2

* Mark bundle as not supporting multiuse
< HTTP/1.1 301 Moved Permanently
< Server: awselb/2.0

* h2h3 [:method: POST]

{"code": 400, "name": "Bad Request"}[jp:~]
```

The 301 allows for / demands for changing the request method from POST to GET, and maybe curl violates the rule here? (on purpose?)

Status code(s) to perform a permanent redirect without changing POST to GET are 307 and 308.

https://github.com/psf/requests/issues/5284

### DIFF
--- a/benchclients/python/benchclients/conbench.py
+++ b/benchclients/python/benchclients/conbench.py
@@ -154,7 +154,12 @@ class ConbenchClient(RetryingHTTPClient):
 
         login_result = self._make_request_retry_until_deadline(
             method="POST",
-            url=self._base_url + "/login",
+            # Trailing slash is important so that we do not get redirected.
+            # Some systems might redirect a POST to GET here.
+            # Interesting topic:
+            # https://github.com/galaxyproject/bioblend/pull/336
+            # https://github.com/aio-libs/aiohttp/issues/6764
+            url=self._base_url + "/login/",
             json=creds,
             expected_status_code=204,
             timeout=self.timeout_login_request,
--- a/benchclients/python/tests/test_conbench_client.py
+++ b/benchclients/python/tests/test_conbench_client.py
@@ -102,7 +102,7 @@ def test_cc_performs_login_when_env_is_set(
     }
 
     httpserver.expect_oneshot_request(
-        "/api/login", method="POST", json=creds
+        "/api/login/", method="POST", json=creds
     ).respond_with_data("", status=204)
 
     # This confirms that the initialization of this object performs an HTTP


austin3dickey (approved):
thank you for looking into this!!

jgehrcke:
OK, mystery solved by more careful log-reading. The flow with curl is fascinating and has revealed two things I wasn't aware of:
- special meaning of the `-X` curl arg in view of 301 redirects (https://github.com/curl/curl/issues/5145#issuecomment-604105088)
- bad scheme in Conbench-emitted 308 redirects

Step by step:

1) POST to `https:..../login` (scheme HTTPS, no trailing slash):
```
$ curl -L --header "Content-Type: application/json"  -v --request POST  --data '{"username":"xyz","password":"xyz"}'  "https://cb-staging.voltrondata.run/api/login"

* Connected to cb-staging.voltrondata.run (3.139.229.187) port 443 (#0)
* SSL connection using TLSv1.2 / ECDHE-RSA-AES128-GCM-SHA256
* h2h3 [:method: POST]
* h2h3 [:path: /api/login]
* h2h3 [:scheme: https]

> POST /api/login HTTP/2
```

2) Flask emits a 308 redirect response with the idea to add the trailing slash, but it also sets the scheme to HTTP. Bad URL scheme change (a bug in our Conbench deployment), client is however instructed to retain the method (308 response):
```
< HTTP/2 308 
< date: Sat, 22 Apr 2023 11:23:51 GMT
< content-type: text/html; charset=utf-8
< content-length: 275
< location: http://cb-staging.voltrondata.run/api/login/
< server: gunicorn

* Connection #0 to host cb-staging.voltrondata.run left intact
* Clear auth, redirects to port from 443 to 80
* Issue another request to this URL: 'http://cb-staging.voltrondata.run/api/login/'
* Connected to cb-staging.voltrondata.run (3.140.70.173) port 80 (#1)
> POST /api/login/ HTTP/1.1
```
now tracked here: https://github.com/conbench/conbench/issues/1151

3) ALB sees HTTP scheme and is configured to emit a 301 redirect response for the HTTP->HTTPS redirect, i.e. to ` https...login/`
```
> Host: cb-staging.voltrondata.run
< HTTP/1.1 301 Moved Permanently
< Server: awselb/2.0
< Location: https://cb-staging.voltrondata.run:443/api/login/
```
4) curl sees the 301 redirect and even logs "Switch from POST to GET" (which is what 301 demands!):
```
* Ignoring the response-body
* Issue another request to this URL: 'https://cb-staging.voltrondata.run:443/api/login/'
* Switch from POST to GET
* Found bundle for host: 0x562f65b37f20 [can multiplex]
* Re-using existing connection #0 with host cb-staging.voltrondata.run
* Connected to cb-staging.voltrondata.run (3.139.229.187) port 443 (#0)
```

5) But then the emitted request _still_ has the POST method:
```
* h2h3 [:method: POST]
* h2h3 [:path: /api/login/]
* h2h3 [:scheme: https]
* h2h3 [:authority: cb-staging.voltrondata.run]
* h2h3 [user-agent: curl/7.85.0]
* h2h3 [accept: */*]
* h2h3 [content-type: application/json]


> POST /api/login/ HTTP/2
> Host: cb-staging.voltrondata.run
> user-agent: curl/7.85.0
> accept: */*
> content-type: application/json
```

The reason is documented, it's as of the `--request POST` CLI argument in the curl command line. This forces the method to be a specific one. From the curl docs:

> The method string you set with [-X, --request](https://curl.se/docs/manpage.html#-X) will be used for all requests, which if you for example use [-L, --location](https://curl.se/docs/manpage.html#-L) may cause unintended side-effects when curl does not change request method according to the HTTP 30x response codes - and similar.

6) Correspondingly, Flask really sees a POST request and responds with 400 Bad Request (custom login handler):
``` 
* TLSv1.2 (IN), TLS header, Supplemental data (23):
< HTTP/2 400 
< date: Sat, 22 Apr 2023 11:23:51 GMT
< content-type: application/json
< content-length: 36
< server: gunicorn
```
----------

requests, however, will go through the 308 -> 301 indirection, and is instructed (by the ALB, in that sense), to switch from POST to GET.

Issue #1156, opened 2023-04-24
gunicorn: increase HTTP 1.1 keep-alive timeout from the short default (2 s)


jgehrcke:
In CI, when POSTing benchmark results to Conbench, we have recently seen an elevated rate of 502 Bad Gateway responses emitted by an AWS ALB in front of Conbench.

The technical nature of the problem is now understood in-depth.

https://github.com/conbench/conbench/issues/1018 had the [side effect](https://docs.gunicorn.org/en/latest/faq.html#why-is-there-no-http-keep-alive) of enabling HTTP 1.1's keep-alive mechanism in gunicorn, i.e. the re-usage of a TCP connection across multiple requests. Gunicorn, however, by default doesn't want to use these connections for a long time, but closes these connections [after just two seconds](https://docs.gunicorn.org/en/latest/settings.html#keepalive).

Most load balancers want to (are configured to) re-use TCP connections for longer than two seconds (minutes, usually).

When using such a load balancer in front of gunicorn, there is a race condition where the load balancer may want to use a TCP connection pretty much in the exact moment that the other end closes it.

[Clean TCP connection shutdown](https://stackoverflow.com/a/23483487/145400) allows for this race condition. During clean TCP shutdown, the party that `close()`s the socket (gunicorn, in our case) triggers an RST packet to be sent to the remote end.  [AWS docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-troubleshooting.html#http-502-issues) specifically list "load balancer received a TCP RST from the target" as a cause for emitting a 502 response. They finish the sentence with  "when attempting to establish a connection" but that's of course technical non-sense; the connection _is_ established, just breaks down earlier than the ALB expected. 

Remarks:

* The 5xx class of HTTP responses almost always means "dear client, please retry this, I had a minor problem". So, the "correct fix" for this type of problem really _is_ retrying on the client side. And we have taken care of that. However, it's of course also important that we have understood this in-depth now and can also conceptually patch this by adjusting two time constants to each other (that's how I like to think of this -- often times in software engineering, bugs happen because two time constants are not matching/adjusted, and this here is a beautiful example).

* This would not have happened with a reverse proxy right in front of gunicorn, tracked by #1109. The most successful deployment model for gunicorn is to have a proper reverse proxy running on the same machine, directly in front of it.

austin3dickey:
amazing analysis, thank you!

jgehrcke:
Closing this here as 'done'.

PR #1158, opened 2023-04-24
k8s: alb annotation: set idle_timeout to 50 s (reduce rate of load balancer 502 responses)

jgehrcke:
For issue #1156. Docs:
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout

> We also recommend that you configure the idle timeout of your
> application to be larger than the idle timeout configured
> for the load balancer. Otherwise, if the application closes
> the TCP connection to the load balancer ungracefully, the
> load balancer might send a request to the application before
> it receives the packet indicating that the connection is
> closed. If this is the case, then the load balancer sends
> an HTTP 502 Bad Gateway error to the client.

> By default, Elastic Load Balancing sets the idle timeout
> value for your load balancer to 60 seconds.


jgehrcke:
`ci / cb-on-minikube` failed as of newly observed CI instability: https://github.com/conbench/conbench/issues/1159

Issue #1155, opened 2023-04-24
benchmark result view: hist plot render: sigfig: `ValueError: invalid input Character "n" (position 1, state A)`


jgehrcke:
Seen for staging/benchmark-results/064458556d0a7178800039f9ef44f47f/
```
File "/app/conbench/app/results.py", line 239, in page
plotinfo = self.get_history_plot(benchmark, run)
File "/app/conbench/app/_plots.py", line 64, in get_history_plot
time_series_plot(samples, benchmark, run),
File "/app/conbench/app/_plots.py", line 476, in time_series_plot
formatted, axis_unit = _should_format([s.mean for s in samples], unit)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/app/conbench/app/_plots.py", line 188, in _should_format
units_formatted.add(unit_fmt(f, unit).split(" ", 1)[1])
^^^^^^^^^^^^^^^^
File "/app/conbench/units.py", line 53, in fmt_unit
return f"{sigfig.round(value, sigfigs=4)} {unit}"
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/sigfig/sigfig.py", line 579, in round
given = _arguments_parse(args, kwargs)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/sigfig/sigfig.py", line 274, in _arguments_parse
given['num'] = _num_parse(args[0])
^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/sigfig/sigfig.py", line 547, in _num_parse
A(num)
File "/usr/local/lib/python3.11/site-packages/sigfig/sigfig.py", line 480, in A
raise ValueError('invalid input Character "%s" (position %d, state A)' % (num[0], i))
ValueError: invalid input Character "n" (position 1, state A)
```

jgehrcke:
Relevant benchmark result stats:
```
  "stats": {
    "data": [
      402.365759388078
    ],
    "iqr": null,
    "iterations": 1,
    "max": null,
    "mean": null,
    "median": null,
    "min": null,
    "q1": null,
    "q3": null,
    "stdev": null,
    "time_unit": "s",
    "times": [],
    "unit": "s",
    "z_improvement": false,
    "z_regression": false,
    "z_score": null
  },
```
[...]

PR #1160, opened 2023-04-24
history sample: add single_value_for_plot() method 

jgehrcke:
This helps for debugging and conceptually addressing #1155.

This helps centralizing and documenting how to pick the main "y axis value" from a HistorySample (which is in general a multi-sample benchmark result with almost no guarantees, in particular in view of legacy database state).

### DIFF
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -185,6 +185,9 @@ def _should_format(floats: List[Optional[float]], unit):
             # I think in legacy code this would have simply blown up in a
             # `float(None)``
             continue
+
+        # Not yet understood. We got here for a non-float non-None value:
+        # https://github.com/conbench/conbench/issues/1155
         units_formatted.add(unit_fmt(f, unit).split(" ", 1)[1])
 
     unit_formatted = units_formatted.pop() if len(units_formatted) == 1 else None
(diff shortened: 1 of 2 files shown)


jgehrcke:
OK, let's see if this helps a tiny bit with #1155. :ship:

alistaire47:
A bit late, sorry, but I'm not quite sure why we're averse to populating the mean (unlike standard deviation) for < 3 data points—a mean is valid with 2 or 1 items, if not as useful. Is there a benefit that makes it worth these workarounds?

jgehrcke:
>  I'm not quite sure why we're averse to populating the mean (unlike standard deviation) for < 3 data points—a mean is valid with 2 or 1 items, if not as useful. 

I fully agree with this sentiment; I wasn't  clear about the depth of the consequences (I probably also wasn't thinking straight :-)), and this was my slow way of rationalizing explicitly that the mean value _is_ a decent "single value summary".

I titled bug #1169 "regression: BenchmarkResult.mean can be None for non-errored results, history analysis assumes numeric value" and am implying that the solution should be "non-errored results are guaranteed to have the mean value populated". Sounds so simple, but it's important to spell it out in code or spec, because I didn't have that crystal :crystal_ball: clarity so far.

Issue #1161, opened 2023-04-24
gunicorn worker shutdown: periodically_set_q_rem does not shut down upon SIGTERM


jgehrcke:
When developing with `make run-app-dev` I noticed that the periodically_set_q_rem thread might delay gunicorn worker process shutdown unnecessarily long, sometimes triggering the  `WORKER TIMEOUT` condition:
```
conbench-app-1  | [230424-10:37:27.379] [4553] [conbench.job] INFO: BMRT cache job (started: True): saw signal 15, set shutdown flag
conbench-app-1  | 
conbench-app-1  | worker_int hook: send myself SIGTERM
conbench-app-1  | [2023-04-24 10:37:27 +0000] [4553] [INFO] Worker exiting (pid: 4553)
conbench-app-1  | [2023-04-24 10:39:27 +0000] [1] [CRITICAL] WORKER TIMEOUT (pid:4553)
conbench-app-1  | Exception ignored in: <module 'threading' from '/usr/local/lib/python3.11/threading.py'>
conbench-app-1  | Traceback (most recent call last):
conbench-app-1  |   File "/usr/local/lib/python3.11/threading.py", line 1583, in _shutdown
conbench-app-1  |     lock.acquire()
conbench-app-1  |   File "/usr/local/lib/python3.11/site-packages/gunicorn/workers/base.py", line 203, in handle_abort
conbench-app-1  |     sys.exit(1)
conbench-app-1  | SystemExit: 1
conbench-app-1  | [2023-04-24 10:39:27 +0000] [4701] [INFO] Booting worker with pid: 4701
```
See, SIGTERM was sent to self.

But then two minutes passed:
```
conbench-app-1  | [2023-04-24 10:37:27 +0000] [4553] [INFO] Worker exiting (pid: 4553)
conbench-app-1  | [2023-04-24 10:39:27 +0000] [1] [CRITICAL] WORKER TIMEOUT (pid:4553)
```

This is a no-go for local dev (the quick app reload upon file saving is an important feature).

jgehrcke:
Done in https://github.com/conbench/conbench/pull/1162.

PR #1162, opened 2023-04-24
metrics/gauge thread: respond to SHUTDOWN signal

jgehrcke:
This is a patch for #1161. It makes the job.SHUTDOWN bool public, and it's consumed in the main loop of said thread. I have confirmed in manual testing that this makes the app reload work again with snappiness.

### DIFF
--- a/conbench/metrics.py
+++ b/conbench/metrics.py
@@ -24,6 +24,8 @@ import flask
 import prometheus_client
 from prometheus_flask_exporter import NO_PREFIX, PrometheusMetrics
 
+from conbench import job
+
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


conbench/metrics.py
```diff
@@ -192,7 +194,17 @@ def periodically_set_q_rem() -> None:
     def func():
         log.info("periodically_set_q_rem(): initiate")
         while True:
-            time.sleep(1)
+            # Delay for a bit until setting the gauge next time. But don't just
+            # sleep, to a busy sleep (responsive sleep loop that inspects
+            # SHUTDOWN often_.
+            deadline = time.monotonic() + 3
+            while time.monotonic() < deadline:
+                if job.SHUTDOWN:
```
jgehrcke, line 202: For the record: a module-level global is just a dictionary value, the key is the name. Threads can coordinates themselves through such simple data structures, i.e. this thread here immediate sees the update (boolean switch) when another thread runs code in job.py which does:

```
    global SHUTDOWN
    SHUTDOWN = True
```

jgehrcke:
:ship:

PR #1164, opened 2023-04-24
units.py: do not call sigfig.round() with math.nan

jgehrcke:
Further follow-up from #1155, after reproducing the exception that we've seen thrown in the wild.

### DIFF
--- a/conbench/units.py
+++ b/conbench/units.py
@@ -1,3 +1,4 @@
+import math
 from typing import Optional
 
 import sigfig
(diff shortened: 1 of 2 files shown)


jgehrcke:
:ship: would love to see if that resolves the crash seen in #1155.

conbench/units.py
```diff
@@ -41,6 +42,11 @@ def fmt_unit(value: Optional[float], unit) -> Optional[str]:
     if value is None:
         return None
 
+    if value == math.nan:
+        # Must not call sigfig.round() with math.nan, see
+        # https://github.com/conbench/conbench/issues/1155
+        return None
```
jgehrcke, line 48: Sigh.

```
>>> math.nan == math.nan
False
```

The identity operator is needed!

PR #1166, opened 2023-04-24
units.py: check for math.nan by identity

jgehrcke:
Interesting! https://github.com/conbench/conbench/pull/1164 did not improve things because:

```
>>> math.nan == math.nan
False

>>> math.nan is math.nan
True
```

Once we have understood which condition really was required for triggering https://github.com/conbench/conbench/issues/1155 we absolutely need to cover that in the test suite.

### DIFF
--- a/conbench/units.py
+++ b/conbench/units.py
@@ -1,7 +1,10 @@
+import logging
 import math
 from typing import Optional
 
 import sigfig
+
+log = logging.getLogger(__name__)
 
 
 def items_per_second_fmt(value, unit):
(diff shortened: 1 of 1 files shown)


jgehrcke:
With this patch, I am now rather confident that this type of exception cannot happen through the same code path anymore:
image

:rocket:

conbench/units.py
```diff
@@ -42,11 +45,17 @@ def fmt_unit(value: Optional[float], unit) -> Optional[str]:
     if value is None:
         return None
 
-    if value == math.nan:
+    if value is math.nan:
```
jgehrcke, line 48: It seems to _not_ be a question of identity :)

> To check whether a number is a NaN, use the [isnan()](https://docs.python.org/3/library/math.html#math.isnan) function to test for NaNs instead of `is` or `==`

Gotcha.

PR #1167, opened 2023-04-24
units.py and history.py: more safeguards in single_value_for_plot() and fmt_unit()

jgehrcke:
We still got this:
image

Throwing more tools and a different approach at this challenge. Also, reading more docs.

https://docs.python.org/3/library/math.html#math.nan
https://docs.python.org/3/library/math.html#math.isnan

### DIFF
--- a/conbench/entities/history.py
+++ b/conbench/entities/history.py
@@ -145,7 +145,7 @@ class HistorySample:
         includes `math.nan`.
         """
         if isinstance(self.mean, float):
-            if not math.isnan(self.value):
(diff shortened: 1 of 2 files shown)


jgehrcke:
While I am rather confident that this will actually help, I used this calm evening to try to build a test that reproduces the problem. Succeeded at that: https://github.com/conbench/conbench/pull/1168. Let's get this in first. And then rebase this branch here.

jgehrcke:
No no, this wasn't a good direction. Superseded by #1172.

PR #1176, opened 2023-04-25
benchalerts: Comparison step for run IDs

### LINKED ISSUE #1031: benchalerts: allow for different comparison entities
Right now `benchalerts` only allows you to use the lookback z-score method to compare a contender commit with its most-recent default-branch baseline commit. But what if you wanted to compare a contender run to the head-of-default-branch baseline commit? We should allow users to be able to specify this.

alistaire47:
Closes #1031. Adds a `GetConbenchZComparisonForRunsStep` as a parent class of `GetConbenchComparisonStep` which operates on a list of run IDs instead of a commit hash.

~As of writing, the new class operates on the new results from the `/api/runs/{run_id}` endpoint with the `candidate_baseline_runs` dict, while the behavior of the old class is completely unchanged. Since versions are pinned everywhere, we probably do want to change the existing class to use the new data (and more of the same code), but for review and testing I haven't changed it yet.~

Both classes now use the new `candidate_baseline_runs` dict from the `/api/runs/{run_id}` endpoint, so this version will depend on those changes being deployed to whatever server they're being used with. Since versions are pinned, this update shouldn't be complicated.

### DIFF
--- a/benchalerts/README.md
+++ b/benchalerts/README.md
@@ -47,7 +47,10 @@ pipeline = AlertPipeline(
             commit_hash=commit_hash,
             baseline_run_type=steps.BaselineRunCandidates.fork_point,
         ),
-        steps.GitHubCheckStep(repo=repo),
+        steps.GitHubCheckStep(
+            comparison_step_name="GetConbenchZComparisonStep",
+            repo=repo
+        ),
     ],
     error_handlers=[
         steps.GitHubCheckErrorHandler(
--- a/benchalerts/README.md
+++ b/benchalerts/README.md
@@ -43,7 +43,10 @@ build_url = (
 # Create a pipeline to update a GitHub Check
 pipeline = AlertPipeline(
     steps=[
-        steps.GetConbenchZComparisonStep(commit_hash=commit_hash),
+        steps.GetConbenchZComparisonStep(
+            commit_hash=commit_hash,
+            baseline_run_type=steps.BaselineRunCandidates.fork_point,
+        ),
         steps.GitHubCheckStep(repo=repo),
     ],
     error_handlers=[
--- a/benchalerts/benchalerts/pipeline_steps/conbench.py
+++ b/benchalerts/benchalerts/pipeline_steps/conbench.py
@@ -81,7 +81,6 @@ class GetConbenchZComparisonForRunsStep(AlertPipelineStep):
 
     def run_step(self, previous_outputs: Dict[str, Any]) -> FullComparisonInfo:
         log.info(f"Getting comparisons from {len(self.run_ids)} run(s)")
-        assert self.baseline_run_type
         return FullComparisonInfo(
             run_comparisons=[
                 self._get_one_run_comparison(run_id) for run_id in self.run_ids
@@ -90,6 +89,7 @@ class GetConbenchZComparisonForRunsStep(AlertPipelineStep):
 
     def _get_one_run_comparison(self, run_id: str) -> RunComparisonInfo:
         """Create and populate one RunComparisonInfo instance."""
+        assert self.baseline_run_type
         run_comparison = RunComparisonInfo(
             contender_info=self.conbench_client.get(f"/runs/{run_id}/")
         )
--- a/benchalerts/benchalerts/pipeline_steps/conbench.py
+++ b/benchalerts/benchalerts/pipeline_steps/conbench.py
@@ -1,6 +1,7 @@
 """Pipeline steps to talk to Conbench."""
 
-from typing import Any, Dict, Optional
+from enum import Enum
+from typing import Any, Dict, List, Optional
 
 from benchclients.conbench import LegacyConbenchClient
 from benchclients.logging import fatal_and_log, log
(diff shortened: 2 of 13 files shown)


alistaire47:
Integration tests are failing because #1078 has not yet been deployed to Velox conbench, so `candidate_baseline_runs` doesn't exist in the `/api/runs/{run_id}` results

benchalerts/benchalerts/pipeline_steps/conbench.py
```diff
@@ -11,17 +12,46 @@
...
+"""
+
+
+class BaselineRunCandidates(Enum):
+    """Types of baselines available from `/api/runs/{run_id}` in the
+    `candidate_baseline_runs` field.
+    """
+
+    fork_point = "fork_point"
+    head_of_default = "head_of_default"
```
austin3dickey: Oops, I changed this name in https://github.com/conbench/conbench/pull/1078. It should be `latest_default`

benchalerts/README.md
```diff
@@ -43,7 +43,10 @@ build_url = (
 # Create a pipeline to update a GitHub Check
 pipeline = AlertPipeline(
     steps=[
-        steps.GetConbenchZComparisonStep(commit_hash=commit_hash),
+        steps.GetConbenchZComparisonStep(
+            commit_hash=commit_hash,
+            baseline_run_type=steps.BaselineRunCandidates.fork_point,
```
austin3dickey, line 48: thanks for updating the README!!

benchalerts/benchalerts/pipeline_steps/conbench.py
```diff
@@ -11,17 +12,46 @@
...
     Parameters
     ----------
-    commit_hash
-        The commit hash of the contender commit to compare. Needs to match EXACTLY what
-        Conbench has stored; typically 40 characters. It can't be a shortened version of
-        the hash.
+    run_ids
+        A list of Conbench run IDs of the runs to compare.
+    baseline_run_type
+        The type of baseline to use. See ``BaselineCandidates`` for options.
```
austin3dickey:
```suggestion
        The type of baseline to use. See ``BaselineRunCandidates`` for options.
```

benchalerts/benchalerts/pipeline_steps/conbench.py
```diff
@@ -11,17 +12,46 @@
...
+    `candidate_baseline_runs` field.
+    """
+
+    fork_point = "fork_point"
+    head_of_default = "head_of_default"
+    parent = "parent"
+
+
+class GetConbenchZComparisonForRunsStep(AlertPipelineStep):
+    (
```
austin3dickey: Unfortunately this doesn't count as a docstring, so the docs don't show up on hover in IDEs or in the rendered HTML on our GitHub Page.
```py
>>> TEST = "string"
>>> class a:
...     """docstring"""
...     pass
...
>>> class b:
...     ("""doc""" + TEST)
...     pass
...
>>> a.__doc__
# 'docstring'
>>> b.__doc__ is None
# True
```
f-strings don't work either, only postprocessing after the class definition. You could do something like
```py
GetConbenchZComparisonForRunsStep.__doc__ += CONBENCH_ENV_VAR_HELP
```
at the end of the file if you'd like.

benchalerts/benchalerts/pipeline_steps/conbench.py
```diff
@@ -11,17 +12,46 @@
...
+    ``CONBENCH_URL``
+        The URL of the Conbench server. Only required if ``conbench_client`` is not
+        provided.
+    ``CONBENCH_EMAIL``
+        The email to use for Conbench login. Only required if ``conbench_client`` is not
+        provided and the server is private.
+    ``CONBENCH_PASSWORD``
+        The password to use for Conbench login. Only required if ``conbench_client`` is
+        not provided and the server is private.
+"""
```
alistaire47: Realized later the github steps just duplicate the equivalent section. Opinions on how DRY we want our docstrings?

benchalerts/benchalerts/pipeline_steps/github.py
```diff
@@ -31,11 +32,14 @@ class GitHubCheckStep(AlertPipelineStep):
         A GitHubRepoClient instance. Either provide this or ``repo``.
     commit_hash
         The commit hash to update. Default is to use the same one that was analyzed in
-        the GetConbenchZComparisonStep earlier in the pipeline.
+        the ``GetConbenchZComparisonStep`` earlier in the pipeline. If
+        ``GetConbenchZComparisonForRunsStep`` is used because the contender lacks commit
+        metadata, ``commit_hash`` must be provided.
```
alistaire47: Pretty sure this is right; pls correct me if not.

benchalerts/benchalerts/pipeline_steps/conbench.py
austin3dickey: This is why I didn't make the docstrings DRY earlier. We could make a separate issue for a decorator like other libraries tend to do.

benchalerts/benchalerts/pipeline_steps/conbench.py
austin3dickey: See my other comment

benchalerts/benchalerts/pipeline_steps/github.py
```diff
@@ -31,11 +32,14 @@ class GitHubCheckStep(AlertPipelineStep):
...
     comparison_step_name
-        The name of the ``GetConbenchZComparisonStep`` that ran earlier in the pipeline.
-        Defaults to "GetConbenchZComparisonStep" (which was the default if no name was
-        given to that step).
+        The name of the ``GetConbenchZComparisonStep`` or
+        ``GetConbenchZComparisonForRunsStep`` that ran earlier in the pipeline.
+        Defaults to "GetConbenchZComparisonStep" (which was the default for that step if
+        no name was given to that step).
```
alistaire47: Don't love this default anymore, but I'm not seeing a simple obviously better option

benchalerts/benchalerts/pipeline_steps/conbench.py
```diff
@@ -109,10 +122,8 @@ def _get_one_run_comparison(self, run_id: str) -> RunComparisonInfo:
 
         else:
             log.warning(
-                "Conbench could not find a baseline run for the contender run "
-                f"{run_id}. A baseline run needs to be on the default branch in the "
-                "same repository, with the same hardware, and have at least one of the "
-                "same benchmark case/context pairs."
+                f"Conbench could not find a {self.baseline_run_type.value} baseline run "
+                f"for the contender run {run_id}. Error: {candidate_baseline_run['error']}"
```
austin3dickey, line 119: Nice!

benchalerts/benchalerts/pipeline_steps/conbench.py
```diff
@@ -121,3 +132,69 @@ def _get_one_run_comparison(self, run_id: str) -> RunComparisonInfo:
                 )
 
         return run_comparison
+
+
+class GetConbenchZComparisonStep(GetConbenchZComparisonForRunsStep):
```
austin3dickey, line 130: This is a very clever way to maintain this feature, thanks!

benchalerts/benchalerts/pipeline_steps/github.py
austin3dickey: This defaulting behavior seems a little too magic now. I wonder if we should remove the default of this arg.

benchalerts/benchalerts/pipeline_steps/github.py
```diff
@@ -31,11 +32,14 @@ class GitHubCheckStep(AlertPipelineStep):
...
+        the ``GetConbenchZComparisonStep`` earlier in the pipeline. If
+        ``GetConbenchZComparisonForRunsStep`` is used because the contender lacks commit
+        metadata, ``commit_hash`` must be provided.
     comparison_step_name
-        The name of the ``GetConbenchZComparisonStep`` that ran earlier in the pipeline.
-        Defaults to "GetConbenchZComparisonStep" (which was the default if no name was
-        given to that step).
+        The name of the ``GetConbenchZComparisonStep`` or
+        ``GetConbenchZComparisonForRunsStep`` that ran earlier in the pipeline.
+        Defaults to "GetConbenchZComparisonStep" (which was the default for that step if
```
austin3dickey: oh, this is messy too, since the default step name is the class name. This argument probably shouldn't have a default anymore either.

benchalerts/tests/integration_tests/test_alert_pipeline.py
```diff
@@ -75,10 +75,16 @@ def test_alert_pipeline(monkeypatch: pytest.MonkeyPatch, github_auth: str):
     # now a real pipeline
     pipeline_steps = [
         steps.GetConbenchZComparisonStep(
-            commit_hash=velox_commit, z_score_threshold=None, step_name="z_none"
+            commit_hash=velox_commit,
+            baseline_run_type=steps.BaselineRunCandidates.fork_point,
```
austin3dickey: I think this velox commit is on the default branch, so this won't be populated. Maybe choose `parent`.

You can `pytest.skip` this integration test for now until we deploy the latest API changes there.

austin3dickey (commented):
Looking great!! Thank you

benchalerts/benchalerts/pipeline_steps/github.py
austin3dickey: ah yeah, see my other comment

benchalerts/benchalerts/pipeline_steps/github.py
alistaire47: That's fine, though we'll need to update all the implementations. I guess we need to anyway to specify `baseline_run_type` though

austin3dickey: Yeah, this is already a breaking change.

benchalerts/benchalerts/pipeline_steps/conbench.py
alistaire47: I was going by [staging redoc](https://cb-staging.voltrondata.run/api/redoc#tag/Runs/paths/~1api~1runs~1%7Brun_id%7D~1/get) which still says `head_of_default`? I'm confused

benchalerts/benchalerts/pipeline_steps/conbench.py
alistaire47: Oh this is annoying and against my expectations; I'll revert. Putting it at the end of the file would make it too inobvious it's there when reading the code imo.

benchalerts/benchalerts/pipeline_steps/github.py
alistaire47: Removed, which meant I had to move the param to first since everything else has defaults, and `None` doesn't mean anything here. In tests we were mostly setting these names anyway, so updating wasn't bad; hopefully implementations are similar. And hopefully we're not relying on positional args in usages anywhere.

benchalerts/benchalerts/pipeline_steps/conbench.py
austin3dickey: Oh no I forgot to change it there! Will do in a followup

alistaire47: I could change it here if it's just docs, searching for `head_of_default` it looks like it's in a fair number of places still...

alistaire47: Wait, mostly `test_history.py`. Should I just fix it?

austin3dickey: I opened https://github.com/conbench/conbench/pull/1179 already.

I'm leaving the phrase `head_of_default` in `test_history.py` on purpose, because it's a slightly different thing. It's not talking about a candidate baseline run, just a strategy for pytesting z-scores.

austin3dickey (approved):
thanks again!

benchalerts/benchalerts/pipeline_steps/conbench.py
```diff
@@ -121,3 +125,80 @@ def _get_one_run_comparison(self, run_id: str) -> RunComparisonInfo:
...
+        The password to use for Conbench login. Only required if ``conbench_client`` is
+        not provided and the server is private.
+    """
+
+    def __init__(
+        self,
+        commit_hash: str,
+        baseline_run_type: BaselineRunCandidates,
+        z_score_threshold: Optional[float] = None,
+        conbench_client: Optional[LegacyConbenchClient] = None,
```
jgehrcke, line 180: What do we think about using the new one (`ConbenchClient` backed by `RetryingHTTPClient`) with a max-retry-time adjusted to the context this here runs in (like: retry for 90 seconds or so)?

benchalerts/tests/integration_tests/test_pipeline_steps/test_conbench.py
```diff
@@ -15,54 +15,71 @@
...
             7,
             True,
         ),
     ],
 )
 def test_GetConbenchZComparisonStep(
-    monkeypatch: pytest.MonkeyPatch, conbench_url, commit, expected_len, expected_bip
+    monkeypatch: pytest.MonkeyPatch,
+    conbench_url,
+    commit,
```
jgehrcke: Let's rename this to `commit_hash`, for consistency and to distinguish this from more complex commit objects.

jgehrcke: Also, maybe a type annotation? I was literally just wondering what kind of object `commit` might be, and of course this might be addressed via type annotation and/or more explicit naming.

jgehrcke (approved):
:rocket:

Nice!

Have left few remarks for potential improvement.

benchalerts/benchalerts/pipeline_steps/conbench.py
austin3dickey: I like this idea but let's do it in a followup PR. I'll make an issue.

What will make this take a while is that this library's approach to mocking a Conbench server in unit tests makes heavy use of the requests `adapter` paradigm, which the new client stopped using. So unfortunately it will be a little more than a one-line change.

austin3dickey: #1186

benchalerts/tests/integration_tests/test_pipeline_steps/test_conbench.py
alistaire47: oh good calls, done

PR #1182, opened 2023-04-27
UI: fix #1153 as well as #1185

### LINKED ISSUE #1153: c-bench: `rel err` column sort order broken when "n/a" values are present
### unsorted:
image

### sorted:
image

8.2 is not the largest!

jgehrcke:
Fixes #1153 as well as #1185.

Reference docs: https://datatables.net/examples/advanced_init/html5-data-attributes.html

See title.

---

### desc
Screenshot from 2023-04-27 13-17-05

### asc

Screenshot from 2023-04-27 13-17-01

### DIFF
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -78,13 +78,22 @@ class TimeSeriesPlotMixin:
         """
         samples = get_history_for_benchmark(benchmark_result_id=benchmark_result.id)
 
+        # Does `samples` include the current benchmark result? Interesting: the
+        # current benchmark result may be failed, but I think all items in
+        # samples (historical benchmark results are guaranteed to not be
+        # failed.
+
         # The number (1, 2, 3?) maybe needs to be tuned further. Also see
         # https://github.com/conbench/conbench/issues/867
         if len(samples) > 2:
             assert isinstance(samples[0], HistorySample)
             jsondoc = json.dumps(
                 bokeh.embed.json_item(
-                    time_series_plot(samples, benchmark_result, run),
+                    time_series_plot(
+                        samples=samples,
+                        current_benchmark_result=benchmark_result,
+                        run=run,
+                    ),
                     f"plot-history-{i}",  # type: ignore
                 )
             )
@@ -571,11 +580,14 @@ def time_series_plot(
         ),
     )
 
-    source_current_bm_mean = _source(
-        [dummy_hs_for_current_svs],
-        unit,
-        formatted=formatted,
-    )
+    # Edge case: this may be failed, in which case we cannot show a data point!
+    source_current_bm_mean = "failed"
+    if not current_benchmark_result.is_failed():
+        source_current_bm_mean = _source(
+            [dummy_hs_for_current_svs],
+            unit,
+            formatted=formatted,
+        )
 
     # Create same dummy structure, only difference: set min as single value
     # summary (SVS).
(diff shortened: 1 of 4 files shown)


conbench/templates/c-benchmark-results-for-case.html
```diff
@@ -44,7 +44,7 @@ <h3>
               <td class="font-monospace">{{ result.context.id[:8] }}</td>
               <td class="font-monospace">{{ result.ui_non_null_sample_count }}</td>
               <td class="font-monospace">{{ result.ui_mean_and_uncertainty }}</td>
-              <td class="font-monospace">{{ result.ui_rel_sem }}</td>
+              <td class="font-monospace" data-order="{{ result.ui_rel_sem[0] }}">{{ result.ui_rel_sem[1] }}</td>
```
jgehrcke, line 47: TODO: make this not be two function calls

jgehrcke:
:ship:

conbench/app/_plots.py
```diff
@@ -571,11 +580,14 @@ def time_series_plot(
...
     )
 
-    source_current_bm_mean = _source(
-        [dummy_hs_for_current_svs],
-        unit,
-        formatted=formatted,
-    )
+    # Edge case: this may be failed, in which case we cannot show a data point!
+    source_current_bm_mean = "failed"
+    if not current_benchmark_result.is_failed():
```
jgehrcke, line 585: Previously, we showed this data point, I think, despite the benchmark result being "failed".

Interesting, CC @jonkeane

conbench/app/_plots.py
```diff
@@ -78,13 +78,22 @@ def get_history_plot(
         """
         samples = get_history_for_benchmark(benchmark_result_id=benchmark_result.id)
 
+        # Does `samples` include the current benchmark result? Interesting: the
```
jgehrcke, line 81: answer is "no" I think, but should probably be spelled out explicitly in the docstring(s) of the history-related function(s).

conbench/app/_plots.py
```diff
@@ -78,13 +78,22 @@ def get_history_plot(
...
+        # current benchmark result may be failed, but I think all items in
+        # samples (historical benchmark results are guaranteed to not be
+        # failed.
```
jgehrcke, line 84: Interesting consideration, CC @austin3dickey -- we should also specify this properly in related docstring(s).

conbench/entities/benchmark_result.py
```diff
@@ -486,23 +486,30 @@ def ui_mean_and_uncertainty(self) -> str:
 
     # maybe make this a cached property
     @property
-    def ui_rel_sem(self) -> str:
+    def ui_rel_sem(self) -> Tuple[str, str]:
+        """
+        The first string is a stringified float for sorting in the table.
+        The second string is for display in the table, with unit.
+        Associate "n/a" with a negative value so that DESC sorting by error
+        is helpful.
```
jgehrcke, line 494: Will need to make this interface better in the (near) future. This works for now, moving on.

conbench/app/_plots.py
austin3dickey: Yeah! We do specify that history samples always consist of non-errored benchmark results:
https://github.com/conbench/conbench/blob/6b12e2b7152840105c379341ed48a94ba451c4ad/conbench/entities/history.py#L188
but I don't think we wrote down that you can still get history _given_ an errored benchmark result

PR #1172, opened 2023-04-25
guarantee BenchmarkResult.mean again, add clarity in many ways, introduce 'single value summary' to feed plot and history api

jgehrcke:
This was motivated by #1169 and #1155. The methodological fixes in here are almost boring, but where I hope the actual value is:

- clarification of assumptions and intent
- introduction of a 'single value summary' concept/interface (naming is hard)
- exposure of the good old BenchmarkResult object in the _plots module (that was soo much needed)
- more deliberate test coverage
- rather significant overhaul of some involved type checks

What's maybe not visible from the patch itself: type checking here was _incredibly_ useful, raising really important questions along the way; it's not just overhead keeping us busy, it really reveals those aspects where our intuition let us down and where thinking just wasn't right. The tighter and more correct the type annotations get, the more productive the feedback gets.

There's more to this patch, please also have a look at commit messages and code comments. They are all not perfect and might appear a little sloppily written here and there. But I figured quickly typing down thoughts is better than not doing it.

And in case it isn't obvious: so much sweat in this PR. Everything fell apart (largely self-inflicted, yes) and it was difficult to find out why.

Edit:

> introduction of a 'single value summary' concept/interface (naming is hard)

This makes it easier to do https://github.com/conbench/conbench/issues/594 and https://github.com/conbench/conbench/issues/640.

PR #1183, opened 2023-04-27
BenchmarkResult single_value_summary: address special case

jgehrcke:
For https://github.com/conbench/conbench/pull/1172#issuecomment-1525421877.

### DIFF
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -441,6 +441,7 @@ class BenchmarkResult(Base, EntityMixin):
             # being failed_. Because of a temporary logic error. Let's remove
             # this code path again for sanity. Since is_failed() returned False
             # we know that `self.data` has only numbers, and at least one.
+            assert self.data is not None
             return float(statistics.mean(self.data))
 
         return float(self.mean)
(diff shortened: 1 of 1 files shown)


conbench/entities/benchmark_result.py
```diff
@@ -435,10 +435,15 @@ def _single_value_summary(self) -> float:
...
-        # Should be implied by the outcome of self.is_failed(), encode this.
-        assert self.mean is not None
+        if self.mean is None:
+            # See https://github.com/conbench/conbench/issues/1169 -- Legacy
+            # database might have mean being None _despite the benchmark not
+            # being failed_. Because of a temporary logic error. Let's remove
+            # this code path again for sanity. Since is_failed() returned False
+            # we know that `self.data` has only numbers, and at least one.
+            assert self.data is not None
+            return float(statistics.mean(self.data))
```
jgehrcke, line 445: Not beautiful, but with all the commentary and the is_failed() abstraction I think this is easy enough to reason about in the future. 

To me, the biggest takeaway is that we want to enhance our test coverage, because so far we had so many non-exercised scenarios in the test suite: https://github.com/conbench/conbench/issues/1184

jgehrcke:
:ship: OK, I am excited about this.

Issue #1165, opened 2023-04-24
API: allow submitting runs with known repository but unknown commit hash


austin3dickey:
Currently, if
- a Conbench instance tracks multiple repositories
- on that instance, someone hits `/api/runs/{run_id}` for a contender run that is not associated with a commit
- they inspect the `latest_default` baseline run of the response

That baseline run could come from any of the repositories tracked, depending on which one has the latest commit. This is confusing. But right now it's the best we can do with the information given, since we were never told which repository the contender run is associated with.

If the contender run _actually_ came from a certain repository, we should be able to tell Conbench that, even if it wasn't associated with a commit hash on the GitHub-backed git graph source of truth for that repository.

Related:
- https://github.com/conbench/conbench/pull/1134#issuecomment-1518240762 (the PR where we allowed for Runs to be unassociated with a Commit object)
- https://github.com/conbench/conbench/pull/1078#discussion_r1175419772 (where we discovered this issue)

jgehrcke:
Just came by this one! Didn't even read properly yet, but let's certainly reconcile this with https://github.com/conbench/conbench/issues/817 :). :rocket: New insights / arguments are of course always great!

jgehrcke:
Unfinished thoughts; maybe not for tackling this issue specifically; but in general:

1) What about adding a `repo_url` property directly to BenchmarkResult, and making an index for that (we know it has few values)? That would imply consistency challenges (consistency not maintained by DB, but by app code)
2) For maximum consistency: what about adding a Repository table, and then relating BenchmarkResult to it, as well as Commit? (removing repo information from the Commit table)?

With (2):
- a BenchmarkResult can be enforced to be associated to a Repo.
- a BenchmarkResult can again optionally be associated with a Commit (for e.g. pre-merge capability)
- a Commit can be enforced to be associated with a Repo
- the triangle between BenchmarkResuit-Repo-Commit can be enforced to be consistent

A Run can also be associated with a just a Repo, or also a Commit (enforcing consistency between Run-Repo-Commit). 

To me this sounds like a sane, powerful data model removing quite a bit of ambiguity from the application code.

Could also again reduce the importance of Run as a grouping construct, and would directly relate the Result entity to Repo / Commit info (as I think we identified as a favorable direction elsewhere).

austin3dickey:
I do really like (2) and how it allows us to reduce the importance of Run. It could potentially slightly speed up certain queries as well.

alistaire47:
[...]

PR #1194, opened 2023-04-28
api/compare: cleanup

### LINKED ISSUE #1075: api/compare: greatly simplify, clarify, and scope the inputs and outputs
This is mostly a continuation of the discussions in #1030.

Full proposal:
- get rid of `/api/compare/batches`
- get rid of `/api/compare/commits` (to be replaced with the workflow described in https://github.com/conbench/conbench/issues/384#issuecomment-1500582537)
- rename `/api/compare/benchmarks` to `/api/compare/benchmark-results`
- in `/api/compare/benchmark-results` and `/api/compare/runs/`, clarify the comparison dict by separating keys into `baseline` (raw info), `contender` (raw info), and `analyses` sub-dicts
- `/api/compare/runs/` should return the full outer join of all results on the baseline and contender run, matched on case/context/hardware/repo

### LINKED ISSUE #1092: api/compare/runs: Return full outer join of all results
Splitting out of #1075 to keep PRs tidy and small. `/api/compare/runs/` should return the full outer join of all results on the baseline and contender run, matched on case/context/hardware/repo

austin3dickey:
This PR cleans up the code for the compare API. Despite the diff ("548 additions and 2,121 deletions"), there is not much behavior change.

- I didn't like the separation of `entities/_comparator.py` from `api/compare.py`. Many lines of code from both files relied implicitly on each other, and it was too easy to make mistakes, so I moved the important things into `api/compare.py` and deleted the rest.
- Added many type hints. Caught a lot of bugs thanks to `mypy`.
- Started using the new single-value summary instead of the hardcoded `mean`.
- Made the analysis objects nullable in the API response if they don't apply to that comparison.
- Deleted a lot of unused code.
- Made the `/compare/runs` API return a simple full outer join of the benchmark results, instead of the very strange deduplicating logic it had before. Fixes https://github.com/conbench/conbench/issues/1092.
- Add lots more documentation to the compare API endpoints.


conbench/api/_docs.py
```diff
@@ -78,7 +78,6 @@ def _201_created(example, schema=None):
 spec.components.response("CommitList", _200_ok([ex.COMMIT_ENTITY]))
 spec.components.response("CompareEntity", _200_ok(ex.COMPARE_ENTITY))
 spec.components.response("CompareList", _200_ok(ex.COMPARE_LIST))
-spec.components.response("CompareBenchmarkResult", _200_ok(ex.COMPARE_BENCHMARK_RESULT))
```
austin3dickey, line 81: This was unused.

conbench/entities/benchmark_result.py
```diff
@@ -332,12 +318,10 @@ def to_dict_for_json_api(benchmark_result):
...
-                # Note(JP): it's interesting that this was added here into
-                # `stats`, that's not a nice separation between input and
-                # output.`
-                "z_score": benchmark_result.z_score,
-                "z_regression": z_regression(benchmark_result.z_score),
-                "z_improvement": z_improvement(benchmark_result.z_score),
+                # TODO: remove these.
+                "z_score": None,
+                "z_regression": False,
+                "z_improvement": False,
```
austin3dickey, line 324: This change makes our previous implicit behavior explicit.

conbench/entities/history.py
```diff
@@ -731,6 +730,12 @@ def _add_rolling_stats_columns_to_df(
     return df
 
 
+def _less_is_better(unit) -> bool:
+    if unit in ["B/s", "i/s"]:
+        return False
+    return True
```
austin3dickey, line 736: I think here is a better place to put this since it's the only place it's used.

austin3dickey:
I'm going to go through the tests one more time to make sure I covered all the test cases I wanted to, but other than that this is ready for review/merge.

conbench/api/_examples.py
```diff
@@ -191,7 +191,7 @@ def _api_compare_entity(
             "benchmark_name": benchmark_name,
             "case_permutation": case_permutation,
             "language": "Python",
-            "value": 0.03637,
+            "single_value_summary": 0.03637,
```
jgehrcke, line 194: :heart:

conbench/entities/benchmark_result.py
jgehrcke: > This change makes our previous implicit behavior explicit.

I might be slow here, but why do we think we  can't we remove these yet?

conbench/tests/api/_expected_docs.py
```diff
@@ -420,6 +359,7 @@
                                 "error": None,
                                 "language": "Python",
                                 "run_id": "some-run-uuid-2",
+                                "single_value_summary": 0.03637,
```
jgehrcke, line 362: thank you for adopting this term here. consistency is great!

conbench/entities/_comparator.py
```diff
@@ -1,241 +0,0 @@
...
-
-    def compare(self):
-        for pair in self.pairs.values():
-            baseline, contender = pair.get("baseline"), pair.get("contender")
-            yield BenchmarkResultComparator(
-                baseline,
-                contender,
-                self.threshold,
-                self.threshold_z,
-            ).compare()
```
jgehrcke, line 241: :rocket: :guitar: This part of the diff looks especially great. For real. :)

conbench/entities/benchmark_result.py
```diff
@@ -285,19 +284,6 @@ def update(self, data):
 
     def to_dict_for_json_api(benchmark_result):
         # `self` is just convention :-P
-        #
-        # Note(JP): Communicating the z-score as part of the benchmark result
-        # object is a legacy approach. I think in the future we want to have a
-        # property called `analyses` (or so, conceptually); and then this is
-        # either empty, or shows details per-method. That could look like:
```
jgehrcke, line 292: Ah right! I didn't remember at all to have written that.

conbench/api/compare.py
```diff
@@ -1,146 +1,273 @@
...
+        """For type hints only. This represents a BenchmarkResult that has been
+        augmented in-place with the following augmentation functions, as is done in this
+        module:
 
-        return threshold, threshold_z
+        - set_z_scores()
+        - set_display_benchmark_name()
+        - set_display_case_permutation()
 
+        TODO: remove this and replace with actual BenchmarkResult properties.
```
jgehrcke, line 78: :+1:

conbench/api/compare.py
```diff
@@ -1,146 +1,273 @@
...
 
-class CompareEntityEndpoint(ApiEndpoint, CompareMixin):
-    def _get_results(self, rid: str) -> List[BenchmarkResult]:
-        raise NotImplementedError
+        display_bmname: str
+        display_case_perm: str
+        z_score: Optional[float]
+
+
+class UnmatchingUnitsError(Exception):
```
jgehrcke, line 86: Nice.

Totally leave this as-is. Just sharing: in my Python life I never really settled on whether I like more:
- exception names with 'constant' prefix (precision towards the end, hierarchy obvious from names) -- that is what most libraries do.
- exceptions with constant suffix (precision in the beginning of the name, grouping still possible via tail end), often shorter names.

Maybe at some point we can look at all the custom exception types we have added to the code base and come up with a neat naming scheme.

conbench/api/compare.py
```diff
@@ -1,146 +1,273 @@
...
+    pass
+
+
+class BenchmarkResultComparator:
+    """Dataclass to hold the comparison of two BenchmarkResults."""
```
jgehrcke: Thanks for this docstring, even simple things like this can make it drastically easier to read code.

Buut is it a dataclass?

conbench/api/compare.py
```diff
@@ -1,146 +1,273 @@
...
+        return benchmark_result
 
     @maybe_login_required
-    def get(self, compare_ids) -> f.Response:
+    def get(self, compare_ids: str) -> f.Response:
         """
         ---
-        description: Compare benchmark results.
+        description: |
+            Compare benchmark results.
```
jgehrcke: two? many?

conbench/api/compare.py
```diff
@@ -1,146 +1,273 @@
...
+
+            Returns basic information about the baseline and contender benchmark results
+            as well as some analyses comparing the performance of the contender to the
+            baseline.
+
+            The `pairwise` analysis computes the percentage difference of the
+            contender's mean value to the baseline's mean value. The reported difference
+            is signed such that a more negative value indicates more of a performance
+            regression. This difference is then thresholded such that values more
```
jgehrcke, line 243: Thank you!

conbench/api/compare.py
```diff
@@ -1,146 +1,273 @@
...
+            extreme than the threshold are marked as `regression_indicated` or
+            `improvement_indicated`. The threshold is 5.0% by default, but can be
+            changed via the `threshold` query parameter, which should be a positive
+            number representing a percent.
```
jgehrcke: just "in percent."?

conbench/api/compare.py
```diff
@@ -1,146 +1,273 @@
...
+
+            The `pairwise` analysis may be `null` if either benchmark result does not
+            have a mean value, or if the baseline result's mean value is 0.
```
jgehrcke, line 250: using the svs term here could be done, but we can also do that later, right?

conbench/api/compare.py
```diff
@@ -1,146 +1,273 @@
...
+
+            The `lookback_z_score` analysis computes the z-score of the contender's mean
+            value compared to a distribution of benchmark results ending with the
+            baseline result. The reported z-score is also signed such that a more
```
jgehrcke: "compares the contender's mean value to a distribution of previously obtained benchmark result mean values via the so-called z-score method"?

>  ending with the  baseline result

"ending with" is interesting. The reader thinks: "Is this the end towards the past or towards the future? :D "

```
(past) -----------------------------------------------------------------------------> time (future)
                 baseline   x  x       x   x       x   x             contender
```
or
```
(past) -----------------------------------------------------------------------------> time (future)
                    x  x       x   x       x   x      baseline       contender
```


The docstring is already great, and you don't need to think about this at all. I just had this thought that I wanted to write down, maybe we can use my questions here in the future if it's useful.

conbench/api/compare.py
```diff
@@ -1,146 +1,273 @@
...
+            negative value indicates more of a performance regression, and thresholded.
+            The threshold z-score is 5.0, but can be changed via the `threshold_z` query
+            parameter, which should be a positive number.
+
+            The `lookback_z_score` may be `null` if the contender benchmark result is
+            missing a z-score, due to not finding a baseline distribution that matches
```
jgehrcke: > The `lookback_z_score` may be `null` if the contender benchmark result is   missing a z-score

Is that like saying "the cow might be black if the cow is black"? :) :cow:

conbench/api/compare.py
```diff
@@ -1,146 +1,273 @@
...
+            the contender benchmark result. More details about this analysis can be
+            found at https://conbench.github.io/conbench/pages/lookback_zscore.html.
```
jgehrcke: Cross-linking to docs: perfect, I was waiting for that. I'd still prefer linking to github-rendered markdown though, especially from within code. I think that's less of a moving target. (just my opinion :shrug:)

conbench/api/compare.py
```diff
@@ -1,146 +1,273 @@
...
+
+            If either benchmark result is not found, this endpoint will raise a 404. If
```
jgehrcke, line 266: with a good err msg right?

conbench/api/compare.py
```diff
@@ -160,53 +287,131 @@ def get(self, compare_ids) -> f.Response:
...
+        there are no results for the run.
+        """
+        result = Session.scalars(
+            s.select(BenchmarkResult).where(BenchmarkResult.run_id == run_id)
+        ).all()
 
-        return comparator.compare()
+        if not result:
+            f.abort(
+                404, description=f"no benchmark results found for run ID: '{run_id}'"
```
jgehrcke, line 341: :+1:

conbench/api/compare.py
```diff
@@ -160,53 +287,131 @@ def get(self, compare_ids) -> f.Response:
...
+            baseline_results_by_cchru[_generate_key(result)].append(result)
+
+        for result in contender_results:
+            contender_results_by_cchru[_generate_key(result)].append(result)
+
+        joined_results: List[
+            Tuple[Optional[BenchmarkResult], Optional[BenchmarkResult]]
+        ] = []
+
+        for cchru in set(baseline_results_by_cchru) | set(contender_results_by_cchru):
```
jgehrcke, line 390: thanks for embracing cchru! we should come up with a pronunciation for it! sounds funny for sure!

jgehrcke (approved):
This is amazing. Thank you. 

> This PR cleans up the code for the compare API. Despite the diff ("548 additions and 2,121 deletions"), there is not much behavior change.
> Added many type hints. Caught a lot of bugs thanks to mypy.

So happy that we're all contributing to making the codebase easier to navigate / more productive to work in over time. This is another important step that you've just contributed here :muscle: 

 I have left a few comments, but as always in our current phase the spirit should be to integrate, learn, iterate. :ship:

conbench/api/_docs.py
jonkeane: I'm a little surprised that this is unused? Do we no longer have the ability to get benchmark result comparisons from the API?

I checked just now, and this might be true (with what's deployed to the arrow instance, so it's not reflective of changes in _this_ PR, of course, but validating my assumptions here): https://conbench.ursa.dev/api/compare/benchmarks/805383b072ac488db96ef1ec6bd941c3...8803a118ab6245b7b244d6227ffd9530/ returns a 404 for me. Or did this move somewhere else? If we've gotten rid of it entirely, that's not great — one should be able to use the API to get this kind of information (and one definitely used to be able to!).

conbench/api/compare.py
```diff
@@ -1,146 +1,273 @@
...
-
-def _get_pairs(baseline_items, contender_items):
+
+def _parse_two_ids_or_abort(compare_ids: str) -> Tuple[str, str]:
+    """Split a string of the form "<id>...<id>" into two strings, or abort if it's
+    not of the correct form.
+
+    The input string is the standard way to specify two IDs for the compare API.
+
+    The two IDs can be either two run IDs or two benchmark result IDs.
```
jonkeane, line 29: very nice!

conbench/api/_docs.py
austin3dickey: We do still have the ability to compare benchmark results today, but that entity is represented by `CompareEntity` in docs, not `CompareBenchmarkResult`. The latter was a legacy object from a while ago!

austin3dickey: >https://conbench.ursa.dev/api/compare/benchmarks/805383b072ac488db96ef1ec6bd941c3...8803a118ab6245b7b244d6227ffd9530/ returns a 404 for me

We also renamed the endpoint - https://conbench.ursa.dev/api/compare/benchmark-results/805383b072ac488db96ef1ec6bd941c3...8803a118ab6245b7b244d6227ffd9530/ works. This rename was done in #1094.

conbench/entities/benchmark_result.py
austin3dickey: I bet we can. But I wanted to limit this PR to only the compare API for now. I had to touch this file since we can't import the `z_regression()` etc functions from `_comparator.py` anymore.

conbench/api/compare.py
austin3dickey: Good idea!

conbench/api/compare.py
austin3dickey: Haha, no it's not. Though it is a class meant to store data and provide an easy interface for accessing it. I can clarify ;)

Hopefully this will be turned into a pydantic (or equivalent) class in the future that can be re-used for docs and validation of API schema.

conbench/api/_docs.py
jonkeane: Awesome, thanks for that! And sorry I forgot about the renaming

conbench/api/compare.py
austin3dickey: Yep I was thinking we do that later.

conbench/api/compare.py
austin3dickey: Changed to "which should be a positive percent value"

conbench/api/compare.py
austin3dickey: Changed to "Compare a baseline and contender benchmark result."

conbench/api/compare.py
austin3dickey: >"ending with" is interesting.

Yes, I wasn't super happy with how that sentence ended up either. Really, the distribution is a somewhat-complicated function of the baseline result's commit, which might not be as simple as "ending with" it. I'll tweak this a bit.

conbench/api/compare.py
austin3dickey: Oops, forgot a few words!

conbench/api/compare.py
austin3dickey: Yes - `no benchmark result found with ID: '{benchmark_result_id}'`

conbench/api/compare.py
austin3dickey: Haha, it's really close to "church". We could backronym that to mean "case/hardware/unit/repository/context/happiness" 🙂

conbench/api/compare.py
```diff
@@ -1,146 +1,275 @@
...
+    @staticmethod
+    def result_info(result: Optional["AugmentedBenchmarkResult"]) -> Optional[dict]:
+        if not result:
+            return None
+
+        return {
+            "benchmark_result_id": result.id,
+            "benchmark_name": result.display_bmname,
+            "case_permutation": result.display_case_perm,
+            "language": result.context.tags.get("benchmark_language", "unknown"),
```
alistaire47, line 145: I think it's validated somewhere (well, by using it) that this has to be populated; results without it fail to post. But the defensiveness is good in case that ever gets relaxed

conbench/api/compare.py
```diff
@@ -160,53 +289,131 @@ def get(self, compare_ids) -> f.Response:
...
+        If a case/context/hardware/repo/unit combination is present in one list but not
+        the other, it will still be included but the other tuple element will be None.
+
+        If there are multiple results for a case/context/hardware/repo/unit combination
+        in both lists, a cartesian product of them all will be returned.
+        """
+        _key_type = Tuple[str, str, str, Optional[str], Optional[str]]
+
+        def _generate_key(result: BenchmarkResult) -> _key_type:
+            repo = result.run.commit.repository if result.run.commit else None
```
alistaire47, line 363: IMO we should accept `"github": {"repo": <repo>}` instead of nothing for commitless results, but that's beyond the scope here probably

alistaire47 (approved):
Nice simplification!

conbench/api/compare.py
austin3dickey: Yeah, we will come back to this. #1165

conbench/entities/benchmark_result.py
jgehrcke: Good choice!

Issue #1181, opened 2023-04-27
webapp exception handler fails with `...Error is not JSON serializable`


jgehrcke:
Seen this so many times in logs, needs addressing:
![image](https://user-images.githubusercontent.com/265630/234837571-2a136b33-67b4-4ce5-a545-a193b9f06726.png)

Not every exception can be turned into a JSON object, of course. 

What we can do:
- handle this type of exception and render into something way more silent / less verbose
- actually expose the orig traceback to the http client, like a debug server (this saves the roundtrip of going to the webapp logs, in our stage of development this might be a productive choice).
- an intermediate approach: do not expose full traceback (generally, there are security concerns associated with such an approach -- I think they don't apply here yet, but well), but then show the core exception error msg.

jgehrcke:
Done :muscle:

PR #1195, opened 2023-04-28
webapp: consolidate exception treatment in HTTP handlers

jgehrcke:
For issue #1181.
Ref docs: https://flask.palletsprojects.com/en/2.2.x/errorhandling/#generic-exception-handlers

### DIFF
--- a/conbench/__init__.py
+++ b/conbench/__init__.py
@@ -17,6 +17,7 @@ import importlib.metadata as importlib_metadata
 import json
 import logging
 import os
+import traceback
 from typing import TYPE_CHECKING
 
 import conbench.logger
@@ -134,6 +135,8 @@ def _init_flask_application(app):
     app.register_blueprint(api, url_prefix="/api")
     app.register_blueprint(api_docs, url_prefix="/api/docs")
     app.register_error_handler(werkzeug.exceptions.HTTPException, _json_http_errors)
+    app.register_error_handler(Exception, _handle_all_other_exceptions)
+
     _init_api_docs(app)
 
     @app.before_request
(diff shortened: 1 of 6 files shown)


conbench/api/index.py
```diff
@@ -207,5 +207,21 @@ def wipe_db():
...
+        f.abort(400, description="foo")
+
+    @api.route("/raise-unexpected", methods=("GET",))
+    def raise_unexpected():
+        """
+        For local development / developer productivity.
+        """
+        raise Exception("doh")
+        # raise AssertionError("doh!")
+        # assert True == False
```
jonkeane: Can we remove these?

jonkeane (approved):
Thanks for this!

conbench/__init__.py
```diff
@@ -166,32 +169,48 @@ def _init_api_docs(application):
...
-    # part of requently called error handler (even if the import machinery is
-    # fast and cached) is a code smell -- let's see about this.
     import flask as f
 
-    data = {"code": exc.code, "name": exc.name}
+    # For now: even in 'prod', expose the traceback corresponding to the
+    # exception in an HTTP 500 response, facilitating debugging; so that users
+    # can send tracebacks right away.
+    tbstr = "".join(traceback.format_exception(exc))
+    return f.make_response((f"unexpected exception, please report this:\n{tbstr}", 500))
```
austin3dickey, line 184: "unexpected exception" is very fun to read and say out loud

conbench/tests/app/_asserts.py
```diff
@@ -52,7 +52,8 @@ def assert_404_not_found(self, r):
         # TODO: How to make this an HTML error page?
         # While still having json error rs for the API.
         assert r.content_type == "application/json", r.content_type
-        assert r.json == {"code": 404, "name": "Not Found"}, r.json
+        assert r.json["code"] == 404, "name"
```
austin3dickey: I think "name" isn't intentional here, right?

austin3dickey (approved):
very nice!

conbench/__init__.py
jgehrcke: > "unexpected exception" is very fun to read and say out loud

Yes and one of the most powerful concepts to embrace is to -- from the get-go when starting into an engineering problem -- distinguish "expected errors" from "unexpected errors" or to at least think in those two categories in _every part of the code_ (regardless of how one wants to call those).

Especially in distributed systems there's a pretty large set of expected errors along the way, but usually always when there is a human/machine boundary (user data validation), etc. As simple as powerful. I've heard people use "error" vs. "failure" but I find that silly :D The important thing is to actively distinguish.

conbench/tests/app/_asserts.py
jgehrcke: Haha, no. I wonder how it got there, and what the facial expression would have been to discover this later.

conbench/api/index.py
jgehrcke: Yes! (forgot to push that commit before :shrimp: (and the `:shrimp:` is what github sometimes suggests when I want to write `:shrug:`))

conbench/tests/api/_asserts.py
```diff
@@ -60,7 +60,8 @@ def assert_400_bad_request(self, r, message):
     def assert_401_unauthorized(self, r):
         assert r.status_code == 401, r.status_code
         assert r.content_type == "application/json", r.content_type
-        assert r.json == {"code": 401, "name": "Unauthorized"}, r.json
+        assert r.json["code"] == 401, r.json
+        assert r.json["name"] == "Unauthorized"
```
jgehrcke, line 64: Note: now the dict also has a `description` key and it's a long  sentence with a description of what this error means; default Flask behavior. Didn't want to put this sentence into the test suite.

conbench/api/index.py
jonkeane: Yum 🦐 , now I'm hungry 😂

PR #1198, opened 2023-04-28
api/run: fix GET /api/runs/{run_id} when the run doesn't have a commit

### LINKED ISSUE #1197: api/run: does not work for a run without a commit
As of #1134, GET `/api/run/{run_id}` won't work for a run when the run's commit is None. Example: this test I just wrote:
https://github.com/conbench/conbench/blob/ee16a7946f88812a996dc9dd53c08e8673b8445e/conbench/tests/api/test_runs.py#L92-L111
fails with
```
conbench/api/runs.py:73: in get
    return self.serializer.one._dump(run, get_baseline_runs=True)
conbench/entities/run.py:417: in _dump
    commit = CommitSerializer().one.dump(run.commit)
conbench/entities/_entity.py:219: in dump
    return self._dump(data)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <conbench.entities.commit._Serializer object at 0xffff885f7610>, commit = None

    def _dump(self, commit):
        url = None
>       if commit.repository and commit.sha:
E       AttributeError: 'NoneType' object has no attribute 'repository'
```

austin3dickey:
Fixes #1197. See discussion there.

### DIFF
--- a/conbench/tests/api/test_runs.py
+++ b/conbench/tests/api/test_runs.py
@@ -89,6 +89,27 @@ class TestRunGet(_asserts.GetEnforcer):
             ),
         )
 
+    def test_get_run_without_commit(self, client):
+        self.authenticate(client)
+        result = _fixtures.benchmark_result(no_github=True)
+        run = result.run
+        response = client.get(f"/api/runs/{run.id}/")
+        self.assert_200_ok(
+            response,
+            _expected_entity(
+                run,
+                candidate_baseline_runs={
+                    "fork_point": DEFAULT_BRANCH_PLACEHOLDER,
+                    "latest_default": DEFAULT_BRANCH_PLACEHOLDER,
+                    "parent": {
+                        "baseline_run_id": None,
+                        "commits_skipped": None,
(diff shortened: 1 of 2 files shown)


conbench/tests/api/test_runs.py
```diff
@@ -17,8 +17,8 @@
 }
 
 
-def _expected_entity(run, candidate_baseline_runs=None):
-    parent = run.commit.get_parent_commit()
+def _expected_entity(run: Run, candidate_baseline_runs=None):
```
austin3dickey, line 20: I suspect this small type hint would have prevented this bug in the first place 🙂

conbench/tests/api/test_runs.py
```diff
@@ -89,6 +89,35 @@ def test_get_run(self, client):
             ),
         )
 
+    def test_get_run_without_commit(self, client):
```
jgehrcke, line 92: :muscle:

jgehrcke (approved):
:ship:

PR #1202, opened 2023-05-01
benchalerts: update to latest API changes

### LINKED ISSUE #1196: benchalerts: handle run comparisons without commits
Today I think this will break when `benchalerts` tries to analyze a run without a commit:
https://github.com/conbench/conbench/blob/1e6e2b603b53cd754e729360606ec88cefaf708e/benchalerts/benchalerts/conbench_dataclasses.py#L267
The "commit" value will be None.

austin3dickey:
This PR updates `benchalerts` to conform to the latest API changes. Notably:
1. Runs can be missing commits now. Handle this case without erroring out. Fixes https://github.com/conbench/conbench/issues/1196
2. Because of (1), enforce that `GitHubCheckStep` and `GitHubStatusStep` must be explicitly given a `commit_hash` to post to, instead of inspecting the commit hash associated with the analyzed runs. No more magic default behavior. I think this is better for the interface anyway.
3. Because of (1), fix the intro message for Checks and PR comments to handle the case that the runs might not be associated with any commit hash.
4. Since we now get `["candidate_baseline_runs"][<type>]["commits_skipped"]`, we don't need to hit `/api/runs/{baseline_run_id}` anymore to figure out whether the "baseline is parent". Get rid of that code and the code that tracks the response.


austin3dickey:
Still need to add a mocked run without a commit and its associated tests.

benchalerts/benchalerts/conbench_dataclasses.py
```diff
@@ -30,26 +30,24 @@ class RunComparisonInfo:
...
+        /benchmark-results/?run_id={contender_run_id}, only if the contender run has
+        errors and there is no baseline run. Contains info about each benchmark result
+        in the contender run, including statistics and tracebacks. Only used when a
+        baseline run doesn't exist because otherwise all this information is already in
+        the compare_results.
     """
 
     contender_info: dict
-    baseline_info: Optional[dict] = None
+    baseline_run_type: str
```
alistaire47, line 50: Should this use the enum?

benchalerts/benchalerts/conbench_dataclasses.py
```diff
@@ -62,32 +60,23 @@ def contender_benchmark_result_info(self) -> List[BenchmarkResultInfo]:
             assert self.compare_link
             return [
                 BenchmarkResultInfo(
-                    name=benchmark_result["contender"]["case_permutation"]
-                    if "analysis" in benchmark_result
-                    else benchmark_result["benchmark"],
```
alistaire47, line 67: Thank you for cleaning up this mess

benchalerts/benchalerts/conbench_dataclasses.py
austin3dickey: There would be a circular dependency if we did so. I think it's okay to use a string since this dataclass is entirely internal. The only time this dataclass is ever constructed is here:
https://github.com/conbench/conbench/blob/a12c8e62708f668ad293712ff4389bd9bd157c1f/benchalerts/benchalerts/pipeline_steps/conbench.py#L92-L95
So it's safe to let this be a string.

benchalerts/benchalerts/conbench_dataclasses.py
austin3dickey: I think it's a great idea to maintain backwards HTTP API compatibility in our clients, but in this case the HTTP API has diverged so much that I thought it was easier to just do the synchronized deployment this one time.

benchalerts/benchalerts/conbench_dataclasses.py
alistaire47: Oh I was assuming it should move into this file, but this is ok

austin3dickey:
Okay, added tests.

benchalerts/benchalerts/conbench_dataclasses.py
```diff
@@ -30,26 +30,24 @@ class RunComparisonInfo:
...
-        /benchmarks?run_id={contender_run_id}, only if the contender run has errors and
-        there is no baseline run. Contains info about each benchmark result in the
-        contender run, including statistics and tracebacks. Only used when a baseline
-        run doesn't exist because otherwise all this information is already in the
-        compare_results.
+        /benchmark-results/?run_id={contender_run_id}, only if the contender run has
+        errors and there is no baseline run. Contains info about each benchmark result
+        in the contender run, including statistics and tracebacks. Only used when a
+        baseline run doesn't exist because otherwise all this information is already in
+        the compare_results.
```
jonkeane, line 46: Thanks for this cleanup

benchalerts/tests/unit_tests/expected_md/summary_nocommit.md
```diff
@@ -0,0 +1,10 @@
+Conbench analyzed the 2 benchmark runs that triggered this notification.
+
```
jonkeane, line 2: Sorry for this comment being in the test and not at all attached to the code that generates this, but I think it's easier to discuss here in context:

I wonder if we should mention something about _why_ the commit hash isn't here. It "should be" relatively rare, but given the setup we are designing for, it'll come up semi-frequently. That might pre-empt questions about why conbench _sometimes_ links to a | describes the commit hash and _sometimes_ it doesn't

jonkeane: Possibly we do this in a footnote or in the details somewhere, if we are worried about adding too much here.

jonkeane (commented):
Thanks for this!

benchalerts/tests/unit_tests/expected_md/details_nocommit.md
```diff
@@ -0,0 +1,3 @@
+This report was not associated with any commit connected to the git graph. This probably means that benchmarks were run on a transient merge-commit.
```
austin3dickey, line 1: @jonkeane To address your latest comment, I added this sentence to the details. Those show up at the bottom of a report (like [this one](https://github.com/conbench/benchalerts/runs/13049239396))

jonkeane: This is fantastic, thank you!

benchalerts/tests/unit_tests/mocked_responses/GET_conbench_compare_runs_some_baseline_no_commit.json
```diff
@@ -0,0 +1,118 @@
...
+                    "compression": "snappy",
+                    "cpu_count": 2,
+                    "dataset": "nyctaxi_sample",
+                    "file_type": "parquet",
+                    "input_type": "arrow",
+                    "name": "file-write"
+                }
+            },
+            "contender": {
+                "benchmark_name": "file-write",
```
jgehrcke, line 80: I love to see "benchmark_name" here :heart:

jgehrcke (commented):
Didn't have a deeper look, but thank you for pushing good things forward! :rocket: Awesome.

PR #1205, opened 2023-05-03
BMRT cache: use __slots__-based minimal typed objects instead of massive and magic SQLAlchemy-flavored objects, misc

jgehrcke:
This is mainly for #1211.

See commit messages.

Need to see impact on mem consumption in staging. Based on brief local analysis I think it might be significant. Curious.

### DIFF
--- a/conbench/hacks.py
+++ b/conbench/hacks.py
@@ -3,7 +3,6 @@ from typing import Dict, List
 
 from conbench.entities.benchmark_result import BenchmarkResult
 
-
 log = logging.getLogger(__name__)
 
 
--- a/conbench/hacks.py
+++ b/conbench/hacks.py
@@ -3,12 +3,6 @@ from typing import Dict, List
 
 from conbench.entities.benchmark_result import BenchmarkResult
 
-# Note(JP): ideally this contains just a single key. I think that this here is
-# purely for display purposes, not for storing in the DB. Now that I think
-# about it: Let's keep things simple: this is part of the case permutation in
-# the database, so show it.
-#  CASE_KEYS_DO_NOT_USE_FOR_STRING = ("id", "name", "suite", "source", "dataset")
-
 
 log = logging.getLogger(__name__)
 
@@ -27,6 +21,8 @@ def get_case_kvpair_strings(tags: Dict[str, str]) -> List[str]:
     Special keys are omitted (does that make sense?)
 
     Keep those with None value.
+
+    TODO: move to different module
     """
     return [f"{k}={v}" for k, v in sorted(tags.items()) if k not in ("name")]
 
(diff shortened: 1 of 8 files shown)


jgehrcke:
OK, let's see the mem consumption plot drop. :ship:

jgehrcke:
container_memory_working_set_bytes for the deployment (as reported by https://github.com/google/cadvisor) dropped:
image

Note that "working set" measurement is an approximation, as always in memory measurement. Also see https://stackoverflow.com/a/66778814.

conbench/job.py
```diff
@@ -72,33 +98,75 @@ def _fetch_and_cache_most_recent_results(n=0.06 * 10**6) -> None:
...
+            ui_hardware_short=str(result.ui_hardware_short),
+            ui_time_started_at=str(result.ui_time_started_at),
+            ui_rel_sem=result.ui_rel_sem,
+            ui_non_null_sample_count=result.ui_non_null_sample_count,
+            ui_mean_and_uncertainty=result.ui_mean_and_uncertainty,
+        )
+
+        # The str() indirections below (and above) are here to quickly make
+        # sure that there is no more SQLAlchemy magic associated to objects we
+        # store here (no more mapping to columns). Maybe that is not needed
```
austin3dickey, line 161: good idea!

conbench/job.py
```diff
@@ -37,10 +37,36 @@ class CacheUpdateMetaInfo:
     n_results: int
 
 
+# CPython 3.10/3.11 feature:  dataclass nicely integrated with slots. __slots__
+# saves memory: prevents the automatic creation of __dict__ and __weakref__ for
+# each instance. See
+# https://docs.python.org/3/reference/datamodel.html#object.__slots__
+# https://www.trueblade.com/blogs/news/python-3-10-new-dataclass-features
+@dataclasses.dataclass(slots=True)
+class BMRTBenchmarkResult:
```
austin3dickey, line 46: Very clean!

conbench/hacks.py
```diff
@@ -127,3 +130,70 @@ def sorted_data(benchmarks):
     new_data = sorted(new_data)
 
     return [row[1] for row in new_data]
+
+
+# Note(JP): Keeping this here because this turned out to be highly useful and
```
austin3dickey, line 135: nice

austin3dickey (commented):
I'm extremely impressed by how much difference this made!

jgehrcke:
Translation into those other objects of course consumes CPU, but I am shocked by how _much_, see how much longer it now takes to process the data:

image

Most of the time is spent in the data processing loop, not in the database query.

Need to do simple profiling.

jgehrcke:
Thanks for looking, @austin3dickey!!

> I'm extremely impressed by how much difference this made!

Yes, it's a good direction! The win was exchanged immediately into a higher N ([60000 -> 80000](https://github.com/conbench/conbench/commit/1b3018f2ee69e88a1155e4fb6fc1b72cebc86da5#diff-c4992dceff3f972d221793e7a3237f6bf66fa402e3591df9bf130edfd87a8ccdL92)), but I find that a little non-impressive. That's OK, this was a tiny start. After thinking a little more about this I think another big win in terms of peak memory will come from processing the DB response [in a streaming fashion](https://docs.sqlalchemy.org/en/20/core/connections.html#using-server-side-cursors-a-k-a-stream-results).

PR #1206, opened 2023-05-03
benchclients: fix oversight, invert connect() and recv() timeout

jgehrcke:
This is towards https://github.com/conbench/conbench/issues/801.

The oversight did not _yet_ result in a fallout. It would, if a response is expected to arrive after more than 3.5 s -- then it can never be fetched.

Ref docs, specifying order:

```
timeout (float or tuple) – (optional)
How many seconds to wait for the server to send data before giving up, as a float, or a (connect timeout, read timeout) tuple.
```
from https://requests.readthedocs.io/en/latest/api/

Seen via careful log reading:
```
230502-09:52:09.739 INFO: error during request/response cycle (treat as retryable, retry soon): 
    HTTPSConnectionPool(host='<redacted>', port=443):
         Read timed out. (read timeout=3.5)
```


jgehrcke:
Landing this :ship:

Issue #800, opened 2023-03-06
client HTTP tooling: plan for retryable errors, set specific goal: how long to retry before accepting data loss (robustness in view of micro outage, maintenance, etc)


jgehrcke:
When an ephemeral CI system generates benchmark results and then uses client tooling that _we_ build in order to _try to_ push the data into a system designed for analysis and long-term storage (Conbench API) we should make a choice about how hard that client tooling is trying to get the data there before giving up.

Giving up probably means _data loss_. We want to have a stance on that.

Let's set goals. We may want to retry (with considerable back-off) for up to O(1 hour); this is what I would like for us to consider at least, for certain CI-powered use cases.

In https://github.com/conbench/conbench/pull/788#discussion_r1122081341 I commented:

> We should have a pretty clear idea about how long recommended client tooling (tooling that we build) attempts to retry, before it drops data and moves on. In similar systems I have always advocated for and also built an explicit, tested design choice: "we want to survive a micro outage / maintenance window of N minutes length; that is our promise". Such a deliberate goal makes it easy to develop/test, and it also makes it easy to communicate guarantees to users.

This ticket here relates to a pattern I have observed over the years. In distributed systems, things often fail in a user-facing way (and lead to frustrating user experience) because developers didn't carefully handle _expected_, retryable errors. For me, it's the number one source of bugs I have seen in the space of distributed systems. Distributed systems are inherently prone to transient problems at the RPC interface; and yet often times individual HTTP calls are built like function calls that never leave machine boundaries.

Key ingredient to achieving a more or less pleasant distributed system _with guarantees_ is to have specific goals in mind  from the start, towards non-ignorant error handling. It helps a lot to handle all expected code paths around how an HTTP request/response cycle can fail, making a clear distinction between _retryable_ errors and non-retryable errors. That distinction is sometimes obvious, but often it also requires application-specific decision-making and goal-setting. Certainly, the API implementation needs to do a decent job communicating errors to clients, and therefore also a good internal error handling job.

When retrying on client side a question is "how long, how often"? Data locality, general context, the value of the ephemeral data, and deliberate goal-setting for the overall user experience help answering these questions.

In the client tooling we build we should make a deliberate choice.

This goes hand-in-hand with making good choices for the involved timeout criteria and timeout constants: https://github.com/conbench/conbench/issues/801.

I want to add: CI tooling is special with respect to retrying because one does not need to optimize for latency so much (bringing feedback quickly to a human), but one can exchange time budget into robustness.

[...]

PR #1208, opened 2023-05-03
retrying http client: better log msg for 5xx etc (add word "retryable")

jgehrcke:
Instead of logging e.g.
```
    unexpected response (code: 503, body bytes: ...
```

put the word 'retryable' word into the log msg, too. When applicable.

Related to #800.


jgehrcke:
:ship:

PR #1210, opened 2023-05-03
api.compare.runs: don't force repos to match

austin3dickey:
The `/api/compare/runs` endpoint currently returns a list of benchmark result comparisons, matching pairs of benchmark results from the contender and baseline runs that have the same case/context/hardware/repo/unit.

But today it's valid for a baseline run to be from a different repo than a contender run, if the contender run doesn't have a commit (and therefore doesn't have a repo):
https://github.com/conbench/conbench/blob/1b3018f2ee69e88a1155e4fb6fc1b72cebc86da5/conbench/entities/run.py#L275-L278

So this PR allows matches among pairs of benchmark results even if the baseline and contender runs do not share a repo.

This is admittedly sort of a hack, but I want to take the time to do #1165 correctly. Solving that issue will remove this hack.


conbench/tests/api/test_compare.py
```diff
@@ -161,6 +166,25 @@ def test_multiple_contexts_for_same_cases_but_they_kinda_line_up(self):
         pairs = CompareRunsAPI._join_results(baselines, contenders)
         assert self._parse_ids_from_pairs(pairs) == {("id1", "id3"), ("id2", None)}
 
+    def test_contender_missing_commit(self):
```
austin3dickey, line 169: This test would fail on `main` with the same issue I was seeing on a Conbench deployment:

```
AssertionError: assert {('id1', None), ('id2', None), (None, 'id5'), (None, 'id4'), ('id3', None), (None, 'id6')} == {('id3', None), ('id1', 'id4'), ('id2', 'id5'), (None, 'id6')
```

jgehrcke:
> But today it's valid for a baseline run to be from a different repo than a contender run, if the contender run doesn't have a commit (and therefore doesn't have a repo):

Okay! Thank you for the thorough explanation. 

We should prioritize working on https://github.com/conbench/conbench/issues/1165. I would like to help make that point, if necessary!

conbench/api/compare.py
```diff
@@ -403,7 +400,7 @@ def get(self, compare_ids) -> f.Response:
 
             This endpoint will return a list of comparison objects, pairing benchmark
             results from the given baseline and contender runs that have the same case,
-            context, hardware, repository, and unit. The comparison object is the same
```
jgehrcke, line 406: Thank you for updating the docstring here. This makes it visible for us how far-reaching a patch like this is, and a diff like this makes it much easier to go back to this topic in the future.

I think we agree that the _desired_ behavior is to keep `repository` constant across compared entities.

From this PR it is obvious that here we changed behavior for practical reasons, but we are _not_ willing to change the spec (the desired behavior), and this is just a temporary band-aid.

Just to state the obvious:

- we bought in to multi-repository support, and a change like this can be the source for really mean bugs
- if we were to limit each Conbench deployment to a single repository, things would be simpler (not saying we made a bad choice -- just saying we made a _far-reaching_ choice)

conbench/tests/api/test_compare.py
jgehrcke: Thanks for that. Really.

Issue #976, opened 2023-03-29
document a method for how to import a sql dump into the docker-compose-managed db


jgehrcke:
Currently I am using something like
```
docker-compose port db 5432
127.0.0.1:32818
```
```
$ psql -h 127.0.0.1 -p 32818 -U postgres -d postgres -1 -f dbdump_20230329-103457.sql 
Password for user postgres: 
```
to use a real-world database state in local dev environment. This is worth documenting properly, potentially worth to support with a Makefile target.

jgehrcke:
By the way -- me doing this (working with a snapshot of the arrow DB locally) is extremely valuable. So many good findings about the data model / DB schema, query pattern, etc.

jgehrcke:
### Load into dev database, from dump file:

Delete old container (DB) state:
```
± make teardown-app
```

Start 'just' the DB container in the dc stack:
```
± docker compose -f docker-compose.yml -f docker-compose.dev.yml  up db
[+] Running 2/2
 ⠿ Network conbench_default  Created                                                                                  0.1s
 ⠿ Container conbench-db-1   Created                                                                                  1.9s
Attaching to conbench-db-1
conbench-db-1  | The files belonging to this database system will be owned by user "postgres".
conbench-db-1  | This user must also own the server process.
conbench-db-1  | 
conbench-db-1  | The database cluster will be initialized with locale "en_US.utf8".
...
```

The log lines above confirm that a _new_ DB was initialized (important).

identify port  for reaching db from host:
```
± docker-compose port db 5432
127.0.0.1:33023
```

perform import:

```
psql -h 127.0.0.1 -p 33023 -no-owner -U postgres -d postgres -1 -f conbench_dbdump.sql
```
type password: postgres defined here: https://github.com/conbench/conbench/blob/5393917be45cb2703e5997d2c73b5cf131e2935b/docker-compose.yml#L36

On my machine that takes ~15 minutes for a ~13 GB dump file.

--
psql args:
- `-1 ("one"), --single-transaction`
- `-f, --file=FILENAME      execute commands from file, then exit`


I just remembered that I also removed access control-related SQL commands from the tail end of the the SQL dump file, to get things working. Need to document that bit.

jgehrcke:
Above, we used `up db` to start just the DB container in the stack.

Now, use `up app` to start the web app container of the stack:
```
± docker compose -f docker-compose.yml -f docker-compose.dev.yml up app
[...]

PR #1213, opened 2023-05-04
BMRT cache: process database response in streaming fashion (reduce peak mem consumption during DB query)

jgehrcke:
Had a deeper look at how `results = Session.scalars(query_statement).all()` behaves in terms of memory consumption. I was wrongly assuming that this is already using a small buffer, but in fact it's using a big buffer where the entire database response is stored in memory before processing in application code can continue. Also see https://github.com/sqlalchemy/sqlalchemy/discussions/9735.

I want to quote from https://docs.sqlalchemy.org/en/20/core/connections.html, corroborating the architectural choices that we're going to make here, and giving a bit more insight on the trade-offs:

> From this basic architecture it follows that a “server side cursor” is more memory efficient when fetching very large result sets, while at the same time may introduce more complexity in the client/server communication process and be less efficient for small result sets (typically less than 10000 rows).

The next obvious thing to do here is to change paradigms, to process the data in chunks (semi "streaming fashion" :-)). Keeping buffers in check, so to say.

Also see my comment here: https://github.com/conbench/conbench/pull/1205#issuecomment-1534359324.

Relevant docs:
- https://docs.sqlalchemy.org/en/20/orm/queryguide/api.html#orm-queryguide-yield-per
- https://docs.sqlalchemy.org/en/20/core/connections.html

### DIFF
--- a/conbench/job.py
+++ b/conbench/job.py
@@ -8,8 +8,6 @@ from typing import Dict, List, Optional, Tuple, TypedDict
 
 import sqlalchemy
 
-# from filprofiler.api import profile as filprofile
-
 import conbench.metrics
 from conbench.config import Config
 from conbench.db import Session
--- a/conbench/job.py
+++ b/conbench/job.py
@@ -7,7 +7,8 @@ from collections import defaultdict
 from typing import Dict, List, Optional, Tuple, TypedDict
 
 import sqlalchemy
-from filprofiler.api import profile as filprofile
+
+# from filprofiler.api import profile as filprofile
 
 import conbench.metrics
 from conbench.config import Config
@@ -259,8 +260,8 @@ def _periodically_fetch_last_n_benchmark_results() -> None:
             t0 = time.monotonic()
 
             try:
-                filprofile(lambda: _fetch_and_cache_most_recent_results(), "fil-result")
-                # _fetch_and_cache_most_recent_results()
+                # filprofile(lambda: _fetch_and_cache_most_recent_results(), "fil-result")
+                _fetch_and_cache_most_recent_results()
             except Exception as exc:
                 # For now, log all error detail. (but handle all exceptions; do
                 # some careful log-reading after rolling this out).
--- a/conbench/job.py
+++ b/conbench/job.py
@@ -117,11 +117,10 @@ def _fetch_and_cache_most_recent_results(n=0.08 * 10**6) -> None:
 
     # Corresponding to the `yield_per` magic, consume the returned value as an
     # iterator. `all()` would consume all results and would defeat the purpose
-    # of the memory-saving exercise. The following line of code does not do the
-    # work yet; that begins once the iterator is consumed.
+    # of the memory-saving exercise. The following line of code does not do
+    # much of the work yet; that begins once the iterator is consumed (maybe it
+    # fetches the first chunk?).
     result_rows_iterator = Session.scalars(query_statement)
-
-    t1 = time.monotonic()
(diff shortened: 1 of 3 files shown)


jgehrcke:
Keeping the `filprofiler` instrumentation (outcommented) in for now. It took a bit of code-reading / docs-reading and creative thinking in order to make this work.

Comparison.

### Before (non-streaming, big DB response buffer): ~460 MB

Screenshot from 2023-05-03 22-26-02

### After  ("streaming", smallish DB response buffer(s)): ~110 MB

Screenshot from 2023-05-03 22-56-29

Notes:
- I show these plots here just for giving a peek into my process -- there's so much to interpreting them, and I don't try to write down  relevant thoughts here.
- Maybe the most important piece: the absolute numbers are pretty meaningless, but the relative change (drop by 75 %) makes me very optimistic that this is a smart move.
- This work was rather quick, and enabled by using a production DB snapshot with a whole lot of heterogeneous data. For work like this I keep using a 13 GB snapshot from the "Arrow DB" from March. Loading it into my local dev database takes ~15 minutes, the process is described here: https://github.com/conbench/conbench/issues/976

jgehrcke:
```
Run make lint-ci
flake8
isort --check .
ERROR: /home/runner/work/conbench/conbench/conbench/job.py Imports are incorrectly sorted and/or formatted.
Skipped 2 files
make: *** [Makefile:76: lint-ci] Error 1
```
is a pain because I  don't have that error locally with the exact branch checked out. Hm. The error message is not revealing which line of code is tripping it up.

jgehrcke:
OK, finally green. Let's get this in :ship: Feedback extremely welcome.

jgehrcke:
Tears of joy.
image

conbench/job.py
```diff
@@ -100,28 +100,39 @@ def _fetch_and_cache_most_recent_results(n=0.08 * 10**6) -> None:
...
     # too, and so that result.run.hardware is a quick lookup.
-    # Note that N here determines peak usage of memory.
     query_statement = (
         sqlalchemy.select(BenchmarkResult)
         .join(Run)
         .order_by(BenchmarkResult.timestamp.desc())
         .limit(n)
-    )
-    results = Session.scalars(query_statement).all()
+    ).execution_options(yield_per=2000)
```
jgehrcke, line 114: I chose the chunk size on the one hand arbitrarily, on the other hand of course this is a number that absolutely tries to keep the iteration overhead minimal while at the same time significantly changing N. That is:
```
1 << 2000 < 60000
```

docker-compose.dev.yml
```diff
@@ -4,6 +4,7 @@
 services:
   app:
     # Add --reload to command, to reload on file change.
+    # command: ["fil-profile", "python", "-m",  "gunicorn", "-c", "conbench/gunicorn-conf.py", "--reload"]
```
jgehrcke, line 7: Note: the fact that this works isn't obvious at all:

- It wasn't obvious that gunicorn can be invoked via `python -m gunicorn ..`
- It wasn't obvious that we can wrap this entire thing into `fil-profile` and things "just work"
- It wasn't obvious that `fil-profile` manages to keep track of the _child process_ spawned by this machinery; the first process that's spawned here is the gunicorn master process, and the work happens in a worker process. This entry point here ensures that even the child is instrumented properly (with fil-profile specifics), and it needs to be combined with the file-profile API-based tracking (i.e. usage of the `filprofile()` function in code)

Issue #532, opened 2022-12-13
input validation: reject only-zero samples?


jgehrcke:
Here, a single zero was submitted:

![image](https://user-images.githubusercontent.com/265630/207322586-8330d282-6497-4446-b995-b2762eea3e67.png)

Maybe this should not survive sanity data validation check, and be rejected with a helpful error message wrapped in a 400 Bad Request response.

jgehrcke:
https://conbench.ursa.dev/benchmarks/df3b604775524876aa1ac9a03cf2b3ad/

![image](https://user-images.githubusercontent.com/265630/223660436-f5ee3ef0-e6fb-4735-b522-d8b1dde45106.png)

It's likely that this is a methodological error. If Conbench were to reject this it's easier to discover.

If this is not a methodological error then I'd be curious to learn what the value of this result is.

jonkeane:
Agreed, separately we should reach out to the Arrow Javascript folks and ask them about this.

We might consider doing this error|warning in the client (benchadapt or benchclient) first to minimize disruptiveness: start off with a warning or error there but still accept the (bad) info into conbench. We would be able to provide slightly more context at the client level + see if that's enough nudge before baking the restriction in to the server. 

I will also say, IMO, looking at that data in conbench it's much easier to see how not-helpful it is compared to the the command line | getting a 400 (even with a specific message)

jgehrcke:
A few more thoughts:

- This is very likely a mean bug in some client/benchmarking tooling. Might have to do with string truncation, even. `0` means infinitely fast or infinitely slow -- I just thought about this again, this can never be a good data point I think. If I were to write tooling that submits these values to Conbench -- I'd _love_ for Conbench to reject my request, to tell me that my tool chain has a bug.
- The advantage of rejecting this would be to let the CI job fail. A failed CI job is more likely to be debugged in a timely fashion than a succeeding one. Enforced communication :-).
- The API server can generate a precise, nice, simple error message.
- Our client tooling should expose a that error message to the user. If it does not, we should also work on that.

But I also see the point that if manyyyyy (independently obtained) results are submitted in the same request, that it might hurt the process a lot if the entire submission request fails. So in that case we may want to partially accept the good data, and reject only the individual bad result -- and _still_ send an error response.

jgehrcke:
I think I might have found the relevant code, somewhere around here:
[...]

PR #1214, opened 2023-05-04
BenchmarkResult: fix SEM calculation (division by zero)

jgehrcke:
https://github.com/conbench/conbench/issues/532 strikes back. Ran into this with real-world DB state (Arrow DB snapshot).

The methods we are working on here are not yet perfect. I would love to be able to iterate on them, and change their names and docs and code over time. But there should be convergence towards a minimal, stable, valuable set of methods that is used across the code base. Methods that hide some of the madness around what `data` can be in the database, methods that buy in to the idea of `is_failed`, etc.

### DIFF
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -446,12 +446,11 @@ class BenchmarkResult(Base, EntityMixin):
         We also may want to instruct SQLAlchemy to return numbers as floats
         directly.
         """
-        values = []
-        if not self.is_failed:
-            assert self.data is not None
-            values = [float(d) for d in self.data]
+        if self.is_failed:
+            return []
 
-        return values
+        assert self.data is not None
+        return [float(d) for d in self.data]
 
     @property
     def ui_mean_and_uncertainty(self) -> str:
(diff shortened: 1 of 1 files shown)


conbench/entities/benchmark_result.py
```diff
@@ -500,26 +499,29 @@ def ui_mean_and_uncertainty(self) -> str:
...
         """
-        samples = self.data
+        if self.is_failed:
+            return (str(-1), "n/a (failed)")
 
-        if samples is None:
-            return (str(-1), "no data")
+        values = self.measurements
+        if all(v == 0 for v in values):
+            return (str(-1), "n/a (bad data)")
```
jgehrcke, line 513: Interesting case, isn't it? See https://github.com/conbench/conbench/issues/532.

conbench/entities/benchmark_result.py
```diff
@@ -500,26 +499,29 @@ def ui_mean_and_uncertainty(self) -> str:
...
 
-        # otherwise: `TypeError: can't convert type 'NoneType' to numerator/denominator`
-        # in statistics.stdev(samples)
-        samples = [s for s in samples if s is not None]
+        if len(values) < 3:
+            return (str(-1), "n/a (< 3)")
```
jgehrcke, line 516: I wasn't sure about `< 3` vs `<3`.
:heart_on_fire:

jgehrcke:
:ship:

PR #1215, opened 2023-05-04
c-benchmarks landing page: case-insensitive sort order, small tweaks

jgehrcke:
@sara-hannigan's eyes are sensitve to case sensitivity!


jgehrcke:
:ship:

PR #1217, opened 2023-05-04
BenchmarkResult: fix `.data` type, think through `.unit` type, use more `.measurements`

jgehrcke:
The type annotation for BenchmarkResult.data we had in place already was useful and helped us fix bugs and move forward. But it wasn't quite correct yet. This patch accounts for the fact that individual iterations may report as `None`.

That is, we go from `Mapped[Optional[List[Decimal]]]` to `Mapped[Optional[List[Optional[Decimal]]]]`. :fire: 

That again provided interesting mypy feedback that I acted on.

I also doubled down on `self.measurements` which has the simple return type signature of `List[float]` which is what we _really_ want further down in application code. That allowed for vast simplification of some other functions.

Think `Mapped[Optional[List[Optional[Decimal]]]]` ->  `List[float]` -- looks useful, right? Of course this needs proper spec/docstrings to prevent misunderstandings from happening, but I think we're on a good path. Key ingredient here is the `is_failed` abstraction that I have been carrying along for a while. I think it really helps.

See code comments, and commit messages.

The bigger methods `BenchmarkResult.ui_mean_and_uncertainty()` and `BenchmarkResult.ui_rel_sem()` I did not just move, but I reworked them using newly gained simplicity.

Along the way I wondered about mypy's type inference and opened https://github.com/python/mypy/issues/15180.

### DIFF
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -70,11 +70,15 @@ class BenchmarkResult(Base, EntityMixin):
 
     # `data` holds the numeric values derived from N repetitions of the same
     # measurement (experiment). These can be empty lists. An item in the list
-    # is of type float, which includes `math.nan` as valid value.
-    data: Mapped[Optional[List[Decimal]]] = Nullable(
+    # is of type float, which includes `math.nan` as valid value. Note(JP): I
+    # think the type annotation `Optional[List[Optional[Decimal]]]` is now
+    # finally correct, reflecting what we allow to be inserted into the DB. I
+    # think we should really make it so that each element in self.data is of
+    # type float.
+    data: Mapped[Optional[List[Optional[Decimal]]]] = Nullable(
         postgresql.ARRAY(s.Numeric), default=[]
     )
-    times: Mapped[Optional[List[Decimal]]] = Nullable(
+    times: Mapped[Optional[List[Optional[Decimal]]]] = Nullable(
         postgresql.ARRAY(s.Numeric), default=[]
     )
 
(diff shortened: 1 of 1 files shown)


conbench/entities/benchmark_result.py
```diff
@@ -559,6 +519,74 @@ def ui_hardware_short(self) -> str:
...
+    The first string in the tuple is a stringified float for sorting in a
+    table. The second string in the tuple is for display in the table, with
+    unit (percent).
+
+    Associate "n/a" with a negative value so that DESC sorting by error is
+    helpful.
+    """
+    if not values:
+        # That is the case for 'failed'
+        return (str(-1), "n/a (no data)")
```
jgehrcke, line 532: here it's not distinguishable anymore if there was partial data or no data, but that's fine. after all it's "no useful data" and "no data" is a bit of an ignorant summary for that, not wrong.

conbench/entities/benchmark_result.py
```diff
@@ -559,6 +518,74 @@ def ui_hardware_short(self) -> str:
...
+
+    if all(v == 0 for v in values):
+        return (str(-1), "n/a (bad data)")
```
jgehrcke, line 535: https://github.com/conbench/conbench/issues/532

conbench/entities/benchmark_result.py
```diff
@@ -559,6 +518,74 @@ def ui_hardware_short(self) -> str:
...
+
+    It's difficult to write code like this because of
+    https://github.com/conbench/conbench/issues/813 and related discussions.
+
+    Note that sigfig.round is incredibly CPU intense (takes 10 seconds to run
+    on 60000 benchmark results on my beefy CPU). Do not call this for many
+    benchmark results, only for a smaller set that must be displayed.
+    """
+    if not values:
+        return "no data"
```
jgehrcke, line 568: Much better than before. We're approaching sanity here.

conbench/entities/benchmark_result.py
```diff
@@ -559,6 +518,74 @@ def ui_hardware_short(self) -> str:
...
+    '3.1 ± 0.7'.
+
+        mean_unc_str = sigfig.round(mean, uncertainty=stdem)
+
+    It's difficult to write code like this because of
+    https://github.com/conbench/conbench/issues/813 and related discussions.
+
+    Note that sigfig.round is incredibly CPU intense (takes 10 seconds to run
+    on 60000 benchmark results on my beefy CPU). Do not call this for many
+    benchmark results, only for a smaller set that must be displayed.
```
jgehrcke, line 565: New insight. More about that in next PR.

jgehrcke:
:ship: 
Need that to move on. Happy about feedback.

jgehrcke:
cb-on-minikube ci failure is a known instability, tracked here: https://github.com/conbench/conbench/issues/794#issuecomment-1534755762

Issue #1218, opened 2023-05-04
BMRT cache: cpu work regression as of `sigfig.round()`


jgehrcke:
Pulling this out of https://github.com/conbench/conbench/pull/1205#issuecomment-1534347430.

In this plot we see that the BMRT cache update time jumped from below 10 seconds to a minute:
![screenshot](https://user-images.githubusercontent.com/265630/236158684-cd438879-d7d7-41de-bbf4-26f9e58a1be2.png) (note: this is a custom application metric :rocket:).

The CPU work regression can also be seen here:
![image](https://user-images.githubusercontent.com/265630/236228579-441404ed-de4d-4e44-a599-d7a95888bf20.png)
Around April 11 the surge was as of BMRT cache introduction. The additional surge on top of that towards end of April/beginning of May was a negative side effect of https://github.com/conbench/conbench/pull/1205.

I did basic profiling work and found that (given real-world data: Arrow DB snapshot) when the BMRT cache update took ~16 seconds, that ~10 seconds were spent in `sigfig.round()`. Wow:

```
conbench-app-1  | name                                                                                                  ncall  tsub      ttot      tavg      
conbench-app-1  | /app/conbench/job.py:99 _fetch_and_cache_most_recent_results                                          1      0.490366  15.88532  15.88532
conbench-app-1  | /app/conbench/entities/benchmark_result.py:455 BenchmarkResult.ui_mean_and_uncertainty                20000  0.074558  9.699357  0.000485
conbench-app-1  | /usr/local/lib/python3.11/site-packages/sigfig/sigfig.py:560 round                                    244..  0.159072  9.666894  0.000395
conbench-app-1  | /app/conbench/entities/benchmark_result.py:482 <genexpr>                                              35514  0.029356  6.600892  0.000186
```

The profiling output tells us that `BenchmarkResult.ui_mean_and_uncertainty()` is what after all resulted in many calls of `sigfig.round()`.

That's an easy fix. Compute that only when needed by UI, not in advance within the BMRT cache update.

jonkeane:
> Compute that only when needed by UI, not in advance within the BMRT cache update.

💯 on this — and not even only from a performance stand point, I've found it routinely best to defer presentation details until the last second (e.g. only in the UI) to avoid various other malignancies like double-rounding, inconsistent styles of soundings etc. 

[...]

Issue #1035, opened 2023-04-04
pragmatic profiling: find out what it takes to fetch many BenchmarkResult objects for building an HTTP response


jgehrcke:
It becomes apparent that the individual `BenchmarkResult` is the core entity, and many UI views and API responses (HTTP responses in general) are built by processing potentially many of those.

Do some (rough) measurements to see where we are today in terms of timing, allowing for some back-of-the envelope calculations.

jgehrcke:
An interesting data point comes from using the Arrow Conbench DB state in my local dev environment (strongish CPU):

- 164298 benchmark results distributed across 100 runs.
- In that case, 657192 calls to `sqlalchemy/orm/loading.py:1028(_instance)` took `3.6 s` wall time.
- and `1.6 s` of those `3.6 s` were spent alone in `json.loads()` (stdlib) with 492895 invocations.

I will look into documenting how I got those numbers.

mariusvniekerk:
may be worth testing swopping the json implementation with cysimdjson or msgspec

jgehrcke:
> may be worth testing swopping the json implementation with cysimdjson or msgspec

Right. For _serializing_ into JSON we have the related ticket https://github.com/conbench/conbench/issues/999 and started using orjson already: https://github.com/conbench/conbench/pull/1002.

For _deserealizing_ JSON we might want to use https://github.com/sqlalchemy/sqlalchemy/blob/6f3c741e41e13624d63f4faa4bdcec8466bda7f4/lib/sqlalchemy/dialects/postgresql/json.py#L198 and then maybe pick a library based on simdjson -- yes. But we will have to see, maybe orjson is also the right thing to use there. I found this rather insightful (emphasis mine):

> pysimdjson / cysimdjson are by far the fastest if you only need to parse documents and access **a few individual keys** (> 2x faster than orjson). As soon as you convert the simdjson result to a dict or iterate over all keys, orjson is the faster option. simdjson.dumps() basically does that, since it's only an alias to json.dumps(), so it's no surprise that orjson is faster in this benchmark. simdjson is primarily a parser, not a serializer, so measuring its serialization speed is not a good performance comparison.

from https://gist.github.com/kgriffs/e34364896ae2f329aa8016ec6f71fdce?permalink_comment_id=4177378#gistcomment-4177378

Here, we see that parsing with orjson is a really good choice:
https://github.com/tktech/json_benchmark#complete-load-of-datatwitterjson
[...]

PR #1219, opened 2023-05-04
BMRT cache: reduce cpu work in update

jgehrcke:
This addresses https://github.com/conbench/conbench/issues/1218.

Started with a Arrow DB snaptho. Tested with 50000 results. Processing took ~ 9.0 s.
- removed ui_mean_and_uncertainty() from hot path -> 5.0 s
- removed ui_rel_sem() from hot path -> 4.0 s

One more profiling round after that showed once again that the stdlib JSON decoder is heavily involved with
```
JSONDecoder.decode                                                                                        300000      0.832742      1.810002 
```

I reported that before in https://github.com/conbench/conbench/issues/1035 where I also mentioned an easy improvement for that.

Quickly replaced that with the orjson decoder (already introduced before). Gain was another 0.5 s.

That is, this patch cuts processing time with my test data from 9.0 to 3.5 seconds.

Just like before, I kept the boilerplate for the yappi-based profiling work in place so that we can quickly go back to it. You will notice that the relevant changes in this PR are very little.

### DIFF
--- a/conbench/job.py
+++ b/conbench/job.py
@@ -14,7 +14,11 @@ import yappi
 import conbench.metrics
 from conbench.config import Config
 from conbench.db import Session
-from conbench.entities.benchmark_result import BenchmarkResult
+from conbench.entities.benchmark_result import (
+    BenchmarkResult,
+    ui_mean_and_uncertainty,
+    ui_rel_sem,
+)
 from conbench.entities.run import Run
 from conbench.hacks import get_case_kvpair_strings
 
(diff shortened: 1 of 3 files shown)


conbench/job.py
```diff
@@ -93,7 +114,7 @@ class CacheDict(TypedDict):
 
 # Fetching one million items from a sample DB takes ~1 minute on my machine
 # (the `results = Session.scalars(....all())` call takes that long.
-def _fetch_and_cache_most_recent_results(n=0.08 * 10**6) -> None:
+def _fetch_and_cache_most_recent_results(n=0.05 * 10**6) -> None:
```
jgehrcke: Oh, I want to revert that to keep changes in plots meaningful. Commit coming. That was just for testing.

requirements-webapp.txt
```diff
@@ -33,3 +33,4 @@ sigfig
 SQLAlchemy>=2
 tenacity
 uuid7
+yappi
```
jgehrcke: removing that, too

jgehrcke:
:ship: OK, I think that will set us free quite a bit, and I think it will after all allow us to bump the size of the cache quite a bit, covering a more interesting timeframe -- making timeseries plots more interesting, etc.

PR #1221, opened 2023-05-04
benchalerts: link to compare/benchmarks when we can

### LINKED ISSUE #1189: benchalerts: Use the comparison URL instead of entity URL (or in addition to)?
Currently with benchalerts, when a benchmark indicates a regression or improvement we link to the benchmark result itself. Is it possible to link to the comparison of that benchmark with the baseline that was used by default? 

It's currently a little tricky to go from the benchmark result to the comparison page to actually see the z-score, what the baseline looked like etc. etc.

And same thing for runs, it's nice to be able to get to the run itself, but most people are probably looking to get to the comparison first from these alerts.

austin3dickey:
Closes #1189. This PR changes the links in the Check summaries and PR comments to link to the compare pages for benchmark results, not just the contender pages, whenever we can.

### DIFF
--- a/benchalerts/benchalerts/conbench_dataclasses.py
+++ b/benchalerts/benchalerts/conbench_dataclasses.py
@@ -57,12 +57,17 @@ class RunComparisonInfo:
         benchmark result on the contender run.
         """
         if self.compare_results:
-            assert self.compare_link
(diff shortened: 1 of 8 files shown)


jonkeane:
Thanks for this

Issue #1222, opened 2023-05-05
BMRT cache population: excessive / slow query pattern (yield_per vs. lazy=joined)


jgehrcke:
The cache population time did not fall back to previous level after #1219 which focused on removing CPU hogs:
![image](https://user-images.githubusercontent.com/265630/236311779-e1804c6b-c587-41a3-9dec-42feacbb4917.png)

This is not unexpected per se, because we _are_ doing more work than before (re-writing objects into smaller ones, for example). But this didn't look 'right' yet.

Important:

> The yield_per execution option is not compatible with [“subquery” eager loading](https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html#subquery-eager-loading) loading or [“joined” eager loading](https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html#joined-eager-loading) when using collections. It is potentially compatible with [“select in” eager loading](https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html#selectin-eager-loading) , provided the database driver supports multiple, independent cursors

Brief log inspection and local testing confirmed: we started emitting much more queries.

This is not a problem per se. Might even be a good thing, because the cache population routine does not need to be particularly fast. However, it makes sense to look into that. The query pattern does not need to be perfect, but it should not be stupid.

On my local system with a big Arrow DB snapshot and containerized Postgres running on my laptop right next to the web application this excessive query pattern btw is not a problem at all; each query is responded to blazingly fast.  A valid conclusion from this observation: the slowdown compared to my local dev environment is a symptom of having mediocre CPU, networking, and memory resources available in the application stack we're looking at here.

I am saying: processing 80.000 benchmark results on my laptop takes 10 seconds, and in our 'cloud setup' it takes over a minute. Near-zero latency as of localhost networking is just one explanation, but nah, the database is still underpowered IMO. This is worth following up upon.

jgehrcke:
This screenshot confirms that we issue many queries for runs:
[Screenshot from 2023-05-05 09-53-14](https://user-images.githubusercontent.com/265630/236406325-1a476451-01ef-4811-8876-82947ec3746b.png)

It also confirms that we use server-side cursers now (FETCH FORWARD), and that these are looking good. This screenshot shows the specific offending query (with its high average rate):
[Screenshot from 2023-05-05 09-53-35](https://user-images.githubusercontent.com/265630/236406498-8a882d63-32e7-4920-9af6-662d54552196.png)

[...]

PR #1223, opened 2023-05-05
BMRT cache: use selectinload strategy

jgehrcke:
For #1222. I started looking at the specific query count with this method:
```
± cat count-queries.patch 
diff --git a/conbench/db.py b/conbench/db.py
index 62c62c8..4c184e4 100644
--- a/conbench/db.py
+++ b/conbench/db.py
@@ -35,7 +35,7 @@ def configure_engine(url):
     engine = create_engine(
         url,
         future=True,
-        echo=False,
+        # echo=True,
         pool_pre_ping=True,
         # As of today some requests take a (too) long while to generate a
         # response for. We want to improve that fundamentally over time. Until
diff --git a/conbench/job.py b/conbench/job.py
index 9991f1f..f10ea3c 100644
--- a/conbench/job.py
+++ b/conbench/job.py
@@ -17,6 +17,7 @@ from conbench.entities.benchmark_result import (
     ui_rel_sem,
 )
 from conbench.entities.run import Run
+from conbench.entities.hardware import Hardware
 from conbench.hacks import get_case_kvpair_strings
 
 # A memory profiler, and a CPU profiler that are both tested to work well
@@ -34,6 +35,7 @@ Central cache data structure are Python dictionaries (with thread-safe(atomic)
 set/get operations).
 """
 
+
 original_sigint_handler = signal.getsignal(signal.SIGINT)
 original_sigquit_handler = signal.getsignal(signal.SIGQUIT)
 
@@ -111,6 +113,28 @@ bmrt_cache: CacheDict = {
 SHUTDOWN = False
 _STARTED = False
 
+import contextlib
+from sqlalchemy import event
+
+
+@contextlib.contextmanager
+def count_queries(conn):
+    queries = []
+
+    def before_cursor_execute(
+        conn, cursor, statement, parameters, context, executemany
+    ):
+        queries.append(statement)
+
+    event.listen(conn, "before_cursor_execute", before_cursor_execute)
+    try:
+        yield queries
+    finally:
+        event.remove(conn, "before_cursor_execute", before_cursor_execute)
+
+
+from sqlalchemy.orm import selectinload, raiseload
+
 
@@ -129,6 +153,9 @@ def _fetch_and_cache_most_recent_results(n=0.08 * 10**6) -> None:
     # too, and so that result.run.hardware is a quick lookup.
     query_statement = (
         sqlalchemy.select(BenchmarkResult)
+        # .options(selectinload(BenchmarkResult.run))
+        # .options(selectinload(Run.hardware))
+        # .options(raiseload(BenchmarkResult.run.commit))
         .join(Run)
         .order_by(BenchmarkResult.timestamp.desc())
         .limit(n)
@@ -280,7 +307,9 @@ def _periodically_fetch_last_n_benchmark_results() -> None:
 
             try:
                 # filprofile(lambda: _fetch_and_cache_most_recent_results(), "fil-result")
-                _fetch_and_cache_most_recent_results()
+                with count_queries(Session.connection()) as queries:
+                    _fetch_and_cache_most_recent_results()
+                print("total number of queries: %s" % len(queries))
             except Exception as exc:
                 # For now, log all error detail. (but handle all exceptions; do
                 # some careful log-reading after rolling this out).

```
props to Mike, as always: https://github.com/sqlalchemy/sqlalchemy/issues/5709#issuecomment-729689097

With the `.join(Run)`: `total number of queries: 225`
With `.options(selectinload(BenchmarkResult.run))`: `total number of queries: 41`

Let's try out which difference this makes in our RDS/cloud deployment scenario.

### DIFF
--- a/conbench/job.py
+++ b/conbench/job.py
@@ -7,6 +7,7 @@ from collections import defaultdict
 from typing import Dict, List, Optional, Tuple, TypedDict
 
 import sqlalchemy
+from sqlalchemy.orm import selectinload
 
 import conbench.metrics
 from conbench.config import Config
@@ -16,7 +17,6 @@ from conbench.entities.benchmark_result import (
     ui_mean_and_uncertainty,
     ui_rel_sem,
 )
-from conbench.entities.run import Run
 from conbench.hacks import get_case_kvpair_strings
 
 # A memory profiler, and a CPU profiler that are both tested to work well
--- a/conbench/job.py
+++ b/conbench/job.py
@@ -126,10 +126,15 @@ def _fetch_and_cache_most_recent_results(n=0.08 * 10**6) -> None:
     # https://docs.sqlalchemy.org/en/20/core/connections.html#using-server-side-cursors-a-k-a-stream-results
     # https://docs.sqlalchemy.org/en/20/orm/queryguide/api.html#fetching-large-result-sets-with-yield-per
     # Also fetch hardware from associated Run, so that result.run is in cache,
-    # too, and so that result.run.hardware is a quick lookup.
+    # too, and so that result.run.hardware is a quick lookup. Use selectinload
+    # for fetching Run data -- "The yield_per execution option is not
+    # compatible with “subquery” eager loading loading or “joined” eager
+    # loading when using collections. It is potentially compatible with “select
+    # in” eager loading , provided the database driver supports multiple,
+    # independent cursors." -- seems to result in overall less queries.
     query_statement = (
         sqlalchemy.select(BenchmarkResult)
-        .join(Run)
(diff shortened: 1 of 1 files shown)


jgehrcke:
:ship:

jgehrcke:
Fascinating, this really had a big impact, the DB was bottlenecking (CPU-bound, as shown in https://github.com/conbench/conbench/issues/1222#issuecomment-1535889563), and that removed the CPU pressure:
image

From the DB's point of view:

### average query pattern + stats over last 24 hours (bad state):
Screenshot from 2023-05-05 10-28-29


### average query pattern + stats over last 5 minutes (better state):
Screenshot from 2023-05-05 10-35-15

PR #1224, opened 2023-05-05
Various minor follow-ups: rename svs, reduce log verbosity, 

jgehrcke:
Main motivator to start a "misc" patch was that I remembered that I never acted on Austin's feedback on https://github.com/conbench/conbench/pull/1172 with respect to renaming the svs prop in the history JSON entity.

I also had a quick look at current staging web application logs and inspected for log statements with a dubious signal/noise ratio.

Also see commit messages.


jgehrcke:
:ship:

Issue #1238, opened 2023-05-08
c-bench small multiples: "scientific notation for values that would be shorter without it"


jgehrcke:
Feedback from @alistaire47 

See https://github.com/conbench/conbench/issues/1154

This is about
![image](https://user-images.githubusercontent.com/265630/236790437-9c8f6419-2af3-44bd-9c90-d2e79713e123.png)

> if there’s a way to turn off scientific notation for values that would be shorter without it, that would be nice

jgehrcke:
Done via https://github.com/conbench/conbench/pull/1239.

PR #1239, opened 2023-05-08
c-benchmarks: small multiples / tiny history plots: change axis labels 

jgehrcke:
This is for https://github.com/conbench/conbench/issues/1238

image

I think from a change like this it's obvious that this code needs to live in its own file, subject to linting, and also testing to a certain extent.

Despite being ugly and appearing non-maintainable, I do think this is a sustainable direction: _if_ we work on making code re-usable, and tested.


jgehrcke:
:ship:

Issue #1243, opened 2023-05-08
Conbench UI: adhere to "Data Visualization Standards"


jgehrcke:
This is an amazing resource: https://xdgov.github.io/data-design-standards. As a physicist, I got taught a 'set of rules', too. As far as I have looked this resource, there is strong overlap between what I have learned, and what is required there. That's probably why I like this resource :). But I have not looked too deeply yet.

We should pick a few categories / rules that we really care about, and then make sure we adhere.

Other resources, while we are here:
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4161295/
https://doctor-pasquale.com/rules-for-graphs/

Issue #1240, opened 2023-05-08
c-bench small multiples: link to "last benchmark result" from a tinyplot


jgehrcke:
@joosthooz and others rightfully said

> I’d love to be able to click through from a plot to its most recent (probably? or the individual point if you want) result

jgehrcke:
Done via https://github.com/conbench/conbench/pull/1244.

Issue #1154, opened 2023-04-24
benchmark-centric UI/UX: ideas, initial observations and feedback


jgehrcke:
After landing https://github.com/conbench/conbench/pull/1086 and follow-up patches, the entire "list benchmarks" UI section is still super unpolished, in terms of presentation and UX paradigms. A lot of work can be done, but I'd like to get feedback.

Observations / ideas, in no particular order:

- :heavy_check_mark:  `data` column: show unit when there are only one or two values (done in #https://github.com/conbench/conbench/pull/1245)
-  :heavy_check_mark:  `rel err` column doesn't sort well when there are "n/a" entries: https://github.com/conbench/conbench/issues/1153
- overview: multi-repo support
- overview: sort oder? alphabetically, by last result, by most results?
- "hardwares" sounds weird, find better plural
- find better presentation for "case permutation" column, the content is huge.
- maybe say "spanning ... case permutation(s)" instead of "spanning 1234 case(s)."
- in plots grid: for each plot,
    - :heavy_check_mark: show the unit(!) basic stuff (https://github.com/conbench/conbench/pull/1244) 
    - :heavy_check_mark: link to the benchmark result overview for the last result (showing history plot, https://github.com/conbench/conbench/pull/1244) 
    - link to hardware detail page (e.g. `hardware/7719d3c95ed84fa99dd46d2a64e413d4`)
- build context detail page? so that we can link from context ID to context detail 
- plots: show errors bars/uncertainty based on multi-sample result
- topN section:
     - highest relative error (in last multi-sample result)
     - highest relative change (over the last few multi-sample results
- :heavy_check_mark: show 'last run timestamp' (done: #1241.  )

I  want to quote @joosthooz @alistaire47 from Slack

>  making the context clickable or showing the context values when hovering or in a side window?

Tracked here: #1242

> we should think about the best way to do the filtering. It seems natural to filter by case params, but because they can differ they can’t be easily be split out into cols…maybe selectable tags or something?

> making the points in the plots clickable to go to that benchmark?

> if there’s a way to turn off scientific notation for values that would be shorter without it, that would be nice 

Tracked here: https://github.com/conbench/conbench/issues/1238

> I’d love to be able to click through from a plot to its most recent (probably? or the individual point if you want) result

The easy version of that (most recent result) is tracked here: #1240

> some kind of “last run timestamp” col might be useful there so outdated permutations are more obvious

Tracked here: #1241

jgehrcke:
[...]

PR #1244, opened 2023-05-08
c-bench tinyplot: show unit, link to last result

jgehrcke:
For issues #1154 and #1240. Two improvements on the tinyplots / small multiples:

### Axis label (benchmark metric unit):

Current state (before this patch, plot not showing unit on y-axis/ordinate:
image

That's of course a no-go. Mentioned that in #1154, but now that we have #1243 I also like to link to that.

With this patch, we show the unit on the ordinate ("axis label"):
image

### Link to "most recent result" shown in plot

For the proper history plot.

Screenshot from 2023-05-06 17-48-36

### DIFF
--- a/conbench/app/benchmarks.py
+++ b/conbench/app/benchmarks.py
@@ -112,7 +112,9 @@ def show_benchmark_cases(bname: str) -> str:
 
 class TypeUIPlotInfo(TypedDict):
     title: str
+    url_to_newest_result: str
     data_for_uplot: List[List]
+    unit: str
 
 
 @app.route("/c-benchmarks/<bname>/<caseid>", methods=["GET"])  # type: ignore
@@ -159,7 +161,7 @@ def show_benchmark_results(bname: str, caseid: str) -> str:
         results_by_hardware_and_context[(hwid, ctxid, hwname)].append(result)
 
     # Make it so that infos_for_uplots is sorted by result count, i.e. show
-    # most busy plots first.
+    # most busy plots first. Instead, it might make sense to sort by recency!
     results_by_hardware_and_context_sorted = dict(
         sorted(
             results_by_hardware_and_context.items(),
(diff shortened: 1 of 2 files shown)


conbench/app/benchmarks.py
```diff
@@ -112,7 +112,9 @@ def show_benchmark_cases(bname: str) -> str:
 
 class TypeUIPlotInfo(TypedDict):
     title: str
+    url_to_newest_result: str
     data_for_uplot: List[List]
+    unit: str
```
jgehrcke, line 118: The beauty of ad-hoc typing. This code is in flux, but it doesn't quickly become a gigantic mess, because of typing constraints like this.

conbench/app/benchmarks.py
```diff
@@ -196,8 +205,25 @@ def show_benchmark_results(bname: str, caseid: str) -> str:
...
+            "url_to_newest_result": flask.url_for(
+                "app.benchmark-result",
+                benchmark_result_id=newest_result.id,
+            ),
+            "unit": newest_result.unit,
         }
 
+    # For now, only emit a warning in the web application log.
+    # TODO: show a user-facing warning on this page.
+
```
jgehrcke, line 217: Interesting trade-off to pay attention to.

jgehrcke:
:ship:

boshek:
A tiny bit of feedback here is that I always try to avoid anything that makes the reader turn their head sideways. Sometimes it is unavoidable (or not worth the extra real estate) but my desired state is a title (sometimes coupled with units directly on axis labels) such that a vertical axis title isn't required. So here one could add a title that is `Benchmark time in seconds...` etc. That way the unit of analysis is the first thing a reader will read. For extra clarity one could also append an "s" to axis labels.

jgehrcke:
>  such that a vertical axis title isn't required.

It's an interesting point. In the plotting/scientific culture I grew up in that is a rather strict no-go. "An axis must have a label". I am generally in favor of this rule because conventions like this often actually do help.

From a pure information/communication point of view of course there are ways to replace that label with something else. That is, I think we all agree that the most important aspect that needs to be fulfilled is that the relevant information _is_ communicated, in some way.

I'll think about this more.

PR #1245, opened 2023-05-08
Controlled `float`->`str`: use numpy's Dragon4 implementation instead of sigfig.round (unleash the dragon 🐉)

jgehrcke:
This introduces a module `numstr.py` which at the core defines a function `numstr()` which uses numpy's `np.format_float_positional()` function. This is documented [here](https://numpy.org/doc/stable/reference/generated/numpy.format_float_positional.html). It uses an established _configurable_ algorithm for float-str conversion ("Dragon4", https://dl.acm.org/doi/pdf/10.1145/93542.93559). The configuration options turn out to be super flexible and seem allow for better formatting than `sigfig.round()`.

I found this because I saw that e.g. `sigfig.round(15272000, 5)` producing `"15272000.0"` with the trailing `.0` being against the goal of producing a short representation.

Side effect: numpy.format_float_positiona() is three orders of magnitude faster. :rocket:  this matters! See https://github.com/conbench/conbench/issues/1218 where we previously identified sigfig.round()'s performance to be ridiculous. 

Notably, sigfig.round() still has its purpose for building those strings that capture measurement value _and_ uncertainty.

I quickly typed up some tests and perf comparsion; from the module docstring of the new `numstr.py` module:

```
A module for turning a floating point number into short, meaningful text, for
representation in a user interface, or in e.g. JSON documents.

Guidelines:

- no unnecessary precision (reasonable number of significant digits)
- shortest possible output (e.g., no trailing ".0" or even ".")


At first we started using sigfig.round() but then we noticed that it's
incredibly unnecessarily slow. See
https://github.com/conbench/conbench/issues/1218

Then I found a test case that fails for sigfig.round(), but does not for

    sigfig.round(15272000, 5) produces "15272000.0", i.e. here we have the
    unnecessary .0 suffix.

Then I researched further options and gladly found np.format_float_positional()
from numpy. It uses the so-called Dragon4 algorithm with all its beautiful
configuration options for stringifying floats. Found a set of configuration
options that achieves the desired results across test cases. Rough timing
experiments show that this function is about three orders of magnitude faster
than sigfig.round(), that's a nice performance win side effect here from trying
to get a more 'correct' result! Beautiful.

Note: for particularly small or large numbers we can start using
npformat_float_scientific().
Timing results on my machine (stdout of this program when invoked as main
module):
sigfig : sigfig.round(0.000123888129341456, sigfigs=5)  -->  0.337580 s
numstr : numstr(0.000123888129341456, 5)  -->  0.002462 s
---
sigfig : sigfig.round(015272063.0, sigfigs=3)  -->  0.244474 s
numstr : numstr(015272063.0, 3)  -->  0.002069 s
---
sigfig : sigfig.round(9.738480924743*10**8, sigfigs=7)  -->  0.317508 s
numstr : numstr(9.738480924743*10**8, 7)  -->  0.002376 s
---
sigfig : sigfig.round(982378347837483, sigfigs=7)  -->  0.318014 s
numstr : numstr(982378347837483, 7)  -->  0.002471 s
```

This PR also
* Takes care of "data column: show unit when there are only one or two values" from #1154.
* changes a code path where I somehow previously made the wrong decision to somehow want to control the precision of numbers _before_ storing them in the database. I think that was short-sighted thinking, no clear advantage, but it might result in super mean situations where e.g. min/max suddenly are not equivalent to an actually provided value. See https://github.com/conbench/conbench/issues/1246.


conbench/entities/benchmark_result.py
```diff
@@ -642,36 +643,44 @@ def validate_and_aggregate_samples(stats_usergiven: Any):
...
+                            key,
+                            stats_usergiven[key],
+                            value,
+                        )
+                except Exception as exc:
                     log.warning(
-                        "key %s, user-given val %s vs. calculated %s",
-                        key,
-                        stats_usergiven[key],
-                        value,
```
jgehrcke, line 674: Also added a bit of safe-guard here, where the `floatcomp_with_leeway()` is just sugar for logging/debugging. It would be sad if a bug in that function was to result in data not being accepted.

conbench/numstr.py
```diff
@@ -0,0 +1,113 @@
...
+than sigfig.round(), that's a nice performance win side effect here from trying
+to get a more 'correct' result! Beautiful.
+
+Note: for particularly small or large numbers we can start using
+npformat_float_scientific().
+
+Timing results on my machine (stdout of this program when invoked as main
+module):
+
+sigfig : sigfig.round(0.000123888129341456, sigfigs=5)  -->  0.337580 s
```
jgehrcke, line 33: :thinking: could have tuned this until it takes 1.337 seconds

jgehrcke:
I think I like the direction that this takes. Feels sustainable. numpy feels like the right tool for the job in general.

Feedback appreciated.

For the record, for the curious:
https://github.com/numpy/numpy/blob/8cec82012694571156e8d7696307c848a7603b4e/numpy/core/src/multiarray/dragon4.c#L1143
https://github.com/numpy/numpy/commit/28a96544c2d523a351895904737c441f2c4fe336

jgehrcke:
Landing this, subsequent patch uses this. :ship:

conbench/entities/benchmark_result.py
```diff
@@ -570,7 +571,7 @@ def ui_mean_and_uncertainty(values: List[float], unit: str):
 
     if len(values) < 3:
         # Show each sample with five significant figures.
-        return "; ".join(str(sigfig.round(v, sigfigs=5)) for v in values)
+        return "; ".join(f"{numstr(v, sigfigs=5)} {unit}" for v in values)
```
jgehrcke, line 574: This line here adds the unit, i.e. this takes care of "data column: show unit when there are only one or two values" tracked in #1154.

conbench/numstr.py
```diff
@@ -0,0 +1,113 @@
...
+- shortest possible output (e.g., no trailing ".0" or even ".")
+
+At first we started using sigfig.round() but then we noticed that it's
+incredibly unnecessarily slow. See
+https://github.com/conbench/conbench/issues/1218
+
+Then I found a test case that fails for sigfig.round(), but does not for
+
+    sigfig.round(15272000, 5) produces "15272000.0", i.e. here we have the
+    unnecessary .0 suffix.
```
alistaire47, line 17: I can't reproduce this—it gives me an int. But that raises a larger point about `sigfig.round()`: it's [mostly] for rounding, not formatting, and so its output is usually of the same type as the input, regardless of whether that makes sense (e.g. `sigfig.round(12345.67890, sigfigs=3)` returns a float `12300.0`, which is annoying, even if understandable from a type-stability perspective). It does have an [`output_type` param](https://sigfig.readthedocs.io/en/latest/api.html#output-type-type) we should always be setting to `str` if using it for display purposes: `sigfig.round(12345.67890, sigfigs=3, output_type=str)` returns `12300`.

jgehrcke: Thanks for trying this out! Woopsie. I copy/pasted the input badly into the docstring. But the output/problem is real, with `str(sigfig.round(nbr, sigfigs=5))` this test breaks:
```
  File "/home/jp/dev/conbench/conbench/numstr.py", line 102, in <module>
    got == expected
AssertionError: input: 15272063.0 -- expected 15272000 but got 15272000.0
```

With `sigfig.round(nbr, sigfigs=5, output_type=str)` I found the following difference between numpy and sigfig:
```
  File "/home/jp/dev/conbench/conbench/numstr.py", line 102, in <module>
    got == expected
AssertionError: input: 0.00123898129341456 -- expected 0.001239 but got 0.0012390
```

edit: in previous judgement, I misread the last one -- it makes sense, if seen strictly, to show `0.0012390` instead of `0.001239` to indicate that these really are five significant digits. So, sigfig.round()'s behavior is OK here in terms of strictness. But in terms of compact representation / 'saving horizontal space', I think numpy's behavior is even a tiny better for us.

jgehrcke: > I can't reproduce this—it gives me an int.

So, thanks for catching this. As said, this was just a copy/paste error. Also mentioned this here: https://github.com/drakegroup/sigfig/issues/14

Will adjust the docstring to show the correct example.

PR #1250, opened 2023-05-08
benchmark result compare page: Rename to benchmark-results and redirect

### LINKED ISSUE #1229: benchmark result compare page: rename to `/compare/benchmark-results/`
In the spirit of the app-wide rename from `benchmarks` to `benchmark-results`, this endpoint should be renamed. Probably keep the old endpoint around for old stable links, at least for a while.

https://github.com/conbench/conbench/blob/7c4968e631ecdc064559c86a1174a1353713b700/conbench/app/compare.py#L288

alistaire47:
Closes #1229. Renames the page to `/compare/benchmark-results/<compare_ids>`, and redirects `/compare/benchmarks/<compare_ids>` to it so existing links will continue to work. Adds some basic tests to ensure the redirect is working. Both the new path and the redirect work in a local instance.

Learned a bit about flask and our testing framework along the way.

### DIFF
--- a/conbench/app/compare.py
+++ b/conbench/app/compare.py
@@ -285,8 +285,15 @@ class CompareRuns(Compare):
 
 
 rule(
-    "/compare/benchmarks/<compare_ids>/",
+    "/compare/benchmark-results/<compare_ids>/",
     view_func=CompareBenchmarkResults.as_view("compare-benchmark-results"),
+    methods=["GET"],
+)
+# legacy route
+rule(
+    "/compare/benchmarks/<compare_ids>/",
+    endpoint="compare-benchmarks",
+    redirect_to="/compare/benchmark-results/<compare_ids>/",
     methods=["GET"],
 )
 rule(
--- a/conbench/tests/app/test_compare.py
+++ b/conbench/tests/app/test_compare.py
@@ -16,7 +16,7 @@ def _emsg_needle(thing, thingid):
 
 
 class TestCompareBenchmarkResults(_asserts.GetEnforcer):
-    url = "/compare/benchmarks/{}/"
+    url = "/compare/benchmark-results/{}/"
     title = "Compare Benchmark Results"
     redirect_on_unknown = False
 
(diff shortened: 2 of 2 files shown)


conbench/app/compare.py
```diff
@@ -285,10 +285,17 @@ class CompareRuns(Compare):
...
-    "/compare/benchmarks/<compare_ids>/",
+    "/compare/benchmark-results/<compare_ids>/",
     view_func=CompareBenchmarkResults.as_view("compare-benchmark-results"),
     methods=["GET"],
 )
+# legacy route
+rule(
+    "/compare/benchmarks/<compare_ids>/",
+    endpoint="compare-benchmarks",
+    redirect_to="/compare/benchmark-results/<compare_ids>/",
```
jgehrcke, line 296: How did you find the `redirect_to` arg? I was looking for reference docs for the `redirect_to` arg. It's not mentioned on this page: https://flask.palletsprojects.com/en/2.2.x/api/

That's a bit suspicious / a bit of a red flag.

Also no result here (as of today):
https://github.com/pallets/flask/issues?q=is%3Aissue+is%3Aopen+%22redirect_to%22+

It is only mentioned here: https://flask.palletsprojects.com/en/2.3.x/patterns/favicon/?highlight=redirect_to -- doesn't seem to be a first-class API citizen.

Specifically, I was wondering about the HTTP response status code that this is specified to emit (there are various different kinds of redirect responses).

I have previously opted for not using a redirect, but to simply have two routes using the same handler. I think that's more pragmatic. :man_shrugging: https://github.com/conbench/conbench/blob/9f31f6b9f90a58ca084e9b8fda0147f69eca9145/conbench/api/results.py#L256

conbench/tests/app/test_compare.py
```diff
@@ -39,6 +39,41 @@ def test_flash_messages(self, client):
         assert "cannot perform comparison:" in response.text
 
 
+class TestCompareBenchmarks(_asserts.AppEndpointTest):
```
jgehrcke, line 42: Is this the first set of tests exercising the UI's compare view?

Can we rename this to TestCompareBenchmarkResults?

conbench/tests/app/test_compare.py
```diff
@@ -39,6 +39,41 @@ def test_flash_messages(self, client):
...
+        assert any(
+            [
+                "/compare/benchmarks/unknown...unknown2/" == res.request.path
+                for res in response.history
+            ]
+        )
+        assert "cannot perform comparison:" in response.text, response.text
+
+        response = client.get("/compare/benchmarks/foo...bar/", follow_redirects=True)
+        assert "/compare/benchmark-results/foo...bar/" in response.request.url
```
jgehrcke, line 67: I mean, it's OK :)

But I think it's wort pointing out that this tests an implementation detail (exhaustively, even!) that we don't care about too much.

We care about that the route works, not really how it's doing so. My preference in future situations would be to not encode implementation details like this in a test.

A bit of background: we already have a number of tests that probe for implementation details, and this gets in the way when changing things, need to fight this trend a bit.

Just for comparison and towards a better future: a simpler, more focused way to test here could be as simple as:

```python
def test_legacy_route():
    resp = requests.get('http://testhost:1337/compare/benchmarks/foo...bar')
    assert res.status_code == 200, ...
    assert "cannot perform comparison:" in resp.text
```

That would test the interface from the user's point of view. It makes use of the fact that most HTTP clients follow redirects by default (requests.get() does).

About that last point. If you think about this, by picking the redirect method _you_ also assume that "most HTTP clients follow redirects by default"! :) And it's worth probing for that assumption in tests. _if_ one decides to build this with a redirect.

jgehrcke (approved):
Thank you! Left a few critical remarks, for the future :). 

But it's good that we have this taken care of for now! Let's get this in now. Most importantly, we should start using these new URL paths in tooling and reports :).

> Learned a bit about flask and our testing framework along the way.

:rocket:

conbench/app/compare.py
alistaire47: > I was looking for reference docs for the `redirect_to` arg. 

The kwargs from [`add_url_rule()`](https://flask.palletsprojects.com/en/2.3.x/api/#flask.Flask.add_url_rule) that this `rule()` helper is calling get sent to a werkzeug [`Rule`](https://werkzeug.palletsprojects.com/en/2.3.x/routing/#werkzeug.routing.Rule) class, where it's documented. 

> Specifically, I was wondering about the HTTP response status code that this is specified to emit (there are various different kinds of redirect responses).

While sorting out the testing for this I discovered it's definitely emitting a 308

> I have previously decided to to not use a redirect, but to simply have two routes using the same handler. I think that's more pragmatic.

That's fine too if this becomes a pain in any way; I had that for a bit in a previous iteration.

conbench/tests/app/test_compare.py
alistaire47: The class testing the view was already renamed `TestCompareBenchmarkResults` at some point; see L18-19.

This is a lighter-weight copy of that test that mostly just tests the redirect.

Issue #1252, opened 2023-05-09
history plot: min from multi-sample result is often below lookback z-score lower leeway bound


jgehrcke:
Relates to https://github.com/conbench/conbench/issues/594

Example: https://conbench.ursa.dev/benchmark-results/47efd7474830460b8e7243c02f117ca1

![image](https://github.com/conbench/conbench/assets/265630/e537142d-00a4-4962-9106-738b1903f8ee)

Here it happens relatively often that the 'best outcome' is outside the 'leeway interval' (what do we call that?) as defined by the lookback z-score method. If it is not: do we display an appropriate interval? I think we also haven open discussions somewhere where we want to treat the lower bound of that interval different from the upper bound. After all, the movement around the mean is never symmetric, because one of the directions is the "better" and the other direction is "worse" -- this breaks symmetry.

CC @jonkeane @alistaire47 @austin3dickey 
would love to read about your opinions.

austin3dickey:
One (clarifying? complicating?) point to make is that these history plots always show the ±5z interval around the mean, _no matter how alerts are configured_. That's a consequence of handling alerting logic on the CI side, independently from the server. So because of that, I frequently take these interval lines with a grain of salt. To me, they're more of a visual aid that may not directly correspond to how users get alerts.

And yes, compounded on top of that is the fact that we always use the `mean(iterations)` for each benchmark result as the raw data when calculating the z-scores. We've proposed using `min` (or "best" depending on unit) in #640. @jonkeane's comment in that issue is particularly illuminating. I like the proposal that we

> build out a small set of these ... and let folks specify which ones they care about.

This implies a small level of user configuration, which also could include the z-width of the interval, which could solve both problems of (1) my first point and (2) asymmetry.

alistaire47:
> One (clarifying? complicating?) point to make is that these history plots always show the ±5z interval around the mean, _no matter how alerts are configured_.

This is a really important note! And if we start displaying configured thresholds, these plots are going to look ridiculous when they've been set high to avoid false positives...

jonkeane:
IMHO, having the mins here while in-the-fullness-of-time (for most benchmarks) is the right answer, given our current set up, having them display so prominently might be more confusing than it is helpful, just like the core of the issue pointed out here. I know that there was lots of work to add these in, but maybe we turn them off for now if they are confusing and re-enable them when we can support mins-as-the-metric more fully?

[...]

Issue #1230, opened 2023-05-05
ui: certain numbers are missing across the webapp


austin3dickey:
I've noticed a few cases across the webapp where numbers are missing that I'd expect to be there (or be handled better).

1. `/compare/runs` - all `Baseline` and `Contender` column values for benchmark results without errors just display `s` without a number.
2. `/compare/runs` - if a benchmark result has an error, there's no indication of that. I believe we used to display a (!) glyph.
3. `/compare/runs` - if a benchmark result is missing a pairwise change, there's just a `%` displayed
4. `/compare/benchmarks` - the `Baseline` and `Contender` cells just display `s` without a number
5. `/compare/benchmarks` - if the comparison is missing a pairwise change, there's just a `%` displayed
6. `/compare/benchmarks` - the z-score-related rows are always empty and do not belong with the stats anymore
7. `/benchmark-results` - the z-score-related rows are always empty and do not belong with the stats anymore

Screenshots:
<img width="1309" alt="Screenshot 2023-05-05 at 11 35 17" src="https://user-images.githubusercontent.com/16600275/236519232-3abc95d2-0bc3-420c-9a91-ce4dce022fb2.png">
<img width="1220" alt="Screenshot 2023-05-05 at 11 43 49" src="https://user-images.githubusercontent.com/16600275/236519234-9adf082e-f8bd-4758-bcec-3d84d7c24994.png">
<img width="1231" alt="Screenshot 2023-05-05 at 11 46 20" src="https://user-images.githubusercontent.com/16600275/236519237-a7d202a6-2182-47ef-9f05-28ee076e219f.png">
<img width="1320" alt="Screenshot 2023-05-05 at 11 44 09" src="https://user-images.githubusercontent.com/16600275/236519239-e5290abe-dcba-425d-bd7b-7e0709295065.png">
<img width="651" alt="Screenshot 2023-05-05 at 11 48 42" src="https://user-images.githubusercontent.com/16600275/236519241-d4f97805-efa0-4bf0-b6cb-243ac10b2cb8.png">

jgehrcke:
Thank you for noting down the individual places. We have accumulated a whole bunch of bugs like this over time. 

Most important conclusion: we really need to work on UI test coverage.

We even lack GETting relevant routes at all. We should GET them, confirm that response contains HTML, and then also to pragmatic grep-like checks if expected values/numbers/words are in there. That will generate a whole lot of confidence without requiring to do complex UI testing.

[...]

Issue #1209, opened 2023-05-03
History plot on the compare benchmark result page?


jonkeane:
We link to the [compare/runs page](https://conbench.ursa.dev/compare/runs/536d3a2929894f629560e66253bb9418...8d0be02fd53c451985848e1288179e55/) when there are regressions from benchalerts (+ also the bespoke Arrow alerts). This then links to [compare/benchmarks (results) page](https://conbench.ursa.dev/compare/benchmarks/9e32fc4c91274dac8012cb5531e069b6...45789e2ee72a402caa1250144272195c/) for a single benchmark result. But from that page, it's not totally obvious how I could see the historical data so that I could see if this is really a true regression, something that will end up just being noise, or what.

One possible solution here is to add the history plot to this page.

A workaround that exists (but is not at all obvious): one can click the contender link up top and that takes you to the contender page which has the history plot. But in order to know to do that you have to know quite a bit about conbench generally. And that's yet another click to get to the data one is interested in.

jgehrcke:
Yes! Thank you, Jon.

I also forgot to create a ticket earlier based on user feedback on Slack ( :-) @thisisnic)

Some quotes from Nic and myself on the thread

> I expected to see an interactive plot of commits at the top, but it's not there

> In those direct comparison views, we find links to the history for the individual benchmark scenarios. Those are most insightful I think.

> Now I see the path from the notifications to where I want to be (i.e. the history plot), I think I have everything I need here! 

> We will need to make this "path from the notifications to where I want to be" better for sure.

jgehrcke:
Thought about this a little more.

> that takes you to the contender page which has the history plot. But in order to know to do that you have to know quite a bit about conbench generally. And that's yet another click to get to the data one is interested in.

And even if you do so, you do not get to see the two specific benchmark results that you actually wanted to compare in wider context: baseline and contender. All you get is a history plot showing a "current benchmark result", and a smart user might remember that "current benchmark result" is contender or baseline, and they might go actively look for the other one in the history plot by clicking through data points. Really bad UX.

We also cannot just "add the history plot to this page". To be considered: The "compare benchmark result page" allows for comparing a wider range of benchmark results than the history plot is (rightfully) designed to show results for. 

- The history plot requires a set of properties H to be equivalent for two benchmark results to be "history-comparable", i.e. to be shown in the same plot.
- The benchmark result comparison page has a set of properties C that it requires to be equivalent before it shows a comparison.

[...]

PR #1253, opened 2023-05-09
UI: hist plot: add "highlight other" feature, compare two benchmark results: show hist plot and highlight other (& misc)

jgehrcke:
This is mainly for https://github.com/conbench/conbench/issues/1209.

This adds a feature to the history plot to highlight one specific result in the plot, besides the currently highlighted "current" / most recent result.

This feature is exposed in the benchmark result view (which tries to render the history plot) with a query parameter of the following shape:
```
?highlight-other=<other_benchmark_id>,description
```
Example: `http://localhost:5000/benchmark-results/47efd7474830460b8e7243c02f117ca1/?highlight-other=2d298663e6b942f38695467987a889f0,123-release`

leading to
image

Notice how the `123-release` term ends up in the legend of the plot.

When I started to try to use that in the comparison view (to generate a link to the result view), I realized that the compare view wanted to render the history plot, but it was just broken. I decided to re-activate that, and _also_ use this "highlight other" feature :shrug: 

Keeping the query parameter technique in for now. I think it's useful.

Error handling: when the "other" cannot be found:

image

Also adjusted the plot style and legend based on recent insights, and me growing more confidence in what is what. See commit messages! Some notes:
* I think that removing the line connecting the min values is a good idea -- it got too busy. Also see https://github.com/conbench/conbench/issues/1252 for an example of that.
* Renamed "benchmark" to "result" where appropriate
* Started using "lookback z-score" instead of "rolling window"

Also: in the view for comparing two results, I don't think we need a bar chart having two bars, comparing two mean values. Even without history plot I find that a little boring when the two mean values are displayed in raw below that. But with history plot I think that we can probably agree that we don't need the bar chart.

Then there's also a table that at least for me in local dev is kind of broken and also @austin3dickey found it broken here and there: https://github.com/conbench/conbench/issues/1230 -- I wanted to get over with this patch, and for now decided to comment the table out. I think the history plot conveys the relevant information even better. Does it? :)

### Comparison view, before:
image

### Comparison view, after:
image

### DIFF
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -722,12 +722,12 @@ def time_series_plot(
         if multisample_count:
             label += f" (n={multisample_count})"
 
-        p.line(
-            source=source_min_over_time,
-            legend_label=label,
-            name="min-over-time",
-            color="#222",
-        )
+        # p.line(
+        #     source=source_min_over_time,
+        #     legend_label=label,
+        #     name="min-over-time",
+        #     color="#222",
+        # )
         p.circle(
             source=source_min_over_time,
             legend_label=label,
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -785,10 +785,8 @@ def time_series_plot(
             size=18,
             line_width=2.5,
             color="#A65DE7",
-            legend_label="current benchmark (mean)"
-            if multisample
-            else "current benchmark",
-            name="benchmark",
+            legend_label="current result (mean)" if multisample else "current result",
+            name="result",
         )
 
         if multisample:
@@ -797,8 +795,8 @@ def time_series_plot(
                 source=source_current_bm_min,
                 size=6,
                 color="#000",
-                name="benchmark",
                 legend_label="current result (min)",
+                name="result",
             )
 
     # further visually separate out distribution changes with a vertical line
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -676,9 +676,9 @@ def time_series_plot(
     p.xaxis.axis_label = "date of commit associated with the benchmark result"
 
     multisample, multisample_count = _inspect_for_multisample(samples)
-    label = "benchmark (n=1)"
+    label = "result (n=1)"
     if multisample:
-        label = "benchmark mean"
+        label = "result mean"
         if multisample_count:
             label += f" (n={multisample_count})"
 
@@ -718,7 +718,7 @@ def time_series_plot(
         # Do not show min-over-time when each benchmark reports at most one
         # sample, i.e. when mean and min are the same.
 
-        label = "benchmark min"
+        label = "result min"
         if multisample_count:
             label += f" (n={multisample_count})"
 
@@ -796,8 +796,8 @@ def time_series_plot(
                 source=source_current_bm_min,
                 size=6,
                 color="#000",
-                legend_label="current benchmark (min)",
                 name="benchmark",
+                legend_label="current result (min)",
             )
 
     # further visually separate out distribution changes with a vertical line
@@ -858,7 +858,7 @@ def time_series_plot(
     )
 
     p.legend.title_text_color = "darkgray"
-    p.legend.title = f"benchmark results: {len(samples)}"
+    p.legend.title = f"number of results: {len(samples)}"
     p.legend.location = "top_left"
 
     # Change the number of expected/desired date x ticks. There is otherwise
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -523,7 +523,7 @@ def time_series_plot(
     samples: List[HistorySample],
     current_benchmark_result: BenchmarkResult,
     run,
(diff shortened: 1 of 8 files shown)


conbench/app/_plots.py
```diff
@@ -17,17 +16,11 @@
...
+from .types import BokehPlotJSONOrError, HighlightInHistPlot
 
 log = logging.getLogger(__name__)
 
 
-@dataclasses.dataclass
-class BokehPlotJSONOrError:
-    # mutually exclusive
-    jsondoc: Optional[str]
-    reason_why_no_plot: Optional[str]
```
jgehrcke, line 28: circular import errors came back, motivated me to create a new types.py module

conbench/app/_plots.py
```diff
@@ -85,28 +89,49 @@ def get_history_plot(
...
+                s.benchmark_result_id: s for s in samples
+            }
+
+            if highlight_other_result.bmrid not in s_by_id:
+                return BokehPlotJSONOrError(
+                    None,
+                    "you asked to highlight the benchmark result with ID "
+                    f"<kbd>{highlight_other_result.bmrid}</kbd> in this view. "
+                    "However, that result is not part of the history of the "
+                    f"result <kbd>{current_benchmark_result.id}</kbd>",
```
jgehrcke, line 115: kbd is styled by bootstrap
https://getbootstrap.com/docs/5.2/content/reboot/#user-input

conbench/app/_plots.py
```diff
@@ -473,11 +498,11 @@ def gen_js_callback_click_on_glyph_show_run_details(repo_string):
 
         // TODO? show run timestamp, not only commit timestamp.
         var newHtml = \
-            '<li>Report: <a href="' + run_report_relurl + '">' + run_report_relurl + '</a></li>' +
+            '<li>Result view: <a href="' + run_report_relurl + '">' + run_report_relurl + '</a></li>' +
             '<li>Commit: ' + commit_repo_string + '</li>' +
             '<li>Commit message (truncated): ' + run_commit_msg_pfx + '</li>' +
             "<li>Commit timestamp: " + run_date_string + "</li>" +
-            "<li>Result value: " + run_result_value_with_unit + "</li>";
+            "<li>Result mean: " + run_result_value_with_unit + "</li>";
```
jgehrcke, line 505: just a tiny bit of cleanup while I was here

conbench/app/_plots.py
```diff
@@ -643,17 +676,17 @@ def time_series_plot(
...
+        label = "result mean"
         if multisample_count:
             label += f" (n={multisample_count})"
 
     scatter_mean_over_time = p.circle(
         source=source_unfiltered_mean_over_time,
         legend_label=label,
         name="samples",
-        size=6,
+        size=5,
```
jgehrcke, line 689: those grey dots became a little big

conbench/app/_plots.py
```diff
@@ -806,8 +857,9 @@ def time_series_plot(
     )
 
     p.legend.title_text_color = "darkgray"
-    p.legend.title = f"benchmark results: {len(samples)}"
+    p.legend.title = f"number of results: {len(samples)}"
     p.legend.location = "top_left"
+    p.legend.label_text_font_size = "12px"
```
jgehrcke, line 862: 13px is default, i.e. this makes the legend contents a tiny little smaller. I felt motivated to do so because now we may have even one more entry in the legend.

conbench/numstr.py
```diff
@@ -13,7 +13,7 @@
 
 Then I found a test case that fails for sigfig.round(), but does not for
 
-    sigfig.round(15272000, 5) produces "15272000.0", i.e. here we have the
+    sigfig.round(15272063, 5) produces "15272000.0", i.e. here we have the
```
jgehrcke, line 16: CC @alistaire47

jgehrcke:
lint failed here:
```
-        <table id="benchmarks"
-               class="table table-striped table-bordered table-hover">
+        <table id="benchmarks" class="table table-striped table-bordered table-hover">
```
That is weird. I am removing this instead of commenting it out. We can bring it back.

alistaire47:
Pretty! This will help a lot in presenting context for comparisons.

> Then there's also a table that at least for me in local dev is kind of broken and also @austin3dickey found it broken here and there: #1230 -- I wanted to get over with this patch, and for now decided to comment the table out. I think the history plot conveys the relevant information even better. Does it? :)

There is some value in seeing the values for lookback z-score and pairwise percent change—we surface those in alerts, but not otherwise here. There's no need to keep the current table, though, if there's a better way to communicate them. Maybe we could make a generalized table to display the subdicts of the `analysis` part of the `/compare` response? Or just stick them in a title or annotation somewhere if it's too much presentation for too little data.

jgehrcke:
Thanks for feedback. I just thought let's maybe merge this now, then we can maybe show it in the Q&A round that's upcoming now. 

Will certainly want to act on any feedback as it's coming.

PR #1254, opened 2023-05-09
c-bench UI: case perms per benchmark: show each case param with observed values

jgehrcke:
Just a tiny step, but an important one:
image

From here, we can make individual entries clickable / toggleable, and then filter the contents in the table! :)


jgehrcke:
:ship: -- I'll add work on top of this.

PR #1256, opened 2023-05-10
c-bench UI: case permutation overview: add workflow for filtering by case param key/value pairs

jgehrcke:
Relates to https://github.com/conbench/conbench/issues/1154.
This was discussed quite a bit, we need a workflow/UX for browsing case permutations. The [previous patch](https://github.com/conbench/conbench/pull/1254) proposed:

image

I took inspiration from the Grafana UI query builder here. See [this screenshot](https://github.com/conbench/conbench/assets/265630/66e97128-d386-4c92-aa70-c3ab99051fb1). That is: one panel per parameter. Each panel shows the parameter name, the "key". And all observed values, including the value count. I added another count, the number of individual benchmark results submitted that have this key/value pair set.

------

Here, I went a little further and added logic for selecting individual key/value pairs. Upon toggle (select/unselect), the table below these panels automatically updates. It only shows those rows (case permutations) that have the selected case parameter key/value pairs set. A video:

https://github.com/conbench/conbench/assets/265630/4bb36922-a480-4ff5-9b63-293d3274c9a2

Work in progress, that I'd like to test out in staging. 

With two known limitations/bugs:
- Based on string.includes() search, i.e. a `foo=1` matches a `foo=10`. Easy to fix, just not today.
- Multiple selections in the same panel are interpreted as `AND` which never applies to any row. Must be `OR`. Fixable, just not today.

### DIFF
--- a/conbench/app/benchmarks.py
+++ b/conbench/app/benchmarks.py
@@ -1,6 +1,6 @@
+import collections
 import logging
 import time
-import collections
 from typing import Dict, List, Tuple, TypedDict
 
 import flask
@@ -225,7 +225,7 @@ def show_benchmark_results(bname: str, caseid: str) -> str:
 
     results_by_hardware_and_context: Dict[
         Tuple, List[BMRTBenchmarkResult]
-    ] = defaultdict(list)
+    ] = collections.defaultdict(list)
 
     # Build up timeseries of results (group results, don't sort them yet).
     for result in matching_results:
--- a/conbench/app/benchmarks.py
+++ b/conbench/app/benchmarks.py
@@ -1,6 +1,6 @@
 import logging
 import time
-from collections import defaultdict
+import collections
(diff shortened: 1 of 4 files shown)


jgehrcke:
Another patch upcoming, want to keep this simple. Merging. Feedback of course much appreciated.

jgehrcke:
Some notes to self:
https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes
https://api.jquery.com/data/
https://stackoverflow.com/a/34913701/145400
https://stackoverflow.com/a/14379304/145400
https://datatables.net/manual/plug-ins/search
https://datatables.net/forums/discussion/55461/searching-filtering-using-custom-predicate-function
https://datatables.net/forums/discussion/63184
https://datatables.net/forums/discussion/comment/81658/#Comment_81658

PR #1259, opened 2023-05-11
c-bench UI: case permutations: use space more wisely

jgehrcke:
This patch is mainly about using UI space more wisely, mainly in the "case permutation" view of the conceptual benchmark UI section.

This uses the Bootstrap 5 flex wrap system with fixed width 'panels'. Used the typical incredients to save space: padding, font size, margin, ... also scrollbar in this case. Also worked on wasting less vertical space, i.e. getting 'right to business' below the navbar (pulled misc info to bottom).

### Before

This is a sanitized screenshot from a prod system benchmark with a largish number of case parameters, showing how much space this can currently consume:

Screenshot from 2023-05-11 15-10-16

### After

This is a local dev screenshot (not the same state of data as above, but still demonstrating the change quite vividly I think:
Screenshot from 2023-05-11 15-02-55

### DIFF
--- a/conbench/static/app.css
+++ b/conbench/static/app.css
@@ -7,8 +7,44 @@ html {
   font-size: 14px;
 }
 
+.c-bench-scrollbar {
+  /* Foreground, Background */
+  scrollbar-color: #812570 #f3f3f3;
+  scrollbar-width: thin;
+}
+.c-bench-scrollbar::-webkit-scrollbar {
+  width: 5px; /* Mostly for vertical scrollbars */
+  height: 7px; /* Mostly for horizontal scrollbars */
+}
+.c-bench-scrollbar::-webkit-scrollbar-thumb { /* Foreground */
+  background: #81257088;
+  border-radius: 2px;
+}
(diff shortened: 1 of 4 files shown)


alistaire47:
Nice! If we need to save more space here, skipping cards for params with one value and handling those separately is an option. We have a good number of those in various contexts, for better or worse reasons.

jgehrcke:
Thank you, Ed!

> skipping cards for params with one value and handling those separately is an option. We have a good number of those in various contexts, for better or worse reasons.

Good feedback.

Yeah, thought about that a bit, too. I agree that there is opportunity to save much more space. One could toggle the selection/dropdown only when needed, for example, and otherwise display the 'currently selected' value.

And just for completeness: The 'currently selected value' can, for all parameters, also be "not defined". Generally, at least. The  "one value scenarios" as they show in the UI right now _may_ actually be two-value scenarios. (I didn't count the "not defined" state yet). It may make sense to explicitly represent the "any value or not defined" state in the UI (it's the default right now: don't select a value, and the table is not constrained).

PR #1260, opened 2023-05-11
benchalerts: handle empty runs/results better

### LINKED ISSUE #1191: benchalerts: notify/fail if there are no results on the contender
Right now I suspect that when running `benchalerts`, if there are no benchmark results associated with any contender run, you will get a PR comment like

>There were no benchmark performance regressions. 🎉

with a successful GitHub Check. I think we definitely want to notify about this case, and probably by default we'd want to fail the GitHub Check. We could make that user-configurable in https://github.com/conbench/conbench/issues/756.

austin3dickey:
Fixes #1191 and some other issues I've noticed while looking through the code.

To sum up this PR, the original stance of `benchalerts` was "if the analysis can't find any contender runs or any contender benchmark results, raise an error." That's turned out to hide some complexities that could be surfaced to consumers of the alerts so that they could take action. This PR makes it so it's okay if we can't find contender runs or results, as long as we surface that detail to the user.

### DIFF
--- a/benchalerts/benchalerts/conbench_dataclasses.py
+++ b/benchalerts/benchalerts/conbench_dataclasses.py
@@ -3,7 +3,7 @@
 from dataclasses import dataclass
 from typing import List, Optional
 
-from benchclients.logging import fatal_and_log
+from benchclients.logging import fatal_and_log, log
 
 
 @dataclass
@@ -25,6 +25,8 @@ class RunComparisonInfo:
     """Track and organize specific info about a comparison between a contender run and
     its baseline run.
 
+    compare_results and benchmark_results are mutually exclusive.
+
     Parameters
     ----------
     contender_info
@@ -39,11 +41,10 @@ class RunComparisonInfo:
         to its baseline, including the statistics and regression analysis.
     benchmark_results
         The list returned from Conbench when hitting
-        /benchmark-results/?run_id={contender_run_id}, only if the contender run has
-        errors and there is no baseline run. Contains info about each benchmark result
-        in the contender run, including statistics and tracebacks. Only used when a
-        baseline run doesn't exist because otherwise all this information is already in
-        the compare_results.
+        /benchmark-results/?run_id={contender_run_id}, only if there is no baseline run.
+        Contains info about each benchmark result in the contender run, including
+        statistics and tracebacks. Only used when a baseline run doesn't exist because
+        otherwise all this information is already in the compare_results.
     """
 
     contender_info: dict
@@ -140,11 +141,6 @@ class RunComparisonInfo:
         return f"{self.app_url}/benchmark-results/{contender_result_id}"
 
     @property
-    def has_errors(self) -> bool:
-        """Whether this run has any benchmark errors."""
-        return self.contender_info["has_errors"]
-
-    @property
     def contender_id(self) -> str:
         """The contender run_id."""
         return self.contender_info["id"]
@@ -190,14 +186,34 @@ class FullComparisonInfo:
     the comparisons to their baselines.
     """
 
+    # Can be empty.
     run_comparisons: List[RunComparisonInfo]
 
-    def __post_init__(self):
-        # Any code that constructs an instance should ensure run_comparisons is non-empty
-        assert self.run_comparisons
+    @property
+    def has_any_contender_runs(self) -> bool:
+        """Whether there are any contender runs available."""
+        return bool(self.run_comparisons)
 
     @property
-    def benchmarks_with_errors(self) -> List[BenchmarkResultInfo]:
+    def has_any_contender_results(self) -> bool:
+        """Whether there are any contender benchmark results available."""
+        for run_comparison in self.run_comparisons:
+            if run_comparison.contender_benchmark_result_info:
+                return True
+        return False
+
+    @property
+    def has_any_z_analyses(self) -> bool:
+        """Whether there are any lookback z-score analyses available."""
+        for run_comparison in self.run_comparisons:
+            if run_comparison.compare_results:
+                for compare_result in run_comparison.compare_results:
+                    if compare_result["analysis"]["lookback_z_score"]:
+                        return True
+        return False
+
+    @property
+    def results_with_errors(self) -> List[BenchmarkResultInfo]:
         """Get information about all benchmark results with errors across runs."""
         out = []
 
@@ -211,7 +227,7 @@ class FullComparisonInfo:
         return out
 
     @property
-    def benchmarks_with_z_regressions(self) -> List[BenchmarkResultInfo]:
+    def results_with_z_regressions(self) -> List[BenchmarkResultInfo]:
         """Get information about all benchmark results across runs whose z-scores were
         extreme enough to constitute a regression.
         """
@@ -227,14 +243,7 @@ class FullComparisonInfo:
         return out
 
     @property
-    def no_baseline_runs(self) -> bool:
-        """Whether all contender runs are missing a baseline run."""
-        return not any(
-            run_comparison.baseline_id for run_comparison in self.run_comparisons
-        )
-
-    @property
-    def z_score_threshold(self) -> float:
+    def z_score_threshold(self) -> Optional[float]:
         """The z-score threshold used in this analysis."""
         z_score_thresholds = set()
         for run_comparison in self.run_comparisons:
@@ -245,8 +254,10 @@ class FullComparisonInfo:
                             compare["analysis"]["lookback_z_score"]["z_threshold"]
                         )
 
+        if len(z_score_thresholds) == 0:
+            return None
         if len(z_score_thresholds) != 1:
-            fatal_and_log(
+            log.warn(
                 f"There wasn't exactly one z_score_threshold: {z_score_thresholds=}"
             )
         return z_score_thresholds.pop()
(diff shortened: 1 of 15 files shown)


benchalerts/benchalerts/conbench_dataclasses.py
```diff
@@ -190,14 +186,34 @@ class FullComparisonInfo:
...
+        for run_comparison in self.run_comparisons:
+            if run_comparison.compare_results:
+                for compare_result in run_comparison.compare_results:
+                    if compare_result["analysis"]["lookback_z_score"]:
+                        return True
+        return False
 
     @property
-    def benchmarks_with_errors(self) -> List[BenchmarkResultInfo]:
+    def results_with_errors(self) -> List[BenchmarkResultInfo]:
```
austin3dickey, line 216: Another clarifying rename (it doesn't really have to do with this PR).

benchalerts/benchalerts/message_formatting.py
```diff
@@ -172,44 +196,61 @@ def pr_comment_link_to_check(
...
-    else:
-        comment += _clean(
-            """
-            There were no benchmark performance regressions. 🎉
+            None of the specified runs had any associated benchmark results.
             """
         )
         comment += "\n\n"
+    else:
+        if full_comparison.results_with_errors:
```
austin3dickey, line 214: This is mostly whitespace diff.

benchalerts/benchalerts/pipeline_steps/github.py
```diff
@@ -42,6 +42,8 @@ class GitHubCheckStep(AlertPipelineStep):
     external_id
         An optional reference to another system. This argument will set the
         "external_id" property of the posted GitHub Check, but do nothing else.
+    build_url
+        An optional build URL to include in the GitHub Check.
```
austin3dickey, line 46: This will be nice to include!

benchalerts/benchalerts/pipeline_steps/github.py
```diff
@@ -117,18 +115,37 @@ def run_step(
...
-        else:
+        elif full_comparison.has_any_z_analyses:
+            # we analyzed z-scores and didn't find errors or regressions
             return CheckStatus.SUCCESS
+        elif full_comparison.has_any_contender_results:
+            # we have results but couldn't analyze z-scores (normal at beginning of history)
+            return CheckStatus.SKIPPED
+        else:
+            # we don't have results; this requires action
+            return CheckStatus.ACTION_REQUIRED
```
austin3dickey, line 132: This change fixes #1191 by failing when there are no results. I think that's a sensible default (it's probably pretty bad if you expected results and didn't get them) but users can change it once we do #774 .

jonkeane:
I haven't yet dug deeply into this, but I wanted to say (1) thank you for following up on this, having more details here is fantastic (2) thank you for the thoughtful comments on various bits to help review — those are super super helpful

benchalerts/tests/unit_tests/expected_md/details_noruns.md
```diff
@@ -0,0 +1 @@
+This report was not associated with any commit connected to the git graph. This probably means that benchmarks were run on a transient merge-commit.
```
jonkeane, line 1: Nice additional detail here!

benchalerts/tests/unit_tests/expected_md/summary_noruns.md
```diff
@@ -0,0 +1,3 @@
+Conbench analyzed the 0 benchmark runs that triggered this notification.
+
+None of the specified runs were found on the Conbench server. See the [build logs](https://austin.something) for more information.
```
jonkeane, line 3: ❤️ for linking out to where one might find a way to figure out what's up. That's fantastic.

benchalerts/benchalerts/conbench_dataclasses.py
```diff
@@ -190,14 +186,34 @@ class FullComparisonInfo:
     the comparisons to their baselines.
     """
 
+    # Can be empty.
     run_comparisons: List[RunComparisonInfo]
```
alistaire47, line 190: Just to double-check what all of this is indicating, this can now be an empty list, but still not `None`, correct?

alistaire47 (approved):

jgehrcke (approved):
>  This PR makes it so it's okay if we can't find contender runs or results, as long as we surface that detail to the user.

Love the sentiment of this. Assuming this kind of perspective is important and highly valuable. Thanks for the great work in here! (I did not look in detail)

benchalerts/benchalerts/conbench_dataclasses.py
jgehrcke: If the type annotation is correct then `run_comparisons` cannot be None (sorry for being caption obvious here haha :ship:)

benchalerts/benchalerts/conbench_dataclasses.py
```diff
@@ -190,14 +186,34 @@ class FullComparisonInfo:
...
+
+    @property
+    def has_any_z_analyses(self) -> bool:
+        """Whether there are any lookback z-score analyses available."""
+        for run_comparison in self.run_comparisons:
+            if run_comparison.compare_results:
+                for compare_result in run_comparison.compare_results:
+                    if compare_result["analysis"]["lookback_z_score"]:
+                        return True
+        return False
```
jgehrcke, line 213: A useless comment on code style.

I _like_ this style here.

Not sure how to put it into words: saving the `else:` clauses, having the catch-all-other-cases return statement.

Often we/people write code in the inverse shape. Which is not wrong, but this one here is just so much more problem-oriented and elegant I think.

benchalerts/benchalerts/conbench_dataclasses.py
jgehrcke: Feels good. :)

benchalerts/benchalerts/conbench_dataclasses.py
```diff
@@ -245,8 +254,10 @@ def z_score_threshold(self) -> float:
                             compare["analysis"]["lookback_z_score"]["z_threshold"]
                         )
 
+        if len(z_score_thresholds) == 0:
+            return None
         if len(z_score_thresholds) != 1:
-            fatal_and_log(
+            log.warn(
```
jgehrcke, line 260: thanks for doing this here, too!

benchalerts/benchalerts/message_formatting.py
```diff
@@ -82,11 +82,35 @@ def _intro_sentence(full_comparison: FullComparisonInfo) -> str:
...
 
-    if full_comparison.benchmarks_with_errors:
+    if not full_comparison.has_any_contender_runs:
+        summary += _clean(
+            """
+            None of the specified runs were found on the Conbench server.
+            """
+        )
+        if build_url:
+            summary += f" See the [build logs]({build_url}) for more information."
```
jgehrcke, line 98: Interesting not so important language question: what do you think is nicer?

"see build logs" vs "see the build logs"

I would often choose "see build logs" because it is shorter, but does it also sound like a weird person talking? :D

benchalerts/benchalerts/message_formatting.py
```diff
@@ -172,44 +196,61 @@ def pr_comment_link_to_check(
...
+                There {were} {len(full_comparison.results_with_errors)} benchmark
+                result{s} with an error:
+                """
+            )
+            comment += _list_results(full_comparison.results_with_errors, limit=2)
+
+        if not full_comparison.has_any_z_analyses:
+            comment += _clean(
+                """
+                There weren't enough matching historic benchmark runs to make a call on
```
jgehrcke: "runs"... hm. 

benchmark executions?
benchmark results?

benchalerts/tests/unit_tests/expected_md/comment_noresults.md
```diff
@@ -0,0 +1,5 @@
+Conbench analyzed the 1 benchmark run on commit `abc`.
```
jgehrcke, line 1: Style/language questions which we can resolve/conclude later.

- 'analyzed 1 run' vs 'analyzed the 1 run' -- as a reader, I don't know what _the_ refers to, and I can live w/o it
- 'one' vs '1'. In scientific publishing, I had once learned that we want to use words for numbers smaller than 11 or ten :D (see? haha). This is not a rule, just a guideline, but here I think the guideline applies, aesthetically.

jgehrcke: That is, I propose:

> Conbench analyzed one benchmark run on commit `abc`

:)  (I think it reads kind of much nicer even :shrug:)

jgehrcke: From here, we can refine further. "run" is a tough concept, as we find out.

> Conbench analyzed one run of benchmarks on ...

But yeah. For later probably.

benchalerts/benchalerts/message_formatting.py
jonkeane: IMO, "see build logs" reads a bit like newspaper-headline style where articles that are required for grammatical sentences are (sometimes) omitted. And the '"the" in "see the build logs" is required for a grammatical sentence in this case (linguistically, the fact that we link to them is a pretty strong sign they are definite).

benchalerts/benchalerts/conbench_dataclasses.py
austin3dickey: > this can now be an empty list, but still not None

Correct!

benchalerts/benchalerts/conbench_dataclasses.py
austin3dickey: I do too! I wrote a few of these almost completely with GitHub Copilot 🙃 🤖

benchalerts/benchalerts/message_formatting.py
austin3dickey: good call, fixed

benchalerts/tests/unit_tests/expected_md/comment_noresults.md
austin3dickey: cool, made #1269!

benchalerts/benchalerts/conbench_dataclasses.py
jgehrcke: >  with GitHub Copilot upside_down_face robot

haha! Thanks for the punch, I still want to be proud of my craft sometimes :cry: Took years to converge to this style (to see more often when it's possible _after_ having written the other variant, or to start with that right away). Anyway :)

PR #1262, opened 2023-05-12
c-bench: case param panels: css improvements based on prod data

jgehrcke:
case parm panels:
- flexible width
- better color contrasts
- misc


jgehrcke:
pure css change, CI isn't doing anything here - skipping CI

Issue #1264, opened 2023-05-15
c-bench UI: current "by results per case" metric easily dominated by old data (favor recency)


jgehrcke:
Follow-up from meeting with Jon.

For a given benchmark (defined by its name) there are C case permutations. For each case permutation, there are $`r_i`$ results.

Currently we derive a metric that is "results per case permutation" as a simple way to normalize:

$$ \frac{1}{C} \sum_{i=1}^C r_i$$

But that's not yet super useful to sort benchmarks by, because this can easily be dominated by "old data". Let's add a bit of recency/age to the mix.

Then:

$$ \frac{1}{C} \sum_{i=1}^C \frac{1}{a_i} r_i$$

where $`a_i`$ is going to be an age metric for the given case permutation. Either the age of the most recent result, or the median age, or something like that.

That changes the unit of the metric to 

$$ \left[  \frac{results}{cases \cdot  days} \right]$$

(if the age metric is used in "days")

Maybe think of this as "results per case and recency" metric, in code `rpcr`

jgehrcke:
After #1265 we can now close this as 'done', but of course we may want to iterate.

PR #1265, opened 2023-05-15
c-bench UI: introduce/tweak results per case per recency metric (RPCR)

jgehrcke:
This is mainly for https://github.com/conbench/conbench/issues/1264.

Arbitrary choices within RPCR calc:
- unit "days" instead of "seconds" so that the metric is more like O(1000) than O(0.1) for common scenarios.
- fraction: 10 % -- played with this quickly with real-world data. The weight should be on the recent fraction of results. If this is for example exclusively the last result then the metric becomes unstable when the result is extremely new. If this is for e.g. the last 50 % of results then the metric is dominated again by too old results.

By writing about this in all detail I don't want to raise the impression that this is overly smart or important or super well thought through. It's meant to be a quick improvement. With #1264 and this description here I just wanted to document what would otherwise be exclusively doddle scribbling scratchy scrawly d00dle on my desk invisible to nobody (probably including myself in a few weeks from now -- where is this piece of paper? :scroll:)  :).

### DIFF
--- a/conbench/app/benchmarks.py
+++ b/conbench/app/benchmarks.py
@@ -1,9 +1,11 @@
 import collections
 import logging
+import math
 import time
 from typing import Dict, List, Tuple, TypedDict
 
 import flask
+import numpy as np
 import orjson
 
 import conbench.numstr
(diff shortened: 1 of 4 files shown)


jgehrcke:
:ship:

conbench/app/benchmarks.py
```diff
@@ -61,22 +63,49 @@ def list_benchmarks() -> str:
...
+                sorted((now - r.started_at for r in cresults), reverse=True)
+            )
+
+            # Take a most recent fraction of this list and build the mean value
+            # (mean age in seconds of the last 10 % of the results)
+            a_i_seconds = np.mean(
+                ages_seconds[-math.ceil(len(ages_seconds) / 10) :]  # noqa
+            )
+            a_i_days = float(a_i_seconds / (24.0 * 60 * 60))
+            rpcr += 1 / a_i_days * len(cresults)
```
jgehrcke, line 99: maybe try to read meaning into these variable names only after reading https://github.com/conbench/conbench/issues/1264

PR #1266, opened 2023-05-15
c-bench UI: case permutation view: better contrasts, make better use of spacing

jgehrcke:
Major goals:
- make better use of macro spacing, overall optimization target is to use vertical space more efficiently
- make better use of micro spacing: tiny padding/margin optimizations for much better readability/tidiness
- make better use of font spacing and color contrast to show _dense_ information (always a little ugly) with an OKish appearance.

No screenshots here, sadly, because I tested against a prod DB snapshot and don't want to show DB contents here.

### DIFF
--- a/conbench/app/benchmarks.py
+++ b/conbench/app/benchmarks.py
@@ -1,5 +1,4 @@
 import collections
-import itertools
 import logging
 import math
 import time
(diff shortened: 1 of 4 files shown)


jgehrcke:
:ship: please keep the feedback/questions coming!

Issue #1271, opened 2023-05-16
UI: remove version number / legacy CLI version number


jgehrcke:
From https://github.com/conbench/conbench/issues/617#issuecomment-1549501784: 

> Action item: remove the Legacy CLI version number from the Conbech web application UI.

jgehrcke:
Done via https://github.com/conbench/conbench/pull/1272.

PR #1272, opened 2023-05-16
API/UI: remove misleading version number; require buildinfo (always show commit hash)

jgehrcke:
This is for #1271.


jonkeane (approved):
Thanks for this

PR #1273, opened 2023-05-16
c-bench ui: fix literal "nan" in array (#1263)

### LINKED ISSUE #1263: c-bench UI: "nan" in uplot data breaks plot
image

from the raw JSON in the HTML doc:

```
        "21.64405",
        "21.30293",
        "nan",
        "21.42239",
        "22.25914",
        "21.36055",
        "21.74071",
```
https://github.com/leeoniya/uPlot/issues/208

> Ah nevermind, you're supposed to use null not NaN.

jgehrcke:
This fixes #1263. 

Ha! :) It was the new `numstr()`, and in there the `np.format_float_positional(float("NaN"))` that would return the literal string 'nan'.

Confirmed locally that we now have a `null` in the JSON array for uplot to not fail:
image


jgehrcke:
:ship:

PR #1274, opened 2023-05-17
webapp runs page: add links to baselines

### LINKED ISSUE #1175: run page: Surface new comparisons in UI
After #1078, each contender run has up to three candidate baseline runs. We should show links to those on the `/runs/{run_id}` page.

Previously we did this:
image

Proposal: display a link for each of
- `± compare to baseline run from parent commit`
- `± compare to baseline run from fork point commit`
- `± compare to latest baseline run on default branch`

but don't display any link for which the baseline run doesn't exist.

austin3dickey:
Fixes #1175.

This PR adds back the capability to compare contender Runs to their candidate baseline runs by clicking a link at the top of the Run page. Now there are a few options, thanks to the new API structure.

Examples:

Run on the default branch:
image

Run on a PR branch:
image

### DIFF
--- a/conbench/app/compare.py
+++ b/conbench/app/compare.py
@@ -75,7 +75,9 @@ class Compare(AppEndpoint, BenchmarkResultMixin, RunMixin, TimeSeriesPlotMixin):
 
         if comparisons and self.type != "benchmark-result":
             comparisons_by_id = {
-                c["contender"]["benchmark_result_id"]: c for c in comparisons
+                c["contender"]["benchmark_result_id"]: c
+                for c in comparisons
+                if c["contender"]
             }
             if self.type == "run":
                 benchmarks, response = self._get_benchmarks(run_id=contender_id)
@@ -272,8 +274,11 @@ class Compare(AppEndpoint, BenchmarkResultMixin, RunMixin, TimeSeriesPlotMixin):
         # Mutate comparison objs (dictionaries) on the fly
         for c in comparisons:
             view = "app.compare-benchmark-results"
-            compare = f'{c["baseline"]["benchmark_result_id"]}...{c["contender"]["benchmark_result_id"]}'
-            c["compare_benchmarks_url"] = f.url_for(view, compare_ids=compare)
+            if c["baseline"] and c["contender"]:
+                compare = f'{c["baseline"]["benchmark_result_id"]}...{c["contender"]["benchmark_result_id"]}'
+                c["compare_benchmarks_url"] = f.url_for(view, compare_ids=compare)
+            else:
+                c["compare_benchmarks_url"] = None
 
             if (
                 c["analysis"]["lookback_z_score"]
--- a/conbench/templates/run.html
+++ b/conbench/templates/run.html
@@ -7,9 +7,22 @@
       {% if run_id %}<li class="breadcrumb-item active" aria-current="page">{{ run_id }}</li>{% endif %}
     </ol>
   </nav>
-  {% if compare_runs_url %}
+  {% if compare_to_baseline_urls['parent'] %}
     <p class="fs-5">
-      <i class="bi bi-file-diff"></i> <a href="{{ compare_runs_url }}">compare to baseline</a>
+      <i class="bi bi-file-diff"></i> <a href="{{ compare_to_baseline_urls['parent'] }}">compare to baseline run from
+      parent commit</a>
+    </p>
+  {% endif %}
+  {% if compare_to_baseline_urls['fork_point'] %}
+    <p class="fs-5">
+      <i class="bi bi-file-diff"></i> <a href="{{ compare_to_baseline_urls['fork_point'] }}">compare to baseline run from
+      fork point commit</a>
+    </p>
+  {% endif %}
+  {% if compare_to_baseline_urls['latest_default'] %}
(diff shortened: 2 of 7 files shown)


conbench/app/compare.py
```diff
@@ -272,8 +274,11 @@ def _compare(
         # Mutate comparison objs (dictionaries) on the fly
         for c in comparisons:
             view = "app.compare-benchmark-results"
-            compare = f'{c["baseline"]["benchmark_result_id"]}...{c["contender"]["benchmark_result_id"]}'
-            c["compare_benchmarks_url"] = f.url_for(view, compare_ids=compare)
+            if c["baseline"] and c["contender"]:
+                compare = f'{c["baseline"]["benchmark_result_id"]}...{c["contender"]["benchmark_result_id"]}'
+                c["compare_benchmarks_url"] = f.url_for(view, compare_ids=compare)
+            else:
+                c["compare_benchmarks_url"] = None
```
austin3dickey, line 281: Small bugfixes due to new API structure.

conbench/templates/run.html
```diff
@@ -73,7 +86,8 @@ <h4>Benchmark results in Run</h4>
               <a href="{{ url_for('app.batch', batch_id=benchmark.batch_id) }}">{{ benchmark.display_bmname }}</a>
             </td>
             <td>
-              <a href="{{ url_for('app.benchmark-result', benchmark_result_id=benchmark.id) }}">{{ benchmark.display_case_perm }}</a>
+              <a href="{{ url_for('app.benchmark-result', benchmark_result_id=benchmark.id) }}">{{
+              benchmark.display_case_perm }}</a>
```
austin3dickey, line 90: Everything here and below this line in this file was caused by my VS Code format-on-save feature, sorry. Our current linting system doesn't have a problem with these changes either before or after.

conbench/tests/populate_local_conbench.py
```diff
@@ -678,5 +678,10 @@ def sample_slowdown(s, offset: float):
     4e7441d72f1d14cf9a40525eca8993f7713937a7
     """
 
+# Add a PR commit to the Conbench commits.
+CONBENCH_COMMIT_HASH_LINES_50 = (
+    "9da0b7a68ea3003eb5431f1a5708222f1c15efa6" + CONBENCH_COMMIT_HASH_LINES_50
+)
```
austin3dickey, line 684: This is nice!

austin3dickey:
temporary CI instability: https://github.com/conbench/conbench/issues/794#issuecomment-1553025039

conbench/app/compare.py
austin3dickey: Just added a test for this file's bugfixes (confirmed that it fails on `main` but passes now)

jonkeane: Thanks for adding tests to this — we should have had them already, but better late than never!

conbench/templates/run.html
```diff
@@ -7,9 +7,22 @@
...
+  {% if compare_to_baseline_urls['fork_point'] %}
+    <p class="fs-5">
+      <i class="bi bi-file-diff"></i> <a href="{{ compare_to_baseline_urls['fork_point'] }}">compare to baseline run from
+      fork point commit</a>
+    </p>
+  {% endif %}
+  {% if compare_to_baseline_urls['latest_default'] %}
+    <p class="fs-5">
+      <i class="bi bi-file-diff"></i> <a href="{{ compare_to_baseline_urls['latest_default'] }}">compare to latest
+      baseline run on default branch</a>
```
jonkeane, line 25: This will complicate this code (and we could defer this to a new issue, it's a bit of scope creep), but it would be fantastic if we had a label or something that marked which comparison we believe most in. Something like a little `default` or something on:

* parent if the commit is on the default branch
* fork point if the commit is on a non-default branch

This way it's clear to folks looking at this which comparison we firmly believe is the best one for the data they are looking at.

jonkeane (approved):
Thanks for this!

conbench/templates/run.html
austin3dickey: This is a good idea and shouldn't be too complicated. Right now,

1. if the run isn't associated with a commit, only `latest_default` can be found
    - `parent` and `fork_point` don't apply because it's not on the git graph
2. if the run is on the default branch, only `parent` can be found
    - `fork_point` doesn't make sense and we [decided](https://github.com/conbench/conbench/pull/1078#discussion_r1174160609) not to find `latest_default` because analysis could get confusingly circular
3. if the run is on a PR branch, all three can be found

So we'll only need to "recommend" one in the last case, in which case the recommendation will always be `fork_point`.

jonkeane: _nods_ I would also expect that a(n oldish) run on default might also have `latest_default` too? If a commit after the current one has benchmarks. That's totally fine, and I can imagine a case where someone might want to do that, but we should definitely warn that that's not our recommended, and instead the parent commit is the recommended one. right?

austin3dickey: Yeah, now that we have an official recommendation in the webapp it's probably less dangerous to return the `latest_default` baseline for default-branch commits too. I'll do this in a followup PR since it'll have a bigger scope.

PR #1275, opened 2023-05-18
api/runs, app/runs: enable latest_default baselines for default branch commits, recommend one in app

austin3dickey:
This PR addresses @jonkeane's feedback in https://github.com/conbench/conbench/pull/1274#discussion_r1197953929. It:

- enables the `latest_default` candidate baseline run type coming from the `/api/runs/{run_id}` endpoint for contender runs that are on the default branch
- enables those in the webapp too
- labels the "recommended" baseline run in the webapp and moves it to the top of the list

Screenshots:

Run on a PR branch:
image

Run on the default branch:
image

### DIFF
--- a/conbench/templates/run.html
+++ b/conbench/templates/run.html
@@ -9,7 +9,7 @@
   </nav>
   {% for link in comparisons %}
     <p class="fs-6">
-      <i class="bi bi-file-diff"></i> <a href="{{ link.url }}">{{ link.full_text }}</a>
+      <i class="bi bi-file-diff"></i> <a href="{{ link.url }}">{{ link.text }}</a> {{ link.badge | safe }}
     </p>
   {% endfor %}
   {% if plot_history %}
--- a/conbench/templates/run.html
+++ b/conbench/templates/run.html
@@ -7,24 +7,11 @@
       {% if run_id %}<li class="breadcrumb-item active" aria-current="page">{{ run_id }}</li>{% endif %}
     </ol>
   </nav>
-  {% if compare_to_baseline_urls['parent'] %}
-    <p class="fs-5">
-      <i class="bi bi-file-diff"></i> <a href="{{ compare_to_baseline_urls['parent'] }}">compare to baseline run from
-      parent commit</a>
+  {% for link in comparisons %}
+    <p class="fs-6">
+      <i class="bi bi-file-diff"></i> <a href="{{ link.url }}">{{ link.full_text }}</a>
     </p>
-  {% endif %}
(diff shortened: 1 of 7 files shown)


conbench/tests/api/test_runs.py
```diff
@@ -79,7 +79,11 @@ def test_get_run(self, client):
                 run,
                 candidate_baseline_runs={
                     "fork_point": DEFAULT_BRANCH_PLACEHOLDER,
-                    "latest_default": DEFAULT_BRANCH_PLACEHOLDER,
+                    "latest_default": {
+                        "baseline_run_id": run.id,
```
austin3dickey: Open question: as this stands, if the contender run is already on the latest commit on the default branch, the `latest_default` baseline run will be itself. Is that okay? Or should I exclude the contender run from the search?

jonkeane: IMO we should exclude that from the list, but if that's a lot of work to do it's ok if it's there. It's definitely cleaner to not have that there + have someone click and wonder why something is being compared to itself

conbench/templates/run.html
```diff
@@ -7,24 +7,11 @@
       {% if run_id %}<li class="breadcrumb-item active" aria-current="page">{{ run_id }}</li>{% endif %}
     </ol>
   </nav>
-  {% if compare_to_baseline_urls['parent'] %}
-    <p class="fs-5">
-      <i class="bi bi-file-diff"></i> <a href="{{ compare_to_baseline_urls['parent'] }}">compare to baseline run from
-      parent commit</a>
+  {% for link in comparisons %}
+    <p class="fs-6">
+      <i class="bi bi-file-diff"></i> <a href="{{ link.url }}">{{ link.full_text }}</a>
```
jonkeane: I know this would involve changing the function that generates this, but would it be possible to have the "(recommended)" be not hyperlinked (so after that `</a>`? It's not a huge deal, but it would be nice if that were slightly set apart (ideally a small badge or something so that it's clear it's an annotation on the link, but I appreciate doing something and getting it out quickly!)

austin3dickey: Cool, got the badge working:
image

conbench/tests/api/test_runs.py
austin3dickey: Done! Will push in a second

conbench/templates/run.html
jonkeane: OOOH nice! that's fantastic

jgehrcke: BS badge: good choice!

PR #1277, opened 2023-05-22
api: /benchmark-results/run_reason: fix sort order

### LINKED ISSUE #1276: api:  /benchmark-results/run_reason?<reason> does not query with ORDER BY (unstable result set)
I missed adding the order_by arg here:
https://github.com/conbench/conbench/blob/736a71fa629d5335e98fe00210129fdfa231601e/conbench/api/results.py#L191

The outcome is an unstable result set, with varying distribution over time.

```
$ curl localhost:5000/api/benchmark-results/?run_reason=nightly | python checkthis.py before
$ curl localhost:5000/api/benchmark-results/?run_reason=nightly | python checkthis.py before2
$ curl localhost:5000/api/benchmark-results/?run_reason=nightly | python checkthis.py before3
```

results_over_time-before
results_over_time-before2

results_over_time-before3

Thanks @boshek for noticing that something is fishy :fish: 

The "quick patch strategy" is once again proven to likely be error-prone :).

jgehrcke:
This fixes https://github.com/conbench/conbench/issues/1276.
```
curl localhost:5000/api/benchmark-results/?run_reason=nightly | python checkthis.py after
```
results in

results_over_time-after

I have increased N so that in this case we go back up to ~December last year.

### DIFF
--- a/conbench/api/results.py
+++ b/conbench/api/results.py
@@ -189,7 +189,10 @@ class BenchmarkListAPI(ApiEndpoint, BenchmarkValidationMixin):
 
         elif run_reason_arg := f.request.args.get("run_reason"):
             benchmark_results = BenchmarkResult.search(
-                filters=[Run.reason == run_reason_arg], joins=[Run], limit=35000
+                filters=[Run.reason == run_reason_arg],
+                joins=[Run],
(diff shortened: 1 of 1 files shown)


jgehrcke:
We should add tests.
For now, I'd like to see this rolled out.

PR #1278, opened 2023-05-23
c-bench UI: /<bname>/trends -- timeseries analysis on BMRT cache, display conspicious canditates

jgehrcke:
Showing timeseries plots / sparklines / small multiples only after selecting benchmark -> case is a little deep, late. Pull this one level up (this is a major takeaway of a conversation I had with Jon). But when going one level up, it becomes important to decide _which_ series to show, plot. Otherwise, it might be hundreds or thousands of them. Recency is one good criterion. But even when showing only recently populated timeseries there's too much. So. It makes sense to do simple/obvious type of analyses on individual timeseries and to then display those that have interesting statistical properties, such as:
- large scattering of data
- a trend (increasing, or decreasing)
- ...

All of these require setting thresholds, but I think it's a powerful approach to show, for example, top N recently populated timeseries by relative change (increasing -> perf regression).

This patch works into that direction.

### DIFF
--- a/conbench/app/benchmarks.py
+++ b/conbench/app/benchmarks.py
@@ -2,7 +2,7 @@ import collections
 import logging
 import math
 import time
-from typing import Dict, List, Tuple, TypedDict
+from typing import Dict, List, Tuple, TypedDict, TypeVar
 
 import flask
 import numpy as np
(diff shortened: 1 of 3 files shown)


jgehrcke:
:ship: this is work in progress! Many ideas in terms of cleaner code structure and methods, etc. Integrating early now, mainly to see how this affects the BMRT cache refresh approach.

PR #1282, opened 2023-05-25
c-bench UI: per-benchmark trend: outlier removal before lin reg

jgehrcke:
Context-specific outlier removal before doing linear regression. See code comment. All this is very much prototyping, work in progress. But I think it goes into a valuable direction judging by the data/plots I see. I keep moving, but discussion / critical feedback is extremely welcome!

Btw, here I decided to immediately go with a testsuite-supported data analysis approach. Need that rigorous kind of feedback, towards making more or less well-founded decisions as we go.

Building against prod state of database (T) -- so everything in here is technically biased by the kind of raw data we are obtaining there. But I think that's fair! Much better than pure synthetic data.


conbench/app/benchmarks.py
```diff
@@ -229,6 +264,12 @@ def show_trends_for_benchmark(bname: TBenchmarkName) -> str:
         caseid, ctxid, hwid = t3
         results = bmrt_cache["by_4t_list"][(bname, caseid, ctxid, hwid)]
 
+        # dfts = bmrt_cache["by_4t_df"][(bname, caseid, ctxid, hwid)]
+        # print()
+        # print()
+        # print((bname, caseid, ctxid, hwid))
+        # print(dfts.to_csv(na_rep="NaN", float_format=numstr8))
```
jgehrcke, line 271: this is the method via which I generated the test scenario time series CSV files

conbench/app/benchmarks.py
```diff
@@ -180,17 +194,38 @@ def show_trends_for_benchmark(bname: TBenchmarkName) -> str:
         # datetimeindex back to float values. Remove 10^15 from these
         # nanoseconds-since-epoch to make the numeric values a little less
         # extreme for printing/debugging.
+
+        remove_outliers_by_iqrdist(df, "svs")
```
jgehrcke, line 198: bug here: no copy before mutation -- this directly affects BMRT cache. fixing in next patch.

conbench/app/benchmarks.py
```diff
@@ -180,17 +194,38 @@ def show_trends_for_benchmark(bname: TBenchmarkName) -> str:
...
+            # Skip if after outlier removal there's not enough history left.
+            continue
+
         tfloats = (
             np.array(df.index.to_pydatetime(), dtype=np.datetime64).astype("float")
             / 10**15
         )
         yfloats = df["svs"].values
 
+        # Note that this is a least squares fit.
```
jgehrcke, line 215: also note: ideally this for loop is in the future a single matrix operation. it doesn't quite make sense to iterate over many timeseries individually and to do the same analysis logic one-by-one within a Python loop. That's cries for optimization via proper usage of linear algebra / pandas dataframe operation, will probably speed this up by several orders of magnitude (and this might be needed, this takes already a second for 1000 timeseries).

conbench/outlier.py
```diff
@@ -0,0 +1,101 @@
+import logging
+
+import numpy as np
+import pandas as pd
+
+log = logging.getLogger(__name__)
+
+
+def remove_outliers_by_iqrdist(
```
alistaire47, line 9: It would be nice to make a little library here of interchangeable  outlier detection methods including this and the one used in the plots, though we'd need consistent input and output data structures, and to separate concerns from change detection/segmentation, which probably has to happen first. But we should keep pushing down the path of modularity as we can.

jgehrcke: Yes, thought similar thoughts!

PR #1284, opened 2023-05-26
c-bench UI: trends: fixes, 3x4 tinyplot grid, also show decreasing trends

jgehrcke:
Quite a bit of work. Things begin to fall into place. Would love to see this in staging/prod.


jgehrcke:
Please keep the feedback coming after merge! :ship:

Issue #1288, opened 2023-06-02
Pragmatic DoS protection / QoS solution: sempaphore protection on api/compare


jgehrcke:
A follow-up from https://github.com/voltrondata-labs/arrow-benchmarks-ci/issues/124.

jgehrcke:
done

PR #1289, opened 2023-06-02
api/compare: process one request at a time

jgehrcke:
For #1288, especially against memory explosion.

### DIFF
--- a/conbench/api/compare.py
+++ b/conbench/api/compare.py
@@ -1,6 +1,7 @@
 import collections
 import logging
 import math
+import threading
 from typing import TYPE_CHECKING, Dict, List, Optional, Tuple
 
 import flask as f
@@ -18,6 +19,17 @@ log = logging.getLogger(__name__)
 
 DEFAULT_PAIRWISE_PERCENT_THRESHOLD = 5.0
 DEFAULT_Z_SCORE_THRESHOLD = 5.0
+
+
+# Context: https://github.com/voltrondata-labs/arrow-benchmarks-ci/issues/124
+# The compare endpoint can be rather resource-heavy. This here is a pragmatic
+# QoS solution / DoS protection: only ever have one of the request-handling
+# threads work on an /api/compare/... request.
+# Under a lot of API pressure this helps keep the system going. In that case,
+# however, individual requests may be responded to after a longer waiting time.
+# In those contexts, it makes sense to apply large HTTP request timeout
+# constants.
+_semaphore_compare_get = threading.BoundedSemaphore(1)
 
 
 def _parse_two_ids_or_abort(compare_ids: str) -> Tuple[str, str]:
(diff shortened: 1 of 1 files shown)


jgehrcke:
:ship: trying this in staging. Please review+provide feedback!

conbench/api/compare.py
```diff
@@ -20,6 +21,17 @@
...
 
+# Context: https://github.com/voltrondata-labs/arrow-benchmarks-ci/issues/124
+# The compare endpoint can be rather resource-heavy. This here is a pragmatic
+# QoS solution / DoS protection: only ever have one of the request-handling
+# threads work on an /api/compare/... request.
+# Under a lot of API pressure this helps keep the system going. In that case,
+# however, individual requests may be responded to after a longer waiting time.
+# In those contexts, it makes sense to apply large HTTP request timeout
+# constants.
+_semaphore_compare_get = threading.BoundedSemaphore(1)
```
jonkeane, line 32: What about instead of hardcoding this to 1, it be something like `n-2` where `n` is number of available nodes? Or `n/2`? That way if a deployment is large and handling multiple compares is possible without totally saturating the system, that will just work?

conbench/api/compare.py
```diff
@@ -227,7 +239,7 @@ def _get_a_result(benchmark_result_id: str) -> BenchmarkResult:
         return benchmark_result
 
     @maybe_login_required
-    def get(self, compare_ids: str) -> f.Response:
+    def get(self, compare_ids) -> f.Response:
```
austin3dickey, line 242: nit: add a type annotation here

conbench/api/compare.py
```diff
@@ -289,6 +301,14 @@ def get(self, compare_ids: str) -> f.Response:
         tags:
           - Comparisons
         """
+        # Note(JP): a threading.BoundedSemaphore acquired and released with a
+        # context manager. Assumes the web application to be deployed in a
+        # model with N request-handling threads per process.
+        with _semaphore_compare_get:
+            return self._get(compare_ids)
+
+    @maybe_login_required
```
austin3dickey, line 310: nit: I don't think you need this decorator

austin3dickey (commented):
nice!

conbench/api/compare.py
jgehrcke: > What about instead of hardcoding this to 1, it be something like n-2 where n is number of available nodes?

This serialization applies once per process. The number of processes / containers to run then is a deployment question.

That is, there is no need to relate this semaphore max count to the number of nodes running a Conbench deployment.

conbench/api/compare.py
jgehrcke: thanks!

conbench/api/compare.py
jgehrcke: I tried to not change anything on that front, not sure what happened :)

conbench/api/compare.py
jgehrcke: (acted on review feedback in https://github.com/conbench/conbench/pull/1290)

Issue #1298, opened 2023-06-07
relax this again: 400 bad request upon 'one data point from one iteration: property q1 is unexpected'


jgehrcke:
Seen here: https://github.com/voltrondata-labs/arrow-benchmarks-ci/issues/129

We should totally work on that, it doesn't make sense to send this kind of ambiguous stats object.

But given that this is the legacy CLI's job and our ability/motivation to change anything related to that -- let's just make the receiving end a little more relaxed. The strictness has paid off now I think -- we know that we're doing this, and we've kept record of that.

jgehrcke:
Taken care of via https://github.com/conbench/conbench/pull/1299.

PR #1299, opened 2023-06-07
api: post bmr: ignore bogus stats, do not fail

jgehrcke:
For issue #1298


austin3dickey (commented):
Thank you! I think this solution makes sense for now.

Issue #1300, opened 2023-06-08
webapp: SQLAlchemy session management: isolate requests / executors


jgehrcke:
Memory growth / memory leak across requests:
https://github.com/voltrondata-labs/arrow-benchmarks-ci/issues/124#issuecomment-1580563578
https://github.com/conbench/conbench/issues/1285

Made me think about our SQLAlchemy database session isolation between requests, and generally between executors.

Current code is messy in that regard, intent and behavior isn't quite clear. We use a single object `Session` _everywhere_: in HTTP requests, in BMRT cache update code, in the test suite. That's quite some chaos :-).

jgehrcke:
Done in https://github.com/conbench/conbench/pull/1301.

jgehrcke:
The overall memory growth profile in view of "CI waves" :surfer:  hasn't changed: 
![image](https://github.com/conbench/conbench/assets/265630/6c8caaa2-6e33-4c7d-b5dd-cb9d8dec4235)

That is, we learned that the memory growth had little or nothing to do with having improper resource cleanup around SQLAlchemy session objects.

PR #1301, opened 2023-06-08
SQLAlchemy database session isolation: per-request and per-BMRT cache update 

jgehrcke:
For https://github.com/conbench/conbench/issues/1300.

The out-of-band database interaction from within the test suite was particularly painful here, but I think it also revealed that we had a problem there: re-using the same session/sessionmaker object without clear teardown/isolation points.

### DIFF
--- a/conbench/db.py
+++ b/conbench/db.py
@@ -5,10 +5,9 @@ import sys
 import orjson
 import psycopg2
 import sqlalchemy.exc
+import sqlalchemy.orm
 import tenacity
 from sqlalchemy import create_engine
-import sqlalchemy.orm
-
 
 from .config import Config
 
--- a/conbench/db.py
+++ b/conbench/db.py
@@ -7,13 +7,18 @@ import psycopg2
 import sqlalchemy.exc
 import tenacity
 from sqlalchemy import create_engine
-from sqlalchemy.orm import scoped_session, sessionmaker
+import sqlalchemy.orm
+
 
 from .config import Config
 
 engine = None
-session_maker = sessionmaker(future=True)
-Session = scoped_session(session_maker)
+session_maker = sqlalchemy.orm.sessionmaker()
+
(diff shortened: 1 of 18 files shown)


jgehrcke:
CI ui-screenshots affected by https://github.com/conbench/conbench/issues/1302

jgehrcke:
Before calling it a day I'd like to get signal from staging, and potentially revert if that breaks things :ship: -- please keep feedback coming.

conbench/db.py
```diff
@@ -5,15 +5,19 @@
...
 import tenacity
 from sqlalchemy import create_engine
-from sqlalchemy.orm import scoped_session, sessionmaker
 
 from .config import Config
 
 engine = None
-session_maker = sessionmaker(future=True)
-Session = scoped_session(session_maker)
+session_maker = sqlalchemy.orm.sessionmaker()
```
austin3dickey, line 15: What was your reasoning for dropping `future=True`? I've found that to be slightly helpful in the past, though admittedly I forget why

austin3dickey (commented):
Excited to see how this works!

conbench/db.py
austin3dickey: Oh, here are [the docs](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine.params.future).

>Use the 2.0 style [Engine](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Engine) and [Connection](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Connection) API. As of SQLAlchemy 2.0, this parameter is present for backwards compatibility only and must remain at its default value of True.

Haha. So that's why it was useful. For when we migrated to 2.0.

jgehrcke: Yes! I wondered about why we have that, read the docs and thought "oh cool then we can remove this now" :)

PR #1305, opened 2023-06-09
bmrt cache: increase size, spreak out cpu work a bit

jgehrcke:
This is interesting for Arrow Conbench because there are so many different types of benchmarks. Increasing the BMRT cache size will make the per-timeseries analysis slightly more interesting.


jgehrcke:
:ship:

Issue #1307, opened 2023-06-13
c-benchmarks: trends: build from 'default branch' results only


jgehrcke:
This should maybe be a boolean on each benchmark result: `default_branch`. Right now it's unnecessarily complicated to filter benchmark results obtained for the default branch, unless I am missing something.

austin3dickey:
The way we filter benchmark results, runs, and commits to the default branch today (in at least six places) is checking whether `commit.sha == commit.fork_point_sha`. This is definitely confusing and maybe [causes slow performance](https://github.com/conbench/conbench/issues/745#issuecomment-1442363191) in some cases. As I proposed in that comment, it's probably better to have a boolean property on the commit object.

Additionally, you're suggesting denormalizing this property of a commit onto the benchmark result, which of course heavily relates to https://github.com/conbench/conbench/issues/1227. It's an interesting idea!

jgehrcke:
Thank you for the feedback, Austin! You were fast haha, I wanted to ping you and ask specifically about your opinion.

>  it's probably better to have a boolean property on the commit object.

I made a tiny (real tiny) bit of progress here: https://github.com/conbench/conbench/pull/1308/files#r1228321980

jgehrcke:
this is done :)

Issue #1226, opened 2023-05-05
UI: "repo-less" runs show up prominently


jgehrcke:
We know that we have to tackle https://github.com/conbench/conbench/issues/1165 properly. I want to mention one specific motivator, fallout: the repo-less runs right now display rather prominently before regular runs on the landing page, and that's not so nice:

![Screenshot from 2023-05-05 15-56-35](https://user-images.githubusercontent.com/265630/236478623-de9ff455-e53e-46da-bfb0-c98c6f3e3f4d.png)

Related: https://github.com/conbench/conbench/issues/885

A bit of an ambiguous question, but I have labeled this as "bug" because I think we agree that this must not look like that; it's kind of landing page UX regression.

jgehrcke:
Raises questions, ideas. Pre-merge stuff centers around pull requests (merge requests in gitlab). Metadata that might be interesting: author of pull request. pull request number. URL to pull request.

jgehrcke:
When clicking on such a run, the details look rather boring / well, even wrong/fishy:
![image](https://user-images.githubusercontent.com/265630/236480685-488fc12f-3968-4057-9453-28fe4523d426.png)

austin3dickey:
All runs are associated with a repo now.

Issue #904, opened 2023-03-20
commit metadata: if acquisition failed once, it is never retried later (DB scan / consistency)


jgehrcke:
Fundamentally when commit metadata has to be obtained from an external system then there is the likelihood for failure. Even when this is minimized with smart decision-making, it's still non-zero. An helpful concept towards best-effort metadata population will have to scan the database every now and then, try to re-obtain. 

Example: when this happens:
```
conbench-app-1  | [230320-12:31:34.789] [8] [conbench.entities.run] INFO: treat as unknown commit: error during get_github_commit(): Hourly GitHub HTTP API quota exhausted
conbench-app-1  | [230320-12:31:34.806] [8] [conbench.entities.run] INFO: commit_fetch_info_and_create_in_db_if_not_exists(5efe7f943ca2b7d8a53c1d8eb91e31da4fe2e4b5, https://github.com/conbench/conbench, None, None) took 0.115 s
```
the commit is 'forever' stored as "unknown commit" in the database. That means it never gets its metadata populated (author, message, but also  what we call today `branch`, `fork_point_sha`, `parent`, `timestamp`, ...:
https://github.com/conbench/conbench/blob/147be4cdcba3c03d5451dcdff55951c6485e18a3/conbench/entities/commit.py#L38

This is retryable, and we could do that.

jgehrcke:
web application log example for when commit info cannot be fetched because commit is unknown:
```
[conbench.entities.run] INFO: treat as unknown context: error during get_github_commit_metadata():
  Unexpected GitHub HTTP API response for URL https://api.github.com/repos/<redacted>/commits/71000bcf4d3f7704b470c0ff381d9449a0b14e66:
  <Response [422]>. Leading bytes of body: '{"message":"No commit found for SHA: 71000bcf4d3f7704b470c0ff381d9449a0b14e66","documentation_url":"https://docs.github.com/rest/commits/commits#get-a ...'
```
this is a permanent error (not recoverable/retryable) that should I think be exposed nicely in the UI / benchmark result view.

snmvaughan:
@jgehrcke I ran into the scenario where commit information couldn't be acquired when the benchmark was submitted (the github token had expired).  It would be great if it could be easily retried.

jgehrcke:
@snmvaughan Thank you for the feedback. It's great to know that you care (I mean that). Can you tell us anything about your Conbench use case? That would be cool! :)

>  It would be great if it could be easily retried.

[...]

PR #1308, opened 2023-06-13
BMRT cache: cpu spread, default branch; c-bench trends: visual tweaks

jgehrcke:
See commit messages.

Among others, this addresses https://github.com/conbench/conbench/issues/1307.

### DIFF
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -4,6 +4,7 @@ import logging
 import os
 import threading
 import time
+from functools import cached_property
 from datetime import datetime, timedelta
 from typing import Dict, List, Optional, TypedDict
 
@@ -62,17 +63,23 @@ class Commit(Base, EntityMixin):
     # that parent/child relationship better?
     parent: Mapped[Optional[str]] = Nullable(s.String(50))
 
-    # This is meant to be the URL to the repository without trailing slash.
-    # Should be renamed to repo_url. Seemingly this can be an empty string,
-    # too. We should transition to a state where this is known to be a URL.
+    # Note(JP): This is meant to be the URL to the repository without trailing
+    # slash. Should be renamed to repo_url. Seemingly this can be an empty
+    # string, too. We should transition to a state where this is known to be a
+    # URL.
     repository: Mapped[str] = NotNull(s.String(300))
 
+    # truncation intended?
     message: Mapped[str] = NotNull(s.String(250))
+
     author_name: Mapped[str] = NotNull(s.String(100))
+
+    # That's probably github-oriented
     author_login: Mapped[Optional[str]] = Nullable(s.String(50))
 
     # I think this is guaranteed to be a URL. Is it?
     author_avatar: Mapped[Optional[str]] = Nullable(s.String(100))
+
     # Note(JP): tz-naive datetime, git commit author date, in UTC.
     # Edit: adding the type Optional[datetime] is not sufficient because
     # further down we use `.label()` which seems to be sqlalchemy-specific
@@ -91,6 +98,25 @@ class Commit(Base, EntityMixin):
             return self
         else:
             return Commit.first(sha=self.fork_point_sha, repository=self.repository)
+
+    @cached_property
+    def on_default_branch(self) -> bool:
+        """
+        Note(JP): weird: we have user-given data about whether a benchmark
+        result has been obtained for a commit on the default branch or not (In
+        the commit info object that a user can provide, we specify that "Set
+        this to `null` or leave this out to indicate that this benchmark result
+        has been obtained for the default branch."). But here we don't make
+        this easily accessible.
+
+        And then we have (potentially?) GitHub's view on whether a commit is on
+        the default branch or not. But we also don't make this easily
+        accessible.
+
+        This method is an attempt to make "this" easily accessible. But: what
+        if both are in conflict? Can they be?
+        """
+        return self.sha == self.fork_point_sha
 
     @property
     def repo_url(self) -> Optional[str]:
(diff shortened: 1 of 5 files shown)


jgehrcke:
:ship:

conbench/entities/commit.py
```diff
@@ -92,6 +99,25 @@ def get_fork_point_commit(self) -> Optional["Commit"]:
...
+        this easily accessible.
+
+        And then we have (potentially?) GitHub's view on whether a commit is on
+        the default branch or not. But we also don't make this easily
+        accessible.
+
+        This method is an attempt to make "this" easily accessible. But: what
+        if both are in conflict? Can they be?
+        """
+        return self.sha == self.fork_point_sha
```
jgehrcke, line 119: @austin3dickey 
thanks so much for adding https://github.com/conbench/conbench/issues/1307#issuecomment-1589409422 -- this here was my quick solution paired with the snippet

```python
        # Skip results that have not been obtained for the default code branch.
        bmrcommit = bmrrun.commit
        if bmrcommit is None:
            continue

        if not bmrcommit.on_default_branch:
            continue
```

(further below in this patch)

conbench/entities/commit.py
```diff
@@ -92,6 +99,25 @@ def get_fork_point_commit(self) -> Optional["Commit"]:
...
+        this to `null` or leave this out to indicate that this benchmark result
+        has been obtained for the default branch."). But here we don't make
+        this easily accessible.
+
+        And then we have (potentially?) GitHub's view on whether a commit is on
+        the default branch or not. But we also don't make this easily
+        accessible.
+
+        This method is an attempt to make "this" easily accessible. But: what
+        if both are in conflict? Can they be?
```
austin3dickey, line 117: A reply to this docstring 🙂 

`sha` is always user-given, and `fork_point_sha` is never user-given. The latter is always generated from the GitHub HTTP API response when we `get_github_commit_metadata()`. So this new property is guaranteed to represent whether the commit is truly on the default branch or not, according to GitHub.

The user-given `branch` and `pr_number` fields are a little trickier. The `pr_number` field (the documentation of which contains the line you quoted above) is ONLY used to autofill the `branch` field if it's not given. Then the `pr_number` is discarded, not stored on the Commit object. So what do we use `branch` for, knowing that it's user-given?

It was initially supposed to be the main way to "check if the commit is on the default branch", but there were two problems: (1) getting the actual default branch name for a given repo from the GitHub HTTP API is slightly expensive/unsafe and we don't cache/store that anywhere, and (2) of course the branch is user-given so we can't fully trust that there's not an error. So I landed on "sha == fork_point_sha" as a better way to make this check.

Where else do we use `branch`? Probably the most important is in the [commit ancestry query](https://github.com/conbench/conbench/blob/03a9c846fc734e2b9bba3b143aaaabf34fd0c96e/conbench/entities/commit.py#L263), when a commit is on a PR branch and we want to find other commits on the same PR branch. Other than that it's basically only displayed in the UI.

So to answer your question: what if they are in conflict? After reviewing the code, I don't think anything bad would happen except that the UI might display confusing branch names.

jgehrcke: Thank you!!

jgehrcke: > is always generated from the GitHub HTTP API response when we get_github_commit_metadata()

caveat for "always": most of the times only -- as of https://github.com/conbench/conbench/issues/904.

In that case we could fall back to trusting user input, but instead it's probably better to improve on #904.

jgehrcke: > Then the pr_number is discarded, not stored on the Commit object

Connecting the dots here, this might be a property we may want to store. From https://github.com/conbench/conbench/issues/1226#issuecomment-1536314705:

> Pre-merge stuff centers around pull requests (merge requests in gitlab). Metadata that might be interesting: author of pull request. pull request number. URL to pull request.

jgehrcke: > So what do we use branch for, knowing that it's user-given?

Thanks for elaborating on this. This is helpful, and will be helpful again in the future.

> So I landed on "sha == fork_point_sha" as a better way to make this check.

I certainly like that today we derive this "on_default_branch" property from
- the commit hash (user-given, but 100 % _trusted_ (somewhere we need to start in terms of trust))
- repository metadata (not user-given, indirectly user-given via repo spec/URL)

> Where else do we use branch? Probably the most important is in the [commit ancestry query](https://github.com/conbench/conbench/blob/03a9c846fc734e2b9bba3b143aaaabf34fd0c96e/conbench/entities/commit.py#L263), when a commit is on a PR branch and we want to find other commits on the same PR branch.

That's interesting! Thanks for clarification.

A few points, thinking out loud:
- A commit can be on more than one branch (correct?). That is the reason for why only the commit hash as user-given input is not sufficient to do branch-local analysis. That is, the user _must_ provide pr number or branch, fundamentally. Correct?
- When the user provides an erroneous branch name or PR number: nothing we can do; that's fully OK (responsibility boundary thinking).

I think we could say "we want to trust user-given branch information". In fact, we _must_ trust user-given branch information. We can/should tell users already today that if they do not provide correct pr_number/branch information then quite a bit of analysis is broken/missed out on. We _could_ conclude from here that we are in fact already going all in on this piece of information.

Then we could do the "on default branch?" check based on the branch name, and this might be a little easier to maintain / reason about.

More brainstorming, not really important, and maybe stupid:

- Make individual benchmark result have `repo_url` and `branch` properties.
- Store repository entity in the DB. A repository entity has a default branch name. It's mutable, but changes rarely. What if it changes? We may try to do a dynamic lookup via GitHub HTTP API upon repo initialization (when it's not yet known), and then every now and then (e.g., upon container start).
- Cache in memory: repo_url->default_branch (DB lookup only needed during app init).

austin3dickey: These are all really good thoughts, thank you! I remember thinking about some of these things before but didn't write them down. 

>In fact, we _must_ trust user-given branch information.

I really like this statement. In practice, I remember it being difficult to get it to work from the start. When setting up various CI code that reports to Conbench, there were a handful of times that the environment variables weren't set up correctly or something else went wrong, and I ended up "polluting" the Conbench databases with commits whose branch disagreed with `sha == fork_point_sha`. But the solution to this problem is to provide users with delightful tools (and DB schemas/validation) that result in a low probability of failure.

>A commit can be on more than one branch (correct?)

Yes, and I think that's why Conbench lives with the simplified view of git where a benchmark result is either 
- not associated with a repo,
- on the default branch,
- or is a PR commit. 

As far as I know, it's undefined what happens when a commit doesn't correspond with exactly one of those. And making this simplification assumption allows us to assume there's just one associated branch.

I like your brainstorming ideas!

Issue #682, opened 2023-02-06
Why does the comparison use "forked from default at" instead of "commit (SHA)" for labelling commit SHAs?


jonkeane:
One example: https://conbench.ursa.dev/compare/benchmarks/a3faf3e6c7f048dab960f24f58618aa0...77aef10139db4f0a8376e1baa93f06b0/

<img width="1170" alt="image" src="https://user-images.githubusercontent.com/700357/216998811-93685316-0cb3-4173-b4cd-3c109796b7c0.png">

I suspect this was a mistake during the history renovation project — these commit SHAs appear to be the legit commit SHAs of the benchmark run itself (which is what we want here). So maybe this is just a label change?

jgehrcke:
@jonkeane I am trying to understand the issue here. 

While doing so, I note some things that I find ... noteworthy :)

- The `commit` field shows a commit message prefix. It's also clearly a URL (leading to said commit details, I tried this out). But the actual _commit hash_ associated with the run is not shown explicitly in that table field. I think it should be shown.

- The `branch` field shows `apache:master`. That is an unusual notation to me. Should this be either `apache/arrow:master` or just `master`? If `apache:master` is expected I'd love to understand what this means.

- So, both commits seem to be on the default branch (`master`). In that sense, the "forked from default at" is surprising, because neither of both seems to be on a patch/PR branch. Is that what you mean @jonkeane? 

Update: I just confirmed that the linked commits (in the `commit` field) show commit has 32c... (left) and 007.. (right). What the "forked from ..." fields show. Is it true that this field "forked from ..." should simply _not_ appear in this table, or what would the expected outcome look like? Keep the table field but show "not applicable, because commit on master branch"? :)

jonkeane:
> So, both commits seem to be on the default branch (master). In that sense, the "forked from default at" is surprising, because neither of both seems to be on a patch/PR branch. Is that what you mean @jonkeane?

Yeah, very surprising that there's any mention of forking since these are both commits on the default branch.

> Update: I just confirmed that the linked commits (in the commit field) show commit has 32c... (left) and 007.. (right). What the "forked from ..." fields show. Is it true that this field "forked from ..." should simply not appear in this table, or what would the expected outcome look like? Keep the table field but show "not applicable, because commit on master branch"? :)

[...]

PR #1310, opened 2023-06-13
webapp: fix small bugs

### LINKED ISSUE #1230: ui: certain numbers are missing across the webapp

austin3dickey:
Fixes #1230. Helps #682.

Bugs (4) and (5) in #1230 were already fixed by #1253. This PR fixes the rest of the bugs; here's an example:
image

### DIFF
--- a/conbench/app/compare.py
+++ b/conbench/app/compare.py
@@ -9,7 +9,7 @@ import flask as f
 from ..app import rule
 from ..app._endpoint import AppEndpoint, authorize_or_terminate
 from ..app._plots import TimeSeriesPlotMixin, simple_bar_plot
-from ..app._util import augment, error_page
+from ..app._util import error_page
 from ..app.results import BenchmarkResultMixin, RunMixin
 from ..app.types import HighlightInHistPlot
 from ..config import Config
@@ -73,35 +73,35 @@ class Compare(AppEndpoint, BenchmarkResultMixin, RunMixin, TimeSeriesPlotMixin):
             )
             # log.info("compare view: generated history plot", plot_history)
 
-        if comparisons and self.type != "benchmark-result":
-            comparisons_by_id = {
-                c["contender"]["benchmark_result_id"]: c
-                for c in comparisons
-                if c["contender"]
-            }
-            if self.type == "run":
-                benchmarks, response = self._get_benchmarks(run_id=contender_id)
-            if response.status_code != 200:
-                self.flash("Error getting benchmark results.")
-                return self.redirect("app.index")
-            for benchmark in benchmarks:
-                augment(benchmark)
-            (
-                biggest_changes,
-                biggest_changes_ids,
-                biggest_changes_names,
-            ) = self.get_biggest_changes(benchmarks)
-            outlier_urls = [
-                comparisons_by_id.get(x, {}).get("compare_benchmarks_url", "")
-                for x in biggest_changes_ids
-            ]
-            # Note(JP): I think this has been broken for a while. Just
-            # rediscovering this -- do we really want to
-            # show multiple history plots when we compare two runs?
-            plot_history = [
-                self.get_history_plot(b, contender_run, i)
-                for i, b in enumerate(biggest_changes)
-            ]
+        # if comparisons and self.type != "benchmark-result":
+        #     comparisons_by_id = {
+        #         c["contender"]["benchmark_result_id"]: c
+        #         for c in comparisons
+        #         if c["contender"]
+        #     }
+        #     if self.type == "run":
+        #         benchmarks, response = self._get_benchmarks(run_id=contender_id)
+        #     if response.status_code != 200:
+        #         self.flash("Error getting benchmark results.")
+        #         return self.redirect("app.index")
+        #     for benchmark in benchmarks:
+        #         augment(benchmark)
+        #     (
+        #         biggest_changes,
+        #         biggest_changes_ids,
+        #         biggest_changes_names,
+        #     ) = self.get_biggest_changes(benchmarks)
+        #     outlier_urls = [
+        #         comparisons_by_id.get(x, {}).get("compare_benchmarks_url", "")
+        #         for x in biggest_changes_ids
+        #     ]
+        #     # Note(JP): I think this has been broken for a while. Just
+        #     # rediscovering this -- do we really want to
+        #     # show multiple history plots when we compare two runs?
+        #     plot_history = [
+        #         self.get_history_plot(b, contender_run, i)
+        #         for i, b in enumerate(biggest_changes)
+        #     ]
 
         return self.render_template(
             self.html,
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -330,10 +330,6 @@ class BenchmarkResult(Base, EntityMixin):
                 "q1": to_float(benchmark_result.q1),
                 "q3": to_float(benchmark_result.q3),
                 "iqr": to_float(benchmark_result.iqr),
-                # TODO: remove these.
-                "z_score": None,
-                "z_regression": False,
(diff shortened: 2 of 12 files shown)


conbench/templates/benchmark-result.html
```diff
@@ -184,12 +184,6 @@ <h3>History plot for benchmark '{{ benchmark.display_bmname }}'</h3>
               <div align="right" style="display:inline-block; float: right;">{{ run.commit.branch }}</div>
             </li>
           {% endif %}
-          {% if run.commit.fork_point_sha %}
-            <li class="list-group-item" style="overflow-y: auto;">
-              <b>forked from default at</b>
-              <div align="right" style="display:inline-block; float: right;">{{ run.commit.fork_point_sha }}</div>
-            </li>
-          {% endif %}
```
austin3dickey, line 192: This helps with #682 by simply removing these rows. We could come back to displaying fork point info if we're interested later

conbench/templates/compare-list.html
```diff
@@ -125,80 +125,66 @@ <h6>Top Outliers</h6>
                   <div>{{ c.baseline.benchmark_name }}</div>
                 </td>
                 <td>
-                  {% if (c.contender.value is not none or c.contender_error) and (c.baseline.value is not none or c.baseline_error) %}
-                    <a href="{{ c.compare_benchmarks_url }}">
```
austin3dickey, line 129: The diff is much easier to parse if you ignore whitespace

conbench/templates/compare-list.html
```diff
@@ -125,80 +125,66 @@ <h6>Top Outliers</h6>
...
-                           data-toggle="tooltip"
-                           data-placement="top"
-                           title="Has error">
-                        </i></a>
-                      </td>
-                    {% else %}
-                      <td style="white-space: nowrap;">{{ c.contender.value }} {{ c.unit }}</td>
-                    {% endif %}
-                  </tr>
+                    <a href="{{ url_for('app.benchmark-result', benchmark_result_id=c.baseline.benchmark_result_id) }}">Error</a>
```
austin3dickey, line 134: The glyph doesn't seem to be working; not sure if it's related to recent dependency upgrades. I chose to just display the word "Error" instead.

austin3dickey:
Will fix tests tomorrow.

conbench/app/compare.py
```diff
@@ -73,35 +73,35 @@ def page(self, comparisons, regressions, improvements, baseline_id, contender_id
...
+        #         comparisons_by_id.get(x, {}).get("compare_benchmarks_url", "")
+        #         for x in biggest_changes_ids
+        #     ]
+        #     # Note(JP): I think this has been broken for a while. Just
+        #     # rediscovering this -- do we really want to
+        #     # show multiple history plots when we compare two runs?
+        #     plot_history = [
+        #         self.get_history_plot(b, contender_run, i)
+        #         for i, b in enumerate(biggest_changes)
+        #     ]
```
austin3dickey, line 104: Commenting this out for now. Made an issue to revisit: #1311

alistaire47 (approved):
Jinja templates aren't the easiest thing to review, but as far as I can tell everything looks sensible and I didn't notice anything out of place. Might be worth a review of staging during conbenchat after merging to see if anything looks amiss or still need tweaking.

conbench/entities/benchmark_result.py
```diff
@@ -330,10 +330,6 @@ def to_dict_for_json_api(benchmark_result):
                 "q1": to_float(benchmark_result.q1),
                 "q3": to_float(benchmark_result.q3),
                 "iqr": to_float(benchmark_result.iqr),
-                # TODO: remove these.
```
jonkeane, line 333: ✅

conbench/templates/benchmark-result.html
jonkeane: I suspect we will want to reintroduce these sooner rather than later — it can be very helpful to understand why the baseline is what it is in this view, but it shouldn't stop merging this PR now if it's too much to keep i tin

conbench/templates/compare-list.html
jonkeane: 💯

jonkeane (approved):
Agreed with Ed: would be good to review on staging since it's mostly templating stuff, but this looks good to me

conbench/templates/compare-list.html
jgehrcke: Yes, and yes. :)

It was a questionable decision to start building this UI with Bootstrap 3. Glyphicons were introduced with Bootstrap 3 in 2013. Since 2015 we have known that glyphicon support is going to be dropped with Bootstrap 4 (first release: 2015).

https://blog.getbootstrap.com/2015/08/19/bootstrap-4-alpha/
https://getbootstrap.com/docs/4.0/migration/#components

I gradually started replacing glyphicons with [Bootstrap Icons](https://icons.getbootstrap.com/).

Pragmatically switching to a simple word here was the right approach.

conbench/tests/app/test_compare.py
```diff
@@ -86,6 +86,13 @@ def _create(self, client):
         run_id = _fixtures.VALID_RESULT_PAYLOAD["run_id"]
         return f"{run_id}...{run_id}"
 
+    def test_data_shows_up(self, client):
+        ids = self._create(client)
+        self.authenticate(client)
+        response = client.get(self.url.format(ids))
+        self.assert_page(response, "")
+        assert "0.03637 s" in response.text
```
jgehrcke, line 94: perfect, thanks!

conbench/tests/app/test_compare.py
```diff
@@ -86,6 +86,13 @@ def _create(self, client):
         run_id = _fixtures.VALID_RESULT_PAYLOAD["run_id"]
         return f"{run_id}...{run_id}"
 
+    def test_data_shows_up(self, client):
+        ids = self._create(client)
+        self.authenticate(client)
+        response = client.get(self.url.format(ids))
+        self.assert_page(response, "")
```
jgehrcke, line 93: I like when test code is as self-explanatory as possible, and this line here is cryptic. Does it after all only confirm that the response status code is 200? :) we don't need to change this now.

jgehrcke (approved):
Thank you, Austin!
:rocket:

conbench/templates/compare-list.html
austin3dickey: Thanks for the explanation! :)

conbench/tests/app/test_compare.py
austin3dickey: Haha, absolutely. When called in a later test in this file, this method has the comment `# Ensure the run page returns a 200 and HTML, not a traceback`, so I didn't copy that up here. Probably should have.

PR #1317, opened 2023-06-15
Increase BMRT cache size, reduce log verbosity, tune floatcomp_with_leeway()

### LINKED ISSUE #1316: address log msgs: pandas warning, agg warning, misc
bug:
```
/app/conbench/app/benchmarks.py:217: DeprecationWarning: parsing timezone aware datetimes is deprecated; this will raise an error in the future
```

bug:
```
[230615-10:40:23.088] [6] [conbench.entities.benchmark_result] WARNING: key stdev, user-given val 0.001408 vs. calculated 0.0014078253916354126
...
[230615-10:38:57.985] [7] [conbench.entities.benchmark_result] WARNING: key iqr, user-given val 0.072374 vs. calculated 0.07237450000000001
```

verbosity:
```
 [conbench.entities.benchmark_result] WARNING: drop tag key/value pair: `cpu_count`, `None`
```

jgehrcke:
This resolves https://github.com/conbench/conbench/issues/1316 and adds misc changes, mainly comments.

### DIFF
--- a/conbench/job.py
+++ b/conbench/job.py
@@ -155,7 +155,7 @@ bmrt_cache: CacheDict = {
 SHUTDOWN = False
 _STARTED = False
 
-BMRT_CACHE_SIZE = 0.5 * 10**6
+BMRT_CACHE_SIZE = 0.8 * 10**6
 if Config.TESTING:
     # quicker update in testing
     BMRT_CACHE_SIZE = 0.05 * 10**6
--- a/requirements-webapp.txt
+++ b/requirements-webapp.txt
@@ -19,7 +19,7 @@ numpy
 oauthlib
 openapi-spec-validator
 orjson
-pandas
+pandas>=2
 prometheus-client
 prometheus-flask-exporter
 psutil
--- a/conbench/app/benchmarks.py
+++ b/conbench/app/benchmarks.py
@@ -164,6 +164,12 @@ def show_trends_for_benchmark(bname: TBenchmarkName) -> str:
 
     reftime = t_newest
     relchange_by_t3: Dict[Tuple[str, str, str], float] = {}
+
+    # Note that below we operate on one DF per each of T timeseries. We call
+    # remove_outliers_by_iqrdist() and Polynomial.fit() T times, and create
+    # many copies of smallish dataframes. It is probably a lot faster by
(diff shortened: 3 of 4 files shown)


requirements-webapp.txt
```diff
@@ -19,7 +19,7 @@ numpy
 oauthlib
 openapi-spec-validator
 orjson
-pandas
+pandas>=2
```
jgehrcke, line 22: For the future I recommend to have all dependencies precisely pinned, but this change has nothing to do with that idea.

This is not an important change. Just a self-documenting aspect. Background: saw 
```
/app/conbench/app/benchmarks.py:217: DeprecationWarning: parsing timezone aware datetimes is deprecated; this will raise an error in the future
```
in Arrow Conbench, could not reproduce locally. I was wondering which pandas version we use/specify. We did not specify any, i.e. in the last few months at some point we implicitly did the pandas 1.x -> 2.x migration. I just wanted to make this a tiny little more explicit, and with this change also 'forced' a rebuild of otherwise cached container image layers on my machine.

conbench/job.py
```diff
@@ -155,7 +155,7 @@ class ResultTimeseries:
 SHUTDOWN = False
 _STARTED = False
 
-BMRT_CACHE_SIZE = 0.5 * 10**6
+BMRT_CACHE_SIZE = 0.8 * 10**6
```
jgehrcke, line 158: see commit msg.

Adding to that: Arrow Conbench has such a high ingest rate of results that the timespan covered here per timeseries is only a small number of days. That's a little boring. A much bigger cache size would help. But that eats a bit of memory (not too much, but we should take this step by step).

jgehrcke:
:ship:

jgehrcke:
Feedback appreciated!

conbench/job.py
austin3dickey: Should we make this a configurable setting by server?

jgehrcke: > Should we make this a configurable setting by server?

Considering today we do not have the need for a per-deployment setting.

It may become be a candidate for user-facing config at some point in the future. 

There are many ways/ideas though to make "do the right thing" / dynamically adjust to any environment, so changes are this never needs to be user-tunable.

Let's see :)

PR #1326, opened 2023-06-19
value/unit/fmt cleanup (fix #1325),  misc

### LINKED ISSUE #1325: benchmark result compare view: baseline result may appear in wrong unit
Seen here: https://conbench.ursa.dev/compare/benchmark-results/0648a88086ae76968000b407a5e934cf...0648b8b03b1d73878000d9dc573b86c5/

Thanks @pitrou for reporting.

image

The value marked with the black cross has not been converted to MiB/s but still shows B/s, I think.

jgehrcke:
While addressing #1325, I was bitten for the Nth time by parts of plotting and unit/formatting code still being too cryptic and brittle.

Adding clarity is so important to make this code section a little easier to maintain.

Input for history plot is now known to be a collection of HistorySample objects. That means that we can remove complexity.

For #1325, this patch removes complexity as of dynamic unit conversion: this technically removes a UI feature (showing e.g. MiB/s instead of B/s). But for now, correctness is better than fanciness. Bokeh for example properly switches to scientific notation for axis tick labels when numbers get big or small.


austin3dickey:
Thank you!

PR #1328, opened 2023-06-20
benchalerts: allow check/comment customization

### LINKED ISSUE #475: benchalerts: Allow wildcard suppression of alerts

### LINKED ISSUE #756: benchalerts: Use different statuses for different levels of regression
Currently we fail the GitHub Check only if there are any regressions above a given z-score threshold. Add a way for users to configure their own logic on what statuses to use when we get multiple different z-scores.

### LINKED ISSUE #1032: benchalerts: allow for custom text in comments and reports
A specific setup may have specific instructions if there are errors, etc. Allow for this.

austin3dickey:
Fixes #475. Fixes #756. Fixes #1032.

This PR allows an "advanced" user to customize `benchalerts` PR comment text, and GitHub Check text and statuses. They do this by subclassing the default formatter class and overriding its methods. Some use cases for this (definitely not exhaustive)
- add internal project links in the text, or specific instructions on what to do next
- certain conceptual benchmarks may be deemed "too noisy" to create a "failure" status, so add custom status logic to handle those
- etc (see the linked issues)

This UX isn't great but it is very flexible. I think that's sufficient for the level of user that wants to implement this.

### DIFF
--- a/benchalerts/benchalerts/message_formatting.py
+++ b/benchalerts/benchalerts/message_formatting.py
@@ -4,11 +4,7 @@ import textwrap
 from typing import List, Optional
 
 from .conbench_dataclasses import BenchmarkResultInfo, FullComparisonInfo
-
-
-def _clean(text: str) -> str:
-    """Clean text so it displays nicely as GitHub Markdown."""
-    return textwrap.fill(textwrap.dedent(text), 10000).replace("  ", "\n\n").strip()
+from .integrations.github import CheckStatus
 
 
 class _Pluralizer:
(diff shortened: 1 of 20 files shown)


benchalerts/benchalerts/message_formatting.py
```diff
@@ -4,11 +4,7 @@
 from typing import List, Optional
 
 from .conbench_dataclasses import BenchmarkResultInfo, FullComparisonInfo
-
-
-def _clean(text: str) -> str:
-    """Clean text so it displays nicely as GitHub Markdown."""
-    return textwrap.fill(textwrap.dedent(text), 10000).replace("  ", "\n\n").strip()
+from .integrations.github import CheckStatus
```
austin3dickey, line 7: This diff is probably better if you ignore whitespace. I didn't change any code, just moved functions into methods.

benchalerts/tests/unit_tests/test_pipeline_steps/test_github.py
```diff
@@ -19,6 +21,20 @@
 )
 
 
+class MockAlerter(Alerter):
+    def github_check_summary(
+        self, full_comparison: FullComparisonInfo, build_url: Optional[str]
+    ) -> str:
+        old_message = super().github_check_summary(full_comparison, build_url)
```
alistaire47: "old" is perhaps not the best adjective here; "original" or "base" is more apt

alistaire47 (approved):
Looks good; the interface is pretty straightforward, although actually overriding any of the methods usefully would be a bit challenging. 

I wonder if templating for creating messages instead of all the `+=` would make it easier? Though all the conditionals would make the templates complicated in their own right, so maybe not.

benchalerts/tests/unit_tests/test_pipeline_steps/test_github.py
austin3dickey: haha true. I'll fix.

>I wonder if templating for creating messages instead of all the += would make it easier? 

Yeah it would be easier, but I'm afraid it would be less flexible. Maybe after a few implementations we can reevaluate if it's possible? I suspect most implementations will just want to overwrite the intro sentence anyway, which one can do by overriding the `intro_sentence()` method.

Issue #1319, opened 2023-06-15
compare runs page: change default sort order?


austin3dickey:
In my opinion, it isn't useful to see null z-score rows first. I think we should by default sort by z-score ascending with nulls last. That way a user sees possible performance regressions first, and can reverse the sort order to see the (potentially less-important) null z-score rows and possible performance improvements.

<img width="1221" alt="Screenshot 2023-06-15 at 08 41 27 copy" src="https://github.com/conbench/conbench/assets/16600275/fc8d666e-1359-49fd-a022-3578d13dbd5f">

alistaire47:
> I think we should by default sort by z-score ascending with nulls last.

Sorting on absolute value would show all the interesting cases at once, but maybe that's too confusing.

I think this may have been this way because usually those were errored cases, but now that we compare everything, they're much more likely to be cases with no baseline. There's arguably still use to putting errored cases first, but we should really do that explicitly instead of implicitly.

jgehrcke:
This screenshot reminds me of https://github.com/conbench/conbench/issues/1063 -- a lot to catch up here, various unfinished ideas. Because it's generally too hard to change this view without breaking everything. This view is still fed by the JSON API. I might want to try to tackle this these days.

> In my opinion, it isn't useful to see null z-score rows first.

Yes :)

Let's ask ourselves: what do we really lose if we remove the z-score column here for now?

For a user it will be easy to understand the percent change between the two things that are compared here, especially with decent column headings / tooltips. The z-score column, if we want to keep it, needs explanation _in the UI_. Making sense of both, percent change as well as z-score (and _really_ making sense of it) is an advanced use case.

We should think harder in terms of actual UX workflows.

I think that for example a big percent change will / should trigger a click that leads to a history plot showing both results in context of the wider history. Once one is in that view it shows/indicates/reveals so much more than a number in the z-score column in the previous view.

So, I vote for removing the z-score column from this view. We can totally re-introduce it when having more clarity.

alistaire47:
I disagree that we should remove z-scores; percent change is a noisy metric that will not necessarily show you the most important cases; z-scores are more robust because they're based on more history. There's a reason we're not alerting based on percent change; those numbers get chaotic. 

Or to take a step back, maybe we really need a way of highlighting "here are the 6 cases you should look at" rather than just letting users figure it out.

jgehrcke:
> There's a reason we're not alerting based on percent change; those numbers get chaotic.

of course :)

[...]

PR #1330, opened 2023-06-23
UI: compare runs view: cleanup

jgehrcke:
This re-works partts of the UI view for comparing two runs.
This is (also) for https://github.com/conbench/conbench/issues/1319, and for https://github.com/conbench/conbench/issues/1063.

- Use the newer paradigms here for HTML and JS init for datatable things.
- Use `data-sort` attr for num sort in % column.
- Set default sort order to by z-score, asc (smallest first, i.e. regressions)
- more and more
- tooltips, docs

---

Screenshot from 2023-06-23 15-44-10

### DIFF
--- a/conbench/app/compare.py
+++ b/conbench/app/compare.py
@@ -303,6 +303,8 @@ class CompareBenchmarkResults(Compare):
 
 class CompareRuns(Compare):
     type = "run"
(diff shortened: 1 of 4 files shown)


jgehrcke:
Let's collect feedback after rolling this out I would say. Can't get all the details right and towards consensus in one PR!

Issue #1335, opened 2023-06-27
benchmark result submission: require to pick from a known set of unit strings (for reliable less/more-is-better determination)


jgehrcke:
Most benchmarks are measuring a duration where smaller values are better. But there are also other types of benchmarks that report values where more is better. These benchmarks measure for example iterations per second or bytes per second.

Example for a benchmark where each result reports the unit i/s (which is iterations per second? items per second?): https://conbench.ursa.dev/benchmark-results/1b98a98dc8ad4d3fa4d236d5049d5eff/

The unit text/string (in case of the example shown: `i/s`) is user-given.

As of today, Conbench allows users to set any unit string with a benchmark result, and Conbench cannot reliably translate that unit to either "more is better" or "less is better". In fact, Conbench assigns "less is better" when it in fact doesn't know. That is historic decision-making, and a problem. Core value proposition is to detect change and to notify with decent signal-to-noise ratio. Getting the directionality wrong (along the "better"-"worse" axis) leads to significant user-facing data reporting bugs.

A conceptually better way could be to have the user also provide the directionality as in less_is_better: true/false, or that the user must pick from a set of pre-defined unit strings where the translation is hard-coded.

Having to pick from a pre-defined list would be a breaking change. For not breaking at least the deployments that we operate, we can of course make sure to handle all so-far known unit strings.

(in a second step) we could make this set of known units configurable to give other users/deployments a migration path.

jgehrcke:
Each supported unit should come with three properties to cover these concepts:
- short version (string)
- long version (string)
- less_is_better definition (boolean)

alistaire47:
It would be convenient to use a standard for units so long versions and abbreviations are consistent. At a glance it looks like ISO/IEC/ISQ-80000 is probably what we're after, but I haven't found a nice machine-readable version. Or Unicode CLDR has some more lightweight unit standardization, e.g. https://www.unicode.org/cldr/charts/42/supplemental/unit_conversions.html

jgehrcke:
As the first step, I think we should simply have a hard-coded map in application code covering what's in today's databases. Then add new validation into the result submission http handler. Emitting 400 Bad Request for an unknown unit.

alistaire47:
Yeah, long-term if we ever get to the point where our userbase needs customizable units we'll have to stick it in the database, but a Python object accessible from the various parts of the codebase that need it is sufficient and faster for now.

[...]

Issue #1331, opened 2023-06-26
UI: run -> CI run


jgehrcke:
I've been trying to wrap my head around "run" for a couple of months now. I think the most important idea here is that it tries to represent a CI run. Personally, I find "CI run" quite a bit clearer in communicating this idea than just 'run'.

austin3dickey:
From @alistaire47 in https://github.com/conbench/conbench/pull/1336#discussion_r1244210947:

>In practice, this is true of all our results. In theory, anyone could run benchmarks manually on a machine and submit the results to Conbench though, and we wouldn't be able to tell the difference [between "CI run" and "manual run"]. Not sure how useful it is to get pedantic, but maybe it's worth a second of thought.

This is a good point. But I think it's such a minor (and potentially unsupported, depending on how people handle the github graph) use case that it's not worth thinking about too much.

PR #1336, opened 2023-06-27
Compare runs view/table: sort z-score None last, better docs, better headings, misc

jgehrcke:
See commit msgs.

### DIFF
--- a/conbench/api/compare.py
+++ b/conbench/api/compare.py
@@ -192,12 +192,13 @@ class BenchmarkResultComparator:
         if self.less_is_better:
             relative_change = relative_change * -1
 
-        percent_change = _round(relative_change * 100.0)
+        percent_change = relative_change * 100.0
         regression_indicated = -percent_change > self.threshold
         improvement_indicated = percent_change > self.threshold
 
+        percent_change_for_output = _round(relative_change * 100.0)
         return {
-            "percent_change": percent_change,
+            "percent_change": percent_change_for_output,
             "percent_threshold": self.threshold,
             "regression_indicated": regression_indicated,
             "improvement_indicated": improvement_indicated,
--- a/conbench/api/compare.py
+++ b/conbench/api/compare.py
@@ -5,10 +5,10 @@ import threading
 from typing import TYPE_CHECKING, Dict, List, Optional, Tuple
 
 import flask as f
-import sigfig
 import sqlalchemy as s
 
 from conbench.dbsession import current_session
+from conbench.numstr import numstr
 
 from ..api import rule
 from ..api._endpoint import ApiEndpoint, maybe_login_required
@@ -72,9 +72,18 @@ def _get_threshold_args_from_request() -> Tuple[Optional[float], Optional[float]
     return threshold, threshold_z
 
 
-def _round(value: float) -> float:
-    """Round a float to 4 significant figures, or NaN if the input is NaN."""
-    return value if math.isnan(value) else sigfig.round(value, sigfigs=4, warn=False)
+def _round(value: float) -> Optional[float]:
+    """
+    Round a float to 4 significant figures, or NaN if the input is NaN
+
+    Note(JP): Returning math.nan translates to `null` in JSON output, but to
+    empty string "" in naive jinja template interpolation. Using `None` achieves
+    `null` in JSON, too, but can better we dealt with in jinja template
+    interpolation. For the UI we should really do
+    https://github.com/conbench/conbench/issues/1334 though.
+
+    """
+    return None if math.isnan(value) else float(numstr(value, sigfigs=4))
 
 
 if TYPE_CHECKING:
@@ -222,6 +231,8 @@ class BenchmarkResultComparator:
             "contender": self.result_info(self.contender),
             "analysis": {
                 "pairwise": self.pairwise_analysis,
+                # Watch out: self.lookback_z_score_analysis can be dict or
+                # None, and both needs to be handled in the UI (for now).
                 "lookback_z_score": self.lookback_z_score_analysis,
             },
         }
--- a/conbench/templates/compare-entity.html
+++ b/conbench/templates/compare-entity.html
@@ -33,17 +33,10 @@ matching fields between baseline and contender are aligned and mismatched fields
   <nav aria-label="breadcrumb">
     <ol class="breadcrumb">
       <li class="breadcrumb-item active">Compare</li>
-      <li class="breadcrumb-item active">
-        <a href="{{ compare_runs_url }}">Runs</a>
-      </li>
-      <li class="breadcrumb-item active">
-        <a href="{{ compare_batches_url }}">Batches</a>
-      </li>
-      <li class="breadcrumb-item active">Benchmarks</li>
       <li class="breadcrumb-item active" aria-current="page">
-        <a href="{{ url_for('app.benchmark-result', benchmark_result_id=baseline_id) }}">Baseline</a>
+        <a href="{{ url_for('app.benchmark-result', benchmark_result_id=baseline_id) }}">baseline result</a>
         ...
-        <a href="{{ url_for('app.benchmark-result', benchmark_result_id=contender_id) }}">Contender</a>
+        <a href="{{ url_for('app.benchmark-result', benchmark_result_id=contender_id) }}">reference result</a>
       </li>
     </ol>
   </nav>
--- a/conbench/templates/compare-entity.html
+++ b/conbench/templates/compare-entity.html
@@ -36,7 +36,7 @@ matching fields between baseline and contender are aligned and mismatched fields
       <li class="breadcrumb-item active" aria-current="page">
         <a href="{{ url_for('app.benchmark-result', benchmark_result_id=baseline_id) }}">baseline result</a>
         ...
-        <a href="{{ url_for('app.benchmark-result', benchmark_result_id=contender_id) }}">reference result</a>
+        <a href="{{ url_for('app.benchmark-result', benchmark_result_id=contender_id) }}">contender result</a>
       </li>
     </ol>
   </nav>
--- a/conbench/templates/compare-entity.html
+++ b/conbench/templates/compare-entity.html
@@ -64,7 +64,13 @@ matching fields between baseline and contender are aligned and mismatched fields
         <ul class="list-group">
           <li class="list-group-item active">Baseline</li>
           <li class="list-group-item" style="overflow-y: auto;">
-            <b>run</b>
+            <b>result ID</b>
+            <div align="right" style="display:inline-block; float: right;">
+              <a href="{{ url_for('app.benchmark-result', benchmark_result_id=baseline.id) }}">{{ baseline.id }}</a>
+            </div>
+          </li>
+          <li class="list-group-item" style="overflow-y: auto;">
+            <b>submitted as part of CI run</b>
             <div align="right" style="display:inline-block; float: right;">
               <a href="{{ url_for('app.run', run_id=baseline.run_id) }}">{{ baseline.run_id }}</a>
             </div>
@@ -77,9 +83,7 @@ matching fields between baseline and contender are aligned and mismatched fields
           </li>
           <li class="list-group-item" style="overflow-y: auto;">
             <b>case permutation</b>
-            <div align="right" style="display:inline-block; float: right;">
-              <a href="{{ url_for('app.benchmark-result', benchmark_result_id=baseline.id) }}">{{ baseline.display_case_perm }}</a>
-            </div>
+            <div align="right" style="display:inline-block; float: right;">{{ baseline.display_case_perm }}</div>
           </li>
           {% if baseline_run %}
             {% if baseline_run.commit.url %}
@@ -155,7 +159,13 @@ matching fields between baseline and contender are aligned and mismatched fields
         <ul class="list-group">
           <li class="list-group-item active">Contender</li>
           <li class="list-group-item" style="overflow-y: auto;">
-            <b>run</b>
+            <b>result ID</b>
+            <div align="right" style="display:inline-block; float: right;">
+              <a href="{{ url_for('app.benchmark-result', benchmark_result_id=contender.id) }}">{{ contender.id }}</a>
+            </div>
+          </li>
+          <li class="list-group-item" style="overflow-y: auto;">
+            <b>submitted as part of CI run</b>
             <div align="right" style="display:inline-block; float: right;">
               <a href="{{ url_for('app.run', run_id=contender.run_id) }}">{{ contender.run_id }}</a>
             </div>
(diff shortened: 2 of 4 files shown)


conbench/templates/compare-entity.html
```diff
@@ -33,17 +33,10 @@
   <nav aria-label="breadcrumb">
     <ol class="breadcrumb">
       <li class="breadcrumb-item active">Compare</li>
-      <li class="breadcrumb-item active">
-        <a href="{{ compare_runs_url }}">Runs</a>
-      </li>
-      <li class="breadcrumb-item active">
-        <a href="{{ compare_batches_url }}">Batches</a>
-      </li>
-      <li class="breadcrumb-item active">Benchmarks</li>
```
jgehrcke, line 42: This hierarchy suggested via breadcrumbs wasn't true/helpful.

conbench/templates/compare-entity.html
```diff
@@ -71,7 +64,13 @@
         <ul class="list-group">
           <li class="list-group-item active">Baseline</li>
           <li class="list-group-item" style="overflow-y: auto;">
-            <b>run</b>
+            <b>result ID</b>
```
jgehrcke, line 67: Imo, one of the more important properties of a result is its ID.

conbench/templates/compare-list.html
```diff
@@ -55,59 +55,44 @@ <h5 class="card-title"></h5>
...
+    </p>
+    <p>
+      Whether the contender does better or worse than the reference is indicated by sign:
+      a negative sign (for both, <code>change</code> and <code>z-score</code>) reflects the direction of degradation, a positive sign reflects the direction of improvement.
+      The directionality is derived from the user-given unit (units other than s, i/s, B/s are <a href="https://github.com/conbench/conbench/issues/1335">not yet well supported</a>).
+    </p>
+    <p>
+      Further methodological details are described in the
+      <a href="https://github.com/conbench/conbench/blob/52cf34f96cb471f493dc201b0a57ece6132eb10d/docs/pages/lookback_zscore.md">lookback z-score method documentation</a>.
+    </p>
```
jgehrcke, line 93: Right now this reads like this:

image


Happy to tweak further. Raises questions about external docs vs in-UI docs. 

I think a bit of docs/text/help is required for non-expert users to not be completely lost on this page. Thanks to @thisisnic for providing a bit of feedback, more is always appreciated! :)

conbench/templates/compare-list.html
```diff
@@ -162,15 +147,15 @@ <h6>Top Outliers</h6>
                   {{ c.analysis.pairwise.percent_change }} %
                 </td>
               {% else %}
-                <td></td>
+                <td data-order="99999999">n/a</td>
               {% endif %}
-              {% if c.analysis.lookback_z_score.z_score is not none %}
```
jgehrcke, line 167: This piece of code assumed that `if c.analysis.lookback_z_score` is never `None` (wrong assumption).

conbench/templates/compare-list.html
```diff
@@ -162,15 +147,15 @@ <h6>Top Outliers</h6>
...
+              {% if c.analysis.lookback_z_score is not none %}
                 <td class="font-monospace"
                     data-order="{{ c.analysis.lookback_z_score.z_score }}">
                   {{ c.analysis.lookback_z_score.z_score }}
                 </td>
               {% else %}
-                <td></td>
+                <td data-order="99999999">n/a</td>
```
jgehrcke, line 158: This ugly trick sorts non-existant z-score value "highest" in the direction of "improvement", i.e. these rows are last in the table when showing 'regression candidates first'.

For https://github.com/conbench/conbench/issues/1319

jgehrcke:
Feedback appreciated! :ship:

conbench/api/compare.py
```diff
@@ -72,9 +72,18 @@ def _get_threshold_args_from_request() -> Tuple[Optional[float], Optional[float]
     return threshold, threshold_z
 
 
-def _round(value: float) -> float:
-    """Round a float to 4 significant figures, or NaN if the input is NaN."""
-    return value if math.isnan(value) else sigfig.round(value, sigfigs=4, warn=False)
```
austin3dickey, line 77: Cool. I think the strict `float` used to be required before I recently refactored the compare endpoint, but now it's fine to be None.

conbench/templates/compare-entity.html
```diff
@@ -33,17 +33,10 @@
...
-      <li class="breadcrumb-item active">
-        <a href="{{ compare_batches_url }}">Batches</a>
-      </li>
-      <li class="breadcrumb-item active">Benchmarks</li>
       <li class="breadcrumb-item active" aria-current="page">
-        <a href="{{ url_for('app.benchmark-result', benchmark_result_id=baseline_id) }}">Baseline</a>
+        <a href="{{ url_for('app.benchmark-result', benchmark_result_id=baseline_id) }}">baseline result</a>
         ...
-        <a href="{{ url_for('app.benchmark-result', benchmark_result_id=contender_id) }}">Contender</a>
+        <a href="{{ url_for('app.benchmark-result', benchmark_result_id=contender_id) }}">reference result</a>
```
austin3dickey: Is this a purposeful vocabulary change you want to make across the webapp, API, and `benchalerts`? We've been fairly strict on using "contender" for a while now, and I want to be sure we remain consistent. Personally I think "reference" is less clear, because I'd guess that a reference result == the "control" result, but I agree that "contender" isn't perfect either.

conbench/templates/compare-entity.html
```diff
@@ -84,9 +83,7 @@
           </li>
           <li class="list-group-item" style="overflow-y: auto;">
             <b>case permutation</b>
-            <div align="right" style="display:inline-block; float: right;">
-              <a href="{{ url_for('app.benchmark-result', benchmark_result_id=baseline.id) }}">{{ baseline.display_case_perm }}</a>
-            </div>
+            <div align="right" style="display:inline-block; float: right;">{{ baseline.display_case_perm }}</div>
```
austin3dickey, line 86: Much better!

conbench/templates/compare-list.html
```diff
@@ -55,59 +55,44 @@ <h5 class="card-title"></h5>
 </div>
 {% endif %}
 <p>
-  The <code>change</code> column below shows the direct comparison between matching benchmark results.
+  <a data-bs-toggle="collapse"
+     href="#collapseHelpblock"
+     role="button"
+     aria-expanded="false"
+     aria-controls="collapseHelpblock">toggle help for using the table below</a>
```
austin3dickey, line 62: Maybe you could make this a button? My eyes skipped right over this even when I was looking for it :)
image

austin3dickey (approved):
Thank you!

conbench/templates/compare-entity.html
alistaire47: Agreed, at a glance I'd associate "reference" more with "baseline" than "contender".

Maybe we could move more towards older vs newer? Right now you could flip those, but we should warn if you do, since it will almost always be a mistake

conbench/templates/compare-entity.html
alistaire47: As a user who hits our API a lot, I agree. I'm less sure that's true for more general users, but since our users are devs who certainly could hit the API if they wanted, I'm not opposed.

conbench/templates/compare-entity.html
```diff
@@ -71,7 +64,13 @@
...
           <li class="list-group-item active">Baseline</li>
           <li class="list-group-item" style="overflow-y: auto;">
-            <b>run</b>
+            <b>result ID</b>
+            <div align="right" style="display:inline-block; float: right;">
+              <a href="{{ url_for('app.benchmark-result', benchmark_result_id=baseline.id) }}">{{ baseline.id }}</a>
+            </div>
+          </li>
+          <li class="list-group-item" style="overflow-y: auto;">
+            <b>submitted as part of CI run</b>
```
alistaire47, line 73: In practice, this is true of all our results. In theory, anyone could run benchmarks manually on a machine and submit the results to Conbench though, and we wouldn't be able to tell the difference. Not sure how useful it is to get pedantic, but maybe it's worth a second of thought.

conbench/templates/compare-list.html
```diff
@@ -55,59 +55,44 @@ <h5 class="card-title"></h5>
...
-          <br>
-          <br>
-          <br>
-        {% endif %}
-      </div>
-    </div>
+<div class="collapse mb-3" id="collapseHelpblock">
+  <div class="card card-body">
+    <p>
+      Each row in the table below represents a comparison. The outcome of that comparison is summarized by two metrics: <code>change</code> and <code>z-score</code>.
```
alistaire47, line 67:
```suggestion
      Each row in the table below represents a comparison. The outcome of that comparison is summarized by two metrics: <code>percent change</code> and <code>z-score</code>.
```
just to clarify this isn't absolute change (in seconds or whatever units)

alistaire47: Er, this implies changing the name of the column. In that case it's clear since we put a `%` on the number. So maybe `percent <code>change</code>`? It's clarified below, but ideally it should be clear from the start

conbench/templates/compare-list.html
```diff
@@ -55,59 +55,44 @@ <h5 class="card-title"></h5>
...
+      The <code>change</code> column compares two matching benchmark results and expresses the difference in percent.
+    </p>
+    <p>
+      The <code>z-score</code> column shows how the contender benchmark result performs in view of a larger historical ensemble of comparable
+      benchmark results (in this case, the reference run contains only the newest of the results in that ensemble;
+      all other results were submitted via different CI runs in the past).
+    </p>
+    <p>
+      Whether the contender does better or worse than the reference is indicated by sign:
+      a negative sign (for both, <code>change</code> and <code>z-score</code>) reflects the direction of degradation, a positive sign reflects the direction of improvement.
```
alistaire47, line 87: For follow-up, we could use to hit this harder visually. E.g. red/green text (or something more colorblind-friendly but still intuitive, if possible), or conditional formatting for the cell background, or something. 

Not sure exactly what the right solution is, but the goal is to minimize the time for a user to realize if a result is good or bad.

alistaire47 (approved):
Definitely moving in the right direction!

conbench/templates/compare-entity.html
austin3dickey: Note that JP didn't switch out run ID for result ID; he just added result ID to the top of each column here.

conbench/templates/compare-entity.html
austin3dickey: Oh let me forward this to #1331.

conbench/templates/compare-list.html
thisisnic: More feedback it is then! :grin: 

When you talk about units, please could you spell out what the are, e.g. "seconds (s)" and the same for the others.  Without looking it up, I wouldn't know what i/s or B/s is.  Given that previous versions of the benchmarks only had duration metrics, I reckon you also may need to spell out what these other units are and what they represent.

conbench/templates/compare-entity.html
jgehrcke: > a reference result == the "control" result

> I'd associate "reference" more with "baseline" than "contender".

Yes! This was a typo / brain fart. Elsewhere I had sometimes used `baseline (reference)`, i.e. we agree that if at all, then "reference" is conceptually close to "baseline", and kind of the inverse of "contender".

> Maybe we could move more towards older vs newer? 

I was also thinking that maybe what matters is "old" vs. "new".

But yeah when comparing just two results I think I like when the presentation is almost neutral as in A vs. B without assigning specific properties / bias. Like when using GitHub's compare/foo...bar we simply get to see the code difference (visualized as patch view), and no judgement about if one is the reference or not. This would be a bigger change of course.

Backing out of this though in this PR, let's just keep things for today

edit: added a commit to restore previous "baseline" and "contender"

conbench/templates/compare-entity.html
jgehrcke: Thanks! Yeah the main point of showing the result ID was to have _that_ link to the result, and not have the case perm link to the result.

conbench/templates/compare-list.html
jgehrcke: > Maybe you could make this a button? My eyes skipped right over this even when I was looking for it :)

Hm. I will keep this for now and we can iterate on it soon/immediately! At first I had this as a button and it was too prominent I thought. I think there's probably a middle ground where it's not what it's now, but also not a button.

conbench/templates/compare-list.html
jgehrcke: Right, here I used the precise column headings, for consistency between text and table.

We can of course also change the column headings. I thought about `relchange` instead of `change`, but thought it's best do not do everything at once.

conbench/templates/compare-list.html
jgehrcke: Yes! There was something like that in place before. I'd love to revisit that when we have it easier to render the jinja template (not using the API json structures, but our more first-class typed objects).

conbench/templates/compare-list.html
jgehrcke: > When you talk about units, please could you spell out what the are, e.g. "seconds (s)" and the same for the others.

This is good feedback, and I'd totally agree in most contexts! I carried this to https://github.com/conbench/conbench/issues/1335#issuecomment-1611197697; which is about _properly_ associating a short version of a unit to a long version.

Here I think it's OK to move on for now. What's [hard-coded as of today](https://github.com/conbench/conbench/blob/f1815a0c99211b1df2407f5f96fe069e0d5ddf89/conbench/entities/history.py#L696C17-L696C29) is literally `"B/s", "i/s"`. As a benchmark author one can use these short versions and really mean anything. For example, I assumed that `i/s` is iterations per second, but just now I saw another piece of code that talks about "items per second" (see below). Italians per salad or instructions per supervisor -- all that is confusing, but it boils down to _today_ both of `B/s` and `i/s` are interpreted as "more is better", all other units are interpreted as "less is better". :shrug:.

That madness must be resolved first, and then this tiny help text that we talk about here does not even need to go into these details.

> Given that previous versions of the benchmarks only had duration metrics

Found https://github.com/voltrondata-labs/benchmarks/commit/ff43a5fced084dc73526578a2f8384b6ce16747c#diff-75f2979db7996df563791d136dd2305730e347e4e4624ef8213c922da64eda64R174 and saw that already more than two years ago we had Arrow benchmarks reporting with the unit `i/s` (what the Conbench API sees) and meaning items per second (as we see in the client code).

conbench/templates/compare-entity.html
jgehrcke: took this to https://github.com/conbench/conbench/issues/1369

Issue #1341, opened 2023-07-04
UI: show benchmarks in run: remove api layer indirection, use first-class objects


jgehrcke:
analogous to #968 and #890.
https://github.com/conbench/conbench/blob/a60e6af944c10bd272fe649a6f128888afce7b53/conbench/app/runs.py#L121

PR #1343, opened 2023-07-04
UI: work on per-run view: details panel, toggle info/error JSON docs, re-structure table

jgehrcke:
- in the results table show the result ID, link to result from there (unlink case perm)
- conceptualize a new "run info" panel, show it at the top, and make it visually separated
- in that panel, allow for toggling error detail, run information
- both, err and info structures are now exposed as what they are: a JSON doc

### DIFF
--- a/conbench/templates/run.html
+++ b/conbench/templates/run.html
@@ -46,43 +46,59 @@
   </div>
   <div class="mt-5">
     <h3>Results</h3>
-    <p>All benchmark results that have submitted for this CI run:</p>
-    <table class="table results-in-run-table table-hover table-borderless small">
-      <thead>
-        <tr>
-          <th scope="col">benchmark start time</th>
-          <th scope="col" style="width: 120px">benchmark name</th>
-          <th scope="col">case permutation</th>
-          <th scope="col" style="width: 55px">mean</th>
-          <th scope="col" style="width: 55px">error</th>
-        </tr>
-      </thead>
-      <tbody>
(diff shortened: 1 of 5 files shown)


jgehrcke:
:sheep: :ship: feedback appreciated!

conbench/templates/run.html
```diff
@@ -46,59 +57,70 @@ <h3>Compare</h3>
...
+            <th scope="col" style="width: 150px">start time (UTC)</th>
+            <th scope="col">result</th>
+            <th scope="col">case permutation</th>
+            <!--<th scope="col" style="width: 10%">measurements</th>-->
+            <th scope="col" style="width: 8%">measurement</th>
+            <th scope="col" style="width: 10%">
+              rel err
+              <sup><i class="bi bi-info-circle"
+   data-bs-toggle="tooltip"
+   data-bs-title=" Relative standard error: standard error of the mean in relationship to the mean value. Only built when at least three samples are reported by this result. ">
```
austin3dickey, line 77: Wonder if we should display the number of iterations here too?

jgehrcke: > Wonder if we should display the number of iterations here too?

yes, my goal was to make a table similar to this:
image

(rel url e.g. c-benchmarks/tpch/063dbebaf0534d68a462c372deb79a3c)

but then I noticed that we should first do https://github.com/conbench/conbench/issues/1341

Issue #1281, opened 2023-05-25
hardware properties: document, review approach


jgehrcke:
I think from a database perspective we want a world of fully user-defined hardware/platform. We can trust users as much :-).

Today we do not allow for easy addition of additional fields.

It is not really clear why we distinguish hardware, machine, cluster via _database schema_.  Instead, one could allow users to use their own convention. Set e.g. a `type=machine`.

There's also an inconsistency with other data we store. Elsewhere (case, context, info) we store as a 'tag cloud' based on user-given key/value pairs. I think we could win a lot if hardware/platform also was a set of str/str key/value pairs. I believe things were simpler (and at the same time more powerful). One could require a few properties to be set, for sure, to provide guidance. One could also index on the set of properties. Certain aspects can be enforced via JSON schema yes.

Docs. For example, I came across "thread count":

https://github.com/conbench/conbench/blob/24f082fc7463cbb6591b1993003c0fe328b49abd/conbench/entities/hardware.py#L92

That's not really a canonical name. Yes, I suspected what it is: the logical core count (which can be different from the physical core count, as of hyperthreading). But needed code search to confirm: https://github.com/conbench/conbench/blob/24f082fc7463cbb6591b1993003c0fe328b49abd/conbench/machine_info.py#L29

jgehrcke:
Future, maybe:  platform has
- main properties: string string mapping. some required keys. hardware checksum is built from here
- additional properties: string string mapping. for later debugging. 
- name

I think
- hash should be maybe be called checksum (disambiguate from commit hash).
- hardware checksum: allow for 'different platforms' to have the same hardware checksum (makes sense)
- purpose of a unique constraint: is maybe to not have unnecessarily many DB rows / to deduplicate

jgehrcke:
quoting austin from https://github.com/conbench/conbench/issues/1340#issuecomment-1623741534

> None of the public Conbench instances use Clusters, only Machines, so this change won't affect them

great opportunity to remove notion of cluster, and rename machine to something more neutral like platform or environment.

PR #1346, opened 2023-07-06
api POST cluster: create new clusters for new optional_infos

### LINKED ISSUE #1340: cluster info: include optional_info in unique constraint
I just noticed this curious behavior of Cluster upserting:

https://github.com/conbench/conbench/blob/2175b9b19c1c53603c109e2790c7401d74e94994/conbench/entities/hardware.py#L115-L122

This is called whenever a Run is created. So it sounds like we're attempting to put a unique constraint on `(name, info)` for Cluster objects. One consequence of this is that we overwrite the cluster `optional_info` whenever we post a new Run with the same `(name, info)`.

One problem with this: in practice, this unique constraint has no technical enforcement. For example, in an internal Conbench instance, we see:
```
=> SELECT COUNT(DISTINCT name || info::text), COUNT(*) FROM hardware;
 count | count
-------+-------
   102 |   181
```
I'm not exactly sure how the database got in that state, but it can't be good that line 117 might return one row or another nondeterministically.

Another problem with this: we have a use case where we want to manage multiple Cluster instances that each have their own `optional_info` (for tracking purposes) but share the same `(name, info)` so that they're still comparable for history via the hashing algorithm: 

https://github.com/conbench/conbench/blob/2175b9b19c1c53603c109e2790c7401d74e94994/conbench/entities/hardware.py#L124-L131

I think that should be fully supported. It seems that `hash` was created to allow for this kind of thing; allowing the flexibility to choose which attributes should be incomparable or not. Therefore I think we should include `optional_info` in the unique constraint (and actually enforce that unique constraint so that the upsert works correctly).

austin3dickey:
Fixes #1340. Now we will create a new Cluster row when a new cluster is POSTed with new optional_info. That way, old benchmark results can still be associated with the "correct" cluster optional info.

### DIFF
--- a/conbench/entities/_entity.py
+++ b/conbench/entities/_entity.py
@@ -180,11 +180,12 @@ class EntityMixin(Generic[T]):
     @classmethod
     def get_or_create(cls: Type[T], props: Dict) -> T:
         """
-        Try to create, but expect conflict (work with unique constraint on
-        name/tags).
+        Try to create, but expect conflict (work with unique constraint on name/tags).
 
-        Return (newly created, or previously existing) object, or raise an
-        exception.
+        Return (newly created, or previously existing) object, or raise an exception.
+
+        If there is no unique constraint on the keys of props, race conditions will
+        result in duplicated rows, but should not fail. This should be rare.
         """
 
         def _fetch_first():
@@ -195,6 +196,7 @@ class EntityMixin(Generic[T]):
             return result
 
         obj = cls(**props)
+        obj.post_init_pre_add()
         current_session.add(obj)
         try:
             current_session.commit()
@@ -210,6 +212,12 @@ class EntityMixin(Generic[T]):
         result = _fetch_first()
         assert result is not None
         return result
+
+    def post_init_pre_add(self):
+        """Override this method to do any initialization before adding to the session
+        when using get_or_create().
+        """
+        pass
 
 
 class EntitySerializer:
--- a/conbench/entities/_entity.py
+++ b/conbench/entities/_entity.py
@@ -64,7 +64,7 @@ def to_float(value):
     return float(value) if value is not None else None
 
 
-T = TypeVar("T")
+T = TypeVar("T", bound="EntityMixin")
 
 
 class EntityMixin(Generic[T]):
--- a/conbench/entities/hardware.py
+++ b/conbench/entities/hardware.py
@@ -114,10 +114,12 @@ class Cluster(Hardware):
 
     @classmethod
     def upsert(cls, **kwargs):
-        cluster = cls.first(name=kwargs["name"], info=kwargs["info"])
-        if cluster:
-            cluster.update(dict(optional_info=kwargs["optional_info"]))
-        else:
+        cluster = cls.first(
+            name=kwargs["name"],
+            info=kwargs["info"],
+            optional_info=kwargs["optional_info"],
+        )
+        if not cluster:
             cluster = cls.create(kwargs)
         return cluster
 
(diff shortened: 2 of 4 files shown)


conbench/entities/hardware.py
```diff
@@ -114,10 +114,12 @@ class Cluster(Hardware):
...
-        if cluster:
-            cluster.update(dict(optional_info=kwargs["optional_info"]))
-        else:
+        cluster = cls.first(
+            name=kwargs["name"],
+            info=kwargs["info"],
+            optional_info=kwargs["optional_info"],
+        )
+        if not cluster:
             cluster = cls.create(kwargs)
```
jgehrcke: There is a tiny chance that this create here can fail. Can you maybe try using https://github.com/conbench/conbench/blob/1f8a7f3a657c6028cb546bdc368af11d1afcb272/conbench/entities/_entity.py#L181?

jgehrcke (approved):
Thanks! :) Made a tiny suggestion for improvement: for more correctness, but also for deduplicating code I had recently added a `get_or_create()` helper that tries to have decent special case handling/error handling/transaction management (review appreciated).

alistaire47 (approved):

conbench/entities/hardware.py
austin3dickey: Unfortunately I can't use this method, because there is no enforced unique constraint on this table. If we put a unique constraint on `(name, info, optional_info)` for `Cluster` we'd effectively be putting a unique constraint on `(name)` for `Machine`, because `Machine` rows are forced into the same `Hardware` table with `null` values for `info` and `optional_info`. (We definitely want to reduce this complexity in the future! related: #1281)

On the "bright" side, this means technically this line won't fail the way you're thinking. If there's a race condition, it'll result in two rows with the same name/info/optional_info and two different IDs. This is messy but won't negatively affect any logic in the rest of the app.

austin3dickey: Oh wait, I just checked your method's code and it should work even without a unique constraint :) it will just fail to handle any race conditions, resulting in duplicate rows, just like my code.

austin3dickey: Okay, I switched to using `get_or_create()`.

The one thing it was missing: it went from `obj = cls(**props)` straight to `current_session.add(obj)`. But the Hardware objects generate the hash (checksum) column _before_ adding the object to the session, since that column is not nullable. So I added an overridable `post_init_pre_add()` method that will allow for custom initialization. It's not super pretty but I think it should handle any future use case.

Let me know what you think; I'm going to 🚢 this for now and we can revisit later.

conbench/entities/_entity.py
```diff
@@ -64,7 +64,7 @@ def to_float(value):
     return float(value) if value is not None else None
 
 
-T = TypeVar("T")
+T = TypeVar("T", bound="EntityMixin")
```
austin3dickey, line 67: This allows mypy to know that `cls` down below can use EntityMixin's methods.

conbench/entities/hardware.py
jgehrcke: > I just checked your method's code and it should work even without a unique constraint :

:) yeah, cool!

> I added [...] for custom initialization

Ah! Could that also be in the `__init__()` method of `cls` so that `obj = cls(**props)` would run that logic? :shrug: :)

conbench/entities/_entity.py
jgehrcke: Thanks for thinking that through, reading docs

conbench/entities/hardware.py
austin3dickey: >Could that also be in the __init__() method

Haha, this is what I get for trying to go too fast on a Friday. Yes, that would be a nice simplification.

Issue #1355, opened 2023-07-12
benchmark result view: present measurement result more clearly


jgehrcke:
User feedback

>  the result table is difficult to read. There are lots of lines, most of which are empty. Units are not showed besides the numbers, but in a separate line (which is which?). It's not obvious what one should focus on.

We have long proposed to show the result with reasonable precision and unit, ideally with the relative error derived from N repetitions. Let's do that.

alistaire47:
Yeah, a topline table would be helpful.

Warm take: I think we should push users towards the relevant compare pages and away from raw result pages, because without context it's hard to tell if 1.6s is good or bad. Basically all of the useful context we're providing on the result page is in the chart. The chart is good, but the compare page is a better place to expand on it with z-scores, percent change, etc. in a topline table (which we had but removed for now).

jgehrcke:
> push users towards the relevant compare pages and away from raw result pages

Interesting! I find the raw result page to be conceptually really important. Of course it depends on the specific use case. When thinking about "comparison" then I find the result-result comparison confusing beyond direct pairwise comparison.

> but the compare page is a better place to expand on it with z-scores

I think the individual result page is where a lookback z-score value can be explained rather easily (we can then say: this is the z-score of this result in view of the history shown in the history plot).

I think the result-result comparison page is maybe not the best place to present a z-score. It's not so intuitive for people understanding z-score. They think: when comparing only two results then one cannot derive a z-score. A z-score metric is the result of comparing one result with many results.

alistaire47:
> I think the individual result page is where a lookback z-score value can be explained rather easily (we can then say: this is the z-score of this result in view of the history shown in the history plot).
> 
> I think the result-result comparison page is maybe not the best place to present a z-score. It's not so intuitive for people understanding z-score. They think: when comparing only two results then one cannot derive a z-score. A z-score metric is the result of comparing one result with many results.

There's some oddness here, though: `history` (and therefore the plot) assume a certain comparison (i.e. straight up the git tree). You can't plausibly define history another way, so that's fine, but it _is_ a comparison. z-scores are a product of that comparison—not between two commits but from the whole lineage.

But we only produce z-scores in the `compare` endpoint. (I know the UI is not the API, but they are largely parallel.) That makes sense for percent change, but less for z-scores, since they can only be produced for baselines on the contender's history, in which case it will be the same regardless of the baseline.

[...]

Issue #829, opened 2023-03-09
benchmark result view: better self-documentation


jgehrcke:
A user asked:

> That "mark this as the first result of a distribution change" button, does that get stored somewhere and do other people also see it if I clicked it? Is there a way to mark it for all the queries, and all the other parameters?

That certainly relates to https://github.com/conbench/conbench/issues/691, but also to the fact that even very basic principles are not obvious when looking at the benchmark result view. We can improve that.

Concepts like benchmark name, case permutation should be self-evident from the page. The history plot shows a time axis, and we don't explain that these are commit creation (authoring) times. We can explain nicely what all the points shown in the plot have in common.

Note: this might be a small epic / milestone / initiative.

jgehrcke:
https://github.com/conbench/conbench/pull/833 was helpful, but @joosthooz's feedback was also largely about the following: where does this button mutate state? Just for him, locally (via cookie?) or in the database? This is a _great_ question and must be answered properly.

jgehrcke:
We should also clarify the relationship between the result properties shown in the output and the result properties delivered in the input. For example, `Additional Context Details` seems to be coming from `info` in a certain schema. Clarify this with a tooltip or comparable. Relates to #838.

PR #1356, opened 2023-07-13
UI: result view: add summary, explain history plot better

jgehrcke:
This is quite a bit of UI work affecting the benchmark result view.

Mainly for #1355 and also for #829.

Quoting from #829:

> Concepts like benchmark name, case permutation should be self-evident from the page. The history plot shows a time axis, and we don't explain that these are commit creation (authoring) times. We can explain nicely what all the points shown in the plot have in common.

Some aspects:
- makes case permutation more obvious in heading
- shows summary with measurement result (reduced precision, and with unit)
- more usage of monospaced font, a little less boldness
- adjustment of terms

There's a bit more. See commit msgs and also screenshots:
[before (Arrow Conbench)](https://github.com/conbench/conbench/assets/265630/c246d69a-0a48-4877-84f4-486e07bd4e28)
[after (internal, sanitized)](https://github.com/conbench/conbench/assets/265630/0725d874-0666-4a71-a3b1-0c6248bfd497)

This probably provokes quite a bit of feedback which would be great. Let's get this in, and then see how this looks like in prod. Super happy to iterate.

### DIFF
--- a/conbench/templates/benchmark-result.html
+++ b/conbench/templates/benchmark-result.html
@@ -133,10 +133,12 @@
             <div align="right" style="display:inline-block; float: right;"><code>{{ benchmark.display_timestamp }}</code></div>
           </li>
           {% for k,v in benchmark.stats.items() %}
-            <li class="list-group-item" style="overflow-y: auto;">
-              <b>{{ k }}</b>
-              {% if v is not none %}<div align="right" style="display:inline-block; float: right;">{{ v }}</div>{% endif %}
-            </li>
+            {% if v is not none %}
+              <li class="list-group-item" style="overflow-y: auto;">
+                {{ k }}
+                <div align="right" style="display:inline-block; float: right;"><code>{{ v }}</code></div>
+              </li>
+            {% endif %}
           {% endfor %}
         {% endif %}
         <li class="list-group-item list-group-item-primary"> Raw tags (case permutation + name)</li>
--- a/conbench/templates/benchmark-result.html
+++ b/conbench/templates/benchmark-result.html
@@ -211,6 +211,10 @@
           {% endif %}
         {% endif %}
         <li class="list-group-item list-group-item-primary">Hardware (checksum is constant between results in history plot above)</li>
+        <li class="list-group-item" style="overflow-y: auto;">
+          <b>checksum</b>
+          <div align="right" style="display:inline-block; float: right;"><code>{{ result.run.hardware.hash }}</code></div>
+        </li>
         {% for k,v in run.hardware.items() %}
           <li class="list-group-item" style="overflow-y: auto;">
             <b>{{ k }}</b>
--- a/conbench/templates/benchmark-result.html
+++ b/conbench/templates/benchmark-result.html
@@ -206,14 +206,15 @@
             </li>
           {% endif %}
         {% endif %}
-        <li class="list-group-item list-group-item-primary">Hardware</li>
+        <li class="list-group-item list-group-item-primary">Hardware (checksum is constant between results in history plot above)</li>
         {% for k,v in run.hardware.items() %}
           <li class="list-group-item" style="overflow-y: auto;">
             <b>{{ k }}</b>
             <div align="right" style="display:inline-block; float: right;">{{ v }}</div>
           </li>
         {% endfor %}
-        <li class="list-group-item list-group-item-primary">Context</li>
+
+        <li class="list-group-item list-group-item-primary">Context (constant between results in history plot above)</li>
         {% for k,v in benchmark.context.items() %}
           <li class="list-group-item" style="overflow-y: auto;">
             <b>{{ k }}</b>
@@ -221,7 +222,7 @@
           </li>
         {% endfor %}
         {% if benchmark.info and benchmark.info|length > 1 %}
-          <li class="list-group-item list-group-item-primary">Additional Context Details</li>
+          <li class="list-group-item list-group-item-primary">Additional context (may vary between results in history plot above)</li>
           {% for k,v in benchmark.info.items() %}
             <li class="list-group-item" style="overflow-y: auto;">
               <b>{{ k }}</b>
--- a/conbench/templates/benchmark-result.html
+++ b/conbench/templates/benchmark-result.html
@@ -141,8 +141,8 @@
         {% if benchmark.stats %}
           <li class="list-group-item list-group-item-primary">Result</li>
           <li class="list-group-item" style="overflow-y: auto;">
-            <b>timestamp (start time)</b>
-            <div align="right" style="display:inline-block; float: right;">{{ benchmark.display_timestamp }}</div>
+            timestamp (usually the start time)
+            <div align="right" style="display:inline-block; float: right;"><code>{{ benchmark.display_timestamp }}</code></div>
           </li>
           {% for k,v in benchmark.stats.items() %}
             <li class="list-group-item" style="overflow-y: auto;">
@@ -226,7 +226,7 @@
         {% for k,v in run.hardware.items() %}
           <li class="list-group-item" style="overflow-y: auto;">
             <b>{{ k }}</b>
-            <div align="right" style="display:inline-block; float: right;">{{ v }}</div>
+            <div align="right" style="display:inline-block; float: right;"><code>{{ v }}</code></div>
           </li>
         {% endfor %}
 
--- a/conbench/templates/benchmark-result.html
+++ b/conbench/templates/benchmark-result.html
@@ -1,43 +1,47 @@
 {% extends "app.html" %}
 {% import 'bootstrap/wtf.html' as wtf %}
 {% block app_content %}
-
-
   <h4>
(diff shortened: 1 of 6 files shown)


jgehrcke:
Quoting myself from before, just be sure this is seen:

> This probably provokes quite a bit of feedback which would be great. Let's get this in, and then see how this looks like in prod. Super happy to iterate.

conbench/templates/benchmark-result.html
```diff
@@ -1,83 +1,119 @@
...
-          <br />
-          <div class="panel panel-success">
-            <div class="panel-heading">Summary for selected benchmark (green data point in plot)</div>
-            <div class="panel-body">
-              <ul class="ul-histplot-run-details">
-              </ul>
-            </div>
-          </div>
+  <h4>
+    <strong>{{ benchmark.display_bmname }}</strong> / result {{ result.id[:15] }}...
```
austin3dickey, line 5: nit: Cutting off the ID makes it impossible to find the full result ID anywhere on the page. (I think you'd have to check the URL to get it.) Plus, since we use UUID7s for the result IDs, the first 12 characters are going to look very similar for results submitted around the same time, making it a little more difficult to distinguish between the headers of two pages. I think we shouldn't cut this off.

conbench/templates/benchmark-result.html
```diff
@@ -1,83 +1,119 @@
...
+          <br>
+          <code>{{ result.timestamp.strftime("%Y-%m-%d") }}</code>
         </div>
+        <div class="cb-bmr-summary-item">
+          <strong>benchmarked code</strong>
+          <br>
+          <code>
+            <a href="{{ result.run.commit.repo_url }}">{{ result.run.commit.repo_url }}</a>,
+            commit {{ result.ui_commit_url_anchor() |safe }}
+            ({{ result.run.commit.timestamp.strftime("%Y-%m-%d") }})
```
austin3dickey, line 39: nit: maybe consider separating these with newlines instead of commas? With the wraparound I think it's a little hard to read

conbench/templates/benchmark-result.html
```diff
@@ -185,22 +227,34 @@ <h3>History plot for benchmark '{{ benchmark.display_bmname }}'</h3>
             </li>
           {% endif %}
         {% endif %}
-        <li class="list-group-item list-group-item-primary">Hardware</li>
+        <li class="list-group-item list-group-item-primary">
+          Hardware (checksum is constant between results in history plot above)
```
austin3dickey, line 231: grammar nit: I think "between" could be viewed as correct here, as it highlights the individual relationships between the results. But I think "among" would feel more natural to me in this phrase, highlighting that they're all of a shared group.

austin3dickey (commented):
Thanks for the changes! Just a few very small nitpicks

conbench/templates/benchmark-result.html
jgehrcke: > find the full result ID anywhere on the page

good feedback!

yes it should totally be in the raw data area below, and also in the title it should be presented longer (maybe untruncated).

conbench/templates/benchmark-result.html
jgehrcke: > But I think "among" would feel more natural to me in this phrase

Thanks!

Yes, "among" is better and probably what I really wanted to write before a Germanism entered the picture :P

conbench/templates/benchmark-result.html
jgehrcke: yes!

jgehrcke: but maybe: let's only show the run ID?

conbench/templates/benchmark-result.html
alistaire47: "between" is between 2 things, "among" is among 3+ things if we want the formal rule. 

Getting flashbacks to teaching SAT grammar in a former life

conbench/templates/benchmark-result.html
```diff
@@ -1,83 +1,119 @@
...
+                <ul class="ul-histplot-run-details">
+                </ul>
+              </div>
+            </div>
+          </div>
+          <div class="text-muted fst-italic">
+            This plot shows the current result in historical context.
+            The time axis shows the commit date (corresponds to code evolution).
+            Think: along the plot everything is held constant except for the benchmarked code.
+            Click a data point (grey) in the plot to see a corresponding result summary.
```
alistaire47, line 73: This text is great, but it (very arguably) feels like it belongs in a tooltip for the plot, like one of those `(i)` info icons next to the "historical evolution" title.

One advantage: We can make tooltip text quite a bit longer if we want without affecting the page's information density

conbench/templates/benchmark-result.html
```diff
@@ -1,83 +1,119 @@
...
-          <b>Run reason</b>
-          <div align="right" style="display:inline-block; float: right;">
-            {% if run.reason %}{{ run.reason }}{% endif %}
+            <a href="{{ url_for('app.run', run_id=benchmark.run_id) }}">{{ benchmark.run_id }}</a> with name
+            <code>
+              {% if run.name %}{{ run.name }}{% endif %}
+            </code> and reason
+            <code>
+              {% if run.reason %}{{ run.reason }}{% endif %}
+            </code>
```
alistaire47, line 116: Two thoughts:

1. This is moving further from the API result structure. Is this something we want to do? If so it's fine, but most of this page is still largely like that schema (albeit with different section names).
2. Hardware is part of what identifies the CI run; there are multiple runs with the same id/name/reason on different hardwares.

conbench/templates/benchmark-result.html
```diff
@@ -102,19 +138,25 @@ <h3>History plot for benchmark '{{ benchmark.display_bmname }}'</h3>
           {% endfor %}
         {% endif %}
         {% if benchmark.stats %}
-          <li class="list-group-item list-group-item-primary">Result</li>
+          <li class="list-group-item list-group-item-primary">Measurement result</li>
```
alistaire47, line 141: Thank you, this annoyed me

conbench/templates/benchmark-result.html
```diff
@@ -185,22 +227,34 @@ <h3>History plot for benchmark '{{ benchmark.display_bmname }}'</h3>
...
           <li class="list-group-item" style="overflow-y: auto;">
             <b>{{ k }}</b>
-            <div align="right" style="display:inline-block; float: right;">{{ v }}</div>
+            <div align="right" style="display:inline-block; float: right;">
+              <code>{{ v }}</code>
+            </div>
           </li>
         {% endfor %}
-        <li class="list-group-item list-group-item-primary">Context</li>
+        <li class="list-group-item list-group-item-primary">Context (constant between results in history plot above)</li>
```
alistaire47, line 247: These parentheticals would also fit nicely as (i) tooltips if we prefer

conbench/templates/benchmark-result.html
alistaire47: Looking at this more, I really think run reason belongs in this table; it's important context, e.g. for why commit info may be null

conbench/templates/benchmark-result.html
jgehrcke: commit upcoming: changing to show full ID

conbench/templates/benchmark-result.html
jgehrcke: > I really think run reason belongs in this table

one argument I think is a future state where the run-related properties are stored in the result table 


commit upcoming: doing the newlines, thought about this for too long now, moving on.

conbench/templates/benchmark-result.html
jgehrcke: doing that in upcoming commit

conbench/templates/benchmark-result.html
jgehrcke: can do that, yes! let's see

conbench/templates/benchmark-result.html
jgehrcke: note: good thoughts, not quickly actionable for me

about (1) we of course need to understand what a persona needs to see

not sure what's implied by (2).

conbench/templates/benchmark-result.html
jgehrcke: yes can do, skipping for now

jgehrcke:
some feedback processed via https://github.com/conbench/conbench/pull/1363

conbench/templates/benchmark-result.html
jgehrcke: https://github.com/conbench/conbench/issues/1432

conbench/templates/benchmark-result.html
jgehrcke: (done, btw)

conbench/templates/benchmark-result.html
jgehrcke: doing that in https://github.com/conbench/conbench/pull/1433

Issue #1353, opened 2023-07-11
c-bench: tinyplots: use hardware hash for grouping,  not hardware ID


jgehrcke:
In the c-bench UI in tinyplots I have been grouping into individual timeseries via hardware ID and not via hardware hash. That creates bugs!

relates to #1281 / https://github.com/conbench/conbench/issues/1340#issuecomment-1618469152

For example, these belong together:

![image](https://github.com/conbench/conbench/assets/265630/6fb09b2c-d8e3-431a-ad42-dd66622a83df)

austin3dickey:
Thanks for the catch! Simplifying the "environment" schema will help us so much.

jgehrcke:
Done via https://github.com/conbench/conbench/pull/1357.

PR #1357, opened 2023-07-13
UI: c-bench/tinyplots: fix grouping of results into timeseries: use hardware checksum instead of ID

jgehrcke:
This was motivated mainly by https://github.com/conbench/conbench/issues/1353.

That is, the main point of this patch is to use the "hardware checksum" instead of the "hardware ID" for grouping results into timeseries groups.

But: turns out what we call "hardware hash" in Conbench internals (and externals!) isn't a canonical hash/checksum. It can be a long string with various pieces of user-given data. That makes this "hash" difficult to work with in core data structures. One core data structure that we have is a 4-tuple where the hardware checksum is one of the four parameters.

Here, I have no opted for (quick and dirty) re-writing the Conbench hardware hash into a 'proper one'. A downside is now that in HTML source and JS source there is a string/id/hash that pops up which does _not_ match database contents.

We can make things more tidy as part of https://github.com/conbench/conbench/issues/1340.

### DIFF
--- a/conbench/job.py
+++ b/conbench/job.py
@@ -1,4 +1,5 @@
 import dataclasses
+import hashlib
 import logging
 import signal
 import threading
@@ -68,7 +69,7 @@ class BMRTBenchmarkResult:
     unit: str
     benchmark_name: str
     started_at: float
-    hardware_id: str
+    hardware_checksum: str
     hardware_name: str
     case_text_id: str
     case_dict: Dict[str, str]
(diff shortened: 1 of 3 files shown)


conbench/job.py
```diff
@@ -275,7 +276,20 @@ def _fetch_and_cache_most_recent_results_guts(
             svs=result.svs,  # float(result.mean) if result.mean else None,
             svs_type=result.svs_type,
             unit=str(result.unit) if result.unit else "n/a",
-            hardware_id=str(bmrrun.hardware.id),
+            # Current `hardware.hash` is a string (not byte sequence), and does
+            # not have a predictable charset. I hoped it would be just the
+            # hexdigest of a popular hash function. What we have contains
+            # user-given data, i.e. the string is brittle to work with in code
+            # and generated documents. E.g. may not work in JavaScript var
```
austin3dickey, line 283: Thanks for this e.g.!

jgehrcke:
Locally I have confirmed that the bug described in https://github.com/conbench/conbench/issues/1353 has been fixed. Let's get this in; feedback appreciated!

conbench/job.py
```diff
@@ -275,7 +276,20 @@ def _fetch_and_cache_most_recent_results_guts(
...
+            # declaration statements). Translate this Conbench business logic
+            # "hardware hash" into one with predictable charset. This is for
+            # grouping/sorting purposes, and for building UI. Use MD5 (fast,
+            # unlikely collision, good enough). Can clean up when reworking
+            # hardware/platform/env:
+            # https://github.com/conbench/conbench/issues/1340
```
austin3dickey, line 289: nit: I think you mean https://github.com/conbench/conbench/issues/1281

austin3dickey (commented):
LGTM!

alistaire47:
Oof, I hadn't looked at the raw data and just assumed "hardware hash" was just a hash. We should definitely clean this up before we expose it through the API (which we should eventually also do; lots of external reporting and analysis is relying on hardware name instead of hash).

jgehrcke:
>  just assumed "hardware hash" was just a hash. We should definitely clean this up before we expose it through the API (which we should eventually also do; lots of external reporting and analysis is relying on hardware name instead of hash).

:rocket:! Thanks for feedback!

conbench/job.py
jgehrcke: yes!

PR #1362, opened 2023-07-17
benchalerts: switch to new ConbenchClient

### LINKED ISSUE #1186: benchalerts: use new ConbenchClient
Instead of the `LegacyConbenchClient`, `benchalerts` should make use of the new `ConbenchClient` backed by `RetryingHTTPClient` introduced in #1148.

This change will involve fixing the unit test mocking framework since that takes advantage of the old `adapter` arg which is not used anymore.

austin3dickey:
Fixes #1186.

This PR switches `benchalerts` off of the `LegacyConbenchClient` onto the new `ConbenchClient` with better logging and retry logic.

### DIFF
--- a/benchalerts/benchalerts/pipeline_steps/conbench.py
+++ b/benchalerts/benchalerts/pipeline_steps/conbench.py
@@ -3,14 +3,12 @@
 from enum import Enum
 from typing import Any, Dict, List, Optional
 
-import requests
-from benchclients.conbench import LegacyConbenchClient
+from benchclients.conbench import ConbenchClient
+from benchclients.http import RetryingHTTPClientNonRetryableResponse
 from benchclients.logging import log
 
 from ..alert_pipeline import AlertPipelineStep
 from ..conbench_dataclasses import FullComparisonInfo, RunComparisonInfo
-
-ConbenchClient = LegacyConbenchClient
 
 
 class BaselineRunCandidates(Enum):
@@ -72,7 +70,7 @@ class GetConbenchZComparisonForRunsStep(AlertPipelineStep):
         run_ids: List[str],
         baseline_run_type: BaselineRunCandidates,
         z_score_threshold: Optional[float] = None,
-        conbench_client: Optional[LegacyConbenchClient] = None,
+        conbench_client: Optional[ConbenchClient] = None,
         step_name: Optional[str] = None,
     ) -> None:
         super().__init__(step_name)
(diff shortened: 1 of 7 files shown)


benchalerts/benchalerts/pipeline_steps/conbench.py
```diff
@@ -97,8 +95,8 @@ def _get_one_run_comparison(self, run_id: str) -> Optional[RunComparisonInfo]:
         """
         try:
             contender_info = self.conbench_client.get(f"/runs/{run_id}/")
-        except requests.HTTPError as e:
-            if e.response.status_code == 404:
+        except RetryingHTTPClientNonRetryableResponse as e:
+            if e.error_response.status_code == 404:
```
austin3dickey, line 99: Thank goodness for a 98%+ coverage test suite.

alistaire47 (approved):
This change is satisfying; lots of unpleasant bits getting wiped away

benchalerts/tests/unit_tests/mocks.py
```diff
@@ -83,6 +84,15 @@ def send(self, *args, **kwargs):
         return MockResponse.from_file(response_path)
 
 
+class MockConbenchClient(ConbenchClient):
```
jgehrcke, line 87: Can we maybe describe the role of this with a code comment?

benchalerts/tests/unit_tests/mocks.py
```diff
@@ -83,6 +84,15 @@ def send(self, *args, **kwargs):
...
+    def __init__(self):
+        super().__init__()
+        self.session.mount("https://", MockAdapter())
```
jgehrcke, line 90: `MockAdapter` appears to have quite some github-flavored implementation. Why is it OK/good to use it in a `MockConbenchClient`?  If MockAdapter tries to be a bit of a silver bullet then maybe it deserves a code comment.

benchalerts/tests/unit_tests/mocks.py
```diff
@@ -83,6 +84,15 @@ def send(self, *args, **kwargs):
...
+
+    def _login_or_raise(self) -> None:
```
jgehrcke, line 92: Let's also add a code comment here to explain why we choose to override this method.

I can imagine more than one reason, and (not but) I'd like to know what the author of this code thought :-). Makes things easier / quicker to change in the future.

jgehrcke (approved):
:rocket: nice thank you! We can land this!

I have added comments and questions that are all about making this a little easier to understand / navigate / maintain in the future.

benchalerts/benchalerts/pipeline_steps/conbench.py
jgehrcke: Note to self: the `get()` method ideally has a nice docstring explaining which types of exceptions the caller must be prepared to handle. Also, ideally mypy can help here.

austin3dickey: Oops, I hope my comment didn't come off as snarky! I was trying to convey my relief that my time spent on unit tests in the past had paid off 🙂

jgehrcke: > I hope my comment didn't come off as snarky

Not at all!

Issue #1365, opened 2023-07-19
moving dependency: gunicorn: `ValueError: Invalid file descriptor: -1`


jgehrcke:
![image](https://github.com/conbench/conbench/assets/265630/2b70dc52-b7e7-4c8c-8870-f97cf8a4badf)

jgehrcke:
We were affected by https://github.com/benoitc/gunicorn/issues/3025.

A moving dependency hit us, the faulty gunicorn releases 21.0.0 and 21.0.1 both released during  2023-07-17.

Also see https://docs.gunicorn.org/en/stable/news.html.

For now @benoitc has done a good job and released a patch already. Formidable :-)

21.1.0 was released on 2023-07-18 and should contain the fix. In that sense there's nothing we absolutely need to do now.

However: let's absolutely pin gunicorn (like everything else) at some point in the future when we care more about predictability.

jgehrcke:
Edit: gunicorn 21.2.0 - 2023-07-19 (released one hour ago) addresses this bug (not as previously claimed the 21.1.0 release).

jgehrcke:
Resolved for now.

PR #1366, opened 2023-07-19
requirements-webapp: pin gunicorn

jgehrcke:
After a bit of forth and back in #1365 let's just get into the habit of pinning. Tiny step into a good direction.


jgehrcke:
ci / ui-screenshots failed knowingly. Let's get this in.

PR #1373, opened 2023-07-20
benchmark result submission: require one of known units

jgehrcke:
This is for #1335.

Next up:
- use `less_is_better` and `long` in code
- work on HTTP API docs to automatically show list of supported units

### DIFF
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -912,19 +912,28 @@ class BenchmarkResultStatsSchema(marshmallow.Schema):
             # present?
             "description": conbench.util.dedent_rejoin(
                 """
-                A list of benchmark results (e.g. durations, throughput). This
-                will be used as the main + only metric for regression and
-                improvement. The values should be ordered in the order the
-                iterations were executed (the first element is the first
-                iteration, the second element is the second iteration, etc.).
-                If an iteration did not complete but others did and you want to
-                send partial data, mark each iteration that didn't complete as
-                `null`.
+                A list of measurement results (e.g. duration, throughput).
 
-                You may populate both this field and the "error" field in the top level
-                of the benchmark result payload. In that case, this field measures
-                the metric's values before the error occurred. These values will not be
-                compared to non-errored values in analyses and comparisons.
+                Each value in this list is meant to correspond to one
+                repetition of ideally the exact same measurement.
+
+                We recommend to repeat a measurement N times (3-6) for enabling
+                systematic stability analysis.
+
+                Values are expected to be ordered in the order the
+                iterations/repetitions were executed (the first element
+                corresponds to the first repetition, the second element is the
+                second repetition, etc.).
+
+                Values must be numeric or `null`: if one repetition failed but
+                others did not you can mark the failed repetition as `null`.
+
+                Note that you may populate both this field and the "error"
+                field in the top level of the benchmark result payload.
+
+                If any of the values in `data` is `null` or if the `error`
+                field is set then Conbench will not include any of the reported
+                data in automated analyses.
                 """
             )
         },
@@ -957,7 +966,9 @@ class BenchmarkResultStatsSchema(marshmallow.Schema):
         # strings are/were allowed to be injected. TODO: be more strict.
         # Require users to provide a unit (non-zero length string)./
         required=True,
-        metadata={"description": "The unit of the data object (e.g. seconds, B/s)"},
+        metadata={
+            "description": "The unit of the measurement result (object (e.g. seconds, B/s)"
+        },
     )
     time_unit = marshmallow.fields.String(
         required=True,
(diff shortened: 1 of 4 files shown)


conbench/entities/benchmark_result.py
```diff
@@ -598,13 +599,40 @@ def ui_mean_and_uncertainty(values: List[float], unit: str) -> str:
...
     Raises BenchmarkResultValidationError upon logical inconsistencies.
 
+    `stats_usergiven` is deserialized JSON, validated against
+    `BenchmarkResultStatsSchema(marshmallow.Schema)`.
+
     Only run this for the 'success' case, i.e. when the input is a list longer
     than zero, and each item is a number (not math.nan).
 
+    Validate the user-given unit string also only in case of success, i.e.
+    allow for 'bad units' to be submitted with 'errored' results.
```
jgehrcke, line 634: Interesting decision here I needed to make: _when_  really to validate the unit string. We allow for submitting messy erroneous results, and so I went with that philosophy.

conbench/entities/benchmark_result.py
```diff
@@ -912,19 +944,28 @@ class BenchmarkResultStatsSchema(marshmallow.Schema):
...
+
+                Values must be numeric or `null`: if one repetition failed but
+                others did not you can mark the failed repetition as `null`.
+
+                Note that you may populate both this field and the "error"
+                field in the top level of the benchmark result payload.
+
+                If any of the values in `data` is `null` or if the `error`
+                field is set then Conbench will not include any of the reported
+                data in automated analyses.
```
jgehrcke, line 968: Used the opportunity to hopefully clarify some things here.

conbench/entities/benchmark_result.py
```diff
@@ -957,7 +998,9 @@ class BenchmarkResultStatsSchema(marshmallow.Schema):
         # strings are/were allowed to be injected. TODO: be more strict.
         # Require users to provide a unit (non-zero length string)./
         required=True,
-        metadata={"description": "The unit of the data object (e.g. seconds, B/s)"},
+        metadata={
+            "description": "The unit of the measurement result (object (e.g. seconds, B/s)"
```
jgehrcke, line 1002: docs update here will come in a separate patch

conbench/entities/benchmark_result.py
```diff
@@ -598,13 +599,40 @@ def ui_mean_and_uncertainty(values: List[float], unit: str) -> str:
...
+    Raise BenchmarkResultValidationError for invalid unit string.
+
+    Return augmented unit string, to be inserted into database.
+    """
+    if u == "b/s":
+        # Rewrite short variant here for a legacy client, see
+        # https://github.com/conbench/conbench/issues/1335
+        # We might need/want to do a database migration where we rewrite
+        # b/s to B/s.
+        u = "B/s"
```
jgehrcke, line 613: I think this means we have to do a DB migration. At first I thought "ah, then I don't do the rewrite". But then, well.. not a good state either, kicking the can down the road. I don't want to describe to users that we allow `b/s`.

conbench/tests/api/test_results.py
```diff
@@ -1196,3 +1196,39 @@ def test_create_result_unexpected_stats_keys(self, client, uekey, samples):
...
         if len(samples) < 3:
             assert bmrdict["stats"][uekey] is None
+
+    def test_create_result_bad_unit(self, client):
+        self.authenticate(client)
+        result = _fixtures.VALID_RESULT_PAYLOAD.copy()
+
+        result["stats"] = {
+            "data": (3, 5),
+            "times": [],  # key must be there as of now, validate more, change this.
```
jgehrcke, line 1206: maybe we should do this as part of this epic: remove the requirement for the `times`.

jgehrcke:
:ship:  Landing this; feedback appreciated!

conbench/entities/benchmark_result.py
austin3dickey: Agreed. Let's make sure to
1. validate with the Arrow Go benchmark authors that it should be bytes, not bits
2. change [this line](https://github.com/apache/arrow/blob/819b7d5ce7febb4af9ea6f69947244f46403ce53/ci/scripts/go_bench_adapt.py#L98)
3. do a DB migration for Arrow Conbench
4. come back and clean this up

austin3dickey (commented):
Thanks!

conbench/entities/benchmark_result.py
alistaire47: Per [this line](https://github.com/apache/arrow/blob/819b7d5ce7febb4af9ea6f69947244f46403ce53/ci/scripts/go_bench_adapt.py#L86) I'm pretty confident `B/s` is correct. But yes, we should have `@zeroshade` review the PR.

conbench/tests/api/test_results.py
alistaire47: And `time_unit`

conbench/entities/benchmark_result.py
alistaire47: We should copy docs improvements to [benchadapt's result object](https://github.com/conbench/conbench/blob/6ab7ff096c2dc31a49fb9b0078935858c11a7c59/benchadapt/python/benchadapt/result.py#L34-L37) as well

Issue #1369, opened 2023-07-20
Present result-result comparison neutrally, 'baseline' & 'contender' is confusing


jgehrcke:
From https://github.com/conbench/conbench/pull/1336#discussion_r1245036329:

> I was also thinking that maybe what matters is "old" vs. "new".

> when comparing just two results I think I like when the presentation is almost neutral as in A vs. B without assigning specific properties / bias.

Antoine's feedback when looking at a result-result compare page:

> contender commit has an earlier date than baseline commit?

austin3dickey:
I agree that we can be more simple with this language in the UI, and probably API.

When it comes to alerts, are we still open to using "baseline" and "contender"? The alerting workflow is a special case where we can potentially be more precise.

alistaire47:
Yeah, I think in situations where you're getting a comparison via a baseline proposed by Conbench the existing language is good.

I even think we could keep the language entirely as long as we warn when the timestamps are backwards; this might occasionally be intentional, but it's usually by mistake. But I'm open to changing the terminology if it's helpful from a UX perspective

jgehrcke:
> are we still open to using "baseline" and "contender"? 

Oh, I feel like this is a great question! And maybe we want to change something about that in the future. For example, I like "reference distribution".

I want to clarify that in this ticket I was thinking exclusively about the UI view at `/compare/benchmark-results/a...b`; I think on that view we do not need the b-word and the c-word :).

austin3dickey:
>And maybe we want to change something about that in the future. For example, I like "reference distribution".

Yes, we'll probably want to re-examine the "baseline" word in the context of the lookback z-score analysis.

---

Thinking about language more, "old" vs "new" might be misleading in some cases. What if the "baseline" results are actually a re-run of benchmarks on an old commit, with a newer start timestamp than the "contender" start timestamp? Or, if "new" vs "old" refers to commit time instead of benchmark start time, what about git graphs where the "baseline" commit is newer than the "contender" commit because we're comparing to a distribution of results on a different branch?

jgehrcke:
> Thinking about language more, "old" vs "new" might be misleading in some cases.

:+1:  let's just make this as neutral as possible, not assigning special meaning to any of the two results.

The UI view basically shows two columns, and here we wonder about the column headings. We can simply have them be `result <ID1>` and `result <ID2>`, order as defined by the two IDs in the URL.

austin3dickey:
That makes sense to me!

[...]

PR #1375, opened 2023-07-21
UI: result-result comparison: more neutral comp, various tweaks

jgehrcke:
This is mainly for #1369, but also:

- show history plot conditionally
- show info/err msg when history plot doesn't show (see screenshot below)
- devprod: use a macro for generating entity heading (just a tiny start)
- more small tweaks, see commit msgs

Screenshot from 2023-07-21 12-40-27

The err msg needs tweaking, as it's now being used in a new context. That's fine, follow-up.

### DIFF
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -112,7 +112,8 @@ class TimeSeriesPlotMixin:
                     "you asked to highlight the benchmark result with ID "
                     f"<kbd>{highlight_other_result.bmrid}</kbd> in this view. "
                     "However, that result is not part of the history of the "
-                    f"result <kbd>{current_benchmark_result.id}</kbd>",
+                    f"result <kbd>{current_benchmark_result.id}</kbd>. The "
+                    "two results are not directly comparable.",
                 )
             highlight_other = (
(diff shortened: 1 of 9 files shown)


jgehrcke:
:ship: let's get this in; feedback much appreciated.

conbench/app/_plots.py
```diff
@@ -112,7 +112,8 @@ def get_history_plot(
                     "you asked to highlight the benchmark result with ID "
                     f"<kbd>{highlight_other_result.bmrid}</kbd> in this view. "
                     "However, that result is not part of the history of the "
-                    f"result <kbd>{current_benchmark_result.id}</kbd>",
+                    f"result <kbd>{current_benchmark_result.id}</kbd>. The "
+                    "two results are not directly comparable.",
```
alistaire47, line 116: I think this needs more nuance. 

There are reasons for making this kind of comparison, e.g. when Arrow devs want to compare a release candidate on a branch to a previous release candidate on another branch. They're not directly connected by git history so we can't do z-score analysis, but pairwise analysis might still be interesting.

I'm not sure what wording best captures the nuance, but maybe just specifying what specific analyses can and can't be done.

Issue #1347, opened 2023-07-07
UI: table showing results in run: data sorting not done by numerically


jgehrcke:
example url /runs/fd4398e69b4646febdc5bf0a92942fc0/

![image](https://github.com/conbench/conbench/assets/265630/8d4b910a-7dab-4969-a263-2d7334da4e47)

jgehrcke:
This column may contain a mixture of units, i.e. sorting by number only is not always leading to a desired outcome.

jgehrcke:
Taken care of for now via https://github.com/conbench/conbench/pull/1377.

PR #1377, opened 2023-07-21
UI: run: make measurement sortable

jgehrcke:
For issue #1347. This is unit-unaware sorting, optimizing for a frequent special case.

When there is more than one unit in this column then this current sorting method doesn't quite make sense. But without this pragmatic patch it never quite makes sense.

### DIFF
--- a/conbench/app/_util.py
+++ b/conbench/app/_util.py
@@ -70,6 +70,8 @@ def set_display_time(benchmark):
 
 def set_display_mean(benchmark):
     """
+    this needs to go away. :)
+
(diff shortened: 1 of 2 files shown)


conbench/app/_util.py
```diff
@@ -70,6 +70,8 @@ def set_display_time(benchmark):
 
 def set_display_mean(benchmark):
     """
+    this needs to go away. :)
+
```
jgehrcke, line 74: This patch is a great reminder that we should do https://github.com/conbench/conbench/issues/1341.

jgehrcke:
:ship: let's get this in

Issue #1370, opened 2023-07-20
history plot: ordinate axis range: do not show negative


jgehrcke:
Example for a waste of space: 

![image](https://github.com/conbench/conbench/assets/265630/cb33e766-655a-4b69-abc3-85f10629773d)

alistaire47:
Really we should clip the bounds at 0

jgehrcke:
Addressed in https://github.com/conbench/conbench/pull/1378.

PR #1378, opened 2023-07-21
hist plot: do not show negative y range

jgehrcke:
For issue #1370.

Before:

Screenshot from 2023-07-21 13-58-58

After:

Screenshot from 2023-07-21 14-05-59

### DIFF
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -887,6 +887,12 @@ def time_series_plot(
     p.legend.location = "top_left"
     p.legend.label_text_font_size = "12px"
 
+    # y range should not go into the negative, but it should also not
+    # always start at 0. Inspect `source_rolling_alert_min_over_time` which
+    # should conceptually hold the minimal value across all sources.
+    if min(source_rolling_alert_min_over_time.data["y"]) < 0:
+        p.y_range.start = 0
+
     # Change the number of expected/desired date x ticks. There is otherwise
     # only very few of them (like 4). Also see
     # https://github.com/bokeh/bokeh/issues/665 and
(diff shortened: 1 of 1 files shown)


conbench/app/_plots.py
```diff
@@ -887,6 +887,12 @@ def time_series_plot(
     p.legend.location = "top_left"
     p.legend.label_text_font_size = "12px"
 
+    # y range should not go into the negative, but it should also not
+    # always start at 0. Inspect `source_rolling_alert_min_over_time` which
+    # should conceptually hold the minimal value across all sources.
+    if min(source_rolling_alert_min_over_time.data["y"]) < 0:
+        p.y_range.start = 0
```
jgehrcke: creative solution with the help of
https://github.com/bokeh/bokeh/issues/8421
https://stackoverflow.com/a/61119600
https://discourse.bokeh.org/t/axis-range-cannot-be-changed-in-bokeh-dashboard-if-data-updated/10284/2

jgehrcke:
:ship:, let's get this in

alistaire47:
This is good, but we should still `np.clip()` the bounds to 0 since negative bounds are statistically invalid. And/or calculating bounds asymmetrically, since distributions will frequently be skew.

deepsihag:
In case of `more-is-better` would this still be true about not showing negative ?

alistaire47:
Those are rates (bytes/second or iterations/second), which can't be negative either. Though it arguably may not be sensible for the y-axis to go all the way to 0 for them.

deepsihag:
Ah ok.. make sense.  I missed looking at values and was just assuming the negative/lower side of mean but this is negative value.

Issue #959, opened 2023-03-28
towards systematic method testing: make it easy to download timeseries for scenario lib


jgehrcke:
For #644 we think about building up a library of scenarios that we want to test methods against. When we look at history plots we often think that 'oh, that's a lovely example'. Next to the plot there should be a download button, yielding a file that can subsequently be added to a collection of scenarios, for automated testing.

(mentioned this idea before, in meetings. Forgot to leave a trace in GitHub -- wanted to quickly do that now; more discussion to come)

alistaire47:
We should standardize this schema a bit more as we get towards this, but this is a nice starting point: https://github.com/conbench/conbench/pull/956/files#diff-187304752e5d15329df2ce11b69c3711b870bcea5192a9d106affb90da3a7004R162-R191 (Linking to the PR because I really want `BenchmarkResult.timestamp` in there)

PR #1383, opened 2023-07-25
HTTP API: add endpoint for downloading a timeseries as CSV file

jgehrcke:
Some meat towards https://github.com/conbench/conbench/issues/959 -- not yet exposed via UI (button or such). There are many choices to make, this type of task raises interesting questions

This here is to get going. Tiny demo, save as `plothist.py`:

```python

import sys

import matplotlib.pyplot as plt
import pandas as pd

plt.style.use("ggplot")
csv_url = sys.argv[1]

df = pd.read_csv(csv_url, comment="#", index_col="commit_time", parse_dates=True)
df.plot()
plt.tight_layout()
plt.show()
```
invocation (note the URL):
```
$ python conbench/plothist.py http://127.0.0.1:5000/api/history/download/0645e82e9bbe7b1b80004c6f2a731cba
...
```

rendering

image


conbench/entities/case.py
```diff
@@ -13,7 +14,7 @@ class Case(Base, EntityMixin["Case"]):
     # The name of the conceptual benchmark (store on BenchmarkResult directly)?
     name: Mapped[str] = NotNull(s.Text)
 
-    # Note(JP): we should work towards guaranteeing str->str mapping
+    # Note(JP): we must work towards guaranteeing a flat str->str mapping
```
jgehrcke, line 17: code relies on that.

conbench/entities/history.py
```diff
@@ -117,6 +126,12 @@ def _dict_for_api_json(self) -> dict:
         # Rename SVS for clarity for external consumption.
         d["single_value_summary"] = d.pop("svs")
         d["single_value_summary_type"] = d.pop("svs_type")
+
+        # Remove new props for now (needs test suite adjustment, and so far
```
jgehrcke, line 130: This is for the external (JSON) view to not change for now.

conbench/tests/api/test_history.py
```diff
@@ -26,3 +31,20 @@ def test_get_history(self, client):
...
+        response = client.get(f"/api/history/download/{benchmark_result.id}/")
+        assert response.status_code == 200
+        assert response.headers["Content-Type"] == "text/csv; charset=utf-8"
+        assert "filename=" in response.headers["Content-Disposition"]
+
+        df = pd.read_csv(
+            StringIO(response.text),
+            comment="#",
+            index_col="commit_time",
+            parse_dates=True,
```
jgehrcke, line 47: Important part of the test; we want to use these artifact for building up a library of scenarios; and I think we want our actual data analysis as well as testing that analysis center around pandas.

conbench/entities/_entity.py
```diff
@@ -134,8 +134,8 @@ def get(cls, _id):
     def one(cls, **kwargs):
         try:
             return current_session.query(cls).filter_by(**kwargs).one()
-        except NoResultFound:
-            raise NotFound()
+        except NoResultFound as exc:
+            raise NotFound from exc
```
jgehrcke, line 138: unrelated, but this seemed useful to not swallow original SQA detail

conbench/api/history.py
```diff
@@ -42,10 +53,146 @@ def get(self, benchmark_result_id):
...
+
+
 history_entity_view = HistoryEntityAPI.as_view("history")
+history_download_endpoint = HistoryDownloadAPI.as_view("history-download")
+
+rule(
+    "/history/download/<benchmark_result_id>/",
+    view_func=history_download_endpoint,
+    methods=["GET"],
+)
```
alistaire47, line 110: Would it be useful to structure this as a query param, e.g. `http://...?format=csv`?

If csv and json contain different data, the answer may be "no", but that in turn raises the question of why they're different.

conbench/api/history.py
```diff
@@ -42,10 +53,146 @@ def get(self, benchmark_result_id):
...
+    # and then we can make analysis on these dataframes (easy-ish) testable.
+    df = pd.DataFrame(
+        # Note(jp:): cannot use a generator expression here, len needs
+        # to be known.
+        {
+            "result_id": [i.benchmark_result_id for i in items],
+            "commit_hash": [i.commit_hash for i in items],
+            "svs": [i.svs for i in items],
+            "min": [min(i.data) for i in items],
+        },
```
alistaire47, line 148: I suspect we may wan to vary what goes in here over time, e.g. standard deviation, booleans for outliers/shifts, etc. I'm not suggesting we should chuck in everything now, but I do think we should mark this as experimental and liable to change.

conbench/api/history.py
```diff
@@ -42,10 +53,146 @@ def get(self, benchmark_result_id):
...
+        # `unit="s"` is the critical ingredient to convert this list of
+        # floaty unix timestamps to datetime representation. `utc=True`
+        # is required to localize the pandas DateTimeIndex to UTC
+        # (input is tz-naive).
+        index=pd.to_datetime([i.commit_timestamp for i in items], utc=True),
+    )
+
+    # Sort by time. old -> new
+    df = df.sort_index()
+    df.index.rename("commit_time", inplace=True)
```
alistaire47, line 159: Validating my reading: In the CSV, this results in the index looking just like any other column, correct? Hopefully?

conbench/types.py
```diff
@@ -0,0 +1,11 @@
+# Do this for case ID, etc, too.
+
+from typing import NewType, Tuple
+
+TBenchmarkName = NewType("TBenchmarkName", str)
```
alistaire47, line 5: `T` = "type"?

conbench/types.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+# This type is used often. It's the famous 4-tuple defining a time series:
+# benchmark name, hardware ID, context ID, case ID. Maybe turn this into a
+# namedtuple or sth like this. Watch out a bit for mem consumption.
+Tt4 = Tuple[TBenchmarkName, str, str, str]
```
alistaire47, line 11: This name is inscrutable to me. Maybe we need a better-defined name for the concept itself.

conbench/types.py
jgehrcke: Yes. This might be the start of introducing a convention (already did that before around job.py).

conbench/types.py
jgehrcke: > This name is inscrutable to me

Yes. Slowly converging to something that's hopefully useful. 

This type (well, the concept) is going to be so important, touching so many parts of the code base. 

It will pop up in other type definitions quite a bit so I like to have a short, recognizable term. 

Example:
```
TDict4tdf = Dict[Tt4, pd.DataFrame]
TDict4tlist = Dict[Tt4, List[BMRTBenchmarkResult]]
```

Totally work in progress.

https://github.com/conbench/conbench/blob/edeaf66832ca2eaa8446d79c3a8dac6b3c3d368d/conbench/job.py#L118

> It's the famous 4-tuple

Maybe in the future this is going to be a 5-tuple:

```
Tuple[TRepoUrl, TBenchmarkName, TCaseID, TContextID, THardwareHash]
```

Again, still various rough edges but we can start passing timeseries objects around that are self-documenting and hopefully easy to work with, sth in the lines of

```
class ResultTimeseries:
    t4: Tt4
    benchmark_name: str
    df: pd.DataFrame
    bmrlist: List[BMRTBenchmarkResult]
```

---

here, in this PR I just moved the definition of `Tt4` out of job.py because it starts to be re-used (as anticipated).

conbench/api/history.py
jgehrcke: Exactly!

This is super intermediate (I trust we want to change a lot here over time, maybe even in this PR), but so far the head of the generated CSV document looks like
```
# original URL: http://127.0.0.1:5000/api/history/download/0645e82e9bbe7b1b80004c6f2a731cba
# generated by conbench, commit 1748f11628aac62f9213046dbbfa88f8212e467e
# generated at 2023-07-25T12:53:33+00:00
# for result 0645e82e9bbe7b1b80004c6f2a731cba
# benchmark name: tpch-snip
# case permutation: dataset_subvariant=default dataset_variant=parquet-float64 snip ptds=False query_id=snip-09-edited-snip scale_factor=1000 transport_protocol=snip
# hardware hash: snip-infra-748d68f991c92244cefcea963cf3f74f
# timeseries fingerprint: 47c516cec101661312f3fb65eaa90d7c
# single value summary (SVS) type: mean
commit_time,result_id,commit_hash,svs,min
2023-05-03 19:20:30+00:00,06452cc6a52579608000d427b742c4c1,7c4a4fb8f78c55cd072d330040f4e8c775bceaa9,48.666006,48.666006
2023-05-04 00:20:08+00:00,06453129610473c480002253378ce1e3,ade9fd3861cf20d34e4b498ef1c5f573a1a8e5eb,50.470924,50.470924
2023-05-04 02:25:54+00:00,064532fa0b3177968000a11fbb525820,a536476ca3d4943adcff54a08dc0522aab9148e4,50.121353,50.121353
```

conbench/api/history.py
jgehrcke: > I suspect we may wan to vary what goes in here over time

absolutely

> we should mark this as experimental and liable to change

absolutely, too. I mean. Everything we do is liable to change, honestly. But this here is an extreme version, yes. 

I have started thinking about treating (and documenting) file contents as implementation detail (and care about stability of corresponding parsing code).

In any case: this is for starters for us, mainly aimed at allowing us to build up a library of test cases. I also think this can be a more widely useful interface in the future. We will see.

conbench/api/history.py
jgehrcke: > Would it be useful to structure this as a query param, e.g. http://...?format=csv?

Maybe! I didn't want to overthink this yet, but I also thought about `/benchmark-results/<id>/histfile?format=<fmt>` where ideally we have a more structured container like parquet or HDF5.

> why they're different.

I think one good (enough) reason is that we can go crazy here and build something for us and not worry about breaking an interface for others.

We can then later zoom out and look at both and think about if and how we want to reconcile this with the current/existing history endpoint.

jgehrcke:
Going ahead with this :ship:

Issue #1382, opened 2023-07-25
app/api compare endpoint semaphore: return 429 when sem acq fails (immediately or within timeout?)


jgehrcke:
about https://github.com/conbench/conbench/blob/edeaf66832ca2eaa8446d79c3a8dac6b3c3d368d/conbench/api/compare.py#L320:

When acquiring the semaphore blocks: either immediately, or after little time return a 429, allowing the client to retry later. This seems to be valuable for those cases where acquisition would take so long that a general request timeout would be hit. That can also result in retry, but it kind of hides the reason. Instead, it would be better to expose the reason (too much concurrent load) via 429, and let the client retry on that. A 429 should be retried with back-off, and therefore it's fine to immediately return 429 (instead of introducing another time constant).

jgehrcke:
relates to #1289 #1288

jgehrcke:
![Screenshot from 2023-08-01 14-07-42](https://github.com/conbench/conbench/assets/265630/f5eb0940-88ee-40c6-8346-5f6a165887ab)

jgehrcke:
'done' for now

PR #1386, opened 2023-07-26
http api: compare endpoints: emit 429 when semaphore blocked

jgehrcke:
This is for #1382.

Picked a really small timeout constant, could have done `(block=False)` instead -- but using the timeout constant I think is a tiny more elegant, because it can trivially be bumped -- maybe in the future we find that instead of forcing clients to go through an active retry cycle we _want_ to keep them waiting in line for a few seconds.

I manually tested that this works; I did not yet bother creating a test case for this (as it does require more or less precise concurrency control; we can easily do this in the future).

### DIFF
--- a/benchclients/python/benchclients/http.py
+++ b/benchclients/python/benchclients/http.py
@@ -398,8 +398,7 @@ class RetryingHTTPClient(ABC):
         status code alone?
         """
         if code == 429:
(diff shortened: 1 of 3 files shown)


jgehrcke:
:ship: Feedback/review appreciated!

Issue #1379, opened 2023-07-24
benchadapt: ArcheryAdapter is reporting suite name for benchmark name


alistaire47:
`ArcheryAdapter` is currently substituting suite name for benchmark name, e.g. https://conbench.ursa.dev/benchmark-results/064be78e31f67168800063884199004a/ 

<img width="662" alt="image" src="https://github.com/conbench/conbench/assets/1693477/e3f55652-3b78-4863-bc7f-5861011e593c">

and the benchmark name is entirely missing.

After fixing this, we'll need to bump the benchadapt version for voltrondata-labs/benchmarks to get it to pick up the change.

austin3dickey:
awesome: this is the root cause of https://github.com/conbench/conbench/issues/1350#issuecomment-1632264749

alistaire47:
Update: I discovered the issue is not actually in `ArcheryAdapter`—the data being sent is correct:

<details>
  <summary>Example result</summary>

From https://conbench.ursa.dev/api/benchmark-results/064c29de44da729a8000da6354e851fa/

``` json
{
  "batch_id": "c6b3f53b5e924822b9983f28097cdb6d",
  "change_annotations": {},
  "error": null,
  "id": "064c29de44da729a8000da6354e851fa",
  "links": {
    "context": "http://conbench.ursa.dev/api/contexts/ee186706d53446a9a8c4dff18cd251c6/",
    "info": "http://conbench.ursa.dev/api/info/064adb5d5a91755a800058ccdbc7d4fd/",
    "list": "http://conbench.ursa.dev/api/benchmarks/",
    "run": "http://conbench.ursa.dev/api/runs/e57e82d6a8c6464a8a43abb70e0020bc/",
    "self": "http://conbench.ursa.dev/api/benchmarks/064c29de44da729a8000da6354e851fa/"
  },
  "optional_benchmark_info": null,
  "run_id": "e57e82d6a8c6464a8a43abb70e0020bc",
  "stats": {
    "data": [
      155600496.22750118
    ],
    "iqr": null,
    "iterations": 1,
    "max": null,
    "mean": 155600496.22750118,
    "median": null,
    "min": null,
    "q1": null,
    "q3": null,
    "stdev": null,
    "time_unit": "us",
    "times": [
      6739.396261624419
    ],
    "unit": "i/s"
  },
  "tags": {
    "name": "MinMaxDoublesGroupedByMediumInt",
    "params": "1048576/1",
    "source": "cpp-micro",
[...]

Issue #1350, opened 2023-07-07
benchalerts: consider using tables instead of bulleted lists


austin3dickey:
Thanks to @pitrou for the feedback and screenshots. We could use tables instead of bulleted lists when posting PR comments and GitHub Checks.

Before:
![conbench](https://github.com/conbench/conbench/assets/16600275/d98a23a3-d73b-40f3-95a9-7cfd92501c40)

After:
![conbench-table](https://github.com/conbench/conbench/assets/16600275/81c06e43-db2a-4ad7-be12-6a7437b78e2f)

alistaire47:
We could go further here and start with the run info on the left, which would be pretty

jgehrcke:
Adding one point: in his Slack msg Antoine rightfully pointed out that the benchmark name isn't presented clearly.

> We could go further here and start with the run info on the left, which would be pretty

> We could use tables instead of bulleted lists when posting PR comments and GitHub Checks

or make the PR comment as minimal as possible, and focus on making _one_ presentation nice, that we can link to :-) (my fav).

austin3dickey:
>in his Slack msg Antoine rightfully pointed out that the benchmark name isn't presented clearly.

Root cause: #1379

Issue #942, opened 2023-03-24
hacks.py: slowly separate signal from noise


jgehrcke:
[hacks.py](https://github.com/conbench/conbench/blob/9237373f99b2aa04906496e0393eabc0ec1ecd45/conbench/hacks.py) contains important logic, but it also contains logic that I think we don't need and want. It's hard to separate signal from noise. Code paths are prepared to deal with various kinds of objects.

So far, this module was pretty much untouchable. Overcome this. Add comments. Add type annotations. Remove logic that we don't think we need. Simplify. After all, remove the module.

PR #1393, opened 2023-07-27
UI: Never use `tags.suite` for benchmark display name

### LINKED ISSUE #1392: UI: Don't prefer `tags.suite` to `tags.name` for benchmark display name
Remove https://github.com/conbench/conbench/blob/038977d8153510c7fef4d6f656229764b809304c/conbench/hacks.py#L93-L95

so `tags.name` is always used for the benchmark display name in the UI even if `tags.suite` is populated. 

Motivation: At present, this is causing Arrow microbenchmarks to show useless suite names for benchmark names, making navigation hard. Example: https://conbench.ursa.dev/benchmark-results/064c29de44da729a8000da6354e851fa/

Full history: https://github.com/conbench/conbench/issues/1379#issuecomment-1654440938

alistaire47:
Closes #1392 so the benchmark display name in the UI is always `tags.name`.

### DIFF
--- a/conbench/hacks.py
+++ b/conbench/hacks.py
@@ -90,13 +90,10 @@ def set_display_benchmark_name(bmresult):
     tags = bmresult["tags"] if is_api else bmresult.case.tags
     name = tags.get("name") if is_api else bmresult.case.name
 
-    # If `suite` is a key in the tags, then use its value for `name`, ignore
-    # name.
-    bmname = tags.get("suite", name)
     if is_api:
-        bmresult["display_bmname"] = bmname
+        bmresult["display_bmname"] = name
     else:
-        bmresult.display_bmname = bmname
+        bmresult.display_bmname = name
 
 
 def sorted_data(benchmarks):


jgehrcke (approved):
Yes, please.

This approach needs to be further phased out anyway, as part of:
https://github.com/conbench/conbench/issues/1394
https://github.com/conbench/conbench/issues/942

austin3dickey (approved):
My question was: does this actually solve https://github.com/conbench/conbench/issues/1350#issuecomment-1632264749?

No, it doesn't! I was wrong in https://github.com/conbench/conbench/issues/1379#issuecomment-1648226911 that #1379 was the root cause of benchmark names missing in `benchalerts` notifications. We blamed #1379 (which wasn't actually a problem) for two other completely separate problems. 🙃 Thanks for fixing one of them in this PR.

I'll make a separate issue for the benchmark names missing in notifications.

jgehrcke:
> No, it doesn't! 

OK. Thank you!
Merging this, as it's at least in my limited view a simplification and therefore hopefully an improvement.

Issue #1380, opened 2023-07-24
UI: display units for tables (big number treatment)


alistaire47:
Feedback:

> vary the result unit depending on the result value ("1772000000.0 B/s" is less readable than either "1.72 GB/s" or "1772 MB/s").

Example: https://conbench.ursa.dev/compare/runs/dd1cbc13065b4e44a998a12aca4cb108...2e63f05bf950435cbc8425fb57304a57/

<img width="451" alt="image" src="https://github.com/conbench/conbench/assets/1693477/acfeebb8-9744-4374-af11-9e2d4690478c">

We just standardized input units, but display units for tables could vary from those to make them more human-readable. We should try not to make our sorting issue worse in the process, though.

jgehrcke:
Yes of course. Indeed, in my last iteration I was prioritizing sortability and correctness. Also: https://github.com/conbench/conbench/issues/1333

One decision we need to do: scientific notation (keeping the basic unit) or putting the order of magnitude into the unit? I think I like the former more in our context, i.e. sth like `1.7e9 B/s`.

More work on this should be enabled by https://github.com/conbench/conbench/issues/1334 where I write 

> I see this as a requirement for better number formatting: https://github.com/conbench/conbench/issues/1333

This is a bit of a philosophical stance; we need to stop putting patch onto patch onto patch in the 'augment' family of functions (if you have worked on those you know what I mean), but get rid of those first and then generate the outcome we want.

jgehrcke:
If we choose the scientific notation path then I think we may want to use https://numpy.org/doc/stable/reference/generated/numpy.format_float_scientific.html#numpy-format-float-scientific and not `sigfig`. Reminder: sigfig is really cool for producing a string representing uncertainty (e.g., `(5.0 +/- 1.2) m/s`), but it's also slow as hell and cannot/should not be used for live-rendering hundreds of numbers (also see https://github.com/conbench/conbench/issues/1218 and https://github.com/conbench/conbench/pull/1245).

jgehrcke:
Tiny note: natural limitation: we cannot assume a single/definite unit or order of magnitude across all values reported within one column. That is, certain brain cycles are always required when quickly browsing numbers and trying to compare them.

Issue #1334, opened 2023-06-26
compare UI: interpolate template with first-class BenchmarkResult objects (not dict)


jgehrcke:
This will be quite a bit of work, but is absolutely required to make code maintainable / changable.

I see this as a requirement for better number formatting: https://github.com/conbench/conbench/issues/1333

PR #1396, opened 2023-07-28
UI: tables: dynamically switch to scientific notation, better spacing

jgehrcke:
This is work towards https://github.com/conbench/conbench/issues/1380. See commit messages!

This does not yet help directly for #1380 because we need to do #1334 first.

It's noteworthy that within a column this may still result in tiny inconsistencies, fundamentally. Nothing we can/want to do against this. That is because we format each number independently, instead of choosing constant formatting for all numbers in a column.

We have to do this because within one of these colums there might be results from a variety of different benchmark methods, i.e. we cannot assume a certain unit or order of magnitude.

I just wanted to have clarified that maximum numeric/formatting/unit homogeneity across a column in the UI is not the goal here, i.e. some comparisons may require really looking at number and unit).

### DIFF
--- a/conbench/numstr.py
+++ b/conbench/numstr.py
@@ -52,9 +52,55 @@ import numpy as np
 # import sigfig
 
 
+def numstr_dyn(v: Union[float, int]) -> str:
+    """
+    Turn number into string. Limit the count of significant figures.
+    Dynamically switch to scientific notation for large numbers (larger 10**6)
+    and small numbers (smaller 10**-6).
+
+    Regarding precision: this is so far meant to be used in
+    overview/summary/tabular representation where the 'rough' measurement
+    result should be presented in a canonical/non-cryptical way. Higher
+    precision / raw data is available to the user elsewhere. precision/sigfigs
+    is no argument to this function at this point to hopefully achieve
+    consistency across UI.
+
+    >>> numstr.numstr_dyn(1.290823987290392)
+    '1.29082'
+    >>> numstr.numstr_dyn(1239127)
+    '1.24e+06'
+    >>> numstr.numstr_dyn(0.000000000023823987290392)
+    '2.38e-11'
+    >>> numstr.numstr_dyn(123912)
+    '123912'
+
+    """
+    if 10**6 > v > 10**-6:
+        return numstr(v, 6)
+    return numstr_exp(v, 2)
+
+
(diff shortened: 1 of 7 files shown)


conbench/numstr.py
```diff
@@ -52,9 +52,55 @@
...
+    consistency across UI.
+
+    >>> numstr.numstr_dyn(1.290823987290392)
+    '1.29082'
+    >>> numstr.numstr_dyn(1239127)
+    '1.24e+06'
+    >>> numstr.numstr_dyn(0.000000000023823987290392)
+    '2.38e-11'
+    >>> numstr.numstr_dyn(123912)
+    '123912'
```
austin3dickey, line 75: Lovely examples!

austin3dickey (approved):
nice!

conbench/numstr.py
```diff
@@ -52,9 +52,55 @@
...
+    overview/summary/tabular representation where the 'rough' measurement
+    result should be presented in a canonical/non-cryptical way. Higher
+    precision / raw data is available to the user elsewhere. precision/sigfigs
+    is no argument to this function at this point to hopefully achieve
+    consistency across UI.
+
+    >>> numstr.numstr_dyn(1.290823987290392)
+    '1.29082'
+    >>> numstr.numstr_dyn(1239127)
+    '1.24e+06'
```
jgehrcke, line 71: note: since generating examples I have changed `exp_digits=1` (see below), i.e. this here would show `e+6` instead of `+06`. I did this to save horizontal space. Without losing horizontal alignment: the numbers we deal with should rarely exceed  9*10^9 and rarely be smaller than 10^-9.

conbench/numstr.py
```diff
@@ -52,9 +52,55 @@
...
+
+    >>> numstr.numstr_sn(123902901380912893, 2)
+    '1.24e+17'
+
+    The example shows that these may be considered three significant digits.
+
+    Trim "-" is documented with "trim trailing zeros and any trailing decimal
+    point"
+    """
+    return np.format_float_scientific(v, precision=sigfigs, trim="-", exp_digits=1)
```
jgehrcke, line 98: we could also set `sign=false` so that this generates `e6` instead of `e+6` but that goes at the cost of horizontal alignment.

PR #1397, opened 2023-07-31
history/compare APIs: use new less_is_better source, introduce TUnit type, tweaks/fixes

jgehrcke:
Use new source of truth for `less_is_better` in analyses.

This is part of https://github.com/conbench/conbench/issues/1335 and a follow-up after https://github.com/conbench/conbench/pull/1373.

Please see commit messages. This changes a small part of the compare JSON HTTP API.

For PFE-1741.

### DIFF
--- a/conbench/api/compare.py
+++ b/conbench/api/compare.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING, Dict, List, Optional, Tuple
 import flask as f
 import sqlalchemy as s
 
+import conbench.units
 from conbench.dbsession import current_session
 from conbench.numstr import numstr
 
@@ -15,7 +16,7 @@ from ..api import rule
 from ..api._endpoint import ApiEndpoint, maybe_login_required
 from ..api._resp import resp429
 from ..entities.benchmark_result import BenchmarkResult
-from ..entities.history import _less_is_better, set_z_scores
+from ..entities.history import set_z_scores
 from ..hacks import set_display_benchmark_name, set_display_case_permutation
 
 log = logging.getLogger(__name__)
@@ -135,18 +136,36 @@ class BenchmarkResultComparator:
         threshold: Optional[float],
         threshold_z: Optional[float],
     ) -> None:
-        if (
-            baseline
-            and baseline.unit
-            and contender
-            and contender.unit
-            and baseline.unit != contender.unit
-        ):
-            raise UnmatchingUnitsError(
-                "Benchmark units do not match. Benchmark result with ID "
-                f"'{baseline.id}' has unit '{baseline.unit}' and benchmark result "
-                f"with ID '{contender.id}' has unit '{contender.unit}'."
-            )
+        # What do we know here? Is one of baseline and contender guaranteed
+        # to not be None?
+        assert any([baseline, contender])
+
+        # Can/should we assume that only non-errored results are added into a
+        # comparator? Then we can assume that the unit is set. Turns out: no,
+        # either might be errored. Thinking about possible code paths. Can
+        # `baseline` for example be an errored benchmark result while contender
+        # is `None`?
+
+        # `self.unit` is `None` for all cases except for when both contender
+        # and baseline are non-failed results and have the same unit; then it's
+        # that unit.
+        self.unit: Optional[conbench.units.TUnit] = None
+
+        # Verify for the case where both are non-failed results that they have
+        # equal units
+        if baseline and contender:
+            if not any([baseline.is_failed, contender.is_failed]):
+                assert baseline.unit
+                assert contender.unit
+                if baseline.unit != contender.unit:
+                    raise UnmatchingUnitsError(
+                        "Benchmark units do not match. Benchmark result with ID "
+                        f"'{baseline.id}' has unit '{baseline.unit}' and benchmark result "
+                        f"with ID '{contender.id}' has unit '{contender.unit}'."
+                    )
+                # Take `baseline.unitsymbol` here to get the TUnit type.
+                assert baseline.unitsymbol
+                self.unit = baseline.unitsymbol
 
         self.baseline = baseline
         self.contender = contender
@@ -160,16 +179,13 @@ class BenchmarkResultComparator:
         )
 
     @property
-    def unit(self) -> str:
-        if self.baseline and self.baseline.unit:
-            return self.baseline.unit
-        if self.contender and self.contender.unit:
-            return self.contender.unit
-        return "unknown"
-
-    @property
-    def less_is_better(self) -> bool:
-        return _less_is_better(self.unit)
+    def less_is_better(self) -> Optional[bool]:
+        """
+        This is only defined when this comparison has a unit.
+        """
+        if self.unit is None:
+            return None
+        return conbench.units.less_is_better(self.unit)
 
     @staticmethod
     def result_info(result: Optional["AugmentedBenchmarkResult"]) -> Optional[dict]:
@@ -242,6 +258,9 @@ class BenchmarkResultComparator:
     @property
     def _dict_for_api_json(self) -> dict:
         return {
+            # How is 'unit' here specified? Let's specify it: If both results
+            # are not failed and have the same unit then this here is the unit
+            # symbol. Else it's null/None.
             "unit": self.unit,
             "less_is_better": self.less_is_better,
             "baseline": self.result_info(self.baseline),
--- a/conbench/api/compare.py
+++ b/conbench/api/compare.py
@@ -138,7 +138,11 @@ class BenchmarkResultComparator:
     ) -> None:
         # What do we know here? Is one of baseline and contender guaranteed
         # to not be None?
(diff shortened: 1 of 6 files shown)


conbench/api/compare.py
```diff
@@ -135,18 +136,36 @@ def __init__(
...
-            and baseline.unit != contender.unit
-        ):
-            raise UnmatchingUnitsError(
-                "Benchmark units do not match. Benchmark result with ID "
-                f"'{baseline.id}' has unit '{baseline.unit}' and benchmark result "
-                f"with ID '{contender.id}' has unit '{contender.unit}'."
-            )
+        # What do we know here? Is one of baseline and contender guaranteed
+        # to not be None?
+        assert any([baseline, contender])
```
jgehrcke: I encoded this assumption now here thinking that it does not make sense to compared two errored results.  Test suite passes. If in practice in fact we create a `BenchmarkResultComparator` for two errored results then I'd argue maybe we just don't want to do that.

What do we think @austin3dickey?

conbench/api/compare.py
```diff
@@ -135,18 +136,36 @@ def __init__(
...
+
+        # Can/should we assume that only non-errored results are added into a
+        # comparator? Then we can assume that the unit is set. Turns out: no,
+        # either might be errored. Thinking about possible code paths. Can
+        # `baseline` for example be an errored benchmark result while contender
+        # is `None`?
```
jgehrcke, line 151: Another interesting one. The combinatorics of this constructor are a little hard to see through.

conbench/api/compare.py
```diff
@@ -135,18 +136,36 @@ def __init__(
...
+
+        # `self.unit` is `None` for all cases except for when both contender
+        # and baseline are non-failed results and have the same unit; then it's
+        # that unit.
```
jgehrcke, line 161: This is specification I added here, I like the simplicity of that.

I couldn't quite see clearly how this was specified before.

conbench/api/compare.py
```diff
@@ -160,16 +179,13 @@ def __init__(
...
-        if self.contender and self.contender.unit:
-            return self.contender.unit
-        return "unknown"
-
-    @property
-    def less_is_better(self) -> bool:
-        return _less_is_better(self.unit)
+    def less_is_better(self) -> Optional[bool]:
+        """
+        This is only defined when this comparison has a unit.
```
jgehrcke, line 199: I also added this specification here because I think it makes sense.

conbench/entities/benchmark_result.py
```diff
@@ -91,15 +91,13 @@ class BenchmarkResult(Base, EntityMixin):
...
-    # "more is better"? I think we derive this from the unit, which is
-    # fundamentally error-prone. Should be a user-given property.
-    unit: Mapped[str] = Nullable(s.Text)
+    # Note(JP): `unit` can I think be null/none for errored benchmark results
+    # (where no Stats structure was provided). When a stats structure was
+    # provided then it is a string. Legacy DB might contain empty strings, but
+    # now we have validation in the insert path that this is one of the allowed
+    # unit symbol strings. Related:
+    # https://github.com/conbench/conbench/issues/1335
+    unit: Mapped[Optional[str]] = Nullable(s.Text)
```
jgehrcke, line 100: Important type update

conbench/entities/benchmark_result.py
```diff
@@ -363,15 +361,18 @@ def is_failed(self):
...
-        if do_iteration_samples_look_like_error(self.data):
+        if self.data is None:
             return True
 
         if self.error is not None:
             return True
 
+        if do_iteration_samples_look_like_error(self.data):
+            return True
+
```
jgehrcke, line 375: changes in this section: least expensive checks first, and now also use absence of unit as criterion for "failed result".

conbench/entities/benchmark_result.py
```diff
@@ -521,6 +522,17 @@ def ui_commit_url_anchor(self) -> str:
...
+    @functools.cached_property
+    def unitsymbol(self) -> Optional[conbench.units.TUnit]:
+        """Return unit symbol or None if result indicates failure."""
+        if self.is_failed:
+            return None
+
+        # Not None, and not an empty string.
+        assert self.unit
+
+        return conbench.units.legacy_convert(self.unit)
```
jgehrcke, line 534: Let's try to use this in the future more than using `self.unit` directly, or make `self.unit` have a more precise type.

conbench/entities/history.py
```diff
@@ -736,18 +737,6 @@ def _add_rolling_stats_columns_to_df(
...
-def _less_is_better(unit) -> bool:
-    # For some units, "more is better":
-    if unit in ["B/s", "i/s"]:
-        return False
-
-    # Legacy decision was to simply return `True` here -- assuming
-    # duration/seconds. However, this is error-prone and may lead to rather
-    # ugly UX in the core value proposition of Conbench. This needs a better
-    # solution. Also see https://github.com/conbench/conbench/issues/1335
-    return True
```
jgehrcke, line 748: The goal of this PR was mainly to remove this function.

conbench/api/compare.py
austin3dickey: In short: it is a good assumption that at least one of these will not be None. But the optionality of each benchmark result here has nothing to do with whether it has an error or not. 

When comparing two runs, we need to join pairs of benchmark results that make sense. We decided to do a full outer join on the "timeseries fingerprint" (i.e. case/context/hardware/repo). There might be some fingerprints where the baseline run has a fingerprint that is not present in the contender run, or vice versa. In those cases, when we construct the `BenchmarkResultComparator` object, the missing result will be None. But a full outer join cannot result in a pair of benchmark results where _both_ results are None. So this line of code is good.

To clarify, an errored benchmark result passed to either `baseline` or `contender` here will not be `None` -- it will be a full-fledged benchmark result where the `.error` attribute is not None. In most cases, that will also mean the `.unit` attribute is None (but not always; e.g. [the case](https://github.com/conbench/conbench/blob/fcb7e451c807a4c113cfcc9adc9bc4112ba5390e/conbench/tests/api/_fixtures.py#L131) where one iteration failed).

conbench/api/compare.py
austin3dickey: I think this is still a bit too restrictive and will cause visual confusion if we deploy. As I mentioned above, it's possible to submit a result with data and an error, and we have many examples in our current databases of that being the case. So I think the statement should be

>`self.unit` is `None` for all cases except for when both contender and baseline ~are non-failed results and~ have the same unit; then it's that unit.

conbench/api/compare.py
```diff
@@ -135,18 +136,36 @@ def __init__(
...
+
+        # `self.unit` is `None` for all cases except for when both contender
+        # and baseline are non-failed results and have the same unit; then it's
+        # that unit.
+        self.unit: Optional[conbench.units.TUnit] = None
+
+        # Verify for the case where both are non-failed results that they have
+        # equal units
+        if baseline and contender:
+            if not any([baseline.is_failed, contender.is_failed]):
```
austin3dickey: Yeah, so here, it would be `if baseline.unit and contender.unit:`

conbench/api/compare.py
austin3dickey: Thank you!

conbench/entities/benchmark_result.py
austin3dickey: lovely!

conbench/api/compare.py
austin3dickey: Ah, I had forgotten that this is what we settled on:
https://github.com/conbench/conbench/blob/fcb7e451c807a4c113cfcc9adc9bc4112ba5390e/conbench/entities/benchmark_result.py#L966-L968
So you're right that if either `baseline` or `contender` is failed, we shouldn't do any further analysis.

conbench/api/compare.py
austin3dickey: Never mind (see my [comment](https://github.com/conbench/conbench/pull/1397/files#r1279483073) above)

conbench/api/compare.py
austin3dickey: Do you want to go even further with this assumption in this PR? Below we have
https://github.com/conbench/conbench/blob/ed875a0f39e547e1f90f71d0cc5c49559e2a152e/conbench/api/compare.py#L211
We could get rid of that and give None if there's an error.

austin3dickey (approved):

conbench/api/compare.py
jgehrcke: > We could get rid of that and give None if there's an error.

Nice, yes. Adding a commit! I think let's do the relevant validity checks in the constructor, and then in the more mathy methods we can inspect the outcome of that validation.

conbench/api/compare.py
jgehrcke: > But a full outer join cannot result in a pair of benchmark results where both results are None. So this line of code is good.

:heart:

PR #1401, opened 2023-08-01
ci: minikube: bump k8s: 1.23 -> 1.26

jgehrcke:
We pinned this old version so far because some prod deployments were on old k8s. That has changed by now. That's good; kube-prometheus plans for more modern k8s.

https://github.com/prometheus-operator/kube-prometheus#compatibility

We currently use (a pinned commit on) the main branch of kube-prometheus, and should be running 1.26 or 1.27:

image


jgehrcke:
Most importantly, `ci / cb-on-minikube` succeeded. Nice. Getting this in.

PR #1404, opened 2023-08-01
benchalerts: add benchmark name/language to notifications

### LINKED ISSUE #1395: benchalerts: display benchmark name and language in notifications
Currently, to identify a benchmark result in `benchalerts` notifications, we use the `case_permutation` value from the compare endpoint's `contender` payload:

https://github.com/conbench/conbench/blob/038977d8153510c7fef4d6f656229764b809304c/benchalerts/benchalerts/conbench_dataclasses.py#L65

That payload also has `benchmark_name` and `language` keys. User feedback: these should be displayed as well.

`benchmark_name` is of course incredibly important, and a big reason we missed it is that before #1114 we didn't have good access to it. IMO it boils down to #942.

austin3dickey:
Fixes https://github.com/conbench/conbench/issues/1395.

This clarifies notifications by adding the benchmark name and language instead of just the case permutation parameters.

### DIFF
--- a/benchalerts/benchalerts/conbench_dataclasses.py
+++ b/benchalerts/benchalerts/conbench_dataclasses.py
@@ -10,7 +10,7 @@ from benchclients.logging import log
 class BenchmarkResultInfo:
     """Track and organize specific info about one BenchmarkResult."""
 
-    name: str
+    display_name: str
     link: str
     has_error: bool
     has_z_regression: bool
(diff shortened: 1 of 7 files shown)


benchalerts/benchalerts/conbench_dataclasses.py
```diff
@@ -53,6 +53,20 @@ class RunComparisonInfo:
...
+
+        Given: the compare endpoint's payload's "contender" key dict:
+        https://conbench.ursa.dev/api/redoc#tag/Comparisons/paths/~1api~1compare~1benchmark-results~1%7Bcompare_ids%7D~1/get
+        """
+        display_name = f'`{result["benchmark_name"]}`'
+        if result["language"] and result["language"] != "unknown":
+            display_name += f' ({result["language"]})'
+        if result["case_permutation"] != "no-permutations":
+            display_name += f' with {result["case_permutation"]}'
+        return display_name
```
jgehrcke, line 68: :+1:  I am sure that in the future there will be more work towards displaying the "main identity" of a benchmark result via human-oriented text even more consistently across the product surface (including docs/glossary).

austin3dickey: Yes!!

PR #1408, opened 2023-08-02
grafana dash: add 'cluster' and 'velox' support

jgehrcke:
Prepare this dashboard for a Grafana instance that has access to Conbench metrics from absolutely independent deployments. These independent deployments then are distinguishable with the `cluster` metric label. Note that this dashboard also needs to work for that label _not_ being set. That's why I fell back to regex matching and the special value `.*`, same as I used for the `namespace` label before.

from commit msg:
```
"Ad-hoc variables" are a Grafana concept that
seemingly allows for addding a label filter
to _all_ queries in a dashboard (applies to
all panels). This is the second time that I
failed making nice usage of that concept.

I fell back to using a custom variable with
hard-coded values, and then manually adjusting
the query expression of each query.
Well well.
```

### DIFF
--- a/k8s/kube-prometheus/conbench-grafana-dashboard.json
+++ b/k8s/kube-prometheus/conbench-grafana-dashboard.json
@@ -117,7 +117,7 @@
             "uid": "${datasource}"
           },
           "editorMode": "code",
-          "expr": "conbench_bmrt_cache_last_update_seconds{namespace=~\"$namespace\"} > 0",
+          "expr": "conbench_bmrt_cache_last_update_seconds{namespace=~\"$namespace\", cluster=~\"$cluster\"} > 0",
           "legendFormat": "{{instance}}",
           "range": true,
           "refId": "A"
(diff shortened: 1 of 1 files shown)


k8s/kube-prometheus/conbench-grafana-dashboard.json
```diff
@@ -1207,6 +1207,39 @@
...
+            "text": "ursa-2",
+            "value": "ursa-2"
+          },
+          {
+            "selected": false,
+            "text": ".*",
+            "value": ".*"
+          }
+        ],
+        "query": "vd-2, ursa-2, .*",
```
jgehrcke, line 1238: Ideally this can be made more dynamic. Looking up label values from the DB works (I used that method in a first iteration), but for such a variable one cannot inject a custom value (a value that is not part of a query result). There is a special value "all", but that doesn't include "not set". So, see above, I resorted to using a regex match and the special value .* includes "not set".

PR #1409, opened 2023-08-02
dashboard: saner defaults, new panel for /compare/runs response time distribution

jgehrcke:
Example screenshot of new panel, for Arrow Conbench:

Screenshot from 2023-08-02 14-54-41

### DIFF
--- a/k8s/kube-prometheus/conbench-grafana-dashboard.json
+++ b/k8s/kube-prometheus/conbench-grafana-dashboard.json
@@ -257,6 +257,7 @@
             "pointSize": 5,
             "scaleDistribution": {
               "log": 2,
+              "log": 10,
               "type": "log"
             },
             "showPoints": "auto",
--- a/k8s/kube-prometheus/conbench-grafana-dashboard.json
+++ b/k8s/kube-prometheus/conbench-grafana-dashboard.json
@@ -24,7 +24,7 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 11,
+  "id": 27,
   "links": [],
   "liveNow": false,
   "panels": [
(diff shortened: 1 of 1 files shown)


k8s/kube-prometheus/conbench-grafana-dashboard.json
```diff
@@ -1051,100 +1153,88 @@
...
       "targets": [
         {
           "datasource": {
             "type": "prometheus",
             "uid": "${datasource}"
           },
           "editorMode": "code",
-          "expr": "conbench_github_httpapi_quota_remaining{cluster=~\"$cluster\",namespace=~\"$namespace\"} > -1",
-          "legendFormat": "{{instance}}",
+          "expr": "sum by (le) (rate(http_request_duration_seconds_bucket{cluster=~\"$cluster\",namespace=~\"$namespace\",method=\"GET\",http_handler_name=\"api.compare-runs\",status!~\"3..\"}[$timewindow]))",
```
jgehrcke, line 1229: I simply used `http_handler_name=\"api.compare-runs\"` here for now. We also have `api.compare-benchmark-results`.

deepsihag: Should we add a panel for `api.compare-benchmark-results`  too?

jgehrcke: > Should we add a panel for api.compare-benchmark-results too?

Not saying 'no'. Yesterday I decided against this because
- I have added panels to this single dashboard conservatively as this is an 'overview' dashboard already having many panels.
- There is a panel for showing the response time distribution of all HTTP API GET requests, i.e. including those ` api.compare-benchmark-results` requests.
- `api.compare-runs` is the "worst case" of both if that makes sense :).

But let's totally build more dashboards in the future and go wild with adding panels.

Issue #1413, opened 2023-08-03
UI: result-result comparison: show hardware checksum


jgehrcke:
That's critical information, we're not displaying it. We recently added it to the raw result view.

jgehrcke:
done

Issue #1412, opened 2023-08-03
hardware considered equal despite obvious differences


jgehrcke:
The screenshot below is insightful. Two equal hardware checksums despite obvious differences and problems:

- bad CPU core count and frequency readout: 0
- model name does not impact checksum -- deliberate choice? Is that smart, in all generality?
- l1/l2/l3 cache size difference does not impact checkum. Is that smart?

![image](https://github.com/conbench/conbench/assets/265630/12d856c9-0f34-424f-893f-138f27261c54)

In this situation these are treated as 'same hardware' because -- well, spoken provocatively: they have an equal amount of RAM. The rest is practically ignored for various weird reasons :fish: 

A Conbench user should submit a flat key/value property mapping, and we should document that _all_ properties in that mapping are fed into the checksum. We can maybe make some properties required.

This strongly relates to

> I think from a database perspective we want a world of fully user-defined hardware/platform. We can trust users as much :-).

from https://github.com/conbench/conbench/issues/1281. Can users trust _us_? Well... here, it's a classical case of thinking in terms of "who should be responsible for what?".

The HTTP API should allow for Conbench users/operators to precisely control which properties go into the hardware checksum and which don't. Then we're off the hook. It's rather obvious that we cannot build a silver bullet solution :).

austin3dickey:
For a bit of background here, when setting up benchmarking on Velox using GitHub Actions, we [purposely chose](https://github.com/facebookincubator/velox/pull/3830#discussion_r1088195716) to use the same hardware name for every run. 

This was because during original testing, each GHA machine was showing up with name `ip-172-28-33-150`, `ip-172-28-9-238`, etc. And many of those machines are actually comparable. So I decided to let the name always be `GitHub-runner-8-core` (to distinguish it from if we ran on machines with more/fewer cores).

But you're right that I put too much trust in the way we generate the checksum today:
https://github.com/conbench/conbench/blob/534762befe9429ae1080e42056e9c681afab9801/conbench/entities/hardware.py#L70-L78

This of course does not catch these differences in L cache size, which could have a big effect on microbenchmarks. We've been wondering about this bimodality for a while, and this could help explain it. cc @assignuser

I totally agree about the importance of #1281 and probably will account for it when I plan for #1227.

austin3dickey:
I was nerd-sniped into downloading the data and coloring the plot by `cpu_model_name`:

[...]

Issue #1394, opened 2023-07-28
UI/app: obliterate all API layer re-serialization


jgehrcke:
A ticket tracking the 'final push'. Do all of this kind of work:

* https://github.com/conbench/conbench/issues/1341
* https://github.com/conbench/conbench/issues/1334
* https://github.com/conbench/conbench/issues/1332
* https://github.com/conbench/conbench/issues/968
* https://github.com/conbench/conbench/issues/890
* https://github.com/conbench/conbench/pull/891

PR #1415, opened 2023-08-04
UI: result-result comparison: show hardware checksum

jgehrcke:
This is for #1413, a follow-up from #1412.

Again I was significantly slowed down while trying to make a seemingly trivial change -- https://github.com/conbench/conbench/issues/1394.

First, I gave in and accepted I 'have to' change the HTTP API. Then I did that and realized that our tests are not quite ready to dynamically 'predict' the hardware checksum. So, I added in one more 'needless' DB query, just to have hardware checksum available _after_ lossz API layer re-serialization. You can follow me backing out of that approach by looking at commits.

### DIFF
--- a/conbench/app/compare.py
+++ b/conbench/app/compare.py
@@ -37,6 +37,8 @@ class Compare(AppEndpoint, BenchmarkResultMixin, RunMixin, TimeSeriesPlotMixin):
         baseline_run, contender_run = None, None
         biggest_changes_names, outlier_urls = None, None
         benchmark_result_history_plot_info = None
+        contender_hardware_checksum = "n/a"
+        baseline_hardware_checksum = "n/a"
 
         if comparisons and self.type == "run":
             baseline_run_id, contender_run_id = baseline_id, contender_id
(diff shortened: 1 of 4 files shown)


jgehrcke:
:ship:  feedback appreciated!

conbench/app/compare.py
```diff
@@ -61,9 +63,14 @@ def page(self, comparisons, regressions, improvements, baseline_id, contender_id
             # `contender` is a dictionary representing the benchmark result.
             # Also inject the 'proper' benchmark result object  into
             # get_history_plot(). This is a mad way to deal with madness.
-            # We need to remove this API layer indirection:
+            # We (really!) need to remove this API layer indirection:
             # https://github.com/conbench/conbench/issues/968
             contender_benchmark_result = BenchmarkResult.one(id=contender["id"])
+            baseline_benchmark_result = BenchmarkResult.one(id=contender["id"])
```
austin3dickey, line 69: Shouldn't this be `baseline["id"]`?

deepsihag, line 69: Should this be ?
```suggestion
            baseline_benchmark_result = BenchmarkResult.one(id=baseline["id"])
```

jgehrcke: Absolutely. Thank you for spotting this!

austin3dickey: Haha, wow, we noticed this at the same time :)

jgehrcke:
repaired typo in https://github.com/conbench/conbench/pull/1422 -- thanks @deepsihag + @austin3dickey

Issue #1417, opened 2023-08-04
benchmark result: make `iterations` optional, clarify intent, remove validation


jgehrcke:
See title, action item as of https://github.com/conbench/conbench/issues/533#issuecomment-1662415311.

jgehrcke:
Closing this as 'done', see linked PRs above.

PR #1418, opened 2023-08-04
benchmark result: after clarified meaning of `iterations` (not repetitions): remove validation

jgehrcke:
This is for #1417.


jgehrcke:
:ship: questions / feedback appreciated!

PR #1421, opened 2023-08-07
benchmark result `stats` object: make `iterations` optional

jgehrcke:
This is for https://github.com/conbench/conbench/issues/1417. I have updated a small set of tests to cover/make use of the new behavior.


jgehrcke:
:ship: feedback / discussion very much appreciated.

PR #1423, opened 2023-08-08
UI: benchmark result view: expose CSV download url

jgehrcke:
This is for #959 / PFE-1971.

It's a subtle change that probably triggers many opinions. Let's iterate.

image


jgehrcke:
:ship: not beautiful, but will be helpful for us/me. Many ideas from here, other feedback/ideas are super welcome of course.

austin3dickey:
I think this is a great start!

Issue #1424, opened 2023-08-09
`info`: re-do api spec (clarify: does not break history!), make it optional; phase out optional_benchmark_info


jgehrcke:
Follow-up from https://github.com/conbench/conbench/issues/838. Current spec:

![image](https://github.com/conbench/conbench/assets/265630/d5f1da90-553c-43d6-9139-169f5dc8fad2)

I find this misleading:

> This information is expected to be the same across a number of benchmarks.

Taking @austin3dickey's lovely ASCII table from a recent Slack convo. Current behavior:
```
API key                   | DB colname              | UI desc                      | breaks history?
--------------------------+-------------------------+------------------------------+-----------------
machine_info/cluster_info | run.hardware_id         | Hardware                     | yes
tags                      | case_id                 | Raw tags                     | yes
optional_benchmark_info   | optional_benchmark_info | Additional Benchmark Details | no
context                   | context_id              | Context                      | yes
info                      | info_id                 | Additional context           | no
```

action items:

- make `info` key optional (do not require providing an empty dict)
- clarify 'does not break history', store anything that you like
- document that `info` should be used instead of `optional_benchmark_info`

We could do a transition, but maybe never will:
- in api implementation: transparently write optional_benchmark_info to `info` (what to do upon conflict?)
- adjust 'our' client tooling to never provide optional_benchmark_info
- disallow key (reject when provided)
- db migration for optional_benchmark_info -> info
edit: that last bit deserves a new ticket (that me might never act upon): https://github.com/conbench/conbench/issues/1426

austin3dickey:
My understanding before was that

`case : optional_benchmark_info :: context : info`

That is, despite having a terrible name, `info` has been useful in the past to store contextual details about a benchmark run that (1) shouldn't have a different timeseries fingerprint, and (2) are more related to benchmark-runner contextual details than the conceptual benchmark, and should be considered separate from conceptual benchmark details, organizationally.

For example, [this Arrow TPC-H benchmark result](https://conbench.ursa.dev/benchmark-results/064d04a5cd367f598000f74f452bf6e4/) looks like it stores details about Arrow version in `info` and details about the result and memory of the TPC-H query in `optional_benchmark_info`.

I just wanted to call this out for full clarity: This issue is making the proposal that (2) is not worth the overhead, correct? That we only need one place to put non-breaking tags?

[...]

Issue #838, opened 2023-03-10
BenchmarkResult schema: `info` property: clarify, document how this should be used


jgehrcke:
Trying to unconfuse myself with your help. Here is what I know about `info`:

- from db schema:
```
    # this info should probably be called something like context-info
    # it's details about the context that are optional | we believe won't
    # impact performance
    info = relationship("Info", lazy="joined")
```

- from api spec:

> Additional information about the context the benchmark was run in that is not expected to have an impact on benchmark performance (e.g. benchmark language version, compiler version). This information is expected to be the same across a number of benchmarks. (free-form JSON)

So, `info` is documented as similar to "context", but seemingly not in a way that it creates new case permutations, right?

If `info` is per-benchmark-result and does _not_ create new case permutations; why do we in addition need `optional_benchmark_info`? Where's the separation of concerns between those two?

Should / must this be a flat (one-level) key-value pair mapping? If yes: let's specify, and test.

The database column type accepts arbitrarily deep JSON structured, I think:
https://github.com/conbench/conbench/blob/fbf5a6ce898532d44327e3ac961f5abb852843bc/conbench/entities/info.py#L17
https://www.postgresql.org/docs/current/datatype-json.html

How does the index behave, really, with respect to nested JSON structures?
```
s.Index("info_index", Info.tags, unique=True)
```
Seems like there's magic in here that we should call out a little more explicitly.

>  This information is expected to be the same across a number of benchmarks. (free-form JSON)

why is this important?

Output side of things: where is `info` shown in the benchmark result view? Ah, that's here: https://github.com/conbench/conbench/blob/fbf5a6ce898532d44327e3ac961f5abb852843bc/conbench/templates/benchmark-entity.html#L225. Should probably add a tooltip that clarifies the relationship to case permutations.

jgehrcke:
I see that this is related:
https://github.com/conbench/conbench/pull/408
https://github.com/conbench/conbench/issues/407

jgehrcke:
> If info is per-benchmark-result and does not create new case permutations; why do we in addition need optional_benchmark_info? Where's the separation of concerns between those two?

It looks like there is no separation, and this was an unfortunate oversight. After understanding more, the high-level takeaway is: remove `optional_benchmark_info`, remove the `Info` table. 

https://github.com/conbench/conbench/issues/407 had a bit of a different intent than https://github.com/conbench/conbench/pull/408 implemented.

Two follow-ups:

[...]

Issue #880, opened 2023-03-15
batch: phase concept out from UI, re-introduce as optional intra-run grouping construct


jgehrcke:
Had a lovely conversation about #873 with @alistaire47 and @austin3dickey.

We agree that the UI should not imply a hierarchy that does not exist. We should phase the Run -> Batch hierarchy out, from the UI.

It may make sense to provide a grouping construct for benchmark results _within a run_. For certain use cases. This can be achieved with an optional property on a benchmark result. Idea:
- Remove required `batch_id` from `BenchmarkResult`
- Add optional `batch_name` or `batch` or `suite` or `suite_name` or something like that. (do not use "id" because that suggests an invariant, like a uniqueness constraint).

Then in the UI _within a run_ we can do client-side visual grouping/filtering with that.

PR #1425, opened 2023-08-09
benchmark result submission API spec: make `info` optional, improve spec for `github`, `context`, misc

jgehrcke:
For #1424 and #838.

While working on tests for c-bench UI I wanted to submit benchmark results and came around a number of smaller topics that we touched on before, but didn't pull through (e.g., https://github.com/conbench/conbench/pull/1134#discussion_r1172771186).

This patch:

- clarifies what the absence of the `github` key means
- makes `info` optional
- recommends to use `info` instead of  `optional_benchmark_info`
- clarifies what `context` is to be used for
- does a few tiny things more, see commit msgs / diff.

### DIFF
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -222,7 +222,7 @@ class BenchmarkResult(Base, EntityMixin):
         # Create related DB entities if they do not exist yet.
         case = Case.get_or_create({"name": benchmark_name, "tags": tags})
         context = Context.get_or_create({"tags": userres["context"]})
-        info = Info.get_or_create({"tags": userres["info"]})
+        info = Info.get_or_create({"tags": userres.get("info", {})})
 
         # Create a corresponding `Run` entity in the database if it doesn't
         # exist yet. Use the user-given `id` (string) as primary key. If the
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -1285,7 +1285,21 @@ class _BenchmarkResultCreateSchema(marshmallow.Schema):
     context = marshmallow.fields.Dict(  # type: ignore[assignment]
         required=True,
         metadata={
-            "description": "Information about the context the benchmark was run in (e.g. compiler flags, benchmark langauge) that are reasonably expected to have an impact on benchmark performance. This information is expected to be the same across a number of benchmarks. (free-form JSON)"
+            "description": conbench.util.dedent_rejoin(
+                """
+                Required. Must be a JSON object (empty dictionary is allowed).
+
+                Relevant benchmark context (other than hardware/platform
+                details and benchmark case parameters).
+
+                Conbench requires this object to remain constant when doing
+                automated timeseries analysis (this breaks history).
+
+                Use this to store for example compiler flags or a runtime
+                version that you expect to have significant impact on
+                measurement results.
+                """
+            )
         },
     )
     info = marshmallow.fields.Dict(
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -36,7 +36,7 @@ from ..entities.commit import TypeCommitInfoGitHub
 from ..entities.context import Context
 from ..entities.hardware import ClusterSchema, MachineSchema
 from ..entities.info import Info
-from ..entities.run import Run, SchemaGitHubCreate
+from ..entities.run import Run, SchemaGitHubCreate, github_commit_info_descr
 
 log = logging.getLogger(__name__)
 
@@ -1120,8 +1120,6 @@ Currently-recognized keys that change Conbench behavior:
 hardware/case/context)? That is, when evaluating whether future results are regressions
 or improvements, should we treat data from before this result as incomparable?
 """
-
-from conbench.entities.run import github_commit_info_descr
 
 
 class _BenchmarkResultCreateSchema(marshmallow.Schema):
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -1121,6 +1121,8 @@ hardware/case/context)? That is, when evaluating whether future results are regr
 or improvements, should we treat data from before this result as incomparable?
 """
 
+from conbench.entities.run import github_commit_info_descr
+
 
 class _BenchmarkResultCreateSchema(marshmallow.Schema):
     run_id = marshmallow.fields.String(
@@ -1298,7 +1300,11 @@ class _BenchmarkResultCreateSchema(marshmallow.Schema):
             "description": "Benchmark results validation metadata (e.g., errors, validation types)."
         },
     )
-    github = marshmallow.fields.Nested(SchemaGitHubCreate(), required=False)
+    github = marshmallow.fields.Nested(
+        SchemaGitHubCreate(),
+        required=False,
+        metadata={"description": github_commit_info_descr},
+    )
     change_annotations = marshmallow.fields.Dict(
         required=False, metadata={"description": CHANGE_ANNOTATIONS_DESC}
     )
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -1276,7 +1276,9 @@ class _BenchmarkResultCreateSchema(marshmallow.Schema):
     optional_benchmark_info = marshmallow.fields.Dict(
         required=False,
         metadata={
-            "description": "Optional information about Benchmark results (e.g., telemetry links, logs links). These are unique to each benchmark that is run, but are information that aren't reasonably expected to impact benchmark performance. Helpful for adding debugging or additional links and context for a benchmark (free-form JSON)"
+            # TODO: remove this.
+            # https://github.com/conbench/conbench/issues/1424
+            "description": "Deprecated. Use `info` instead."
         },
     )
     # Note, this mypy error is interesting: Incompatible types in assignment
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -1289,9 +1289,29 @@ class _BenchmarkResultCreateSchema(marshmallow.Schema):
         },
     )
     info = marshmallow.fields.Dict(
-        required=True,
+        required=False,
         metadata={
-            "description": "Additional information about the context the benchmark was run in that is not expected to have an impact on benchmark performance (e.g. benchmark language version, compiler version). This information is expected to be the same across a number of benchmarks. (free-form JSON)"
+            "description": conbench.util.dedent_rejoin(
+                """
+                Optional.
+
+                Arbitrary metadata associated with this
+                benchmark result.
+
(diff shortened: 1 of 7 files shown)


conbench/entities/run.py
```diff
@@ -698,18 +721,7 @@ class _RunFacadeSchemaCreate(marshmallow.Schema):
...
         required=False,
-        metadata={
-            "description": conbench.util.dedent_rejoin(
-                """
-                Use this object to tell Conbench with which commit the
-                Run/BenchmarkResult is associated.
-
-                This field can be left out for testing purposes. Note however
-                that commit information is required for powering valuable
-                Conbench capabilities.
```
jgehrcke, line 709: So far we had this (insufficient) spec for `run.github`. Updated spec for both `run.github` and `result.github`. This is of course again a temporary state with obvious quirks, but one that is better than before.

jgehrcke:
I got some clarity here which I summarized in https://github.com/conbench/conbench/issues/1426.

jgehrcke:
Going ahead with this, we have been stuck in analysis paralysis here for a too long time (cf. https://github.com/conbench/conbench/issues/838). Please review this and keep the feedback coming though!

conbench/entities/benchmark_result.py
```diff
@@ -1283,13 +1288,47 @@ class _BenchmarkResultCreateSchema(marshmallow.Schema):
...
+
+                Relevant benchmark context (other than hardware/platform
+                details and benchmark case parameters).
+
+                Conbench requires this object to remain constant when doing
+                automated timeseries analysis (this breaks history).
+
+                Use this to store for example compiler flags or a runtime
+                version that you expect to have significant impact on
+                measurement results.
```
austin3dickey, line 1303: Another thing to call out here: in a few places, Conbench looks for a special key in this dict.

https://github.com/conbench/conbench/blob/2411b0efb70036dc3563c19b12f6bd3ffd482594/conbench/api/compare.py#L214

Specifically, in the compare API (above), and the [result list](https://conbench.ursa.dev/benchmark-results/) and [batch](https://conbench.ursa.dev/batches/ef84837df9e04facbedc0de7c4cce759-1p/) pages (that, to be fair, aren't linked to from other places).

I'm still in favor of first-classing the "language" idea of a benchmark result, because IMO a core strength of Conbench is its ability to manage different benchmarking languages. (For example, the `benchalerts` package  consumes the `language` output of the compare API for clearer notifications.) Putting it in `context` like this might not be the best place, but it's how we manage it today, so I think we should call that out here.

austin3dickey (commented):
Thank you!

conbench/entities/benchmark_result.py
jgehrcke: Oh yes, great point.

conbench/entities/benchmark_result.py
```diff
@@ -1159,7 +1159,10 @@ class _BenchmarkResultCreateSchema(marshmallow.Schema):
             )
         },
     )
-    batch_id = marshmallow.fields.String(required=True)
+    batch_id = marshmallow.fields.String(
+        # this lacks specification and should probably not be required
+        required=True
+    )
```
alistaire47, line 1165: This is user-customizable, but in almost all cases this is the interaction of benchmark name and run ID, i.e. the collection of cases for a benchmark on a run.

I see value in that idea (e.g. when one case is weird, look at the others), but we can get the same result by grouping by name and run ID, which could be the default if it's not set.

conbench/entities/benchmark_result.py
```diff
@@ -1283,13 +1288,47 @@ class _BenchmarkResultCreateSchema(marshmallow.Schema):
...
+                Optional.
+
+                Arbitrary metadata associated with this
+                benchmark result.
+
+                Ignored when assembling timeseries across results (differences
+                do not break history).
+
+                Must be a JSON object if provided. A flat string-string mapping
+                is recommended (not yet enforced).
```
alistaire47, line 1322: In some cases users did want to stick non-flat / non-string things in `optional_benchmark_info`, e.g. R benchmarks currently pass the benchmarking script generated as an array of lines, and the parameters used for running including a jsonified dataframe of installed libraries: https://conbench.ursa.dev/benchmark-results/064d39ad32387b0780009ecd5b68851f/

I'm not convinced either is necessary, but Jon was adamant about the script in particular. If we want to keep it while flat we could store it as a very long string with newlines.

conbench/entities/benchmark_result.py
alistaire47: Yeah, users frequently screw this up when assembling result JSON, so it should definitely be documented. Previously I've advocated for validating it too so users get better errors, but I think we made it optional in certain cases now?

conbench/entities/benchmark_result.py
jgehrcke: Yes. We've had this conversation, and noted down our findings :) https://github.com/conbench/conbench/issues/880.

My code comment there was once again JP trying to 'quickly do something' while I went over this.

jgehrcke: https://github.com/conbench/conbench/pull/1427/files#diff-9d75d1bf6ac23bfc056a410721cf4c9bccd8b302f8e3e83e4b4bf1698d0eaad7R1164

conbench/entities/benchmark_result.py
jgehrcke: Yes: let's certainly document that if a `benchmark_language` key is present in `context` then UI and API do/may emphasize it various places :).

side-note, almost offtopic:

> IMO a core strength of Conbench is its ability to manage different benchmarking languages

Honestly, I can see where this came from, but I find it to be not the strongest selling point, kind of distracting from actual value :) Already on day 1 I was confused that this is part of the tagline. 

We provide a generic interface for ingesting measurement data. When we say "you, please measure, and give us the results in the following format" then to me this implies infinite flexibility in what to measure and how to measure (water level of garden pond in the morning, time it takes to compile brainfuck code with a compiler written in turbo pascal). It may even be slightly misleading: we do not offer benchmark runners / benchmark frameworks in all the languages.

conbench/entities/benchmark_result.py
jgehrcke: yeah :) let's see if this ever becomes relevant

conbench/entities/benchmark_result.py
alistaire47: Benchmark-language agnosticism is part of the _raison d'être_ of Conbench—alternatives like ASV are locked to a single language, while Conbench is a generic service for tracking benchmarks.

> It may even be slightly misleading: we do not offer benchmark runners / benchmark frameworks in all the languages.

For the most part, we don't have any runners or frameworks at all; everything is left to users. Exceptions:

1. We/I support arrowbench, though the runner is all conflated with a bunch of Arrow benchmarks. 
2. We do have the legacy runner stuff in Python, but we'd like to get rid of it.
3. benchrun exists, but is not used. If we create a robust adapter for raw ASV data or pytest-benchmark or another Python runner, we could remove it.

Long-term, a set of paved roads for existing runners in common languages (probably mostly adapters for common frameworks) would help adoption, if that's a thing we care about.

austin3dickey: >a generic interface for ingesting measurement data

Here's my roundabout logic for caring about code language. Your quote above is definitely a core value of Conbench. But I think we agree that another core value is Conbench's ability to aid in the analysis of changes in that data, and further, connect those changes to code changes in a git repository. Making the assumption that every Conbench user deeply cares about this last point is what enables me to think about tailoring our interfaces to that use case (instead of, say, measuring water level 😉). Therefore I think a great "tailorization" is to embrace the fundamentals of coding in a git repository, like code language.

alistaire47: 💯 That's why I get uneasy when we accept results with missing git metadata—because our analysis indexes by commit, these results largely get ignored unless users analyze them themselves.

jgehrcke: You're all right of course and I think we have a misunderstanding here :) I come from a product description / marketing / messaging perspective.

> I think a great "tailorization" is to embrace the fundamentals of coding in a git repository, like code language.

I agree! Of course. We know that. Right? :)

When I wrote above "I find it to be not the strongest selling point" I literally meant that it is not our strongest selling point. Yes it is a selling point, just in my personal opinion not particularly strong relative to others.

A prospect reads

> ability to manage different benchmarking languages

> Conbench allows you to write benchmarks in any language

> Language-independent Continuous Benchmarking

That raises pretty big expectations with respect to this language-independent nature; and it raises the question "what do they mean by that, what's the secret sauce here, how much work was it to achieve this?"

As of today I think this is what we mean:

- we provide a generic interface (for other good reasons than just language agnosticism)
- when having `benchmark_language` in `context` then the UI ...

That's about it. Right? What do I miss?

> Conbench's ability to aid in the analysis of changes in that data, and further, connect those changes to code changes in a git repository

That type of stuff is what I was referring to with "actual value " when writing above

> distracting from actual value 

That 'actual value ' was of course non-ideal wording because it implies that the other thing has little value. I should have said "from our stronger selling points" :).

jgehrcke: > I get uneasy when we accept results with missing git metadata—because our analysis indexes by commit, these results largely get ignored unless users analyze them themselves.

Haha. Well, one can add some structure to this and then one can be almost worry-free?

Can be reconciled:

- low getting-started/testing barrier (not requiring all the fields)
- documenting what's needed for achieving what (we've made good progress here in the recent months)

>  these results largely get ignored unless users analyze them themselves.

that's pretty much what we document as of today

https://github.com/conbench/conbench/blob/b03b1ffa0716098143ac0cbe354aa9169205a275/conbench/entities/run.py#L57


```
    Not associating a benchmark result with commit information has special,
    limited purpose (pre-merge benchmarks, testing). It generally means that
    this benchmark result will not be considered for time series analysis along
    a commit tree.
```

not perfect. docs / guides can and should of course be improved over time.

austin3dickey: These are great thoughts as well! Agreed that our marketing pitch should be better-tuned.

alistaire47: > Can be reconciled:
> 
> * low getting-started/testing barrier (not requiring all the fields)
> * documenting what's needed for achieving what (we've made good progress here in the recent months)

Agreed! The hidden cost here of starting with incomplete metadata is that adding it later will often break histories. (Not applicable to git info, without which there can be no history, but certainly `context`, `tags`, hardware, etc.) Lowering this impact would be good. 

Idle thought: I wonder if we should allow the modification of metadata for all results in a timeseries fingerprint (e.g. add/remove/modify a case tag) so it merges with another? I can think a lot of existing use-cases, e.g. adding batch name to Archery microbenchmark benchmark name, populating CPU count in Arrow R benchmarks 😭, removing vestigial tags when arrowbench benchmarks are run via adapters, etc. etc.

PR #1427, opened 2023-08-09
BMRT cache population, c-bench UI: start adding test infra and tests

jgehrcke:
Before we do more or less significant work on DB schema I wanted to add a bit of test coverage for the c-bench UI which is fed from the BMRT cache.

Tracked by PFE-2013.

### DIFF
--- a/conbench/job.py
+++ b/conbench/job.py
@@ -364,7 +364,7 @@ def _fetch_and_cache_most_recent_results_guts(
     )
 
 
-def _periodically_fetch_last_n_benchmark_results() -> None:
+def _periodically_fetch_last_n_benchmark_results() -> threading.Thread:
     """
     Return right after having spawned a thread that triggers periodic action.
     """
--- a/conbench/job.py
+++ b/conbench/job.py
@@ -416,14 +416,6 @@ def _periodically_fetch_last_n_benchmark_results() -> threading.Thread:
             delay_s = max(min_delay_between_runs_seconds, 5 * last_call_duration_s)
             log.info("BMRT cache: trigger next fetch in %.3f s", delay_s)
 
-    if not Config.CREATE_ALL_TABLES:
-        # This needs to be done more cleanly -- when running the DB migration,
-        # the app should not even initialize so far.
-        log.info(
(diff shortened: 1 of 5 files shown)


conbench/job.py
```diff
@@ -519,10 +518,50 @@ def _generate_tsdf_per_4tuple(
...
+
     log.info("start job: metrics.periodically_set_q_rem()")
-    conbench.metrics.periodically_set_q_rem()
+    _THREADS.append(conbench.metrics.periodically_set_q_rem())
+
+
+def wait_for_first_bmrt_cache_population(timeout=20):
+    """
+    Wait (block) until the first BMRT cache population loop iteration to has
+    completed, or raise a timeout error after deadline.
```
jgehrcke, line 538: will address typo/grammar oversights in follow-up

conbench/job.py
```diff
@@ -519,10 +518,50 @@ def _generate_tsdf_per_4tuple(
...
+    return None
+
+
+def stop_jobs_join():
+    """
+    For clean shutdown, explicitly waits for threads to terminate.
+
+    I've added this for the test suite where we may go through multiple
+    start/stop cycles. This is not really great because this mode of operation
+    deviates from "prod".
```
jgehrcke, line 555: but it's also not too bad. :)

conbench/tests/app/test_conceptual_benchmark_views.py
```diff
@@ -9,6 +12,34 @@ def assert_response_is_login_page(resp):
...
+    "stats": {
+        "data": [
+            "1.1",
+        ],
+        "unit": "s",
+    },
+    "tags": {
+        "name": "fun-benchmark",
+    },
+}
```
jgehrcke, line 40: off-topic, side note.

With less properties being required such a dictionary becomes more and more expressive/explicit on its own. 

I envision a state where `batch_id`, `run_id`, `context` are not required here, too.

Minimal set of properties for a benchmark result to start to make sense:
- name
- hardware/platform
- data
- time

jgehrcke:
:shipit: feedback appreciated!

Issue #1434, opened 2023-08-14
 job.py: split orthogonal concepts; decouple BMRT cache logic


jgehrcke:
Obvious TODO; kept pushing until later: it's time to make job.py more tidy; contains a number of non-orthogonal concepts. Separate the 'job handling' and the 'BMRT-specific logic' into distinct modules.

jgehrcke:
Done in https://github.com/conbench/conbench/pull/1435.

PR #1435, opened 2023-08-14
c-bench UI: more tests, BMRT cache: add reinit() interface, module docstring, move to bmrt.py

jgehrcke:
This patch 

- adds more tests for the c-bench UI (PFE-2013)
- adds a `reinit()` interface for the BMRT cache, required for testability

These changes once again asked for a clearer separation between job management and bmrt-specific logic; so I have mixed the changes above with a bigger change of moving logic to new bmrt.py module (#1434), and adjusting consuming interfaces.

I have also added more commentary so that it can be easier for new devs to understand the origins and trade-offs of the current bmrt cache methodology.


jgehrcke:
:ship:

conbench/bmrt.py
```diff
@@ -0,0 +1,579 @@
...
+# from filprofiler.api import profile as filprofile
+# import yappi
+
+
+log = logging.getLogger(__name__)
+
+BMRT_CACHE_SIZE = 0.8 * 10**6
+if Config.TESTING:
+    # quicker update in testing
+    BMRT_CACHE_SIZE = 0.05 * 10**6
```
jgehrcke, line 67: I also re-ordered code for clarity a bit.

PR #1436, opened 2023-08-14
webapp: remove compare run picker

austin3dickey:
I don't believe anyone uses this picker from the `/compare/runs/` page, especially now that the "candidate baseline runs" functionality is in the UI.

### DIFF
--- a/conbench/templates/compare-list.html
+++ b/conbench/templates/compare-list.html
@@ -165,36 +165,6 @@
     </table>
   </div>
 </div>
-<hr />
-<h4>Compare current contender run with another run:</h4>
-<br>
-<form>
-  <div class="form-group">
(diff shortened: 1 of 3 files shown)


conbench/templates/compare-list.html
```diff
@@ -253,45 +223,5 @@ <h4>Compare current contender run with another run:</h4>
           },
         });
 
-    // Note(JP): maybe this is for the "compare current contender run to
-    // another run" form. Who uses that? Why is this workflow important?
```
jgehrcke, line 257: Ah :)

Thanks for pulling through. Big fan of removing UX complexity.

We can add things like that back if we ever want to.

jgehrcke (approved):
:rocket: good one

Issue #1227, opened 2023-05-05
Explore removing Run table (in favor of run_id on Benchmark_Result, also commit, hardware, etc)


jgehrcke:
We recently had another discussion about the pros and cons of having a distinct Run table. This ticket here is supposed to briefly account for that discussion. It's also non-exhaustive.

In this discussion I found one more time that there might be a huge opportunity for simplification (as previously identified in related discussions, e.g. around implicit run creation).

For most types of analyses one needs to look at individual results anyway, and places in code where we do `result.run.commit` / `result.run.hardware` hide a lot of database interaction complexity. The BMRT cache-based work goes into that direction: here, run is just an indirection to get to commit/hardware info.

We identify that a ‘CI run’ perspective is probably where the ‘Benchmark Run’ came from, and where it provides value from a UX perspective. Run as a UX primitive can be implemented as a 'virtual view', in the database it would be sufficient to have a something like a `run_id` property on each BenchmarkResult object/row. That allows for grouping results by run_id, and UI as well as API routes could maintain current behavior.

There is desire to mark a Run as started and finished and failed. That would be difficult to build. But JP notes that this is error-prone and does not yield to the desired outcome anyway:
* Creating a ‘failed Run’ seems to provide value for reporting.
* But we have to plan for / with the scenario that ‘even just creating the Run object’ will fail, and then there’s a void. Conbench is _not_ the primary CI system.
* That is, consuming tooling needs to inspect the CI job that is supposed to submit benchmark results to Conbench anyway.
* The CI job may fail _before_ it is able to submit a “new Run” to Conbench. And it will fail doing so, leaving no trace in Conbench (may be rare, but it will happen).
* At the highest level, this is a ‘classic’ “optimization-first” architecture error: The reporting tooling needs to consult the CI runner anyway (e.g., buildkite), and be prepared for unexpected CI job failure (before Conbench can even be contacted). That is so-to-speak the error handling foundation that must be built. On top of that, as an optimization, one could in theory do more, but the value isn’t clear if the foundation is built nicely.

In practice, we see that Runs are not marked as finished.

Strongly related:
* https://github.com/conbench/conbench/issues/1165
* https://github.com/conbench/conbench/issues/863
* https://github.com/conbench/conbench/issues/864

austin3dickey:
ah, this is done :)

Issue #607, opened 2023-01-19
run_name: deprecate? document intent, adjust DB schema, remove decision-making burden from users


austin3dickey:
Now that run name is not used anywhere (except for a certain [picker](https://github.com/conbench/conbench/blob/main/conbench/templates/compare-list.html#L295) that we could change), we could consider deprecating it. All the projects using Conbench that I'm aware of populate run name by concatenating run reason and commit hash, both of which are easily accessible otherwise. It also can lead to [complications](https://github.com/voltrondata-labs/arrowbench/pull/127#discussion_r1080613598) in benchmark runners trying to provide a good UX with sensible defaults. It would reduce code complexity in conbench tools if we didn't have to worry about it.

What deprecation might look like:
- stop displaying it in the webapp
- if we need to uniquely, human-readably identify a run (in alerts or the webapp), we could auto-concatenate commit + hardware + reason (+ timestamp if necessary)
- have all conbench tools stop supplying it to the API when posting runs/results
- keep it around in the POST schema and database schema for backwards compatibility, with code comments

jonkeane:
Can `run_name` currently be `""` or `null`?

austin3dickey:
>Can run_name currently be "" or null?

Yes, it could be either. There are some older rows in Conbench databases we know about where this is the case.

(This currently may cause the aforementioned picker at the bottom of the `/compare/runs/A...B` page ([example](https://conbench.ursa.dev/compare/runs/8f36663cfd3f44e49405b9fe83811b7c...a0af231eecf746c8b60891054b1cf84c/)) to behave weirdly for those runs, but I'm not sure how widely-used that picker is, so it's never come up.)

This ticket is suggesting we go a step further than just allowing it to be empty, and actually remove most references to it, in order to reduce code complexity.

jonkeane:
> This ticket is suggesting we go a step further than just allowing it to be empty, and actually remove most references to it, in order to reduce code complexity.

Yes, I understood the proposal. Could you say more about the code complexity we would be removing? Besides displaying it in the UI, and accepting it in `POST`s, and the picker, do we use it elsewhere (we shouldn't anymore after #507 yeah?)?

[...]

PR #1437, opened 2023-08-14
DB: Remove run table

austin3dickey:
This PR removes the Run table but attempts to keep all user interfaces as similar as possible.

Addresses #1227.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ tests: require-env-ghtoken set-build-info
 	docker compose down --remove-orphans && \
 	docker compose build app && \
 	docker compose run app \
-		pytest -vv -s --durations=20 conbench/tests/
+		pytest -vv --durations=20 conbench/tests/
 
 
 # Similar to `make run-app`, but with the `docker-compose.dev.yml` extension
--- a/conbench/api/runs.py
+++ b/conbench/api/runs.py
@@ -1,25 +1,19 @@
+from typing import Dict
+
 import flask as f
 import flask_login
 
 from ..api import rule
 from ..api._docs import spec
 from ..api._endpoint import ApiEndpoint, maybe_login_required
-from ..entities._entity import EntityExists, NotFound
 from ..entities.benchmark_result import BenchmarkResult
 from ..entities.commit import Commit
-from ..entities.run import Run, RunFacadeSchema, RunSerializer
+from ..entities.run import RunFacadeSchema, RunSerializer
 
 
 class RunEntityAPI(ApiEndpoint):
     serializer = RunSerializer()
     schema = RunFacadeSchema()
-
-    def _get(self, run_id):
-        try:
-            run = Run.one(id=run_id)
-        except NotFound:
-            self.abort_404_not_found()
-        return run
 
     @maybe_login_required
     def get(self, run_id):
@@ -69,16 +63,18 @@ class RunEntityAPI(ApiEndpoint):
         tags:
           - Runs
         """
-        run = self._get(run_id)
-        return self.serializer.one._dump(run, get_baseline_runs=True)
+        any_result = BenchmarkResult.first(run_id=run_id)
+        if not any_result:
+            self.abort_404_not_found()
+        return self.serializer.one._dump(any_result, get_baseline_runs=True)
 
     @flask_login.login_required
     def put(self, run_id):
         """
         ---
-        description: Edit a run.
+        description: Deprecated. This endpoint is a no-op, despite returning a 200.
         responses:
-            "200": "RunEntityWithoutBaselines"
+            "200": "RunCreated"
             "401": "401"
             "404": "404"
         parameters:
@@ -93,16 +89,13 @@ class RunEntityAPI(ApiEndpoint):
         tags:
           - Runs
         """
-        run = self._get(run_id)
-        data = self.validate(self.schema.update)
-        run.update(data)
-        return self.serializer.one.dump(run)
+        return {}
 
     @flask_login.login_required
     def delete(self, run_id):
         """
         ---
-        description: Delete a run.
+        description: Deprecated. This endpoint is a no-op, despite returning a 204.
         responses:
             "204": "204"
             "401": "401"
@@ -115,11 +108,6 @@ class RunEntityAPI(ApiEndpoint):
         tags:
           - Runs
         """
-        benchmark_results = BenchmarkResult.all(run_id=run_id)
-        for benchmark_result in benchmark_results:
-            benchmark_result.delete()
-        run = self._get(run_id)
-        run.delete()
         return self.response_204_no_content()
 
 
@@ -131,7 +119,7 @@ class RunListAPI(ApiEndpoint):
     def get(self):
         """
         ---
-        description: Get a list of runs.
+        description: Get a list of runs corresponding to the last 1000 benchmark results
         responses:
             "200": "RunList"
             "401": "401"
@@ -145,16 +133,33 @@ class RunListAPI(ApiEndpoint):
         """
         if sha_arg := f.request.args.get("sha"):
             shas = sha_arg.split(",")
-            runs = Run.search(filters=[Commit.sha.in_(shas)], joins=[Commit])
+            filters = [Commit.sha.in_(shas)]
+            joins = [Commit]
         else:
-            runs = Run.all(order_by=Run.timestamp.desc(), limit=1000)
-        return self.serializer.many.dump(runs)
+            filters = []
+            joins = None
+
+        results = BenchmarkResult.search(
+            filters=filters,
+            joins=joins,
+            order_by=BenchmarkResult.timestamp.desc(),
+            limit=1000,
+        )
+        random_result_for_each_run_id: Dict[str, BenchmarkResult] = {}
+        for result in results:
+            if result.run_id not in random_result_for_each_run_id:
+                random_result_for_each_run_id[result.run_id] = result
+
+        return [
+            self.serializer.one._dump(result, get_baseline_runs=False)
+            for result in random_result_for_each_run_id.values()
+        ]
 
     @flask_login.login_required
     def post(self):
         """
         ---
-        description: Create a run.
+        description: Deprecated. This endpoint is a no-op, despite returning a 201.
         responses:
             "201": "RunCreated"
             "400": "400"
@@ -166,14 +171,7 @@ class RunListAPI(ApiEndpoint):
         tags:
           - Runs
         """
-        data = self.validate(self.schema.create)
-
-        try:
-            run = Run.create(data)
-        except EntityExists as exc:
-            return f.jsonify(error=409, description=str(exc)), 409
-
-        return self.response_201_created(self.serializer.one.dump(run))
+        return self.response_201_created({})
 
 
 run_entity_view = RunEntityAPI.as_view("run")
--- a/conbench/api/runs.py
+++ b/conbench/api/runs.py
@@ -1,4 +1,5 @@
-from typing import Dict
+import datetime
+from typing import Dict, List, Optional, Tuple
 
 import flask as f
 import flask_login
@@ -9,6 +10,49 @@ from ..api._endpoint import ApiEndpoint, maybe_login_required
 from ..entities.benchmark_result import BenchmarkResult
 from ..entities.commit import Commit
 from ..entities.run import RunFacadeSchema, RunSerializer
+
+
+def get_all_run_info(
+    min_time: datetime.datetime,
+    max_time: datetime.datetime,
+    commit_hashes: Optional[List[str]] = None,
+) -> List[Tuple[BenchmarkResult, int, bool]]:
+    """Get info about each run in a time range.
+
+    Query the database for all benchmark results (optionally filtered to certain commit
+    hashes) in a time range, and return a list of tuples, each corresponding to one
+    run_id. The tuples contain:
+
+    - the earliest BenchmarkResult for that run (by its timestamp)
+    - the number of benchmark results seen for that run
+    - whether any benchmark results are "failed" in that run
+
+    This powers the "list runs" API endpoint and the "recent runs" list on the home page.
+    """
+    run_info: Dict[str, Tuple[BenchmarkResult, int, bool]] = {}  # keyed by run_id
+
+    joins = []
+    filters = [
+        BenchmarkResult.timestamp >= min_time,
+        BenchmarkResult.timestamp <= max_time,
+    ]
+    if commit_hashes:
+        joins.append(Commit)
+        filters.append(Commit.sha.in_(commit_hashes))
+
+    for result in BenchmarkResult.search(filters=filters, joins=joins):
+        run_id = result.run_id
+        if run_id not in run_info:
+            run_info[run_id] = (result, 1, result.is_failed)
+        else:
+            old_result, old_count, old_failed = run_info[run_id]
+            run_info[run_id] = (
+                result if result.timestamp < old_result.timestamp else old_result,
+                old_count + 1,
+                old_failed or result.is_failed,
+            )
+
+    return list(run_info.values())
 
 
 class RunEntityAPI(ApiEndpoint):
@@ -119,7 +163,7 @@ class RunListAPI(ApiEndpoint):
     def get(self):
         """
         ---
-        description: Get a list of runs corresponding to the last 1000 benchmark results
+        description: Get a list of runs from the last 30 days of benchmark results.
         responses:
             "200": "RunList"
             "401": "401"
@@ -131,28 +175,16 @@ class RunListAPI(ApiEndpoint):
         tags:
           - Runs
         """
-        if sha_arg := f.request.args.get("sha"):
-            shas = sha_arg.split(",")
-            filters = [Commit.sha.in_(shas)]
-            joins = [Commit]
-        else:
-            filters = []
-            joins = None
-
-        results = BenchmarkResult.search(
-            filters=filters,
-            joins=joins,
-            order_by=BenchmarkResult.timestamp.desc(),
-            limit=1000,
-        )
-        random_result_for_each_run_id: Dict[str, BenchmarkResult] = {}
-        for result in results:
-            if result.run_id not in random_result_for_each_run_id:
-                random_result_for_each_run_id[result.run_id] = result
+        sha_arg: Optional[str] = f.request.args.get("sha")
+        commit_hashes = sha_arg.split(",") if sha_arg else None
 
         return [
             self.serializer.one._dump(result, get_baseline_runs=False)
-            for result in random_result_for_each_run_id.values()
+            for result, _, _ in get_all_run_info(
+                min_time=datetime.datetime.utcnow() - datetime.timedelta(days=30),
+                max_time=datetime.datetime.utcnow(),
+                commit_hashes=commit_hashes,
+            )
         ]
 
     @flask_login.login_required
--- a/conbench/api/runs.py
+++ b/conbench/api/runs.py
@@ -1,5 +1,7 @@
+import dataclasses
 import datetime
-from typing import Dict, List, Optional, Tuple
+import functools
+from typing import Dict, List, Optional
 
 import flask as f
 import flask_login
@@ -10,26 +12,58 @@ from ..api._endpoint import ApiEndpoint, maybe_login_required
 from ..entities.benchmark_result import BenchmarkResult
 from ..entities.commit import Commit
 from ..entities.run import RunFacadeSchema, RunSerializer
+from ..util import short_commit_msg
+
+
+@dataclasses.dataclass
+class RunAggregate:
+    """Aggregated information about a run."""
+
+    earliest_result: BenchmarkResult
+    result_count: int
+    any_result_failed: bool
+
+    def update(self, new_result: BenchmarkResult):
+        """Update the aggregate information based on a new result."""
+        self.earliest_result = (
+            new_result
+            if new_result.timestamp < self.earliest_result.timestamp
+            else self.earliest_result
+        )
+        self.result_count += 1
+        self.any_result_failed = self.any_result_failed or new_result.is_failed
+
+    @functools.cached_property
+    def display_commit_time(self) -> str:
+        return self.earliest_result.timestamp.strftime("%Y-%m-%d %H:%M:%S UTC")
+
+    @functools.cached_property
+    def display_commit_msg(self) -> str:
+        return short_commit_msg(
+            self.earliest_result.commit.message if self.earliest_result.commit else ""
+        )
 
 
 def get_all_run_info(
     min_time: datetime.datetime,
     max_time: datetime.datetime,
     commit_hashes: Optional[List[str]] = None,
-) -> List[Tuple[BenchmarkResult, int, bool]]:
+) -> List[RunAggregate]:
     """Get info about each run in a time range.
 
     Query the database for all benchmark results (optionally filtered to certain commit
-    hashes) in a time range, and return a list of tuples, each corresponding to one
-    run_id. The tuples contain:
+    hashes) in a time range, and return a list of RunAggregates, each corresponding to
+    one run_id.
 
-    - the earliest BenchmarkResult for that run (by its timestamp)
-    - the number of benchmark results seen for that run
-    - whether any benchmark results are "failed" in that run
+    This powers the "list runs" API endpoint and the "recent runs" list on the home
+    page.
 
-    This powers the "list runs" API endpoint and the "recent runs" list on the home page.
+    Note: At the min_time boundary, the count is permanently wrong (single run partially
+    represented in DB query result, true for just one of many runs). That's fine, the UI
+    does not claim real-time truth in that regard. In the vast majority of the cases we
+    get a correct result count per run.
     """
-    run_info: Dict[str, Tuple[BenchmarkResult, int, bool]] = {}  # keyed by run_id
+    run_info: Dict[str, RunAggregate] = {}  # keyed by run_id
 
     joins = []
     filters = [
@@ -42,14 +76,13 @@ def get_all_run_info(
 
     for result in BenchmarkResult.search(filters=filters, joins=joins):
         run_id = result.run_id
-        if run_id not in run_info:
-            run_info[run_id] = (result, 1, result.is_failed)
+        if run_id in run_info:
+            run_info[run_id].update(result)
         else:
-            old_result, old_count, old_failed = run_info[run_id]
-            run_info[run_id] = (
-                result if result.timestamp < old_result.timestamp else old_result,
-                old_count + 1,
-                old_failed or result.is_failed,
+            run_info[run_id] = RunAggregate(
+                earliest_result=result,
+                result_count=1,
+                any_result_failed=result.is_failed,
             )
 
     return list(run_info.values())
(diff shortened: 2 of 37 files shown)


conbench/entities/benchmark_result.py
```diff
@@ -223,38 +226,27 @@ def create(userres) -> "BenchmarkResult":
...
+            commit = commit_fetch_info_and_create_in_db_if_not_exists(
+                user_given_commit_info
             )
-            # The above's `create()` might fail (race condition), in which case
-            # we can re-read, but then we also have to call
-            # `validate_run_result_consistency(userres)` one more time (because
-            # _we_ are not the ones who created the Run).
 
         result_data_for_db["run_id"] = userres["run_id"]
+        result_data_for_db["run_tags"] = userres.get("run_tags") or {}
```
jgehrcke, line 247: Do we have a thought process on `None` (nullable) vs `{}`?

austin3dickey: I think I'd rather it always be stored in the database as `{}` so it's a little easier to work with in code because it's always a `dict`. So I made the column non-nullable. 

However, IMO it shouldn't be required to POST `"run_tags": {}` if you don't have any run tags. So this line turns whatever was posted into `{}` if it was missing (or otherwise falsey). Does this reasoning make sense to you?

jgehrcke:
About "let's fetch data about the last N runs from the DB by fetching the last M benchmark results", and how to decide how many results to fetch/inspect: Talked about that in meeting; and we are all aware. Quickly persisting our thoughts here.

The current WIP `limit=1000` triggers questions at the heart of this patch; towards reaching our goals. This needs to be adjusted if we want to retain the behavior of the GET /runs endpoint because with O(1000) results per run this would yield only O(1) run. 

Hence, we need to inspect a much larger number of results.

There's many ways to achieve that; one could maybe even do a smart query and get N results with unique runs instead of the M most recent results.

As noted in the meeting, the BMRT cache holds a rather big M of most recent results and can serve to drive a quick solution here that might be equally mediocre as the one before (with a cut-off).

Obvious downsides for using the cache instead of the database:
- empty result set before initial cache population (might be a few minutes after container startup, or hold back app until cache is warm); might be ok
- non-realtime-ness; might be ok

The BMRT cache as of today comprises 800.000 results in the deployments we run. For Arrow this is ~14 days worth of time; for other deployments this is 'everything to date'.

conbench/api/runs.py
```diff
@@ -145,16 +133,33 @@ def get(self):
...
-        return self.serializer.many.dump(runs)
+            filters = []
+            joins = None
+
+        results = BenchmarkResult.search(
+            filters=filters,
+            joins=joins,
+            order_by=BenchmarkResult.timestamp.desc(),
+            limit=1000,
+        )
```
austin3dickey: >As noted in the meeting, the BMRT cache holds a rather big M of most recent results and can serve to drive a quick solution here that might be equally mediocre as the one before (with a cut-off).

@jgehrcke This is a good idea; I'll look into it here. It'll probably end up being even faster than before with more runs returned!

conbench/entities/benchmark_result.py
```diff
@@ -534,6 +522,22 @@ def unitsymbol(self) -> Optional[conbench.units.TUnit]:
...
+        This is for those consumers that absolutely need to have a string type
+        representation.
+        """
+        if self.commit and self.commit.repo_url is not None:
+            return self.commit.repo_url
+
+        # This means that the result is not associated with any commit, or it is
+        # associated with a legacy/invalid commit object in the database, one
+        # that does not have a repository URL set.
+        return "n/a"
```
austin3dickey, line 558: @jgehrcke As we talked about earlier, I rearranged things so that the diff now only shows code that I changed. We can move code around later.

The only exception to this is this `associated_commit_repo_url ` property and some of the `Run.create()` internals, which are now in `BenchmarkResult.create()`.

conbench/entities/run.py
```diff
@@ -84,241 +76,169 @@ def _dict_for_api_json(self) -> dict:
...
-    # Ignorantly importing BenchmarkResult results in circular import err.
-    results: Mapped[List["BenchmarkResult"]] = relationship(  # type: ignore # noqa
-        back_populates="run", lazy="select"
-    )
+def _search_for_baseline_run(
+    contender_run_id: str,
+    contender_run_reason: str,
+    contender_hardware_checksum: str,
+    baseline_commit: Optional[Commit],
+    commit_limit: int = 20,
```
austin3dickey, line 83: Hiding whitespace differences will allow the diff of this file to be under the threshold that GitHub uses for auto-collapsing it!

conbench/api/runs.py
alistaire47: A more generalized way to do this would be something like 

``` sql
SELECT *
FROM (
  SELECT 
    *, 
    ROW_NUMBER() OVER (PARTITION BY "run_id") AS "_row_index"
  FROM "result"
) "subquery1"
WHERE ("_row_index" = 1)
```

(Or selecting down and using `DISTINCT`, but that relies on results having identical run metadata.) It would surely be slower than the BMRT cache, but it would let us `LIMIT`/`OFFSET` paginate over everything reliably. 

I'm not sure it's the implementation we want now, but it seems worth running some timings on it to see what the cost of denormalization is without workarounds.

austin3dickey: @alistaire47 Yeah I considered something like that, but even on the Velox DB (which has at least an order of magnitude fewer benchmark results than Arrow) this query takes 45 seconds. Using the cache will be almost instantaneous. The only downside will be potential inaccuracies around the boundaries. But I think that's okay for our purposes. The cache refresh job runs every 2 minutes.

---

@jgehrcke in #1307 you decided to only populate the cache with results from the default branch. What, in your opinion, will be the fallout if I change this to allow results on all branches? Will we have to adjust the c-bench UI code as well?

I'd like the `GET /api/runs/` endpoint and the Runs UI page to return runs that weren't on the default branch too.

Now I'm kind of thinking this is a blocker to using the BMRT cache for this purpose, so if you have some time in your morning and you want to make a quick patch outside of this PR, go for it. Else I will look into it later.

alistaire47: > Yeah I considered something like that, but even on the Velox DB (which has at least an order of magnitude fewer benchmark results than Arrow) this query takes 45 seconds.

This has me concerned; I'm not sure we'll always be able to get around these types of queries as we take next steps like denormalizing hardware, querying for fingerprints, etc., especially if we want/need completeness for paginated APIs. I don't see easy solutions, but am worried about dragons in our unknowns. Am I repeating myself? I feel like I'm repeating myself, sorry.

conbench/entities/benchmark_result.py
jgehrcke: > IMO it shouldn't be required to POST "run_tags": {} if you don't have any run tags

Fully agree.

I don't know the answer really here yet; I just think it might be an important decision to make.

What might matter:

- In the DB, a JSON `null` value or a database `NULL` value (not the same) might make certain queries potentially more efficient than storing an empty JSON object (this is speculation).
- Easy-to-reason about (we agree on this goal). Example, I think when working in code I could like the following simplicity: when it's a dictionary then it it's a flat string/string mapping with at least one key (I would love for us to add this constraint if we can, to work against the nesting). And then there should be a precise, explicit way to check for "not set"; and here I often like the `is None` identity check in Python.

In what I wrote above I think the question of strictness in the JSON document might be more important to figure out initially. _Can_ we do a flat str/str mapping?

conbench/entities/_entity.py
```diff
@@ -139,7 +139,7 @@ def one(cls, **kwargs):
 
     @classmethod
     def first(cls, **kwargs):
-        return current_session.scalars(select(cls).filter_by(**kwargs)).first()
+        return current_session.scalars(select(cls).filter_by(**kwargs).limit(1)).first()
```
jgehrcke, line 142: This change got me curious.

Naive assumption was that `first()` does some LIMIT under the hood.

in the legacy query API this was [documented](https://docs.sqlalchemy.org/en/20/orm/queryguide/query.html#sqlalchemy.orm.Query.first):

> first() applies a limit of one within the generated SQL

But seemingly that was removed; from the migration guide (that's probably where you have it from?):

image

conbench/entities/benchmark_result.py
```diff
@@ -52,24 +54,23 @@ class BenchmarkResult(Base, EntityMixin):
...
-    # (which did bite us, in particular for those Run entities associated with
-    # O(1000) Result entities, i.e. hundreds of thousands of SELECT queries
-    # were emitted for building the landing page. A joined query can always be
-    # performed on demand.
-    run: Mapped[Run] = relationship("Run", lazy="select", back_populates="results")
+    # An arbitrary string to group results by CI run. There are no assertions that this
+    # string is non-empty.
+    run_id: Mapped[str] = NotNull(s.Text)
+    # Arbitrary tags for this CI run. Can be empty but not null.
+    run_tags: Mapped[dict] = NotNull(postgresql.JSONB)
```
jgehrcke: In another thread I asked

> Can we do a flat str/str mapping?

and if we can then we could switch this from type `dict` to something more precise which could help the code base.

conbench/entities/benchmark_result.py
```diff
@@ -223,38 +223,27 @@ def create(userres) -> "BenchmarkResult":
...
+        user_given_commit_info: Optional[TypeCommitInfoGitHub] = userres.get("github")
+        commit = None
+        if user_given_commit_info is not None:
+            commit = commit_fetch_info_and_create_in_db_if_not_exists(
+                user_given_commit_info
             )
-            # The above's `create()` might fail (race condition), in which case
-            # we can re-read, but then we also have to call
-            # `validate_run_result_consistency(userres)` one more time (because
-            # _we_ are not the ones who created the Run).
```
jgehrcke, line 255: It's good to see this type of concern go away. (I mean, why were' doing this :)).

conbench/entities/benchmark_result.py
```diff
@@ -508,20 +500,17 @@ def ui_hardware_short(self) -> str:
         """
         Return hardware-representing short string, including user-given name
         and ID prefix.
-
-        The indirection here through `self.run` is interesting, might trigger
-        database interaction.
         """
-        hw = self.run.hardware
+        hw = self.hardware
```
jgehrcke, line 522: :rocket:

conbench/entities/benchmark_result.py
jgehrcke: Thank you!

conbench/entities/benchmark_result.py
```diff
@@ -886,53 +891,6 @@ def validate_and_augment_result_tags(userres: Any):
...
-            )
-
-        # Cannot do this yet, this is too complicated as of None/empty
-        # string confusion repo_url_run = run.commit.repo_url
-        # repo_url_result = userres["github"]["repository"] if
-        # repo_url_run != repo_url_result: raise
-        #     BenchmarkResultValidationError( f"Result refers to
-        #         repository URL '{repo_url_result}', but Run
-        #         '{run.id}' " f"refers to repository URL
-        #     '{repo_url_run}'" )
```
jgehrcke, line 933: the change we want to see

conbench/entities/benchmark_result.py
```diff
@@ -1128,33 +1087,58 @@ class _BenchmarkResultCreateSchema(marshmallow.Schema):
         metadata={
             "description": conbench.util.dedent_rejoin(
                 """
-                Identifier for a Run (required). This can be the ID of a known
-                Run (as returned by /api/runs) or a new ID in which case a new
-                Run entity is created in the database.
+                Unique identifier for a "run" of benchmarks (typically a single run of a
```
jgehrcke: - Can we make this property already optional now? I'd love that in the short term :). But yeah, can do later.
- "Unique identifier" might sound slightly misleading (like the database helps maintaining uniqueness; technically from the system's point of view it's an arbitrary identifier).

We should specify:

> the API does not ensure uniqueness (and, correspondingly, does not detect collisions). If your use case relies on this grouping construct then use a client-side ID generation scheme with negligible likelihood for collisions (e.g., UUID type 4 or similar).


Overall, let's maybe introduce this with

> Arbitrary identifier that you can use to group benchmark results.

conbench/entities/benchmark_result.py
```diff
@@ -1128,33 +1087,58 @@ class _BenchmarkResultCreateSchema(marshmallow.Schema):
...
+                CI pipeline) on a single commit and hardware. Required.
+
+                The Conbench UI and API make several assumptions that all benchmark
```
jgehrcke: maybe just

> ... assumes that ...

conbench/entities/benchmark_result.py
```diff
@@ -1128,33 +1087,58 @@ class _BenchmarkResultCreateSchema(marshmallow.Schema):
...
-                useful for implementing a custom naming convention. For
-                organizing your benchmarks, and for enhanced search &
-                discoverability. Ignored when Run was previously created.
+                An optional mapping of arbitrary keys and values that describe the CI
+                run. These are used to group and filter runs in the UI and API. Do not
+                include `run_reason` here; it should be provided below.
+
+                For legacy reasons, if `run_name` is given when POSTing a benchmark
+                result, it will be added to `run_tags` automatically under the `name`
+                key. This will be its new permanent home.
```
jgehrcke, line 1137: `run_name` frustrates me a bit :)  https://github.com/conbench/conbench/issues/607

> The linked issues/PRs show that run_name creates work, while the value is still unclear to me.

but yeah, in this patch we still have to drag this on.

conbench/entities/benchmark_result.py
```diff
@@ -1128,33 +1087,58 @@ class _BenchmarkResultCreateSchema(marshmallow.Schema):
...
+                technical enforcement of this on the server side, so some behavior may
+                not work as intended if this assumption is broken by the client.
                 """
             )
         },
     )
+    run_name = marshmallow.fields.String(
+        required=False,
+        metadata={
+            "description": "A legacy attribute. Use `run_tags` instead. Optional."
```
jgehrcke, line 1150: :+1:

conbench/entities/run.py
```diff
@@ -436,50 +356,59 @@ def _guts(cinfo: TypeCommitInfoGitHub) -> Tuple[Commit, bool]:
...
 class _Serializer(EntitySerializer):
-    def _dump(self, run: Run, get_baseline_runs: bool = False):
-        if run.commit:
-            commit = CommitSerializer().one.dump(run.commit)
-            commit.pop("links", None)
+    def _dump(
+        self, benchmark_result: "BenchmarkResult", get_baseline_runs: bool = False
+    ):
+        if benchmark_result.commit:
+            commit_dict = CommitSerializer().one.dump(benchmark_result.commit)
```
jgehrcke, line 362: thanks for picking a variable name that is not just `commit` :).

conbench/entities/run.py
```diff
@@ -436,50 +356,59 @@ def _guts(cinfo: TypeCommitInfoGitHub) -> Tuple[Commit, bool]:
...
-            "error_type": run.error_type,
-            "commit": commit,
-            "hardware": hardware,
-            "has_errors": run.has_errors,
+            "info": benchmark_result.run_tags.get("info"),
+            "error_info": benchmark_result.run_tags.get("error_info"),
+            "error_type": benchmark_result.run_tags.get("error_type"),
+            "commit": commit_dict,
+            "hardware": hardware_dict,
+            "has_errors": benchmark_result.run_tags.get("has_errors"),
```
jgehrcke: This will also be interesting. In planning, we asked

> how is the definition of ‘errored run’ affected?

but we didn't answer so far. Current behavior: a run has `has_errors` set to true when any benchmark result in the run is failed. We do this here:

https://github.com/conbench/conbench/blob/42ef379825e0bb8d7cedc00f0d143e84f102c60b/conbench/entities/benchmark_result.py#L237

For maintaining that behavior we cannot look at just one result.

(I am super open to discussing if we want to change that behavior; if maintaining the behavior might be too expensive compared to the value it delivers)

conbench/api/runs.py
jgehrcke: > What, in your opinion, will be the fallout if I change this to allow results on all branches?

I made this change (https://github.com/conbench/conbench/issues/1307, https://github.com/conbench/conbench/pull/1308) so that the plots (the 'tinyplots') contain data points only for the default branch (and at the same time to not populate that cache with 'useless' data; to save memory).

This can of course be achieved with other means, too. Let me think ...

conbench/entities/run.py
```diff
@@ -436,50 +356,59 @@ def _guts(cinfo: TypeCommitInfoGitHub) -> Tuple[Commit, bool]:
...
             "timestamp": conbench.util.tznaive_dt_to_aware_iso8601_for_api(
-                run.timestamp
-            ),
+                benchmark_result.run_tags["timestamp"]
+            )
+            if "timestamp" in benchmark_result.run_tags
+            else None,
             "finished_timestamp": conbench.util.tznaive_dt_to_aware_iso8601_for_api(
-                run.finished_timestamp
+                benchmark_result.run_tags["finished_timestamp"]
```
jgehrcke: As part of this project I would have said that it's the right thing to remove the notion of 'finished' state, i.e. I think we can go a little further here and either remove the `finished_timestamp` key from the JSON doc, or set the value always to None.

conbench/entities/run.py
```diff
@@ -436,50 +356,59 @@ def _guts(cinfo: TypeCommitInfoGitHub) -> Tuple[Commit, bool]:
...
+        hardware_dict.pop("links", None)
+        out_dict = {
+            "id": benchmark_result.run_id,
+            "tags": benchmark_result.run_tags,
+            "reason": benchmark_result.run_reason,
+            "name": benchmark_result.run_tags.get("name"),
             "timestamp": conbench.util.tznaive_dt_to_aware_iso8601_for_api(
-                run.timestamp
-            ),
+                benchmark_result.run_tags["timestamp"]
```
jgehrcke: Maybe this should be the first /earliest-inserted benchmark result for this run ID (instead of a random / arbitrary one). Then this timestamp can be documented with 'when the first result came in'.

conbench/api/runs.py
jgehrcke: Gave this a tiny bit more thought.

I'd love to understand if it's possible to find a reasonably efficient method to get data straight from the database.

Driving the /runs HTTP API endpoints from the cache will impose a new challenge for the unit test suite; where the cache would need to be re-initialized between tests. That's of course solvable; if we have to.

Also, I didn't make this too explicit above but for Arrow Conbench, something like "the last 1000 runs" would become something like "the last 100 runs". Would that be OK?

As enthusiastic as I was yesterday about the BMRT cache potentially providing a short cut here; today I think we may need/want to force ourselves to build this using the DB only; and then really see how this behaves.

But yeah, still thinking. :).

conbench/entities/benchmark_result.py
austin3dickey: >might make certain queries potentially more efficient 

Yes, there [may be](https://stackoverflow.com/a/36286954) a slight performance increase using a database NULL value.

>And then there should be a precise, explicit way to check for "not set"

I was thinking that the only ways the server will have to work with this attribute are:
1. displaying the run tags nicely in the UI
2. filtering benchmark results to certain keys and values in the run tags for user-submitted queries

Originally I thought neither of these would necessarily require a check for "not set." For 1, a `for` loop would generalize to the empty object. For 2, it might actutally make the SQL a little more complicated to assert that this attribute is non-null.

This is not a strong opinion, just explaining why I made this choice originally 🙂 Happy to change it.

>_Can_ we do a flat str/str mapping?

I believe so. 

As of this PR, the only thing that will be put in here on the server side is the run name, which would be str/str. If users are interested in populating this field with more tags from the old Run object, they are going to have to do a code migration anyway, so during this they can probably flatten/stringify the tags they want to, like `timestamp`.

There's only one instance I know of using the `info` field, and it's a fairly nested JSON object with arrays and such. At worst we could simply store that same JSON as a string.

jgehrcke: Nice! Cool, the only real wish I have from this convo is to settle on a flat string/string mapping then :) (and to also validate that on the input side).

austin3dickey: Great, I'll add some validation then!

>like `timestamp`

@alistaire47 reminded me that clients couldn't ever set `run.timestamp` anyway, so really the only migration people might have to do is for the `info` field. (We're dropping the notion of the other mutable fields in this PR; see my upcoming comment)

conbench/api/runs.py
austin3dickey: We talked in-person and decided to power these "list runs" use cases using a variant of @alistaire47 's query that filters to all benchmark results from the last 30 days, no matter how many or few there are. This takes ~1-3 seconds on the Velox database and returns ~1200 runs. (I think this paradigm can set us up better for time-based pagination in the future anyway.)

conbench/entities/_entity.py
austin3dickey: Yes, isn't this wild? I went down the same rabbit hole and should have left a comment.

conbench/entities/benchmark_result.py
austin3dickey: >Can we make this property already optional now?

Yeah I think I'll do this later because I'm afraid of wading through our test suite. :) you know the feeling

Good docs ideas!

conbench/entities/benchmark_result.py
austin3dickey: I considered making it a no-op in this PR, but now it's tied up in the history endpoint:

https://github.com/conbench/conbench/blob/42ef379825e0bb8d7cedc00f0d143e84f102c60b/conbench/entities/history.py#L110

I still think there's value in cleaning that up in #607.

conbench/entities/run.py
austin3dickey: In this PR I'm basically dropping all first-class support for these old Run fields:
- finished_timestamp
- error_info
- error_type
- has_errors

So an "errored run" is not an official thing anymore. (People should still be able to hit `/api/benchmark-results/?run_id=ABC` and get `[]`, or `/api/runs/ABC` and get a 404, which may indicate something bad happened if they expected to find benchmark results.)

If people really want to submit those fields under `run_tags` with a BenchmarkResult, they can, but to me that doesn't make sense anymore. I think I'm actually going to drop those from this API run JSON instead of looking in `run_tags` for them.

(I'm also dropping support for `Run.timestamp`, which was never mutable but instead an insertion time. I'll keep that in the API run JSON for legacy reasons, by using the earliest benchmark result's timestamp.)

conbench/entities/run.py
austin3dickey: Agreed, see https://github.com/conbench/conbench/pull/1437#discussion_r1296123618

conbench/entities/run.py
austin3dickey: Good idea. At least this will make the API result deterministic.

conbench/api/runs.py
alistaire47: > time-based pagination

oh that's an interesting idea

conbench/entities/run.py
austin3dickey: Okay, update: with the new way we query for runs from the database when populating the Runs UI page, it's basically trivial to aggregate them. So I decided to keep the notion of an "errored run," except now it's scoped to
- viewing the Runs UI page (nowhere else)
- being True if and only if there's at least one associated benchmark result that `is_failed`

austin3dickey:
I responded to everyone's feedback. Thanks for all your time so far. Check out the last two commits if you want to see what changed after your reviews.

The only thing left (besides more feedback) is to fix the test suite, including an Alembic migration script and local sanity-checking.

conbench/api/runs.py
```diff
@@ -1,26 +1,64 @@
...
+    hashes) in a time range, and return a list of tuples, each corresponding to one
+    run_id. The tuples contain:
+
+    - the earliest BenchmarkResult for that run (by its timestamp)
+    - the number of benchmark results seen for that run
+    - whether any benchmark results are "failed" in that run
+
+    This powers the "list runs" API endpoint and the "recent runs" list on the home page.
+    """
+    run_info: Dict[str, Tuple[BenchmarkResult, int, bool]] = {}  # keyed by run_id
```
austin3dickey: Hm, I probably should change this into a dataclass. It's pretty similar to the RunForDisplay dataclass in app/index.py. Maybe I can consolidate.

conbench/api/runs.py
```diff
@@ -1,26 +1,64 @@
...
+        run_id = result.run_id
+        if run_id not in run_info:
+            run_info[run_id] = (result, 1, result.is_failed)
+        else:
+            old_result, old_count, old_failed = run_info[run_id]
+            run_info[run_id] = (
+                result if result.timestamp < old_result.timestamp else old_result,
+                old_count + 1,
+                old_failed or result.is_failed,
+            )
```
alistaire47: You could write this with `functools.reduce()`. Not sure it's better, really, but for me as someone who's written a lot of functional code it's a bit quicker to mentally parse

Makefile
```diff
@@ -34,7 +34,7 @@ tests: require-env-ghtoken set-build-info
 	docker compose down --remove-orphans && \
 	docker compose build app && \
 	docker compose run app \
-		pytest -vv -s --durations=20 conbench/tests/
+		pytest -vv --durations=20 conbench/tests/
```
austin3dickey, line 37: This is just a preference thing, but I noticed that when we printed the stdout while running the tests, it didn't print it next to the traceback when the tests failed. This is cleaner to me and makes more sense

conbench/tests/api/test_results.py
```diff
@@ -1064,30 +956,6 @@ def test_create_benchmark_timestamp_timezone(self, client, timeinput, timeoutput
 
         assert resp.json["timestamp"] == timeoutput
 
-    def test_create_result_mismatch_run_commit_hash(self, client):
```
austin3dickey, line 1067: Good riddance.

conbench/tests/app/test_runs.py
```diff
@@ -119,41 +118,7 @@ def test_baseline_run_links(self, client):
         self._assert_baseline_link(
             response.text,
             "latest_default",
-            benchmark_results[15].run_id,  # the latest default branch run in general
+            # earliest benchmark result on the latest default-branch commit
+            benchmark_results[9].run_id,
```
austin3dickey, line 122: This and similar changes to tests are because we're now tiebreaking the contender baseline runs by earliest benchmark result timestamp instead of run timestamp (which no longer exists). It still makes sense conceptually.

migrations/versions/2ee66194b9eb_remove_run_table.py
```diff
@@ -0,0 +1,104 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "2ee66194b9eb"
+down_revision = "dd6597846acf"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # TODO: data migration from run to benchmark_result goes here!
```
austin3dickey, line 20: will do this in a followup PR on Monday!

austin3dickey:
Merging into the [run_table_removal](https://github.com/conbench/conbench/tree/run_table_removal) branch. Will attempt to deploy that branch to staging on Monday.

conbench/api/runs.py
jgehrcke: Getting back to this; https://github.com/conbench/conbench/issues/1466.

PR #1441, opened 2023-08-18
legacy code: Separate out conbenchlegacy package

### LINKED ISSUE #1438: legacy package: Create separate package for legacy CLI and library
Breaking this out of #617 to be the first step: Duplicate the legacy code (`conbench` CLI, benchmark runner, `Conbench` class) into a new package. This does _not_ include removing that code from its current locations yet; that will happen separately after users have been switched over to the new legacy package.

alistaire47:
Closes #1438. Duplicates all the legacy code into a new package (currently named `conbenchlegacy`; happy to change that if there are opinions), but does not yet remove that code from the main codebase. Mostly does not change code (with a couple exceptions I'll call out in comments), but rearranges a lot of imports and makes necessary modifications to the package structure bits like `setup.py`, GHA testing, etc.


legacy/requirements-cli.txt
```diff
@@ -0,0 +1,7 @@
+click
+numpy
+psutil
+py-cpuinfo
+pytest>=7.0.0
```
alistaire47: I think this can be removed; I moved the new test dir outside of the package dir (which is a weird structure imo)

legacy/setup.py
```diff
@@ -0,0 +1,48 @@
...
+    },
+    classifiers=[
+        "Programming Language :: Python :: 3.8",
+        "License :: OSI Approved :: MIT License",
+    ],
+    python_requires=">=3.8",
+    maintainer="Apache Arrow Developers",
+    maintainer_email="dev@arrow.apache.org",
+    url="https://github.com/conbench/conbench",
+    install_requires=requirements_cli,
```
alistaire47, line 48: I didn't create a pared-down version of `requirements_dev` here; it's still depending on the top-level one, which will always be installed when this is running tests afaik. Might change this if we want total separation of concerns

legacy/tests/benchmark/test_runner.py
```diff
@@ -0,0 +1,246 @@
...
+    SimpleBenchmarkThatFails,
+    SimpleBenchmarkWithClusterInfo,
+)
+
+
+def validate_benchmark_result(result: dict) -> bool:
+    os.environ["CONBENCH_INTENDED_BASE_URL"] = "http://127.0.0.1:5000/"
+    from conbench.entities.benchmark_result import BenchmarkResultFacadeSchema
+
+    return BenchmarkResultFacadeSchema.create.validate(result)
```
alistaire47, line 23: Added this; hate it. This is the one place `conbenchlegacy` still depends on the `conbench` package. I'm not sure there's a good way to separate in unit tests here, as copying the file would require copying a significant part of the webapp, and would risk this getting out of sync with the real one anyway, defeating the purpose.

A better solution would be integration tests that use a live Conbench instance to validate the created json is correct.

Open to other ideas.

legacy/tests/conftest.py
```diff
@@ -0,0 +1,7 @@
+import flask
+import pytest
+
+
+@pytest.fixture
+def runner():
+    return flask.Flask("test").test_cli_runner()
```
alistaire47, line 7: The old version sets up a whole Conbench Flask app for this purpose, but this minimal version seems to work fine. I think `click` has testing utils we could switch to instead, but switching doesn't seem worth the time.

legacy/setup.py
```diff
@@ -0,0 +1,48 @@
...
+    version=__version__,
+    description="Continuous Benchmarking (CB) Framework Legacy Library and CLI",
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    packages=setuptools.find_packages(),
+    entry_points={
+        "console_scripts": [
+            "conbench=conbenchlegacy.cli:conbench",
+        ],
+    },
```
alistaire47, line 39: The new package name is currently `conbenchlegacy`, but this keeps the CLI name as `conbench` so installs and any imports will change slightly, but CLI calls will not.

We could call this package `conbench` and rename the main one to `conbenchwebapp` or whatever, but that would require changing a bunch of imports and would generally break a lot of things and be a lot more effort.

legacy/conbenchlegacy/_version.py
```diff
@@ -0,0 +1,2 @@
+# Do not touch this autogenerated file
+__version__ = "2023.4.10"
```
alistaire47: IMO we should not stick this package on PyPI; users are mostly installing from GitHub anyway. But maybe this version number should be incremented?

.github/workflows/actions.yml
```diff
@@ -54,7 +54,7 @@ jobs:
             -e ./benchadapt/python \
             -e ./benchconnect \
             -e ./benchrun/python \
-            -e .
+            -e ./legacy
```
austin3dickey, line 57: Can you make this same change here?
https://github.com/conbench/conbench/blob/42ef379825e0bb8d7cedc00f0d143e84f102c60b/.github/workflows/publish_docs.yml#L26

legacy/conbenchlegacy/_version.py
austin3dickey: Agreed we shouldn't put this on PyPI. So the version number doesn't matter. Could you remove the comment in this file?

legacy/tests/benchmark/test_runner.py
austin3dickey: yeah I agree this is the best option for now. We'll set up integration tests eventually. #720

austin3dickey (approved):
Thank you!

PR #1443, opened 2023-08-21
DB: data migration for removing run table

austin3dickey:
This PR fills out the data migration necessary for deploying this branch.

### DIFF
--- a/migrations/versions/2ee66194b9eb_remove_run_table.py
+++ b/migrations/versions/2ee66194b9eb_remove_run_table.py
@@ -17,11 +17,9 @@ depends_on = None
 
 
 def upgrade():
-    # TODO: data migration from run to benchmark_result goes here!
-
     op.add_column(
         "benchmark_result",
-        sa.Column("run_tags", postgresql.JSONB(astext_type=sa.Text()), nullable=False),
+        sa.Column("run_tags", postgresql.JSONB(astext_type=sa.Text()), nullable=True),
     )
     op.add_column("benchmark_result", sa.Column("run_reason", sa.Text(), nullable=True))
     op.add_column(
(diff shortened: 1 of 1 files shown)


austin3dickey:
The `db-migrations` job passed so I'll merge.

migrations/versions/2ee66194b9eb_remove_run_table.py
```diff
@@ -17,24 +17,36 @@
...
 
 def upgrade():
-    # TODO: data migration from run to benchmark_result goes here!
-
     op.add_column(
         "benchmark_result",
-        sa.Column("run_tags", postgresql.JSONB(astext_type=sa.Text()), nullable=False),
+        sa.Column("run_tags", postgresql.JSONB(astext_type=sa.Text()), nullable=True),
     )
     op.add_column("benchmark_result", sa.Column("run_reason", sa.Text(), nullable=True))
```
alistaire47, line 24: Should've brought this up on the previous PR, sorry, but I don't think it's a good idea from a user perspective for `run_reason` to be optional. It would be fine for users on a new project where there's only one type of run, but for all our current users, a missing `run_reason` is a mistake, and if we don't require it we won't warn them they're sending bad data.

We can have this conversation elsewhere, as this shouldn't hold up this PR, but we should take a minute for it.

migrations/versions/2ee66194b9eb_remove_run_table.py
```diff
@@ -17,24 +17,36 @@
...
+        sa.Column("hardware_id", sa.String(length=50), nullable=True),
     )
     op.drop_constraint("summary_run_id_fkey", "benchmark_result", type_="foreignkey")
     op.create_foreign_key(None, "benchmark_result", "hardware", ["hardware_id"], ["id"])
     op.create_foreign_key(None, "benchmark_result", "commit", ["commit_id"], ["id"])
 
+    op.execute(
+        """
+        UPDATE benchmark_result SET
+            run_tags = jsonb_build_object('name', coalesce(run.name, '')),
```
alistaire47, line 39: This is fine here, but going forward if `run_name` isn't specified, `name` just isn't in `run_tags`, right? Or `run_tags` is `{"name": null}`? I'd rather store nulls and have the `coalesce()` on the query side if we can

migrations/versions/2ee66194b9eb_remove_run_table.py
```diff
@@ -17,24 +17,36 @@
...
+            run_reason = run.reason,
+            commit_id = run.commit_id,
+            hardware_id = run.hardware_id
+        FROM run
+        WHERE benchmark_result.run_id = run.id
+        """
+    )
+    op.alter_column("benchmark_result", "run_tags", nullable=False)
```
alistaire47, line 47: So if `run_name` isn't specified this will just be `{}`? I thought JP was against that for perf reasons, though we're already doing that elsewhere, so if we want to change maybe we should do a larger cleanup

alistaire47 (commented):
A couple questions, but mostly about the larger schema changes; they shouldn't slow this down

migrations/versions/2ee66194b9eb_remove_run_table.py
austin3dickey: Yeah as you said I tried to minimize the number of changes in #1437 so I didn't want to touch nullability here.

migrations/versions/2ee66194b9eb_remove_run_table.py
austin3dickey: Yes, if `run_name` isn't posted with the benchmark result then there's no `name` key in run_tags. I think this is completely fine because I want to phase out run name in favor of users thinking about how to store better run tags.

migrations/versions/2ee66194b9eb_remove_run_table.py
austin3dickey: See my response [here](https://github.com/conbench/conbench/pull/1437#discussion_r1296011456) -- basically I think storing `null`s here would make our lives harder rather than easier.

PR #1444, opened 2023-08-21
list runs API: change default to 14d, add query param

austin3dickey:
This PR lowers the default time window searched during GET `/api/runs/` from 30d to 14d. It also adds a `?days=X` query parameter so that users can customize it.

### DIFF
--- a/conbench/api/runs.py
+++ b/conbench/api/runs.py
@@ -196,7 +196,9 @@ class RunListAPI(ApiEndpoint):
     def get(self):
         """
         ---
-        description: Get a list of runs from the last 30 days of benchmark results.
+        description: |
+            Get a list of runs from the last few days of benchmark results (default 14
+            days).
         responses:
             "200": "RunList"
             "401": "401"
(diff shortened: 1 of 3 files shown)


conbench/api/runs.py
```diff
@@ -205,17 +207,26 @@ def get(self):
...
         tags:
           - Runs
         """
         sha_arg: Optional[str] = f.request.args.get("sha")
         commit_hashes = sha_arg.split(",") if sha_arg else None
 
+        days_arg: Optional[int] = f.request.args.get("days")
+        days = int(days_arg) if days_arg else 14
+        if days > 30:
+            self.abort_400_bad_request("days must be no more than 30")
```
alistaire47, line 223: Long-term we definitely want a way to get older data, but an offset to provide pagination here is probably the solution we want

alistaire47 (approved):

conbench/api/runs.py
deepsihag: Yeah, pagination is the way!

deepsihag:
How does the home page keep it 30 days?

austin3dickey:
>How does the home page keep it 30 days?

Here's where it calls the same function with different parameters:

https://github.com/conbench/conbench/blob/6e228c5eefdca266af3ff7c65a27fe56601374ab/conbench/app/index.py#L47-L52

PR #1445, opened 2023-08-21
benchmark result API: add run link back

austin3dickey:
This PR adds the run link back to the benchmark result API.


alistaire47 (approved):

boshek (approved):
Thanks!

PR #1447, opened 2023-08-22
devprod: reorganize run code

austin3dickey:
After removing the Run table, this PR moves all the code from `entities/run.py` into `api/run.py` and `entities/results.py`. I think this is much easier for devs to reason about.

There are no code changes in this PR. Just moving things around.


alistaire47 (approved):
Makes sense, thanks!

PR #1448, opened 2023-08-23
api/results: document run_reason arg

austin3dickey:
This PR documents the api/results `run_reason` arg we forgot to document earlier.


alistaire47 (approved):
Lovely, thank you!

PR #1451, opened 2023-08-24
benchalerts: stop using /api/runs links key

austin3dickey:
I noticed that `benchalerts` was using one key from the runs API ("links") that was removed in #1450.

### DIFF
--- a/benchalerts/benchalerts/pipeline_steps/conbench.py
+++ b/benchalerts/benchalerts/pipeline_steps/conbench.py
@@ -104,6 +104,7 @@ class GetConbenchZComparisonForRunsStep(AlertPipelineStep):
             raise
 
         run_comparison = RunComparisonInfo(
+            conbench_api_url=self.conbench_client._base_url,
             contender_info=contender_info,
             baseline_run_type=self.baseline_run_type.value,
         )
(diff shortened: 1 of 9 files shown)


benchalerts/benchalerts/pipeline_steps/conbench.py
```diff
@@ -104,6 +104,7 @@ def _get_one_run_comparison(self, run_id: str) -> Optional[RunComparisonInfo]:
             raise
 
         run_comparison = RunComparisonInfo(
+            conbench_api_url=self.conbench_client._base_url,
```
alistaire47, line 107: Maybe we should un-underscore this? Or we could get it from the env var, which is always set when this is run, I'm pretty sure, but that's less direct.

alistaire47 (approved):

benchalerts/benchalerts/pipeline_steps/conbench.py
austin3dickey: Yeah I considered the env var but the client has some nice handling built in: 
https://github.com/conbench/conbench/blob/be2f5978c6af0291bab0accda1acfc65eb4a724d/benchclients/python/benchclients/conbench.py#L98

Someday we'll set up a nice integration test suite and won't have to worry about our libraries using each others' private attributes.

PR #1453, opened 2023-08-25
DB: store commit repo URL on benchmark_result

austin3dickey:
Part of #1165. Steps 1 and 2 of [this plan](https://github.com/conbench/conbench/issues/1165#issuecomment-1690557411).

This PR adds a nullable `commit_repo_url` column to BenchmarkResult. This is populated by the user-given `["github"]["repository"]` in the `POST /api/benchmark-results` payload if given.

It also makes the `["github"]["commit"]` property optional in the same payload. This is so we can switch over clients that today submit `"github": null` in transient commit use cases to `"github": {"repository": "..."}` instead.

### DIFF
--- a/conbench/api/_examples.py
+++ b/conbench/api/_examples.py
@@ -36,6 +36,7 @@ def _api_benchmark_entity(
     commit_id,
     parent_id,
     commit_type,  # "none", "known", "unknown"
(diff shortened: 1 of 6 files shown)


austin3dickey:
🚢

PR #1457, opened 2023-08-28
API POST result: require github prop again

austin3dickey:
This PR is part of #1165 and makes the `github` property required to post a benchmark result. This is so that each benchmark result is associated with a repository, even if it's not associated with a reproducible commit in that repository.

### DIFF
--- a/conbench/tests/api/test_results.py
+++ b/conbench/tests/api/test_results.py
@@ -517,6 +517,7 @@ class TestBenchmarkResultPost(_asserts.PostEnforcer):
         "run_id",
         "tags",
         "timestamp",
+        "github",
     ]
 
     def test_create_benchmark(self, client):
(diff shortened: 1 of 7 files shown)


conbench/tests/api/test_results.py
```diff
@@ -791,25 +783,6 @@ def _assert_commit_repo_without_hash(self, response):
...
-        location = "http://localhost/api/benchmarks/%s/" % new_id
-        self.assert_201_created(response, _expected_entity(benchmark_result), location)
-
-        # create another benchmark without commit context
-        # (test duplicate key duplicate key -- commit_index)
-        response = client.post("/api/benchmarks/", json=data)
-        benchmark_result, new_id = self._assert_none_commit(response)
-        location = "http://localhost/api/benchmarks/%s/" % new_id
-        self.assert_201_created(response, _expected_entity(benchmark_result), location)
-
```
deepsihag, line 812: Do we have test case to cover 'github' key validation?

austin3dickey: Yes, in the parent class of this test class we try to submit an object without each required key and assert that there's an error.

PR #1459, opened 2023-08-28
DB: make `benchmark_result.commit_repo_url` not null

### LINKED ISSUE #1165: API: allow submitting runs with known repository but unknown commit hash

austin3dickey:
Addresses the core problem of #1165.

This PR makes the `benchmark_result.commit_repo_url` column not null. It also switches the `get_candidate_baseline_runs` logic to find the latest default-branch commit for the given benchmark result's repository, since that is now always known.

### DIFF
--- a/conbench/api/results.py
+++ b/conbench/api/results.py
@@ -285,9 +285,9 @@ class BenchmarkListAPI(ApiEndpoint, BenchmarkValidationMixin):
             return resp400(str(exc))
 
         # Rely on the idea that the lookup
-        # `benchmark_result.commit.repo_url` always succeeds
+        # `benchmark_result.commit_repo_url` always succeeds
         conbench.metrics.COUNTER_BENCHMARK_RESULTS_INGESTED.labels(
-            repourl=benchmark_result.associated_commit_repo_url
+            repourl=benchmark_result.commit_repo_url
         ).inc()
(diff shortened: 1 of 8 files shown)


alistaire47 (approved):

deepsihag:
Was there any manual migration done to populate the `commit_repo_url` column for old records ?

austin3dickey:
>Was there any manual migration done to populate the commit_repo_url column for old records ?

Yes, I did this for some instances, but since we chose to hold off deploying to Arrow until the compare endpoint is sped up, that column doesn't exist there yet.

PR #1462, opened 2023-08-29
legacy: Remove legacy code from webapp

### LINKED ISSUE #1440: legacy package: Remove legacy code from webapp
This story is the final part of #617: to finally remove the legacy code from the main `/conbench` codebase once #1438 and #1439 are done.

There is a decision to be made here of whether to remove or leave the `setup.py` so the webapp can still be installed as a package. In theory I think it could be removed, but in practice it may be simpler to leave it (pared down to remove legacy bits).

alistaire47:
Closes #1440 by removing legacy code from the webapp library. This also allows some simplification of dependencies.


conbench/tests/api/_fixtures.py
```diff
@@ -290,7 +293,15 @@ def benchmark_result(
...
-        data["stats"] = Conbench._stats(results, unit, [], "s", one_sample_no_mean)
+        data["stats"] = validate_and_aggregate_samples(
+            {
+                "data": results,
+                "unit": unit,
+                "times": [],
+                "time_unit": "s",
+                "iterations": len(results),
+            }
+        )
```
alistaire47, line 304: Notable change; the `Conbench` class got moved to `conbenchlegacy`. This (which writes straight to the db) now uses the same code the API uses which is...better? Less bad?

austin3dickey: haha wow

austin3dickey (approved):
awesome!!

jgehrcke (commented):
:rocket: we talked in person, but I also still want to leave a big thank you here. Important work.

requirements-webapp.txt
```diff
@@ -22,15 +21,11 @@ orjson
...
-py-cpuinfo
 pytest>=7.0.0
 python-dotenv
-PyYAML
 requests
 sigfig
 SQLAlchemy>=2
 tenacity
 uuid7
-
```
jgehrcke, line 36: so good!

PR #1463, opened 2023-08-29
DB: formalize history_fingerprint concept

austin3dickey:
Addresses #862.

This PR adds a `history_fingerprint` column to BenchmarkResult, which is a formal identifier for a group of results that can be compared to each other. It also refactors some code to use this concept instead of rolling its own version of this concept.


migrations/versions/9bee3b519bf1_add_history_fingerprint.py
```diff
@@ -0,0 +1,36 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "9bee3b519bf1"
+down_revision = "167a97c81739"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # TODO before I merge this PR: fix this migration code!!
```
austin3dickey: I marked this PR as draft because I need to fix this migration before we merge. But I'd appreciate some eyes on the PR while I'm on PTO, if you have the time.

conbench/entities/history.py
```diff
@@ -419,27 +379,23 @@ def _query_and_calculate_distribution_stats(
...
+            ).all()
         )
-        history = history.join(
-            these_cases_and_contexts,
-            s.and_(
-                these_cases_and_contexts.c.case_id == BenchmarkResult.case_id,
-                these_cases_and_contexts.c.context_id == BenchmarkResult.context_id,
-            ),
+        history = history.filter(
+            BenchmarkResult.history_fingerprint.in_(these_fingerprints)
```
austin3dickey, line 395: I think just this change is going to speed up the compare endpoint by a lot.

austin3dickey:
Note: like I said above, this will be in draft for a couple of weeks while I'm on PTO. I still want to
- fix the data migration
- fix up the `api/compare.py` code (just a small change; could be a different PR)
- consider fixing up the BMRT cache code (this is a lot; could be a different PR)

conbench/entities/benchmark_result.py
```diff
@@ -87,6 +89,13 @@ class BenchmarkResult(Base, EntityMixin):
     hardware_id: Mapped[str] = NotNull(s.String(50), s.ForeignKey("hardware.id"))
     hardware: Mapped[Hardware] = relationship("Hardware", lazy="joined")
 
+    # The "fingerprint" (identifier) of this result's "history" (timeseries group).
+    # Two results with the same history_fingerprint should be directly comparable,
+    # because the relevant experimental variables have been controlled.
+    # Right now, those variables are benchmark name & case permutations, context,
```
alistaire47: Probably obvious for all of us, but this should emphasize it's a _single_ case permutation; it sounds like all of them. Maybe "set of case parameter values" or something

conbench/entities/benchmark_result.py
```diff
@@ -280,6 +289,12 @@ def create(userres) -> "BenchmarkResult":
         result_data_for_db["hardware_id"] = hardware.id
         result_data_for_db["commit_id"] = commit.id if commit else None
         result_data_for_db["commit_repo_url"] = repo_url
+        result_data_for_db["history_fingerprint"] = generate_history_fingerprint(
+            case_id=case.id,
+            context_id=context.id,
```
alistaire47, line 294: A future step, but this highlights it would be advantageous for these to be hashes too instead of UUID7s so the fingerprint is reproducible from the data posted without looking up anything.

conbench/entities/history.py
```diff
@@ -312,10 +273,13 @@ def set_z_scores(
     contender_benchmark_results: List[BenchmarkResult], baseline_commit: Commit
 ):
     """Set the "z_score" attribute on each contender BenchmarkResult, comparing it to
-    the baseline distribution of BenchmarkResults that share the same
-    case/context/hardware/repo, in the git ancestry of the baseline_commit (inclusive).
+    the baseline distribution of BenchmarkResults that share the same history
+    fingerprint, in the git ancestry of the baseline_commit (inclusive).
```
alistaire47, line 277: There's still some `cchr` naming in this file that could use cleaning up

conbench/entities/history.py
```diff
@@ -419,27 +379,23 @@ def _query_and_calculate_distribution_stats(
...
-    # Filter to the correct case(s)/context(s)
-    if case_id and context_id:
-        # filter to the given case/context
+    # Filter to the correct history fingerprint(s)
+    if history_fingerprint:
+        # filter to the given history_fingerprint
         history = history.filter(
-            BenchmarkResult.case_id == case_id,
-            BenchmarkResult.context_id == context_id,
+            BenchmarkResult.history_fingerprint == history_fingerprint
```
alistaire47, line 383: I was wondering why hardware wan't in here, then realized it's filtered in the query above. Can't we skip the hardware filtering there now, since this (or the version in the `else` clause below) will handle it already? 

I guess we still need to join to `hardware` to get `hardware_hash`, annoyingly; I wish that was just the ID so it was already on `benchmark_result`.

conbench/entities/history.py
```diff
@@ -783,21 +734,14 @@ def _detect_shifts_with_trimmed_estimators(
...
-        ],
+        ["history_fingerprint", "timestamp", "result_timestamp"],
         inplace=True,
         ignore_index=True,
     )
 
     # split / apply
     out_group_df_list = []
-    for _, group_df in tmp_df.groupby(["case_id", "context_id", "hash", "repository"]):
+    for _, group_df in tmp_df.groupby(["history_fingerprint"]):
```
alistaire47, line 741: yesss

conbench/types.py
```diff
@@ -1,11 +1,6 @@
...
+from typing import NewType
 
 TBenchmarkName = NewType("TBenchmarkName", str)
-
-
-# This type is used often. It's the famous 4-tuple defining a time series:
-# benchmark name, hardware ID, context ID, case ID. Maybe turn this into a
-# namedtuple or sth like this. Watch out a bit for mem consumption.
-Tt4 = Tuple[TBenchmarkName, str, str, str]
+THistFingerprint = NewType("THistFingerprint", str)
```
alistaire47, line 6: lovely

alistaire47 (approved):
Love it! Going forward I do think we should take steps to make this a deterministic hash rather than dependent on UUID IDs, but this is definitely a step towards where we want to go.

conbench/entities/benchmark_result.py
```diff
@@ -1025,6 +1041,30 @@ def _guts(cinfo: TypeCommitInfoGitHub) -> Tuple[Commit, bool]:
...
+
+    # The conventional field for checking whether two hardwares are compatible is
+    # "hash". This might change in the future:
+    # https://github.com/conbench/conbench/issues/1281
+    hash.update(hardware_hash.encode("utf-8"))
+
+    # Don't mix two repositories
+    hash.update(repo_url.encode("utf-8"))
+
+    return hash.hexdigest()
```
jgehrcke, line 1069: Good!

Future readers / devs might wonder about the choice of MD5 -- it's fine, it's fast, and I think collisions are practically impossible (but that back-of-the-envelope calculation should at some point maybe be done, so that at least current assumptions are documented: e.g., less than 10^6 individual fingerprints or something like that :shrug: ).

conbench/entities/benchmark_result.py
```diff
@@ -1040,6 +1080,11 @@ def _guts(cinfo: TypeCommitInfoGitHub) -> Tuple[Commit, bool]:
 # and index!
 s.Index("benchmark_result_timestamp_index", BenchmarkResult.timestamp)
 
+# An important index. "Give me all comparable results" is a very common query.
```
jgehrcke, line 1087: Thanks for adding that comment!

conbench/entities/benchmark_result.py
austin3dickey: Agreed; I chose the UUIDs for speed today. But it would be nice to clean up/possibly denormalize these attributes, and then it would be easy to make it a function of the tag set.

conbench/entities/history.py
austin3dickey: Hm, I don't see it. Where?

conbench/entities/history.py
austin3dickey: Ah yeah, we could skip the filter above. That's nice.

conbench/entities/benchmark_result.py
austin3dickey: I copied this comment directly into a code comment :) thanks!

conbench/entities/history.py
alistaire47: Nm, I don't see it now. Either I got confused by the gh diff or I'm missing it now 🤷

austin3dickey:
Will merge now; the only outstanding cleanup is in `conbench/bmrt.py` which @jgehrcke volunteered to do later. 🙂

migrations/versions/9bee3b519bf1_add_history_fingerprint.py
```diff
@@ -0,0 +1,57 @@
...
+
+    print(f"{datetime.now()} backfilling")
+    op.execute(
+        """
+        UPDATE benchmark_result SET history_fingerprint = MD5(
+            benchmark_result.case_id ||
+            benchmark_result.context_id ||
+            hardware.hash ||
+            benchmark_result.commit_repo_url
+        )
```
jgehrcke, line 34: Oh, did we confirm that this method yields the exact same output as https://github.com/conbench/conbench/pull/1463/files#diff-9d75d1bf6ac23bfc056a410721cf4c9bccd8b302f8e3e83e4b4bf1698d0eaad7R1044?

It's a bit of a risky move because yes the input to a hash function is a raw byte sequence and here we let the database use some of its implementation detail for how to store text, I think. This can only work (result in the same hash) if the database under the hood also stores text with the UTF-8 encoding and maybe that's true.

austin3dickey: Yes, I should have added a code comment. I agree that this is risky but I validated that it works for at least one real-world history fingerprint.

PR #1465, opened 2023-09-06
ui: compare view: adjust help text for z-score sign

jgehrcke:
After the consolidation work around units we can remove this caveat from the help text. This is for PFE-2025.


jgehrcke:
Landing this as a docs-only change.

Issue #1466, opened 2023-09-07
OOM on landing page: 'most recent runs' cannot be built based on results from last 30 days


jgehrcke:
I have not seen logs and I don't know where / how this was tried but @deepsihag mentioned that I think he has seen an OOM kill when loading the landing page for Arrow Conbench.

I looked into the current state of things. Yes, makes sense; we picked a bit of a bad method here.

Current method for fetching 'most recent runs' for the landing page: https://github.com/conbench/conbench/blame/a494e4a45818369580a3463cd71779239fa55737/conbench/app/index.py#L48C24-L48C39, rebuilt as part of https://github.com/conbench/conbench/commit/cced8f31997dd00b39aaa38df2783f698906ee1d.

I've mentioned this every now and then: it's helpful when we have these numbers roughly in our heads when building solutions. Arrow Conbench

- has a large number of results per run: O(10^3)
- ingests a large number of results per day: O(10^5)

Querying results of the last 30 days means asking the database for about 3*10^6 results. And that brings us back to the deeper investigation we did about memory consumption of "DB query responses" in the context of SQLAlchemy: https://github.com/conbench/conbench/pull/1213. Everything is stored at once on the heap before application code gets to process the data. A small number of millions of results are as of now expected to consume many gigabytes of heap, and an OOM kill is entirely expected.

So, processing results in a streaming-like fashion is one solution but in that case we would still spend most of the time processing a gigantic load of results for no good reason.

Let's not enter solution space too early, but I do believe we indeed need a DISTINCT-like method (as discussed [before](https://github.com/conbench/conbench/pull/1437#discussion_r1294986934)) to let the database throw away tons of data before it reaches application code.

Appendix, anecdotal evidence showing that Arrow Conbech ingested more than 100.000 results in the last 24 hours:
![image](https://github.com/conbench/conbench/assets/265630/bc455f2e-1621-4334-aab7-0f3c77d41cd2)

jgehrcke:
> I do believe we indeed need a DISTINCT-like method (as discussed https://github.com/conbench/conbench/pull/1437#discussion_r1294986934)

We want a fast 'skip scan' or a 'loose indexscan', seemingly. Some good resources:

https://wiki.postgresql.org/wiki/Loose_indexscan#Loose_Indexscan_versus_Index_Skip_Scan

https://dba.stackexchange.com/questions/177162/how-to-make-distinct-on-faster-in-postgresql

[...]

PR #1467, opened 2023-09-11
UI: landing page: query distinct runs from DB (address issue 1466)

jgehrcke:
This feeds the 'recent runs' listing on the landing page using a 'skip scan' query for efficiency, see #1466.

This patch mainly addresses the heap size / OOM challenges described in #1466: the result of this query (communicated to the web application) is a small number of result rows: `O(10^2)` instead of `O(10^6)`.

The result for this query is also generated rather quickly, which helps address a significant page load time issue (that we have not tracked in its own ticket here).

Cost of this approach (for now, maybe not fundamentally):

- unknown result count per run
- no guarantee about which result represents the run
- unknown "any result failed in this run?"

Future / thoughts:
- Measure: how does the page load time look for Arrow Conbench after this is deployed?
- Think: the method is still not ideal because for I think one DB query per result is emitted for fetching commit data. This will be addressed in future patches.
- The approximate result count per run can be fed from the BMRT cache
- The skip scan query result can be cached with small TTL (ideally with `result.commit` already being populated?)

### DIFF
--- a/conbench/api/runs.py
+++ b/conbench/api/runs.py
@@ -39,6 +39,7 @@ class RunAggregate:
 
     @functools.cached_property
     def display_commit_time(self) -> str:
+        # Note(JP): this is not the commit time
         return self.earliest_result.timestamp.strftime("%Y-%m-%d %H:%M:%S UTC")
 
     @functools.cached_property
--- a/conbench/app/index.py
+++ b/conbench/app/index.py
@@ -1,4 +1,3 @@
-import datetime
 import logging
 from collections import defaultdict
 from typing import Dict, List, Optional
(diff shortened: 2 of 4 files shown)


conbench/api/runs.py
```diff
@@ -39,6 +39,7 @@ def update(self, new_result: BenchmarkResult):
 
     @functools.cached_property
     def display_commit_time(self) -> str:
+        # Note(JP): this is not the commit time
```
austin3dickey, line 42: Ha, whoops. Nice catch

austin3dickey: As we wondered in-person: I wonder if we see similar bottlenecks using the code in this file for the query powering the `/runs` endpoint. After we measure, we could look at replacing this code with the new query.

austin3dickey (approved):
This looks great!

jgehrcke:
Let's see how this goes!

PR #1468, opened 2023-09-12
DB: history fingerprint migration: switch to CTAS

austin3dickey:
The previous migration was taking way too long on the Arrow data. Thanks to @jgehrcke 's advice and helpful reading like https://stackoverflow.com/a/3361903, we'll try switching the migration logic to create a completely new table first.

### DIFF
--- a/migrations/versions/9bee3b519bf1_add_history_fingerprint.py
+++ b/migrations/versions/9bee3b519bf1_add_history_fingerprint.py
@@ -7,8 +7,13 @@ Create Date: 2023-08-29 19:33:34.654753
 """
 from datetime import datetime
 
-import sqlalchemy as sa
 from alembic import op
+
+
+def p(msg):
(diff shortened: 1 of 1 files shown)


austin3dickey:
🚢

jgehrcke:
:rocket:

migrations/versions/9bee3b519bf1_add_history_fingerprint.py
```diff
@@ -18,36 +23,75 @@
...
+    p("-- recreating foreign keys --")
+    for tablename in ["hardware", "commit", "case", "info", "context"]:
+        p(tablename)
+        op.create_foreign_key(
+            None,
+            "benchmark_result",
+            tablename,
+            [f"{tablename}_id"],
+            ["id"],
+        )
```
jgehrcke, line 92: Of course creation of indices etc might also take quite a while. We will see :)

Issue #895, opened 2023-03-17
ci instability: Exception: Hourly GitHub HTTP API quota exhausted


jgehrcke:
forgot previously to create a ticket tracking this precise symptom (which is still new-ish):
```
            if reqquota == 0:
                metrics.COUNTER_GITHUB_HTTP_API_REQUEST_FAILURES.inc()
>               raise Exception("Hourly GitHub HTTP API quota exhausted")
E               Exception: Hourly GitHub HTTP API quota exhausted

conbench/entities/commit.py:745: Exception
```

jgehrcke:
Prior occurrence: https://github.com/conbench/conbench/pull/784#issuecomment-1448345472

jgehrcke:
Prior occurrence: https://github.com/conbench/conbench/pull/667#issuecomment-1415952159

jgehrcke:
happened here: https://github.com/conbench/conbench/pull/893#issuecomment-1473696950

austin3dickey:
In #882 I [added](https://github.com/conbench/conbench/pull/882/files#diff-f6c5e9c7299d96d92fc5e6a8d0a84579587bc876777f23d9a780010f3efae70fR93) an integration test that posts [GitHub comments](https://github.com/conbench/benchalerts/pull/5#issuecomment-1473995381) every time the test suite runs in CI. That wasn't my intention; it causes lots of noise and probably adds to our rate limit headache. I will skip those tests in CI.

jonkeane:
Even after #899 we're still running into this pretty consistently. Has anyone started to poke at how we might resolve this?

A few ideas (none are perfect or great):
* a githubapp _might_ help, though that might only be for those that are on an enterprise account: https://docs.github.com/en/apps/creating-github-apps/creating-github-apps/rate-limits-for-github-apps#installation-access-tokens-on-github-enterprise-cloud
* we could mock some of the interactions (i.e. the responses) we use (downside: we have to care for the mocks if and as the GH API changes; upside: no HTTP interaction at all when we use them)
* refactor some of these tests in ways that isolate the GH interactions, test those, and then mock the GH-interacting functions when testing so that we don't hit that API).

Thoughts?

jgehrcke:
> Thoughts?

Quick fix: use more than one API token, rotate upon receiving expected error, retry

More fundamentally, I look at #655 as the mid term change that will qualitatively change the way we interact with the GitHub HTTP API.

austin3dickey:
[...]

PR #1470, opened 2023-09-12
api/runs/run_id: speed up query

austin3dickey:
This PR is a ~small~ internal change to speed up the query executed to find baseline runs for a given contender run.

It takes advantage of the new history fingerprint.

### DIFF
--- a/conbench/api/runs.py
+++ b/conbench/api/runs.py
@@ -1,7 +1,7 @@
 import dataclasses
 import datetime
 import functools
-from typing import Dict, List, Optional, Sequence
+from typing import Dict, List, Optional, Sequence, Set
 
 import flask as f
 import flask_login
@@ -15,7 +15,8 @@ from ..dbsession import current_session
 from ..entities._entity import EntitySerializer
 from ..entities.benchmark_result import BenchmarkResult
 from ..entities.commit import CantFindAncestorCommitsError, Commit, CommitSerializer
-from ..entities.hardware import Hardware, HardwareSerializer
+from ..entities.hardware import HardwareSerializer
+from ..types import THistFingerprint
 from ..util import short_commit_msg, tznaive_dt_to_aware_iso8601_for_api
 
 
@@ -115,7 +116,7 @@ class _CandidateBaselineSearchResult:
 def _search_for_baseline_run(
     contender_run_id: str,
     contender_run_reason: Optional[str],
-    contender_hardware_checksum: str,
+    contender_history_fingerprints: Set[THistFingerprint],
     baseline_commit: Optional[Commit],
     commit_limit: int = 20,
 ) -> _CandidateBaselineSearchResult:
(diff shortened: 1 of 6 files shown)


austin3dickey:
From what I can tell, on Arrow-scale data, this brings the endpoint back to perform about as well as before the run table removal.

austin3dickey:
working on tests (I don't *think* this is a user-facing change; I think it's just another case where inserting benchmark results directly into the database during tests rather than using the API is hurting devprod)

conbench/tests/api/_fixtures.py
```diff
@@ -268,6 +268,7 @@ def benchmark_result(
     data["run_id"] = run_id if run_id else _uuid()
     data["batch_id"] = batch_id if batch_id else _uuid()
     data["tags"]["name"] = name if name else _uuid()
+    data["timestamp"] = datetime.now(timezone.utc).replace(microsecond=0).isoformat()
```
austin3dickey, line 271: I think it makes more sense in this function to use `now()` by default as the timestamp, instead of a hardcoded one. This caused most of the test changes in this PR but also fixed the tests that were broken originally, since the new query uses the result timestamp.

jgehrcke:
`ci/ui-screenshots` is happy again in `main` now.

conbench/api/runs.py
```diff
@@ -115,7 +116,7 @@ def _dict_for_api_json(self) -> dict:
 def _search_for_baseline_run(
     contender_run_id: str,
     contender_run_reason: Optional[str],
-    contender_hardware_checksum: str,
+    contender_history_fingerprints: Set[THistFingerprint],
```
jgehrcke, line 119: Thanks for making this a set!

conbench/api/runs.py
```diff
@@ -142,72 +143,64 @@ def _search_for_baseline_run(
...
+            s.desc("commit_order")
+        ).limit(commit_limit)
     except CantFindAncestorCommitsError as e:
         return _CandidateBaselineSearchResult(
             error=f"could not find the baseline commit's ancestry because {e}"
         )
 
+    commits = commit_query.all()
+    commit_ids = [commit.ancestor_id for commit in commits]
+    commit_hashes = [commit.ancestor_hash for commit in commits]
```
jgehrcke, line 156: In terms of code readability and head scratchability it might be worth explaining why we use `[commit.ancestor_id for commit in commits]` instead of `[commit.id for commit in commits]` :) Same for hash.

There probably is a good reason, it looks like one of those really important off-by-one considerations!

jgehrcke: Maybe this is mainly about the `commits`  variable name vs `commit_ids` which kind of suggests that they both refer to the same set/collection of commits. But that's probably not true. Maybe we want `commit_ids` and then `ancestor_commit_ids`? Well. :shrug: 

And not important but warming up the single-letter var conversation: in list comprehensions I am super cool with single letter loop vars, e.g. `[c.id for c in commits]`.

Sorry, moving on :D.

conbench/tests/api/_fixtures.py
jgehrcke: Yes!!

conbench/api/runs.py
austin3dickey: yeah, good questions! This confusion comes from the fact that `commits` is the result of executing the [commit_ancestry_query](https://github.com/conbench/conbench/blob/9ca3c6d84ec3a0ab47c2f3368b4bbe2dcca5479b/conbench/entities/commit.py#L180), which prefixes its colnames with `ancestor_` to disambiguate them from when we join it to the actual `Commit` table.

But we don't join it anymore in this query, and I'm planning to remove the join in the z-score query in my next PR. So I'll clean this up then.

austin3dickey:
CI failures are due to #895. Merging.

PR #1475, opened 2023-09-14
api/compare/runs: speed up

austin3dickey:
This PR is an internal change to speed up the compare endpoint by simplifying the query used to get historic results for the lookback z-score calculation.

I'm using this to test:
https://conbench.ursa.dev/api/compare/runs/5f3caa45241e469db3fb9f791cf6670e...f7fcdc05191c4cc4892a761f2208f7ea/

which has about 3500 results per run, each having a different history fingerprint, and therefore needs to pull about 350K results to do the z-score calculations (with a lookback of 100 commits). 

- On the current Arrow deployment (before we removed the run table), assuming no query timeout, this call takes about 75 seconds. 
- On the current head of `main` on staging with Arrow data, it takes about 210 seconds. 
- After this PR, it takes about 30 seconds. 

That's still long, and I have an idea for a user-facing change to speed it up (pagination), but this is drastically better than this endpoint has been since #636 in February.

### DIFF
--- a/conbench/entities/history.py
+++ b/conbench/entities/history.py
@@ -348,32 +348,43 @@ def _query_and_calculate_distribution_stats(
     ``_add_rolling_stats_columns_to_df()``.
     """
     try:
-        commits = baseline_commit.commit_ancestry_query.subquery()
+        commit_ancestry_query = baseline_commit.commit_ancestry_query.order_by(
+            s.desc("commit_order")
+        ).limit(Config.DISTRIBUTION_COMMITS)
     except CantFindAncestorCommitsError as e:
         log.debug(f"Couldn't _query_and_calculate_distribution_stats() because {e}")
         return {}
 
-    # Get the last DISTRIBUTION_COMMITS ancestor commits of the baseline commit
-    commits = (
-        current_session.query(commits)
-        .order_by(commits.c.commit_order.desc())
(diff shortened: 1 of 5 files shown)


austin3dickey:
Pleasantly surprised. I'll probably merge this soon, but please give feedback, especially on the tradeoff of not using our canonical SVS property for this code.
image

conbench/entities/history.py
```diff
@@ -348,32 +348,43 @@ def _query_and_calculate_distribution_stats(
...
+        BenchmarkResult.commit_id,
+        BenchmarkResult.history_fingerprint,
+        BenchmarkResult.timestamp.label("result_timestamp"),
+        BenchmarkResult.change_annotations,
+        # Note[austin]: Okay. Pros and cons here. Of course the SVS is not exactly
+        # equivalent to the mean in all cases because of "errored results", and that's
+        # especially true if we change the default definition of SVS in the future (in
+        # which case let's revisit this!). But after careful analysis of the code, I
+        # believe that they are equivalent for BenchmarkResults where "error" is None
+        # and "mean" is not None, at least since #1127 or previous.
```
jgehrcke, line 377: :+1:  good enough for now, of course interesting aspect for future evolution. Thanks for the comment.

PR #1476, opened 2023-09-15
landing page: fetch less runs (more reasonable load time on cache miss)

jgehrcke:
This patch

- does a rename from `fetch_one_result_per_n_recent_runs()` to `fetch_one_result_per_each_of_n_recent_runs()` which is logically more expressive/correct
- adds the `n` parameter to the function
- sets the default to 250 (previously: 500)


jgehrcke:
:ship:

PR #1477, opened 2023-09-15
big arrow migration branch

austin3dickey:
A temporary PR to combine and deploy our latest changes to Arrow.


migrations/versions/9bee3b519bf1_big_arrow_migration.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        CREATE TABLE temp AS
+        WITH result AS (
+            SELECT
+                benchmark_result.*,
+                jsonb_build_object('name', coalesce(run.name, '')) AS run_tags,
+                run.reason AS run_reason,
+                run.commit_id AS commit_id,
+                run.hardware_id AS hardware_id,
+                COALESCE(commit.repository, 'https://github.com/apache/arrow') AS commit_repo_url
```
jgehrcke, line 35: oh! :) nice.

alembic.ini
austin3dickey, line 1: The changes to this file are probably worth a separate PR; I think they'll enable using `logging.getLogger()` in migration scripts, which is helpful for seeing live migration progress.

migrations/versions/9bee3b519bf1_big_arrow_migration.py
```diff
@@ -0,0 +1,179 @@
...
+
+"""
+import logging
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "9bee3b519bf1"
```
austin3dickey, line 15: Not 100% sure that manually making this ID the same as the most recent one on `main` will work. Wonder if there's some magic hashing behind the scenes. Probably not; this will probably be fine.

migrations/versions/9bee3b519bf1_big_arrow_migration.py
austin3dickey: Yeah! I manually validated that all `commit` rows have a URL filled in for the `repository` column in the Arrow DB. So the COALESCE is just to cover any benchmark results not associated with a commit (of which I don't think there are any, but it's slow to check).

jgehrcke (approved):
Amazing work, good luck!

alembic.ini
austin3dickey: nice yep the log changes worked 
image

austin3dickey:
Note: I forgot that we had a few corresponding non-breaking changes to clients too (https://github.com/conbench/conbench/pull/1451, https://github.com/conbench/conbench/pull/1456). I can bump those for Arrow when the server is in downtime.

migrations/versions/9bee3b519bf1_big_arrow_migration.py
austin3dickey: This was fine.

austin3dickey:
Cool, done.

PR #1492, opened 2023-09-26
Increase DB statement timeout

austin3dickey:
To address https://github.com/voltrondata-labs/arrow-benchmarks-ci/issues/164. A certain compare runs query against the Arrow database is taking 97 seconds to complete. Let's increase the timeout before we fix this for real.


austin3dickey:
CI instability: #1493

PR #1494, opened 2023-09-27
ci: kube-prom: jb install from local dir (respect pyrra version pin), bump to kube-prometheus 0.13

### LINKED ISSUE #1493: deployment pipeline: cannot `jb install kube-prometheus`
Seen, for example, [here](https://github.com/conbench/conbench/actions/runs/6318525930/job/17157695989).

When running this command:

https://github.com/conbench/conbench/blob/89ddc4b5745c8bbef74900a5a8ff7d4f51caa0d9/Makefile#L316-L317

we see

```
GET https://github.com/pyrra-dev/pyrra/archive/f2052cce0c848bb8c4c4d5d964a97b5038237487.tar.gz 200
panic: rename vendor/.tmp/jsonnetpkg-github.com-pyrra-dev-pyrra-config-crd-bases-main336769959/config/crd/bases vendor/github.com/pyrra-dev/pyrra/config/crd/bases: no such file or directory
```

I think there might be something wrong with the vendored `pyrra` repo? We got a different tar the [last time](https://github.com/conbench/conbench/actions/runs/6303291179/job/17112233182) the run succeeded.

jgehrcke:
This is to resolve #1493. High-level root cause is a typical moving target problem triggered by the [pyrra 0.7 release from about one day ago](https://github.com/pyrra-dev/pyrra/commit/f2052cce0c848bb8c4c4d5d964a97b5038237487). Note that kube-prometheus tries to version-pin the precise version of pyrra to be used in the build. But that version pin does not seem to apply using the installation method we have picked so far (following docs). I have reported that via https://github.com/prometheus-operator/kube-prometheus/issues/2224 and also documented my initial findings over there.

This patch here makes use of [this finding](https://github.com/prometheus-operator/kube-prometheus/issues/2224#issuecomment-1737247869).

Note that while doing this I have also taken liberty to bump to the kube-prometheus 0.13 release. Let's see if this builds fine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -311,10 +311,10 @@ set-build-info:
 .PHONY: jsonnet-kube-prom-manifests
 jsonnet-kube-prom-manifests:
 	mkdir -p _kpbuild && cd _kpbuild  && mkdir -p cb-kube-prometheus
-	cd _kpbuild/cb-kube-prometheus && echo '{"version": 1, "dependencies": [], "legacyImports": true}' > jsonnetfile.json
+	cd _kpbuild/cb-kube-prometheus && git clone https://github.com/prometheus-operator/kube-prometheus . &&	git checkout 7fafc4cadc1
 	cd _kpbuild/cb-kube-prometheus && \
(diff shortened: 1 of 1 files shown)


jgehrcke:
Looking good; let's see how well this deploys to vd-2!

PR #1495, opened 2023-09-28
webapp: plot benchmark results missing commits

### LINKED ISSUE #1489: webapp: result, compare results: plots should show contender results missing commits
Now that https://github.com/conbench/conbench/issues/1460 is closed, we show timeseries plots on the following pages, even if the contender result is not associated with a commit:
- `/benchmark-results/{contender_id}/`
- `/compare/benchmark-results/{baseline_id}...{contender_id}/`

The plots show the non-errored results on the default branch in the same repo as the contender (with the same history fingerprint).

Often, when users inspect these plots, they want to see the contender result *in context of* these historic results. Today we show the mean and min of the contender result in magenta if it is associated with a commit. But we don't if it isn't, because there's no associated commit timestamp. I think we should try to do so anyway.

There are a few choices for how we'd choose the x-axis value of this point. I think the most intuitive choice is the result's associated user-given timestamp. We could put a disclaimer somewhere that it's not a commit timestamp, but since it will probably be in the same ballpark, I don't think that will affect user analysis much.

austin3dickey:
Closes #1489. Adds the "current result" to the plots even if they're missing a commit. I chose to use the result timestamp as the x-axis value in this case.

Deserves more thinking in #633, but this is a nice-to-have for now.

### DIFF
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -905,27 +905,33 @@ def time_series_plot(
 
 
 def get_source_for_single_benchmark_result(current_benchmark_result, cur_run, unit):
-    # Edge case: this may be failed or the commit time might be unknown, in
-    # which case we do not show a data point (for now).
-    # source_current_bm_min = None
-    # source_current_bm_mean = None
+    # Use the commit timestamp for the x-axis, unless it doesn't exist, in which case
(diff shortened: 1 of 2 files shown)


deepsihag (approved):
🚀

Issue #871, opened 2023-03-15
API endpoints: allow for reading data in chunks (introduce pagination)


jgehrcke:
First, we need to optimize the DB schema for a tail view lookup: #789, and we will need to track insertion time: #790. Then we can have each endpoint that emits a collection of items by default limit the count and/or the covered time window (`now - now-D days`).

Count/Time window properties can be set by the client. This will be an important step forward.

Next, for covering the use case of reading out larger counts/timeframes (resulting in larger object count), chunking (pagination) becomes important. A good overview over the variety of methods is in this paragraph: https://microservice-api-patterns.org/patterns/quality/dataTransferParsimony/Pagination#sec:Pagination:Variants

Between people, we seem to be aligned that pagination will be important:

Quoting myself from #786

> It appears that in the foundation of the API we did not really consider pagination so far.

Quoting @alistaire47 from the same issue:

> Pagination here seems obvious (assuming we can sort sweat_smile). As this is really a results endpoint anyway, maybe we should recreate it paginated as /results or /benchmark-results and deprecate this?

Also @boshek:

> +1 to pagination. 

From #746, @austin3dickey wrote

> add pagination to GET endpoints

jonkeane:
Another place that pagination (especially of runs) is super important: letting folks walk back deep in history. There is currently a user trying to compare today to 3 months ago. The compare picker at the bottom of conbench is limited to the first 1000 commits (and they also can't find the run id to construct the compare URL by hand easily since they can't page back in history).

We won't solve this specific use case with only pagination, but pagination (on runs) would make it slightly less painful.

austin3dickey:
>A good overview over the variety of methods is in this paragraph: https://microservice-api-patterns.org/patterns/quality/dataTransferParsimony/Pagination#sec:Pagination:Variants

This is nice, thank you! With the UUID7 primary keys we enjoy today, and the semi-constant arrival of data, I think the best option for us is "cursor-based pagination". Since we care about the tail view, we'd iterate backwards in time/IDs. The query to get any non-first page would look something like 
```sql
SELECT *
FROM benchmark_result
WHERE id < $cursor
ORDER BY id DESC
LIMIT $page_size
```
This of course takes advantage of the serial, unique, and indexed properties of the UUID7 primary key. After grabbing that chunk of data we could also return a `"next_page_cursor": "..."` key in the response that would be the `min(id)` of the chunk of data.

[...]

Issue #968, opened 2023-03-28
webapp: skip API layer re-serialization of benchmark results


jgehrcke:
Analogous to https://github.com/conbench/conbench/issues/890.

Affects app/benchmarks.py (module name wants to be renamed, too).

jgehrcke:
In an in-person meeting we had an important discussion about this topic.

We talked about the idea the feature parity between UI and API is important, and we fully agree on this.

We discussed whether this kind of architectural change here is endangering the idea above. Maybe! But this can be contained by paying attention to this 'constraint' / paradigm.

I want to stress that I am certain that the most productive argument _for_ doing this architectural change is rather _orthogonal_ to the feature parity argument.

This indirection is a major reason (among other, independent reasons) for why navigating the code base feels like walking in glue. I don't want to judge past decisions and actions but for our planning and decision-making we have to absolutely sync on and respect the current state of maintainability and developer productivity across the code base.

The following example is just a tiny one, out of hundreds. Exercise: identify behavior of this in terms of what it is doing and what it returns: https://github.com/conbench/conbench/blob/26dad7900eaa8abac18999da4e9677bab575dfe6/conbench/app/compare.py#L173

I think you will be occupied for a while :lotus_position: .

The _cost_ of having this indirection is large (I mean it). It's hard to measure and hard to talk about objectively, but when you _do_ work in the code base you see it.

I have seen and built systems where the UI that consumes (only) an HTTP API to build views. A key property of such an architecture is that the HTTP client runs on the user's device (single-page application, typically. Built with React or comparable). It's certainly an option for us, but I think the current approach is actually quite flexible and nice. We should go all-in for now and make it _easy_ for us to build UI :).

Issue #872, opened 2023-03-15
UI & API endpoints for collections: build time range filters / "The Time Window Paradigm"


jgehrcke:
It is more or less obvious to us that we have to first-class the idea of "I am looking at a time slice worth of data", in the UI and also in the API. Related quotes:

> we can have each endpoint that emits a collection of items by default limit the count and/or the covered time window (now - now-D days).

(https://github.com/conbench/conbench/issues/871)

> This is important for being able to query an endpoint that emits a collection of items; to reveal data only from a specific time window

(https://github.com/conbench/conbench/issues/790)

From #746, @boshek wrote:

> Ideally it would be by time range and then by some other filters but a discrete win here could be the time range filter.

#790 (`db schema: store insertion/creation timestamp for entities`) is a requirement, and there we wrote:

> This is required for properly building time window / range filters for entities (exposed via API).

What do we do instead of doing this, you might ask? Well, we provide a "tail end view", with different limits depending on where you look. The UI landing page, for example, is populated with "the last 1000 runs".

------

I am updating this to emphasize that the "The Time Window Paradigm" really seems to be foundational to our problem space here. How others expose this via UI:

### Datadog log viewer

Top-right corner:
![image](https://user-images.githubusercontent.com/265630/234577500-25ac6f25-466b-4b1a-92d3-98016c1c2346.png)

### AWS Cloudwatch dashboard

Top-right corner:
![image](https://user-images.githubusercontent.com/265630/234578364-a853045e-f193-42d2-8c12-c97c90213939.png)

### Grafana dashboard

Top-right corner:

![image](https://user-images.githubusercontent.com/265630/234578512-a4bbb79c-c255-4784-84ee-aa215b84f1a2.png)

Alternatively, providing a specific time window:

![image](https://user-images.githubusercontent.com/265630/234578679-b1116a83-d76c-419c-a572-3b2d927440cf.png)

jgehrcke:
@joosthooz from Slack:

> We've been posting profiling results to the staging conbench every week, but after a couple of days they disappear from the radar.

@jonkeane from #871 

> There is currently a user trying to compare today to 3 months ago. The compare picker at the bottom of conbench is limited to the first 1000 commits

austin3dickey:
I'm not opposed to the idea of using the user-given benchmark result timestamp for these views instead of the DB insertion time. We've found those queries can still be fast.

Issue #746, opened 2023-02-15
Streamlining API calls for performance (faster&smaller HTTP responses via filters etc)


boshek:
:wave: Thanks for everyone's efforts on conbench. It is really so great. Some thoughts and use patterns of the API.

I am frequently making use of the conbench API (via the [conbenchcoms](https://github.com/conbench/conbenchcoms) R package) to grab run and benchmark data. For static reporting this works well enough since all the data is pulled in a single event and used to generate a static report. However for dynamic reports and dashboards, the current api doesn't have enough flexibility to tailor queries making interactive documents and apps that use conbench data slow to load. 

This is a typical this pattern: 

``` r
tictoc::tic()
library(conbenchcoms)
library(dplyr)

last_1000_runs <- runs()

## Do some manipulations on the recent runs. Here so I'd like only those runs that
## use the ursa-i9-9960x machine
i9_df <- last_1000_runs |> filter(hardware.name == "ursa-i9-9960x")

## query length limits means I have to split this up into separate calls
split_id <- split(i9_df$id, ceiling(seq_along(i9_df$id) / 50))

bench_df <- lapply(split_id, benchmarks) |>
  bind_rows()

## join back with the benchmark info
results_raw <- bench_df %>%
  left_join(i9_df, by = c("run_id" = "id")) |>
  select(id, commit.timestamp, commit.sha, hardware.architecture_name)

tictoc::toc()
#> 117.784 sec elapsed
```

In this pattern I am bringing along so much data that I really don't need which really does slow things down. Waiting 117 seconds for app to grab data isn't going to work. Certainly I could streamline that a bit by using the GitHub api to first tailor the commits I am after etc. But since conbench already has done that it would be nice to make use of what we have already. Ideally I'd be able to have more query patterns than just the id|commit. IIRC you can query on a z-score threshold too. Things like time would be particularly nice but really the more variables we can tune a query with the better. 

In addition to reducing the number of rows returns, it would also be nice to be able to make api calls that reduce the number of columns that are returned. Each column is enormously useful but I don't need them all the time. For example `results_raw` has 105 columns. 

I am also finding that I rarely need benchmarking data without runs data - if we were able to implement the above, then doing that join on the server side would be _amazing_.

jgehrcke:
Thank you so much! I have not given this deeper thought yet, but will so in the future!

Edit: added "(faster&smaller HTTP responses via filters etc)" to the title to reflect what I understood is the proposal here.

[...]

PR #1496, opened 2023-10-11
api/benchmark-results: add pagination

austin3dickey:
Helps with #871. Adds pagination to the `GET /api/benchmark-results` endpoint. See that issue for the justification for the method that I chose, and the docstrings for the details.


conbench/app/batches.py
```diff
@@ -43,6 +43,9 @@ def get(self, batch_id):
             self.flash("Error getting benchmarks.")
             return self.redirect("app.index")
 
+        # TODO: paginate
+        benchmarks = benchmarks["data"]
```
austin3dickey, line 47: For now I made this quick fix (and the corresponding changes in `app/results.py` and `app/runs.py`) to get the test suite green. However, I think this will limit these webapp pages to only showing the last 100 matching results. That's not super desirable so I want to look into how we can take better advantage of pagination on these webpages.

conbench/api/results.py
```diff
@@ -157,96 +137,108 @@ def get(self) -> f.Response:
...
+                Cursor for pagination. To get the first page of results, leave out this
+                query parameter or submit `null`. The response's `metadata` key will
+                contain a `next_page_cursor` key, which will contain the cursor to
+                provide to this query parameter in order to get the next page. (If there
+                is expected to be no data in the next page, the `next_page_cursor` will
+                be `null`.)
+
+                The first page will contain the `page_size` most recent results matching
+                the given filter(s). Each subsequent page will have up to `page_size`
+                results, going backwards in time in DB insertion order, until there are
```
alistaire47, line 175: I think this point is important enough it deserves to be in the top-level description; it's getting buried here

conbench/api/results.py
```diff
@@ -157,96 +137,108 @@ def get(self) -> f.Response:
...
-
-        # setting limit value
-        limit: Optional[int] = f.request.args.get("limit")
+        filters = []
+        joins = []
+
+        if run_id_arg := f.request.args.get("run_id"):
+            filters.append(BenchmarkResult.run_id == run_id_arg)
+        else:
+            filters.append(BenchmarkResult.timestamp >= "2023-06-03")
```
alistaire47, line 221: A comment here for why this date would be helpful

conbench/api/results.py
```diff
@@ -157,96 +137,108 @@ def get(self) -> f.Response:
...
+        query = select(BenchmarkResult)
+        for join in joins:
+            query = query.join(join)
+        query = (
+            query.filter(*filters).order_by(BenchmarkResult.id.desc()).limit(page_size)
+        )
+        benchmark_results = current_session.scalars(query).all()
+
+        if len(benchmark_results) == page_size:
+            next_page_cursor = benchmark_results[-1].id
```
alistaire47, line 261: There's a corner case here where the last page has exactly `page_size` results. I guess the call with that cursor would just return an empty list and maybe that's fine? But it's a wasted call.

conbench/api/results.py
```diff
@@ -157,96 +137,108 @@ def get(self) -> f.Response:
...
-                datetime.timezone.utc
-            ).date() - datetime.timedelta(days=int(days))
-            days_filters = [
-                BenchmarkResult.timestamp >= min_time,
-            ]
-
-        # setting limit value
-        limit: Optional[int] = f.request.args.get("limit")
+        filters = []
+        joins = []
```
alistaire47, line 210: I love this structure for this method; it's much easier to follow now

conbench/app/batches.py
alistaire47: 1. Should we just be querying the db directly instead of hitting the API here?
2. We could at least set the `page_size` to 1000 for the moment to probably nobody is affected

conbench/tests/api/test_results.py
```diff
@@ -321,167 +319,104 @@ class TestBenchmarkList(_asserts.ListEnforcer):
...
-        response = client.get("/api/benchmarks/?name=bbb")
-        self.assert_200_ok(response, [_expected_entity(benchmark_result)])
-
-    def test_benchmark_list_filter_by_batch_id(self, client):
-        self.authenticate(client)
-        benchmark_result = _fixtures.benchmark_result(batch_id="20")
-        response = client.get("/api/benchmarks/?batch_id=20")
-        self.assert_200_ok(response, [_expected_entity(benchmark_result)])
+    def _make_request(self, client, **kwargs) -> dict:
+        query_args = [f"{key}={value}" for key, value in kwargs.items() if value]
```
alistaire47: 1. The `if value` makes me nervous if we try to use this to override a non-null value with null
    1. If this gets removed, we need to jsonify `None` into `null`
2. Do strings need quoting here?

If nothing is breaking, maybe just a comment for future debugging

conbench/tests/api/test_results.py
```diff
@@ -321,167 +319,104 @@ class TestBenchmarkList(_asserts.ListEnforcer):
...
+        # the data, each filter should filter out half the remaining results)
+        expected_num_results = 16
+        if name_arg:
+            expected_num_results /= 2
+        if batch_id_arg:
+            expected_num_results /= 2
+        if run_reason_arg:
+            expected_num_results /= 2
+        if run_id_arg:
+            expected_num_results /= 2
```
alistaire47: Kind of a pain, but it would be more robust to use a different number of values for each arg here, so a filter applied when it shouldn't be can't hide a missing filter. Not sure whether it's worth the effort, though.

alistaire47 (approved):
Looks good! 

I do think we should talk and make more of a plan for how we want the app to behave here though; for batches I don't think it's that big of a deal, but for results in a run it really is. Should these be skipping the API and going straight to the DB? Or just call the full set of pages, but that's just adding latency. Or we could link the datatables pagination to API pagination, but that would be a pain and mess up searching.

conbench/app/batches.py
austin3dickey: Yeah we should just hit the DB directly. I think this is the spirit of https://github.com/conbench/conbench/issues/968.

In an effort to keep PRs reviewable, I'll probably do this in a followup PR. It's okay if staging is "broken" in the meantime. I'll bump page size to 1000 like you suggested.

>we could link the datatables pagination to API pagination, but that would be a pain and mess up searching.

This was my original plan but you're right, we'd really have to rethink how searching works. It would probably be annoying to break search. Especially for those Arrow runs with 3300 results.

conbench/api/results.py
austin3dickey: Yep! I thought about this and decided it's worth the tradeoff. Either:
1. the client makes one redundant request whenever the last page has `page_size` results (probably on average about 1 in `page_size` times? so 1% of the time, or 0.1% for clients maxing out the size)
2. the server makes another DB query every single time just to make sure the next page is empty

(1) sounds like a lot less server load. I'll add a comment about this though.

conbench/tests/api/test_results.py
austin3dickey: Maybe I'll change this to `if value is not None` and clarify with a docstring. Since this is a test, I knew exactly all the possible kwargs that would be provided to this method, so it's fine now, but you're right we should make sure it's easy to debug if someone changes the test.

conbench/tests/api/test_results.py
austin3dickey: Sure I'll change to different prime numbers!

conbench/api/results.py
```diff
@@ -157,96 +137,120 @@ def get(self) -> f.Response:
...
+                Filter results to one specific `run_id`. Using this argument allows the
+                response to return results from before `2023-06-03 UTC`.
           - in: query
             name: run_reason
             schema:
               type: string
+            description: Filter results to one specific `run_reason`.
           - in: query
-            name: days
+            name: cursor
```
deepsihag, line 159: Why is days being replaced, user can still ask to get last N days and keep repeating calls untill there is cursor in the result. Cursor should just be an additional optional parameter where we decided to support it and all other filters should just leverage this while returning responses.

austin3dickey: I got rid of `days` for a few reasons:
- It's redundant with `cursor`: using `cursor`, a client can choose to keep scrolling until they receive the min timestamp of data they want, and then stop.
- `days` is stateful in that the response changes depending on when you call the API, even if the underlying data doesn't change. I like the reproducibility of a call using `cursor` better.
- There have been numerous calls to request time ranges better in this endpoint; see #746, #872, and the feedback below from an internal user. If we implement `min_timestamp` and `max_timestamp` filters, that would also supercede the `days` argument, providing its functionality while also being more flexible.

image

If those reasons aren't convincing, I could put `days` back in this PR and add tests for it. But to me it's a little harder to maintain than the other "stateless" filters.

austin3dickey:
(by the way - let's not merge this until I'm sure clients hitting staging won't be affected)

conbench/api/results.py
deepsihag: - I don't think it's right to consider cursor/pagination-token param duplicate to a time filter. Clients can always filter but that doesn't remove the need for keeping filters in API calls let alone a time based filter.
- Totally agree with the fact around the `days` filter implementation which would result different result sets, again not a reason to get rid of a time based filter, definitely a good reason to improve it.
- I'm all pro changing the current available filter `days` to timestamps , but let's make sure not to loose a time based filter for this change.

deepsihag (changes requested):
I'm good with getting rid of days filter but let's try to keep a time based filter.

conbench/api/results.py
austin3dickey: Cool I'll add the timestamp filters then!

deepsihag (approved):

PR #1503, opened 2023-10-17
api/benchmark-results: speed up

### LINKED ISSUE #1502: api/benchmark-results: big performance regression on large data
With Arrow-size data, the `/api/benchmark-results/` endpoint is returning 500s no matter what the query parameters are. The traceback indicates this query is timing out:
https://github.com/conbench/conbench/blob/eb668e7b563bd9d07983ea44e0383bd7fb7edbe8/conbench/api/results.py#L258

I tried executing this query with a 4-hour timeout and it still timed out. An EXPLAIN of this query (visualized with https://tatiyants.com/pev) shows this plan:
image

austin3dickey:
This PR adds a couple of partial indexes to BenchmarkResult that speed up the `GET /api/benchmark-results/` list endpoint. Fixes #1502. See discussion there.

It also removes the `name` and `batch_id` query parameters from that endpoint. I didn't see those being used in any clients, and in order to make sure that using those arguments wouldn't cause 500s on Arrow-scale data, we'd have to do a lot more careful indexing. I chose to get rid of them instead.

### DIFF
--- a/conbench/api/results.py
+++ b/conbench/api/results.py
@@ -18,7 +18,6 @@ from ..entities.benchmark_result import (
     BenchmarkResultSerializer,
     BenchmarkResultValidationError,
 )
-from ..entities.case import Case
 from ._resp import json_response_for_byte_sequence, resp400
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


austin3dickey:
🚢 feedback welcome!

Issue #926, opened 2023-03-22
code, docs: "sha" -> "commit hash"


jgehrcke:
I have been trying to nudge us to say "commit hash" more often instead of "sha". This is an ongoing initiative. This ticket here should make that initiative a little more visible.

Especially in public-facing resources (docs, API), we should work ourselves away from "sha". By itself it can be cryptic, and may waste our user's time:
![image](https://user-images.githubusercontent.com/265630/226838395-792e0533-54ab-4db5-925f-ab8f8e5d67d2.png)

Related:
- https://github.com/conbench/conbench/issues/478#issuecomment-1335031629
- https://github.com/conbench/conbench/pull/811#discussion_r1127840131
- https://github.com/conbench/conbench/pull/771#discussion_r1122995067

PR #1509, opened 2023-10-19
api/runs: paginate

### LINKED ISSUE #1505: api/runs: introduce pagination if it makes sense
See https://github.com/conbench/conbench/issues/871#issuecomment-1769242761.

austin3dickey:
Fixes #1505.

This PR:
- makes it impossible to GET /api/runs/ without using the `commit_hash` parameter
- adds pagination to GET /api/runs/?commit_hash=abc, changing the schema
- removes the POST, DELETE, and PUT endpoints from /api/runs/ (which have been no-ops for a while)


conbench/api/runs.py
```diff
@@ -386,110 +324,129 @@ def get(self, run_id):
...
+
+            This endpoint implements pagination; see the `cursor` and `page_size` query
+            parameters for how it works.
         responses:
             "200": "RunList"
             "401": "401"
         parameters:
           - in: query
-            name: sha
+            name: commit_hash
```
austin3dickey, line 345: in service of #926, better make one breaking change instead of two

conbench/api/runs.py
```diff
@@ -386,110 +324,129 @@ def get(self, run_id):
...
+            if run_id in run_info:
+                run_info[run_id].update(result)
+            else:
+                run_info[run_id] = RunAggregate(
+                    earliest_result=result,
+                    result_count=1,
+                    any_result_failed=result.is_failed,
+                )
+
+        this_page_run_ids = sorted(run_info.keys())[:page_size]
```
austin3dickey: Yeah, so we only limit the page size after executing the full query. Maybe this is dumb and we should never limit the size? or else clients will ask the server to make the same query over and over in those rare cases that page size is less than the total number of runs.

Or I could use [this skip query](https://github.com/conbench/conbench/blob/1f141414edefc721408e723cb5d978ad8e88dfe0/conbench/entities/benchmark_result.py#L942), modified for this endpoint, but it really seems like overkill

austin3dickey: Actually it'll be a slightly smaller query each time since it's filtered on the cursor, but still redundant.

conbench/api/runs.py
```diff
@@ -386,110 +324,129 @@ def get(self, run_id):
...
+            description: |
+                Required. A commit hash or a comma-separated list of commit hashes.
           - in: query
-            name: days
+            name: cursor
+            schema:
+              type: string
+              nullable: true
+            description: |
+                A cursor for pagination through matching runs in alphabetical order by
```
boshek, line 356: Thanks all this text. Very helpful.

boshek (approved):
Thanks for letting me know about this!

conbench/api/runs.py
austin3dickey: This morning I tested this code against the Arrow data. The query is actually extremely fast, but then the server takes 5 seconds to deaggregate by going through the loop.

I think after all I'll refactor this to use the skip scan query, which should remain quick and require no postprocessing. This should speed this up and also enable querying only `page_size` runs at once.

austin3dickey: Never mind, didn't need to use the skip scan query. We can use the much simpler DISTINCT ON idea, because we order by run_id, not benchmark result timestamp, and we don't need to worry about the results scaling (as JP discussed in https://github.com/conbench/conbench/issues/1466#issuecomment-1711571921).

deepsihag (approved):

austin3dickey:
>test against Arrow server for performance

I'm seeing a ~2x speedup from this change. 🙂

Issue #460, opened 2022-11-25
benchmark details: auto-render URL lists as `a href`


jgehrcke:
Picking up a question from @joosthooz -- when a property is a list of strings, then try to auto-render each of these as clickable link. See `grafana_trace_url` example in this screenshot:
![conb_clickable_url](https://user-images.githubusercontent.com/265630/203988455-d05ee238-15d8-4992-8d78-19ae22276f7c.png)

jonkeane:
Thanks for this! We _do_ have some URLizing already added (https://github.com/conbench/conbench/pull/427 was relatively recent). I suspect what's going on is that those string representations of lists need to either be treated differently (there might be a URLize setting to look at embedded strings?) or we should iterate through items like that and treat them as more than just a string of the list representation of them

PR #1523, opened 2023-10-30
webapp: auto-render URL lists as a href

johanpel:
This PR causes lists of urls in the optional benchmark info part of the web app to render as links.

Solves #460


johanpel:
Closed because we're going to do this from a branch here rather than from a fork to help CI

PR #1524, opened 2023-10-30
webapp: auto-render URL lists as a href

johanpel:
This PR causes lists of urls in the optional benchmark info part of the web app to render as links.

Solves https://github.com/conbench/conbench/issues/460

### DIFF
--- a/conbench/templates/benchmark-result.html
+++ b/conbench/templates/benchmark-result.html
@@ -206,15 +206,15 @@
               {{ k }}
               {% if v is not none %}
                 <div align="right" style="display:inline-block; float: right;">
-                  {% if not isinstance(v, list) %}
-                    <code>{{ v | urlize(target="_blank") }}</code>
-                  {% else %}
+                  {% if v is iterable and (v is not string and v is not mapping) %}
                     <code>[
                       {% for i in v %}
                         {{ i | urlize(target="_blank") }}
                         {% if not loop.last %},{% endif %}
                       {% endfor %}
                     ]</code>
+                  {% else %}
+                    <code>{{ v | urlize(target="_blank") }}</code>
                   {% endif %}
                 </div>
               {% endif %}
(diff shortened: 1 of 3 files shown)


conbench/templates/benchmark-result.html
```diff
@@ -206,7 +206,16 @@ <h4>raw data</h4>
               {{ k }}
               {% if v is not none %}
                 <div align="right" style="display:inline-block; float: right;">
-                  <code>{{ v | urlize(target="_blank") }}</code>
+                  {% if v is iterable and (v is not string and v is not mapping) %}
+                    <code>[
+                      {% for i in v %}
+                        {{ i | urlize(target="_blank") }}
+                        {% if not loop.last %},{% endif %}
```
austin3dickey, line 213: I'm fine with this. But would you rather put a newline between elements than a comma? Up to you!

austin3dickey (approved):
Looks great! Thanks for dealing with our CI!

johanpel:
Thanks for the review @austin3dickey. Is the `ui-screenshots` check something I can inspect somewhere?

Edit: nevermind, found it under the artifacts button!

conbench/templates/benchmark-result.html
johanpel: If the list always contained links, that would be quite nice indeed, but if the list contains other things it may appear rather weird. For this reason I tried to keep the default way of rendering lists as much as possible but still have the clickable URLs.

johanpel:
@austin3dickey this seems to work:

image

Although the additional spacing is a bit weird. Is this acceptable? Not sure how to prevent this.

austin3dickey:
>Although the additional spacing is a bit weird. Is this acceptable?

Yes I think this is fine. 🙂 It's probably caused by the whitespace in the template getting collapsed to a single space, but I sort of like how it keeps things spread apart. Thanks again for the PR!

Issue #1345, opened 2023-07-05
compare endpoint: UI, API: make this much faster


jgehrcke:
Doing z-score analysis for an individual benchmark result (plus "baseline 'start' commit") involves building a time series: walk the commit tree, and for each commit look up a matching benchmark result. This is just a rough mental model that clarifies that there's some quadratic complexity in there.

In case of Arrow Conbench there may be thousands of results per run, and that leads to run-to-run comparison being rather complex: can take minutes with the current setup. Sometimes the HTTP requests time out after ~two minutes. See https://github.com/conbench/conbench/issues/745.

We need to make this fundamentally faster, with some smarter data structures. Via caching, or via smart DB work. Somewhere, state should be 'persisted', for quick lookup. I think for this it would help to cleanly define the n-tuple that is required for time series identification, and for precisely identifying an individual time series (along commits). Caching can be in mem but also on disk if we worry that this gets big.

Profiling output showing that `set_z_scores()` dominates response time generation for compare/runs...:
```
conbench-app-1  | Stats for (Thread) (0)
conbench-app-1  | 
conbench-app-1  | Clock type: CPU
conbench-app-1  | Ordered by: totaltime, desc
conbench-app-1  | 
conbench-app-1  | name                                                                                                                                                    ncall       tsub          ttot          tavg          
conbench-app-1  | /app/conbench/app/_endpoint.py:158 CompareRuns.api_get                                                                                                  3           0.000038      2.605300      0.868433    
conbench-app-1  | /usr/local/lib/python3.11/site-packages/werkzeug/test.py:1138 FlaskClient.get                                                                           3           0.000027      2.599406      0.866469    
conbench-app-1  | /usr/local/lib/python3.11/site-packages/flask/testing.py:192 FlaskClient.open                                                                           3           0.000074      2.599379      0.866460    
conbench-app-1  | /usr/local/lib/python3.11/site-packages/werkzeug/test.py:1032 FlaskClient.open                                                                          3           0.000037      2.595528      0.865176    
conbench-app-1  | /usr/local/lib/python3.11/site-packages/werkzeug/test.py:952 FlaskClient.run_wsgi_app                                                                   3           0.000037      2.595404      0.865135    
conbench-app-1  | /usr/local/lib/python3.11/site-packages/flask/app.py:2546 Flask.__call__                                                                                4           0.000033      2.595333      0.648833    
[...]

PR #1527, opened 2023-10-31
api/compare/runs: paginate

### LINKED ISSUE #1507: api/compare/runs: introduce pagination if it makes sense
See https://github.com/conbench/conbench/issues/871#issuecomment-1769242761.

austin3dickey:
Fixes #1507. Adds pagination to the /api/compare/runs endpoint.

Also changes the /app/compare/runs page to hit the DB directly instead of the API so it doesn't need to paginate.


austin3dickey:
TODO:

- [x] check performance against Arrow
- [x] maybe optimize page size (limits, default)
- [ ] fix benchalerts

austin3dickey:
I tried executing this strategy against https://conbench.ursa.dev/api/compare/runs/d3348296b93943a0892fc2607c8f7781...988d757bfe5541519bc704cace69a711/, which should return about 3400 comparisons. This takes about 85 seconds on the current deployment.

With this PR, that drops to 75 seconds, but that's for _any page size_[^1]. Based on my testing, due to the way the DB is indexed today, even a page size of 10 will take 75 seconds per page. That's pretty bad. I think we need to look at the DB again before we can merge this PR.

[^1]: Except a page size of 1 which takes 1 second. Definitely something weird with the query plan.

austin3dickey:
>75 seconds, but that's for _any page size_

Oops this was a bug; I didn't do this yet:

https://github.com/conbench/conbench/blob/6a6935e16b0f46cb0217f60e7d756475e7e51137/conbench/api/compare.py#L685

The way this PR was working, it was calculating distribution stats for ALL fingerprints every time, no matter what the actual page of fingerprints was. I fixed this bug, and now we do see shorter times for smaller page sizes.

I'll run a few tests to see if there's a "best" page size for these big runs. I think it will depend on how the Postgres query planner decides which indexes to scan, which of course also depends on the actual Conbench instance and run size (not just page size). But it'll be good to have some guidance. As part of #1345 I'll probably test out new multi-indexes (since PG thinks commit and timestamp aren't correlated but they are) which could change this logic anyway.

austin3dickey:
That was a very instructive exercise:
```
 pg size | num pgs | mean (s) | total (s) | indexes used
---------+---------+----------+-----------+------------------------
    5000 |       1 |    75.7  |      75.7 | timestamp, commit
    1000 |       4 |    20.0  |      80.2 | timestamp, commit
     500 |       7 |     5.76 |      40.3 | timestamp, fingerprint
     100 |      35 |     1.27 |      44.4 | fingerprint, timestamp
      50 |      69 |     0.81 |      55.7 | fingerprint, timestamp
      10 |     341 |     0.58 |     197   | fingerprint, timestamp
       5 |     681 |     0.55 |     374   | fingerprint, timestamp
       1 |    3403 |     1.03 |    3518   | fingerprint
```
As we can see, using a page size around O(100) is big enough to avoid the overhead cost of too many requests but small enough to avoid a costly filter scan. Looking at the query plans was particularly instructive: Postgres thinks the fingerprint index scan will be too costly once there are >500 fingerprints, and so it reverts to using the timestamp and commit indexes—but those are extremely correlated! So it doesn't filter out nearly as much data as when it can use the timestamp and fingerprint indexes, and thus big page sizes are a lot slower per page.

I think for now I'll adjust the max page size to be 500 with a default page size of 100. Using the default will give an overall 2x speedup for the big Arrow run use cases, which is awesome. And this table will be helpful for taking care of #1345 in the future.

austin3dickey:
I will fix benchalerts in a followup PR. This is ready to go.

Issue #1525, opened 2023-10-30
devprod: CI fails on forks


austin3dickey:
Example: https://github.com/conbench/conbench/pull/1523/checks

If someone opens a PR from a fork, various errors occur. This should be a supported workflow as we can't give write permissions to everyone who wants to open a PR.

PR #1532, opened 2023-11-09
devprod: Fix CI on forks

austin3dickey:
Addresses #1525. Some edits to the token given to the workflow, and fixing tests to only hit this repo's API.


austin3dickey:
Going to merge for now and revisit; CI stopped running because I changed the `on` target I think

PR #1546, opened 2023-12-04
webapp: speed up landing page by adding index

### LINKED ISSUE #1482: landing page: 'recent runs' query may still be slow occasionally (> 10 s)
Seen for Arrow Conbench, response generation time of ~25 seconds:

Screenshot from 2023-09-18 12-12-48

Also represented in the response generation time histogram over time:

image

(population of the 20s - 40s bucket on the right-hand side).

austin3dickey:
Fixes #1482. This PR adds a partial index that speeds up `fetch_one_result_per_each_of_n_recent_runs()`, which powers the landing page. 

That function was fast only when PG's shared buffer cache was hit, but when it was missed it was very slow to read through all the data in the `benchmark_result` table. This PR should improve the cache miss time by limiting the files to look through to only data from the last 2 weeks. (For all the instances we track, 2 weeks is more than enough time to get the last 250 runs.)

I validated that on Arrow-scale data this speeds up the cache-hit time from 1130 ms to 40 ms. Unfortunately I couldn't test the cache-miss time (which used to be 70s or so!) but I have confidence that it should be much faster as well, since according to EXPLAIN ANALYZE the new query is reading 100x fewer blocks from the cache.

The write time does not slow down much either.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ alembic-new-migration: teardown-app
 		exit 1; \
 	fi
 	docker compose run --detach db
-	CREATE_ALL_TABLES=false docker compose \
+	CREATE_ALL_TABLES=false docker compose -f docker-compose.yml -f docker-compose.dev.yml \
 		run -e CREATE_ALL_TABLES app alembic upgrade head
 	CREATE_ALL_TABLES=false docker compose -f docker-compose.yml -f docker-compose.dev.yml \
 		run -e CREATE_ALL_TABLES  app alembic revision --autogenerate -m ${ALEMBIC_MIGRATION_NAME}
--- a/alembic.ini
+++ b/alembic.ini
@@ -63,7 +63,7 @@ keys = console
 keys = generic
 
 [logger_root]
-level = WARN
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -108,7 +108,7 @@ alembic-new-migration: teardown-app
 		exit 1; \
 	fi
 	docker compose run --detach db
-	CREATE_ALL_TABLES=false docker compose \
+	CREATE_ALL_TABLES=false docker compose -f docker-compose.yml -f docker-compose.dev.yml \
```
austin3dickey, line 111: was getting a bug where my app and my dev-app were out of sync

alembic.ini
```diff
@@ -63,7 +63,7 @@ keys = console
 keys = generic
 
 [logger_root]
-level = WARN
+level = INFO
```
austin3dickey, line 66: I tried this a while ago in https://github.com/conbench/conbench/pull/1477 (which wasn't merged). It allows for stdlib logging from migration files.

PR #1550, opened 2023-12-11
api/benchmark-results: speed up

### LINKED ISSUE #1549: api/benchmark-results: still seeing large query times
An internal Conbench instance is seeing timeouts when hitting `/api/benchmark-results` with `page_size=1000`, `run_reason=SOME_REASON`, `earliest_timestamp=SOME_DATE`, `latest_timestamp=21_DAYS_LATER`, and `cursor=SOME_CURSOR`.

This only happens sometimes. The query seems to last for over 5 minutes. Upon retry, it typically executes in 12 seconds or 1 second depending on how recently it was last run. When it does run for a long time, according to RDS Performance Insights it typically spends all its time reading data files. When it runs for a shorter time, it's typically able to get most of its data from the shared buffer cache.

The query plan is what I expected: it scans the partial `benchmark_result_run_reason_id_idx` with the index condition being `((benchmark_result.run_reason = 'SOME_REASON'::text) AND ((benchmark_result.id)::text < 'SOME_CURSOR'::text))`. Then it filters the matching data to `((benchmark_result."timestamp" >= 'SOME_DATE'::timestamp without time zone) AND (benchmark_result."timestamp" <= '21_DAYS_LATER'::timestamp without time zone))`.

Upon a recent execution of this query, the index condition returns 80327 rows, and then the timestamp filter filters out all except 653 of them. My hunch says that if we attempt to read much fewer files we'll be in better shape, especially since we throw out 99.2% of the matching index data. One way we could do that is by providing a lower bound on `benchmark_result.id` in addition to the upper bound. We could get this by converting the `earliest_timestamp` into a UUID7.

It's a currently-known limitation that the ID corresponds to DB insertion time, which could be different from the user-given result timestamp. However, if we make the assumption that the user-given timestamp will never be _later_ than the insertion timestamp (i.e. it's unlikely that a client would submit data with a timestamp from the future; at least I don't know of any that do this), then this change should not cause any problems. Proof:
- assume insertion_time >= user_time
- today we return data where user_time >= SOME_DATE
- therefore we already return data where insertion_time >= SOME_DATE

austin3dickey:
Fixes #1549. See discussion there.

### DIFF
--- a/conbench/api/results.py
+++ b/conbench/api/results.py
@@ -3,7 +3,9 @@ import logging
 import flask as f
 import flask_login
 import orjson
+import pandas as pd
 from sqlalchemy import select
+from uuid_extensions import uuid7
 
 import conbench.metrics
 from conbench.dbsession import current_session
@@ -186,6 +188,12 @@ class BenchmarkListAPI(ApiEndpoint, BenchmarkValidationMixin):
                 The earliest (least recent) benchmark result timestamp to return. (Note
                 that this parameter does not affect the behavior of returning only
                 results after `2023-06-03 UTC` without a `run_id` provided.)
+
+                This parameter will also filter out results that were inserted into the
+                database before the given timestamp, independently of the user-given
+                benchmark result timestamp. This should not be a problem if you never
+                upload data with a timestamp in the future, but if you do, you may get
+                unexpected results around this boundary.
           - in: query
             name: latest_timestamp
             schema:
@@ -210,6 +218,12 @@ class BenchmarkListAPI(ApiEndpoint, BenchmarkValidationMixin):
 
         if earliest_timestamp_arg := f.request.args.get("earliest_timestamp"):
             filters.append(BenchmarkResult.timestamp >= earliest_timestamp_arg)
+            # Speed up the query by also filtering out results that were inserted into
+            # the database before the given timestamp
+            earliest_timestamp_ns = (
+                pd.Timestamp(earliest_timestamp_arg).timestamp() * 10**9
+            )
+            filters.append(BenchmarkResult.id >= uuid7(earliest_timestamp_ns).hex)
 
         if latest_timestamp_arg := f.request.args.get("latest_timestamp"):
             filters.append(BenchmarkResult.timestamp <= latest_timestamp_arg)
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -364,7 +364,7 @@ class BenchmarkResult(Base, EntityMixin):
             tags.update(case.tags)
 
             if benchmark_result.commit:
-                commit_dict = CommitSerializer().one.dump(benchmark_result.commit)
+                commit_dict = CommitSerializer().many._dump(benchmark_result.commit)
                 commit_dict.pop("links", None)
             else:
                 commit_dict = None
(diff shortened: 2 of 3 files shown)


austin3dickey:
I'm testing this out locally and ran into something else -- on my last attempt the query took 0.1s but then 24.3s to serialize the results. Oops. Will look into how to speed this up. We're already using `orjson`. May be due to sqlalchemy extra joins.

austin3dickey:
Yeah, looks like it's executing a SELECT statement for each different commit in the response. Commenting out this line gets rid of those queries and speeds it up a ton:

https://github.com/conbench/conbench/blob/4eddeecf4a33e26703b71e86e416b76062c5dce6/conbench/entities/benchmark_result.py#L367

But this is weird because the original query definitely is joining commit info to each BenchmarkResult object:

```
SELECT benchmark_result.id, benchmark_result.case_id, benchmark_result.info_id, benchmark_result.context_id, benchmark_result.run_id, benchmark_result.run_tags, benchmark_result.run_reason, benchmark_result.commit_id, benchmark_result.commit_repo_url, benchmark_result.hardware_id, benchmark_result.history_fingerprint, benchmark_result.data, benchmark_result.times, benchmark_result.optional_benchmark_info, benchmark_result.unit, benchmark_result.time_unit, benchmark_result.batch_id, benchmark_result.timestamp, benchmark_result.iterations, benchmark_result.mean, benchmark_result.min, benchmark_result.max, benchmark_result.median, benchmark_result.stdev, benchmark_result.q1, benchmark_result.q3, benchmark_result.iqr, benchmark_result.error, benchmark_result.validation, benchmark_result.change_annotations, commit_1.id AS id_1, commit_1.sha, commit_1.branch, commit_1.fork_point_sha, commit_1.parent, commit_1.repository, commit_1.message, commit_1.author_name, commit_1.author_login, commit_1.author_avatar, commit_1.timestamp AS timestamp_1, hardware_1.id AS id_2, hardware_1.name, hardware_1.type, hardware_1.hash, case_1.id AS id_3, case_1.name AS name_1, case_1.tags, info_1.id AS id_4, info_1.tags AS tags_1, context_1.id AS id_5, context_1.tags AS tags_2  
FROM benchmark_result 
LEFT OUTER JOIN commit AS commit_1 ON commit_1.id = benchmark_result.commit_id 
LEFT OUTER JOIN hardware AS hardware_1 ON hardware_1.id = benchmark_result.hardware_id 
LEFT OUTER JOIN "case" AS case_1 ON case_1.id = benchmark_result.case_id
LEFT OUTER JOIN info AS info_1 ON info_1.id = benchmark_result.info_id 
LEFT OUTER JOIN context AS context_1 ON context_1.id = benchmark_result.context_id  
WHERE benchmark_result.timestamp >= '2023-06-03' AND benchmark_result.timestamp >= '2023-11-20 10:57:46' AND benchmark_result.id >= '0655b3baa0007000800095036ba291a9' AND benchmark_result.timestamp <= '2023-12-11 10:57:46' AND benchmark_result.run_reason = 'nightly' AND benchmark_result.id < '0655d7ff5524700a800076e6f5ef0769' 
ORDER BY benchmark_result.id DESC   
LIMIT 1000
```

austin3dickey:
Aha, I narrowed the slowness down to these lines:
https://github.com/conbench/conbench/blob/4eddeecf4a33e26703b71e86e416b76062c5dce6/conbench/entities/commit.py#L347-L352

That's great because we pop off commit `links` when serializing the benchmark result! So we can skip these slow lines entirely without changing what the output looks like.

boshek (approved):
amazing. thank you.

jgehrcke:
Great work!

PR #1566, opened 2024-01-09
commit graph: ensure we start backfill in the past

### LINKED ISSUE #1565: bug: `backfill_default_branch_commits` doesn't work with commits arriving faster than benchmark results
I noticed a rare bug. Suppose the default branch of a repo gets the following commits:
```
commit | commit timestamp
-------+----------
A      | 01:00
B      | 03:00
C      | 05:00
D      | 05:05
```
Say it takes 10 minutes for benchmark results to be posted after commits. Also suppose that commit B was a minor commit that was purposely not benchmarked.

The intent of `backfill_default_branch_commits()` is that once a benchmark result for commit C is posted, the function will hit the GitHub API to find that commit B should exist in the database, and insert a corresponding record for it. But the current code doesn't do this. In fact, commit B will be completely skipped and never inserted into the database. Here's why:

https://github.com/conbench/conbench/blob/14b2a6df786403eaab6d780cf002e83c42e9e407/conbench/entities/commit.py#L488-L494

When this code is run with `new_commit` being commit C, the time is 05:10. So `last_tracked_commit`, which sets the start time for the backfill, will pick up commit D. There will be no backfill because the start time is after the end time.

The intent is that this query always picks up commit A, no matter what out-of-order data might already be in the database. Then there will be a backfill from commit A to C, picking up B. So I believe that the `Commit.sha != new_commit.sha` filter should change to `Commit.timestamp < new_commit.timestamp`.

austin3dickey:
Fixes https://github.com/conbench/conbench/issues/1565. See discussion there.

### DIFF
--- a/conbench/entities/commit.py
+++ b/conbench/entities/commit.py
@@ -486,7 +486,10 @@ def backfill_default_branch_commits(repo_url: str, new_commit: Commit) -> None:
     default_branch = _github.get_default_branch(repospec)
 
     last_tracked_commit = Commit.all(
-        filter_args=[Commit.sha != new_commit.sha, Commit.timestamp.isnot(None)],
+        filter_args=[
+            Commit.timestamp < new_commit.timestamp,
+            Commit.timestamp.isnot(None),
+        ],
         branch=default_branch,
         repository=repo_url,
         order_by=Commit.timestamp.desc(),


boshek:
Thanks for this :)

PR #1571, opened 2024-01-10
results: change SVS to 'best run' by default

### LINKED ISSUE #640: Use "best run" as the "raw" data when calculating z-scores
We have discussed using `min` instead of `mean` in the plots in https://github.com/conbench/conbench/issues/594, https://github.com/conbench/conbench/issues/535, etc.

This issue is scoped to changing the `get_history()` and `append_z_scores()` functions to use a BenchmarkResult's best-case iteration instead of the mean of all iterations. This could be `min` if `_less_is_better(benchmark_result.unit)` or `max` if not.

(Alternatively we could use something else, like weighted mean, but since we've seen a warm-up effect before that mightn't be the most accurate.)

austin3dickey:
Fixes #640 . Changes the single value summary to represent the best run of one benchmark result. This should reduce false positives during regression alerting.

(Also leaves it possible to use the old mean strategy.)


austin3dickey:
Still have to fix tests and check the impact on plots.

austin3dickey:
>check the impact on plots

Yes, I believe the plots now have both the grey and black dots show the min. We'll have to fix that. I kind of lean toward plotting the raw data for the grey dots instead of the mean.

boshek:
> Yes, I believe the plots now have both the grey and black dots show the min. We'll have to fix that. I kind of lean toward plotting the raw data for the grey dots instead of the mean.

Totally agree. It does make me wonder if a label for which replicate it is would be worthwhile.

austin3dickey:
This PR is big enough already so I'll fix the plots in a followup. Reviews welcome but I'll probably merge this soon.

PR #1572, opened 2024-01-12
fix config

austin3dickey:
I noticed after deploying #1571 to staging that some webapp pages returned a 500 with the traceback ending with:
```
  File "/app/conbench/entities/benchmark_result.py", line 510, in _single_value_summary
    assert Config.SVS_TYPE == "best"
           ^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```
My intention is that if the environment variable is empty or unset, this should default to `"best"`, and I assume this AssertionError is thrown because the environment variable is empty. This should fix that.


austin3dickey:
This worked.

PR #1576, opened 2024-01-17
webapp plots: display raw data

### LINKED ISSUE #1573: webapp: display raw repetition data in plots
After merging #1571, plots look strange and legends are incorrect:

image

This plot shows a benchmark that has 3 repetitions per result. The mean used to be represented by the grey dot, and the min by the black dot. Now, both dots show the best value (min if less is better, max otherwise). 

I propose we plot the raw repetition data with the grey dot, and have the visually distinctive black dot represent the best run. This may result in a lot of grey points, but after some experience sorting through plots I think it's more instructive to be able to see the raw data. This helps identify outliers and persistent warmup effects. We can downsample in the future if performance becomes a problem.

@boshek in https://github.com/conbench/conbench/pull/1571#issuecomment-1885714749 also recommended that the hover tooltip displays the repetition number, to help with identifying patterns.

austin3dickey:
Fixes #1573. This PR reorganizes the plotting code a bit and enables plotting of the raw data.
Before:

Screenshot 2024-01-17 at 14 27 22

After:

Screenshot 2024-01-17 at 14 27 26

### DIFF
--- a/conbench/app/_plots.py
+++ b/conbench/app/_plots.py
@@ -694,18 +694,22 @@ def time_series_plot(
     p.xaxis.axis_label = ""
 
     multisample, multisample_count = _inspect_for_multisample(samples)
-    label = "result (n=1)"
-    if multisample:
-        svs_type = current_benchmark_result.svs_type
-        label = f"result {svs_type}"
-        if multisample_count:
-            label += f" (n={multisample_count})"
+    if not multisample:
+        n_label = "(n=1)"
+        svs_type = ""
+    elif multisample_count:
+        n_label = f"(n={multisample_count})"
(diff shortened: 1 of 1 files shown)


boshek (approved):
This is a nice approach. My only comments are with respect to the legend. Here is one suggested edit:

- result replicates 
- result (min)
- outlier result (min)
- lookback z-score (mean)
- lookback z-score leeway
- highlighted result: baseline (min) 
- current result (min)
- current result replicates (n = 3) 
- distribution change

austin3dickey:
>My only comments are with respect to the legend

Thanks for the feedback! Implemented! Will merge soon.

Issue #1577, opened 2024-01-18
BMRT cache: takes too long to populate, being a noisy neighbor


austin3dickey:
I've been noticing a slowdown for internal Conbench instances on the `vd-2` cluster. I checked the RDS Performance Insights dashboard and found this for our production instance:

![Screenshot 2024-01-16 at 08 39 40](https://github.com/conbench/conbench/assets/16600275/13c75c28-4790-4be0-9bd8-7e13328f9dcf)

Scanning the code, these FETCH FORWARD 2000 queries can only be coming from here:

https://github.com/conbench/conbench/blob/76365a5952de4a827f4b6ec2996e03acf2aebe43/conbench/bmrt.py#L245-L249

That's the query that populates the BMRT cache. Sure enough, checking our Grafana dashboard:

![Screenshot 2024-01-18 at 11 59 27](https://github.com/conbench/conbench/assets/16600275/165dabcf-32db-4252-861f-be2835926371)

I see a big variance in how long it takes to populate that cache, with an unfortunately long tail (as big as 50 minutes!) Those individual FETCH FORWARD queries are quite often taking 5-20 minutes to complete, and I notice that other queries that get executed at the same time end up being much slower than I'd ordinarily expect. (This is probably clobbering the shared buffer cache as well, which would otherwise help speed up common queries.)

Interestingly, over the same time range the `ursa-2` time-to-populate dashboard has been remarkably-consistently hovering around 4 minutes. Perhaps the reason: for the respective benchmarking frequencies, the default BMRT_CACHE_SIZE of 800_000 goes back only a few days for Arrow and a few years for the internal database. Due to the index on timestamp, it makes sense to me that the Arrow query would go faster, since the index provides a much smaller search space.

If that's the case, I propose filtering this query to the last 2 weeks. This matches the filter on the recent-runs query that we recently added in #1546, and I think a tail-end of 2 weeks in this cache will be acceptable to users.

austin3dickey:
When I deployed to staging yesterday, the time-to-populate went from about 4 minutes to 2 seconds on average. Good sign. Will deploy to the problematic instance soon.

<img width="837" alt="image" src="https://github.com/conbench/conbench/assets/16600275/7f7bab0d-68e1-44c4-ad23-8cc8cd19de40">

austin3dickey:
Deployed 20 minutes ago. Might be too early to draw conclusions but this is promising to me.
[...]

PR #1578, opened 2024-01-18
bmrt: limit to last 2 weeks

austin3dickey:
See discussion in #1577. We'll see if this cuts down the cache population time.


jgehrcke:
Makes lots of sense!

PR #1579, opened 2024-01-19
lookback z-score: use contender SVS

austin3dickey:
This is a fix to https://github.com/conbench/conbench/pull/1571. The intention has always been that the lookback z-score method uses the contender SVS, not necessarily its mean.


conbench/tests/entities/test_history.py
```diff
@@ -72,7 +72,7 @@ def assert_equal_leeway(comparison, reference, figs=4):
     ("mean", "closest_defaultbranch_ancestor"): [
         None,
         None,
-        28.48,
+        None,
```
austin3dickey, line 75: one outcome of this: now contenders with errors won't have z-scores (which was always the intention anyway)

PR #1583, opened 2024-01-29
Ensure z-scores are populated for one-rep results

austin3dickey:
I recently noticed that the z-score analysis was always `null` when the history only consists of benchmark results that were only run with one repetition. This is due to a bug in how I wrote the history query: the `min` and `max` fields are not populated for these benchmark results, so they were being excluded.

### DIFF
--- a/conbench/tests/entities/test_history.py
+++ b/conbench/tests/entities/test_history.py
@@ -386,3 +386,19 @@ def test_detect_shifts_with_trimmed_estimators():
     assert not np.any(result_df.is_step[:50])
     assert result_df.is_step[50]
     assert not np.any(result_df.is_step[51:])
+
+
+def test_set_z_scores_one_rep():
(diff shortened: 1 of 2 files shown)


conbench/tests/entities/test_history.py
```diff
@@ -386,3 +386,19 @@ def test_detect_shifts_with_trimmed_estimators():
     assert not np.any(result_df.is_step[:50])
     assert result_df.is_step[50]
     assert not np.any(result_df.is_step[51:])
+
+
+def test_set_z_scores_one_rep():
```
austin3dickey, line 391: Validated that this test would have failed before because the `z_score` was None. Now it passes.

austin3dickey:
image

Happy new year; `black` has a breaking change. Will fix (and pin) in a followup PR.

PR #1603, opened 2024-07-26
history API: add more fields

### LINKED ISSUE #1602: api/history: add `run_tags` and `result_timestamp` fields
User request from @boshek: let's add these fields to the history endpoint. (We should consolidate these objects with the /benchmark-results endpoint, but this gets us a bit closer.)

austin3dickey:
Fixes #1602. This PR adds `run_tags` and `result_timestamp` to the history API for user consumption (not plots).


conbench/entities/benchmark_result.py
```diff
@@ -775,8 +775,8 @@ def validate_and_aggregate_samples(stats_usergiven: Any):
             "q1": q1,
             "q3": q3,
             "median": float(np.median(samples)),
-            "min": np.min(samples),
-            "max": np.max(samples),
+            "min": float(np.min(samples)),
+            "max": float(np.max(samples)),
```
austin3dickey, line 779: These `float()` changes are unrelated to the rest of this PR. When I tested locally, I was getting some test errors that sqlalchemy was submitting SQL to the DB that looked something like
```sql
INSERT INTO benchmark_result (..., median, min, ...)
VALUES (..., 5.3, np.float64(5), ...)
```
This obviously isn't valid SQL. It's been a while since we developed in this repo! 🙂 I guess some dependency was updated (probably sqlalchemy or numpy). So this fixes that problem.

austin3dickey:
classic, another stale moving target problem: `AttributeError: np.NaN was removed in the NumPy 2.0 release. Use np.nan instead.. Did you mean: 'nan'?`

benchadapt/python/tests/adapters/test_asvbench.py
```diff
@@ -301,17 +301,17 @@
...
-            [9.3473e-06, 0.00080488, np.NaN],
-            [9.4343e-06, 0.00082726, np.NaN],
-            [1082, 13, np.NaN],
-            [10, 10, np.NaN],
+            [9.2976e-06, 0.00079774, np.nan],
+            [9.5915e-06, 0.00083937, np.nan],
+            [9.3473e-06, 0.00080488, np.nan],
+            [9.4343e-06, 0.00082726, np.nan],
+            [1082, 13, np.nan],
+            [10, 10, np.nan],
```
austin3dickey, line 314: fixed here, see previous comment

boshek (approved):

PR #1606, opened 2024-08-27
/api/compare: fix edge case bug in z-scores

### LINKED ISSUE #1605: z-score algorithm: bug when there are exactly 2 repetitions and SVS_TYPE is "best"
During `set_z_scores()` (used during the `/api/compare/` endpoints), we call this code when SVS_TYPE is "best":
https://github.com/conbench/conbench/blob/da7fb4c2b4874a33789243a60887a8b432359c7f/conbench/entities/history.py#L358-L365

What I forgot when I wrote that code was that min/max are also missing when there are two repetitions, not only when there's one!
https://github.com/conbench/conbench/blob/da7fb4c2b4874a33789243a60887a8b432359c7f/conbench/entities/benchmark_result.py#L846-L854

So in cases like this, we are using the mean of the 2 repetitions when gathering the distribution mean/SD for z-scores, instead of the better repetition. But then we compare that to the contender's better repetition. In cases where the two repetitions typically are far apart (due to a big warm-up effect or something), this means the z-score analysis may _always_ return an improvement indicated.

austin3dickey:
Fixes #1605. See discussion there.

### DIFF
--- a/conbench/tests/entities/test_history.py
+++ b/conbench/tests/entities/test_history.py
@@ -1,5 +1,5 @@
-from datetime import datetime
-from typing import Callable, cast
+from datetime import datetime, timedelta
+from typing import Callable, List, cast
 
 import numpy as np
 import pandas as pd
@@ -11,6 +11,7 @@ from conbench.types import TBenchmarkName
 
 from ...config import Config
 from ...db import _session as Session
+from ...entities.benchmark_result import BenchmarkResult
 from ...entities.commit import Commit
 from ...entities.history import (
     _detect_shifts_with_trimmed_estimators,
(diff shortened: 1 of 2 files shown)


conbench/tests/entities/test_history.py
```diff
@@ -347,6 +348,52 @@ def test_set_z_scores_with_distribution_change(
...
+def test_set_z_scores_with_small_reps(data: List[List[int]], unit: str, svs_type: str):
+    expected_z_scores = {
+        (1, "s", "mean"): [None, None, -4.95],
+        (1, "s", "best"): [None, None, -4.95],
+        (1, "i/s", "mean"): [None, None, 4.95],
+        (1, "i/s", "best"): [None, None, 4.95],
+        (2, "s", "mean"): [None, None, -41.72],
+        (2, "s", "best"): [None, None, -55.86],
+        (2, "i/s", "mean"): [None, None, 41.72],
+        (2, "i/s", "best"): [None, None, 34.65],
```
austin3dickey, line 363: I did confirm that the tests where reps=2 and svs_type=best fail before this PR and pass now.

austin3dickey:
Failing tests seem unrelated; will open an issue.

PR #1614, opened 2025-10-22
Improve landing perf speed

pitrou:
This should hopefully improve the performance of the landing page by making the 14 days timestamp boundary appear as a constant to Postgres (which I assume was the goal of the hardcoded `'2023-11-19'` comparand, but that value is obsolete, and why not compute it dynamically?)

We've been seeing some "503 Service Temporarily Unavailable" errors on https://conbench.ursa.dev/ recently, and that page has become generally very slow to load anyway.

### DIFF
--- a/conbench/entities/benchmark_result.py
+++ b/conbench/entities/benchmark_result.py
@@ -984,32 +984,18 @@ def fetch_one_result_per_each_of_n_recent_runs(n: int = 250) -> List[BenchmarkRe
     commit since there can be multiple runs per commit.
     """
     query_text = f"""
-        WITH RECURSIVE run_results AS (
-            (
-                SELECT *
-                FROM benchmark_result
-                WHERE timestamp > now() - INTERVAL '14 days'
-                AND timestamp > '2023-11-19'
-                ORDER BY run_id
-                LIMIT 1
-            )
-            UNION ALL
-            SELECT next_run.*
-            FROM run_results
-            CROSS JOIN LATERAL (
-                SELECT *
-                FROM benchmark_result
-                WHERE run_id > run_results.run_id
-                AND timestamp > now() - INTERVAL '14 days'
-                AND timestamp > '2023-11-19'
-                ORDER BY run_id
-                LIMIT 1
-            ) next_run
-        )
-        SELECT * FROM run_results
-        ORDER BY timestamp desc
+        SELECT r.* FROM (
+            SELECT DISTINCT ON (run_id) *
+            FROM benchmark_result
+            ORDER by run_id
+        ) r
+        WHERE timestamp > (now() - INTERVAL '14 days')
+        AND timestamp > '2023-11-19'
+        ORDER BY timestamp DESC, run_id DESC
         LIMIT {n}
     """
+    # ana = current_session.execute(s.text(f"EXPLAIN ANALYZE {s.text(query_text)}"))
+    # print("\n".join("".join(*tup) for tup in ana.fetchall()))
     query = s.select(BenchmarkResult).from_statement(s.text(query_text))
     # Need to type hint this again because from_statement() overrides the type hints.
     bmrs: List[BenchmarkResult] = list(current_session.scalars(query, {"n": n}).all())


pitrou:
@austin3dickey I don't know if you could spare some time to look at this, especially as I'm not sure I'm making sense of the SQL query here.

conbench/entities/benchmark_result.py
```diff
@@ -975,21 +975,23 @@ def fetch_one_result_per_each_of_n_recent_runs(n: int = 250) -> List[BenchmarkRe
...
+    partial index: hence the `timestamp_bound` filter. (This probably would be better
+    handled with something like TimescaleDB.)
 
     Since this doesn't include joins, as of the time of writing we're still having to
     reach out to the database to get commit and hardware information: one query per
     result, I think. That SELECT query (one per result) will sometimes refer to the same
     commit since there can be multiple runs per commit.
     """
+    # This should allow for much better query optimization then computing this in
+    # the WHERE clause using e.g. `now() - INTERVAL '14 days'`.
```
pitrou: I don't know if I got this explanation right, but otherwise what was the point of the redundant `timestamp` comparisons in the original query?

pitrou:
Also cc @prmoore77

amoeba:
Hi @pitrou: @assignUser actually already reached out to staff at VoDa and someone said they'll take a look. I also asked about expanding permissions on this repo and will report back.

prmoore77:
> Also cc @prmoore77

Hi @pitrou - I am no longer with VoDa - was there something I can help with on here?

conbench/entities/benchmark_result.py
austin3dickey: If I recall correctly, a couple years ago (https://github.com/conbench/conbench/pull/1546) we added a special [partial index](https://www.postgresql.org/docs/current/indexes-partial.html) to all Conbench database instances (Arrow, VoDa, etc). The partial index has [WHERE timestamp >= '2023-11-19'](https://github.com/conbench/conbench/pull/1546/files#diff-911e29331f989dc0911f4678c067f56e8fa3b4a530cd944584ac6c49c8988f19R24) in it, so this SQL query takes advantage of that by explicitly hardcoding that date. When I was messing around with the query, I noticed that Postgres would not use this partial index unless we had the hardcoded date in there. That's why there were 2 timestamp filters in the query.

I'm guessing this query is slow now because this partial index has grown considerably large over the last 2 years. When we initially merged that PR, we planned to semi-regularly adjust this date and refresh the partial index in the future (or even better, switch to a better partitioning scheme like TimescaleDB).

I don't have access to administrating the databases anymore so I'm not sure that I could personally refresh this index. If someone were interested in doing so, that PR would probably look like
- adding an alembic migration to delete the 2023-11-19 index and add a new one
- switching the hardcoded 2023-11-19 dates in the codebase to the new one

Maybe merging a PR like that would just work? But since I can't see the Buildkite builds anymore, I wouldn't be able to help with failures during the migration.

---

Re: this PR, I'm not sure changing `now() - INTERVAL '14 days'` to a hardcoded date would help, since computing the former takes almost no time on the database side. I think the main problem is that this query just has too much data to sort through, and needs a better DB organization to do so.

pitrou: > The partial index has [WHERE timestamp >= '2023-11-19'](https://github.com/conbench/conbench/pull/1546/files#diff-911e29331f989dc0911f4678c067f56e8fa3b4a530cd944584ac6c49c8988f19R24) in it

Yuck. 

> I'm guessing this query is slow now because this partial index has grown considerably large over the last 2 years.

That's indeed probably the case.

> Re: this PR, I'm not sure changing `now() - INTERVAL '14 days'` to a hardcoded date would help, since computing the former takes almost no time on the database side. I think the main problem is that this query just has too much data to sort through, and needs a better DB organization to do so.

Perhaps. I don't really understand what is so special about what this query is doing and why it is written that way. And I have never done any advanced SQL anyway :)

pitrou:
> Hi @pitrou: @assignUser actually already reached out to staff at VoDa and someone said they'll take a look.

Well, according to @austin3dickey 's comment above, this PR probably won't be useful.

> I also asked about expanding permissions on this repo and will report back.

+1

conbench/entities/benchmark_result.py
pitrou: Do you think something like the following would work? It seems to pass the tests but I can't tell whether it would be faster:
```python
    query_text = f"""
        SELECT r.* FROM (
            SELECT DISTINCT ON (run_id) *
            FROM benchmark_result
            WHERE timestamp > (now() - INTERVAL '14 days')
            AND timestamp > '2023-06-03'
            ORDER BY run_id DESC
        ) r
        ORDER BY timestamp DESC
        LIMIT {n}
    """
```

austin3dickey: It might work, but you're right that it could be slower. I remember @jgehrcke spent a long time trying to optimize this: see https://github.com/conbench/conbench/issues/1466#issuecomment-1710323173 and below.

Ultimately we need to be able to test these against the database.

pitrou: Hmm, I don't know but the comment points to this:
https://stackoverflow.com/questions/66893968/select-distinct-is-slower-than-expected-on-my-table-in-postgresql

And there seems to be a fundamental difference. In the SO question, the DISTINCT ON involves a _subset_ of the primary key, which requires a "index skip scan". In this case, the DISTINCT ON involves an entire key (the `benchmark_result_run_id_index` is on the sole `run_id` column). So hopefully at least the DISTINCT part would use the index.

This discussion seems more directly related:
https://stackoverflow.com/a/7630564/10194

pitrou: But, yes, would need to try on an actual populated database...

pitrou: I updated the query now that I might understand things a bit better.

wesm:
This issue is not present on the experimental-v2 branch. From Codex: "PR 1614: the exact legacy issue appears addressed in v2 at the query-shape level. The old Flask/Python landing query from #1614 is gone. The v2 home page uses GET /api/runs/recent, backed by internal/db/query/benchmark_result.sql:100 and internal/service/runs.go:47. That query samples newest result rows by timestamp, selects recent run IDs, then aggregates only those runs. It does not use the legacy now() - interval '14 days' plus hardcoded 2023-11-19 predicate shape.

pitrou:
@wesm When you say the issue is not present, you mean that the perf problem is solved, or that the landing page actually displays something different? If the former then great!

pitrou:
Also cc @rok

jgehrcke:
I missed most of this here. Nostalgia!

wesm:
> you mean that the perf problem is solved

I believe it's this =) I'm working on getting the new version stood up against the current prod database so they can be examined side by side

